PR #78, opened 2018-09-13
Minor fix on the doc of `vec_method_register()`

yutannihilation:
It seems there's no function like `vec_method_dynamic()`, I guess they should be `vec_method_register()`.


hadley:
Thanks!

PR #85, opened 2018-09-17
Use `generics` package for generic flavors of coercion functions from base

DavisVaughan:
Following the example of `as.POSIXct.vctrs_vctr`, I implemented these to cast to a `new_*()` object with the correct (hopefully?) arguments.

Merge this after merging r-lib/generics#12 (failing builds are expected until this is merged)


hadley:
Can you make generics suggested for now? And then register the S3 methods dynamically in onLoad?

DavisVaughan:
`vec_method_register()` can't parse the function name correctly because of the `.` in `as.factor`.

``` r
vctrs:::vec_method_register("generics::as.factor.vctrs_vctr")
#> Error in vctrs:::vec_method_register("generics::as.factor.vctrs_vctr"): length(pieces) == 3 is not TRUE

strsplit("generics::as.factor.vctrs_vctr", "::|[.]")
#> [[1]]
#> [1] "generics"   "as"         "factor"     "vctrs_vctr"
```

I would say you could look for the first `.` from the right, but if this is going to be a general purpose function then I think you could still hit the case where an S3 class name could have a `.` in it, messing things up.

hadley:
Oh shoot, I forgot about that. Can you please file an issue?

DavisVaughan:
Alright, I think this is working right.

For classes that want to use this, it seems the correct behavior is still to follow what the `generics` readme says. Import and reexport `generics::as.factor` and then when you use `as.factor()` on an object of class `c("my_class", "vctrs_vctr")` the correct behavior happens from it dispatching on `as.factor.vctrs_vctr`, which calls `vec_cast`, which calls `vec_cast.factor.my_class`.

Lots of hoops that it jumps through along the way! 

This might be something to include in a `vctrs` help doc somewhere? If I was a developer trying to extend `vctrs`, I would wonder why `as.character.vctrs_vctr` worked out of the box but `as.factor.vctrs_vctr` didnt. I don't think my first thought would be to look to `generics`.

hadley:
I think the best place would be another bullet in the docs for `new_vctr()`. Would you mind adding?

DavisVaughan:
I've added a bullet point in `new_vctr`.
Also updated `as.difftime` to use `new_duration`.

I'm having trouble with the tests though. I tried adding tests to ensure that `vec_cast()` is called (see below), but the generics don't seem to actually get registered. Even if I add `library(generics)` inside the test (or outside the test and just somewhere in the `test-vctr.R` script) the generics don't seem to get registered when I run `devtools::test()`. (library-ing `generics` is enough to get the tests to run interactively)

[tests](https://github.com/DavisVaughan/vctrs/blob/feature/use-generics-package/tests/testthat/test-vctr.R#L321)

hadley:
Using `::` in the tests is the right thing to do, unless you want to importFrom generics.

I just fixed build failure; can you please rebase?

DavisVaughan:
Up to date now. fyi r-lib/generics#12 still needs to be merged for travis to pass

PR #102, opened 2018-09-25
Use `generics` package for generic versions of coercion functions from base

DavisVaughan:
This is a cleaner version (in terms of git commits) of #85


hadley:
For future reference, I'll usually do a squash commit, so no reason to create a new PR .

PR #127, opened 2018-10-26
Fix) Use `x` not `.data`

### LINKED ISSUE #126: new_vctr() docs mix `x` and `.data`
The first argument to `new_vctr()` is currently `.data`, but `x` is documented. This hasn't been caught by a check because `@keywords internal` is being used here.

https://github.com/r-lib/vctrs/blob/master/R/vctr.R#L53

DavisVaughan:
- Updated `new_vctr()` and `check_attr()` to consistently use `x` rather than `.data` in argument names and error messages.
- Updated 2 tests to use `x` rather than `.data` in their description.
- Updated 1 test that tried to use `x` as an attribute, which failed after this change.

Closes #126 

(getting rate limited on travis)

### DIFF
--- a/R/vctr.R
+++ b/R/vctr.R
@@ -56,26 +56,26 @@
 #' @export
 #' @keywords internal
 #' @aliases vctr
-new_vctr <- function(.data, ..., class = character()) {
-  if (!is_vector(.data)) {
-    stop("`.data` must be a vector type", call. = FALSE)
+new_vctr <- function(x, ..., class = character()) {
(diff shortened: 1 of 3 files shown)


hadley:
I should've been clearer 😞 — can you change the docs? `x` is too commonly used as a field.

DavisVaughan:
Woops. I just assumed `x`, should have asked. I'm going to nuke this and restart since that's a much simpler change.

PR #128, opened 2018-10-26
Use .data rather than x in the docs of new_vctr()

### LINKED ISSUE #126: new_vctr() docs mix `x` and `.data`

DavisVaughan:
* Updated docs of `new_vctr()` to use `.data` rather than `x`

Closes #126


hadley:
Thanks!

PR #131, opened 2018-11-06
Typos and vignette proofreading

echasnovski:
I really enjoyed reading vignettes of this package and can't wait for it to reach "stable" lifecycle. Along the way found these improvement opportunities. Sorry for mixing in one PR code and vignette updates.

Extra thoughts:

- In 's3-vector.Rmd' vignette [there is a description](https://github.com/r-lib/vctrs/blob/4d848c977556d19af320e9dbfd2a3487eb62999b/vignettes/s3-vector.Rmd#L820) about arithmetic between meters and numbers. I would argue that only both multiplications and division by a number are acceptable, as, strictly speaking, meter and number are object of different nature and can't be used with `+` and `-`.
- In 'stability.Rmd' vignette `c()` [is called length-stable](https://github.com/echasnovski/vctrs/blob/1b165d0c0e78634a3dace8e649a793a4a2fd9a95/vignettes/stability.Rmd#L90) (using "always") with [later attempt](https://github.com/echasnovski/vctrs/blob/1b165d0c0e78634a3dace8e649a793a4a2fd9a95/vignettes/stability.Rmd#L169) to rightfully refute this. That is an attempt, because it doesn't actually show the breaking behaviour (see [pkgdown site](https://vctrs.r-lib.org/articles/stability.html#non-vectors)), so extra objects in global environment needed. Maybe it is better to use "almost length-stable" in the examples and [later "Data frames" section](https://github.com/echasnovski/vctrs/blob/1b165d0c0e78634a3dace8e649a793a4a2fd9a95/vignettes/stability.Rmd#L274)

### DIFF
--- a/R/prototype.R
+++ b/R/prototype.R
@@ -163,7 +163,7 @@ vec_ptype <- function(...) {
       step_lines <- apply(steps, 2, paste0, collapse = "")
     }
 
-    cat_line("Protoype: ", out_full[[n]])
+    cat_line("Prototype: ", out_full[[n]])
     cat_line(step_lines)
   }
 
--- a/vignettes/s3-vector.Rmd
+++ b/vignettes/s3-vector.Rmd
@@ -75,7 +75,7 @@ In this section you'll learn how to create a new vctrs class by calling `new_vct
   data-independent, and can easily be customised when the attributes do 
   depend on the data.
 
-* Default subset-assignment methods (`[<-`, `[[<-`, `$<-`) follow theprinciple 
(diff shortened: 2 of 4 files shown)


hadley:
Thank you!!

hadley:
Do you want to do a PR for the other issues? I think the additional restriction on meters makes sense (and makes the problem a bit more interesting). For `c()` it might be better to create a custom environment, rather than being at the mercy of what ever is in the global env.

echasnovski:
Sure. But not right away, as this needs a fully operating git terminal and not only GitHub in the browser :) I hope to get to one in couple of hours.

PR #134, opened 2018-11-06
Fix) replacement op for `vctrs_vctr`

### LINKED ISSUE #133: Use `...` not `i` in `[[<-.vctrs_vctr`
Currently, you are using 

```
`[[<-.vctrs_vctr`(x, i, value)
```

I propose that this would work "better"

```
`[[<-.vctrs_vctr`(x, ..., value) 
```

My use case is in extending `vctrs_vctr` with an array class on top of it, `vctrs_rray`. I need to be able to do:

```
x[[1, 1, 1]] <- 4
```

but right now I can't because only `i` is passed through. I would bypass it with my own `[[<-.vctrs_rray` but I am not sure what it should do, as calling `NextMethod()` will send me straight into `[[<-.vctrs_vctr`.

I altered the `vctrs` code to use `...` rather than `i` and no tests break, so that's encouraging.

For whatever reason, `factor`s do this:

```
> args(`[[<-.factor`)
function (x, ..., value) 
NULL
```

DavisVaughan:
Closes #133 

* `i` is also used for `vctrs_sclr`, `vctrs_list_of`, and `vctrs_rcrd`, but only `vctrs_list_of` calls `NextMethod()` so it would be the only other one where I think this might apply. I did not make the change for it though.

* I added a test that somewhat shows what I was trying to do.

### DIFF
--- a/R/vctr.R
+++ b/R/vctr.R
@@ -214,7 +214,7 @@ diff.vctrs_vctr <- function(x, lag = 1L, differences = 1L, ...) {
 # Modification -------------------------------------------------------------
 
 #' @export
-`[[<-.vctrs_vctr` <- function(x, i, value) {
+`[[<-.vctrs_vctr` <- function(x, ..., value) {
   if (!is.list(x)) {
     value <- vec_cast(value, x)
   }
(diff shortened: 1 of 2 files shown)


hadley:
Thanks!

PR #143, opened 2018-11-27
Fix) Recursively call `vec_slice()` when subsetting data frames

### LINKED ISSUE #142: vec_slice() doesn't work correctly with data frame columns
I'm not sure if you were trying to support these yet, but data frame columns (and matrix columns) don't work with `vec_slice()` currently. One solution might be to call `vec_slice()` instead of `[` when you iterate over the columns of the data frame. I don't know if this greatly slows things down or not.

``` r
library(vctrs)
library(tibble)

x <- tibble(x = c(1,2), y = tibble(z = c(2, 5), y = c(3, 4)))
x
#> # A tibble: 2 x 2
#>       x   y$z    $y
#>   <dbl> <dbl> <dbl>
#> 1     1     2     3
#> 2     2     5     4

res <- vec_slice(x, 1L)
res
#> # A tibble: 1 x 2
#>       x   y$z
#>   <dbl> <dbl>
#> 1     1     2

str(res)
#> Classes 'tbl_df', 'tbl' and 'data.frame':    1 obs. of  2 variables:
#>  $ x: num 1
#>  $ y:Classes 'tbl_df', 'tbl' and 'data.frame':   2 obs. of  1 variable:
#>   ..$ z: num  2 5

vec_slice2 <- function (x, i) {
  if (is.logical(i)) {
    i <- which(i)
  }
  stopifnot(is.integer(i) || is.character(i))
  if (is.null(x)) {
    NULL
  }
  else if (is.data.frame(x)) {
    out <- lapply(x, vec_slice, i)
    #out <- lapply(x, `[`, i)
    vec_restore(out, x)
  }
  else if (is_vector(x)) {
    d <- vec_dims(x)
    if (d == 1) {
      if (is.object(x)) {
        x[i]
      }
      else {
        x[i, drop = FALSE]
      }
    }
    else if (d == 2) {
      x[i, , drop = FALSE]
    }
    else {
      eval_bare(expr(x[i, !!!miss_args, drop = FALSE]))
    }
  }
  else {
    stop("`x` must be a vector", call. = FALSE)
  }
}

vec_slice2(x, 1L)
#> # A tibble: 1 x 2
#>       x   y$z    $y
#>   <dbl> <dbl> <dbl>
#> 1     1     2     3
```

Created on 2018-11-26 by the [reprex
package](http://reprex.tidyverse.org) (v0.2.0).

DavisVaughan:
I also added a test for data frame columns (I believe that is the correct base R approach). Feel free to drop it if you don't think it is necessary.

Closes #142

### DIFF
--- a/tests/testthat/test-size.R
+++ b/tests/testthat/test-size.R
@@ -50,12 +50,10 @@ test_that("can subset object of any dimensionality", {
 })
 
 test_that("can subset data frame columns", {
-  df <- data.frame(x = c(1, 2), y = I(data.frame(z = c(1, 2), w = c(1, 2))))
+  df <- data.frame(x = 1:2)
+  df$y <- data.frame(a = 2:1)
 
-  base_slice <- df[1, , drop = FALSE]
-  rownames(base_slice) <- NULL
-
-  expect_equal(vec_slice(df, 1L), base_slice)
+  expect_equal(vec_slice(df, 1L)$y, vec_slice(df$y, 1L))
 })
 
 test_that("can modify subset", {
--- a/tests/testthat/test-size.R
+++ b/tests/testthat/test-size.R
@@ -49,6 +49,11 @@ test_that("can subset object of any dimensionality", {
   expect_equal(vec_slice(x4, 1L), ones(1, 3, 4, 5))
 })
 
+test_that("can subset data frame columns", {
+  df <- data.frame(x = c(1, 2), y = I(data.frame(z = c(1, 2), w = c(1, 2))))
+  expect_equal(vec_slice(df, 1L), df[1, , drop = FALSE])
+})
+
 test_that("can modify subset", {
   x1 <- c(2, 1)
   vec_slice(x1, 1L) <- 1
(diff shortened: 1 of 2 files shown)


tests/testthat/test-size.R
```diff
@@ -49,6 +49,15 @@ test_that("can subset object of any dimensionality", {
   expect_equal(vec_slice(x4, 1L), ones(1, 3, 4, 5))
 })
 
+test_that("can subset data frame columns", {
+  df <- data.frame(x = c(1, 2), y = I(data.frame(z = c(1, 2), w = c(1, 2))))
```
hadley: I think it's slightly better to add after the fact:

```R
df <- data.frame(x = 1:2)
df$y <- data.frame(a = 2:1)
```

tests/testthat/test-size.R
```diff
@@ -49,6 +49,15 @@ test_that("can subset object of any dimensionality", {
...
+
+  base_slice <- df[1, , drop = FALSE]
+  rownames(base_slice) <- NULL
+
+  expect_equal(vec_slice(df, 1L), base_slice)
```
hadley: I think you could be more direct here:

```R
expect_equal(vec_slice(df, 1L)$y, vec_slice(df$y, 1L))
```

tests/testthat/test-size.R
DavisVaughan: That makes more sense, especially considering I couldn't even print `df` before

tests/testthat/test-size.R
DavisVaughan: 👍

hadley:
Can you resubmit this somehow?

PR #154, opened 2019-01-07
Alter the examples in vec_cast() to demonstrate the lossy-cast condition system

DavisVaughan:
I think the original example was incorrect because it says a warning is thrown when in reality an error is thrown because none of the conversions are lossless. I think it may have been a warning in an early version of vctrs.

### DIFF
--- a/R/cast.R
+++ b/R/cast.R
@@ -79,9 +79,14 @@
 #' # x is a double, but no information is lost
 #' vec_cast(1, integer())
 #'
-#' # Information is lost so a warning is generated
+#' # Information is lost in all conversions so an error is generated
 #' \dontrun{
 #' vec_cast(1.5, integer())
+#' }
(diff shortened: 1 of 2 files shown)


R/cast.R
```diff
@@ -79,11 +79,16 @@
...
 #' vec_cast(1, integer())
 #'
-#' # Information is lost so a warning is generated
+#' # Information is lost in all conversions so an error is generated
 #' \dontrun{
 #' vec_cast(1.5, integer())
 #' }
 #'
+#' # Information is lost in some conversions so a warning is generated
+#' \dontrun{
```
hadley: I think this can run, since it's only a warning?

DavisVaughan: ✅

hadley:
Thanks!

PR #174, opened 2019-01-28
All error messages and warnings end with a full stop.

krlmlr:
(Except one instance that conflicts with an open branch of mine and will be fixed there.)


hadley:
Thanks

Issue #171, opened 2019-01-28
vec_slice<-() should not add names


krlmlr:
Currently, can assign a named slice to an unnamed vector. This should be an error.

Related to #140?

``` r
library(vctrs)

a <- new_vctr(1:3)
vec_slice(a, "x") <- 4L
a
#> <vctrs_vctr[4]>
#>       x 
#> 1 2 3 4

new_vctr(c(1:3, a = 4))
#> Error: If any elements of `.data` are named, all must be named
```

<sup>Created on 2019-01-28 by the [reprex package](https://reprex.tidyverse.org) (v0.2.1.9000)</sup>

krlmlr:
@hadley: How do we ensure that all vectors are either named everywhere or unnamed, even after updates?

hadley:
We should only allow slicing with a character vector if the vector has names?

PR #176, opened 2019-01-28
Required changes for blob support

krlmlr:
Commits are good for individual review.


krlmlr:
Will add more tests.

R/c.R
```diff
@@ -39,8 +39,8 @@ vec_c <- function(..., .ptype = NULL) {
     return(NULL)
 
   ns <- map_int(args, vec_size)
-  out <- vec_na(ptype, sum(ns))
-  if (is.null(names(args))) {
+  out <- vec_na(unname(ptype), sum(ns))
+  if (is.null(names(args)) && all(map_lgl(map(args, vec_names), is.null))) {
```
hadley: Maybe 

```R
has_inner_names <- function(x) {
  !all(map_lgl(map(x, vec_names), is.null))
}
```

R/list_of.R
```diff
@@ -120,6 +120,13 @@ as.list.vctrs_list_of <- function(x, ...) {
 
 #' @export
 `[[<-.vctrs_list_of` <- function(x, i, value) {
+  if (is.null(value)) {
+    # Setting to NULL via [[ shortens the list! Example:
```
hadley: I think if we're going to do this  we should also add a check for `rlang::zap()`, so there's still some way to remove elements.

R/size.R
```diff
@@ -138,11 +138,14 @@ vec_slice <- function(x, i) {
     d <- vec_dims(x)
     if (d == 1) {
       x[i] <- value
+      check_names_all_or_nothing(names(x))
```
hadley: Can't we move this check outside of the `if`?

R/vctr.R
```diff
@@ -563,7 +567,8 @@ levels.vctrs_vctr <- function(x) {
 
 #' @export
 `is.na<-.vctrs_vctr` <- function(x, value) {
-  stop_unsupported(x, "is.na<-")
+  vec_slice(x, value) <- vec_na(x)
```
hadley: I think this needs to assert that `value` is a logical vector and is size-compatible with `x`

krlmlr: This is done in `vec_slice<-()` . Integers and characters are also supported in base:

``` r
x <- 1:3
is.na(x) <- 2
x
#> [1]  1 NA  3

x <- c(a = 1, b = 2, c = 3)
is.na(x) <- "b"
x
#>  a  b  c 
#>  1 NA  3
```

<sup>Created on 2019-01-29 by the [reprex package](https://reprex.tidyverse.org) (v0.2.1.9000)</sup>

hadley: That feels overly broad to me; I think it makes more sense to only accept logical vectors.

R/list_of.R
krlmlr: Currently:

``` r
library(vctrs)

x <- list_of(1, 2, 3)
x[[2]] <- rlang::zap()
#> Error: Can't cast <rlang_zap> to <double>
#> Backtrace:
#>     █
#>  1. ├─base::`[[<-`(`*tmp*`, 2, value = structure(list(), class = "rlang_zap"))
#>  2. └─vctrs:::`[[<-.vctrs_list_of`(`*tmp*`, 2, value = structure(list(), class = "rlang_zap"))
#>  3.   ├─vctrs::vec_cast(value, attr(x, "ptype")) /home/kirill/git/R/vctrs/R/list_of.R:130:2
#>  4.   ├─vctrs::vec_cast.double(value, attr(x, "ptype")) /home/kirill/git/R/vctrs/R/cast.R:102:2
#>  5.   └─vctrs:::vec_cast.double.default(value, attr(x, "ptype")) /home/kirill/git/R/vctrs/R/cast.R:219:2
#>  6.     └─vctrs::stop_incompatible_cast(x, to) /home/kirill/git/R/vctrs/R/cast.R:251:2
#>  7.       └─vctrs:::stop_incompatible(...) /home/kirill/git/R/vctrs/R/conditions.R:53:2
```

<sup>Created on 2019-01-29 by the [reprex package](https://reprex.tidyverse.org) (v0.2.1.9000)</sup>

Do we need to add a special type, akin to `unspecified` ?

hadley: No, just special case it for this usage.

R/vctr.R
krlmlr: I see the following invariant:

```r
identical(`is.na<-`(x, i), `[<-`(x, i, NA))
```

It makes perfect sense to me to just pass through to `vec_slice<-()` , given that special-casing just adds complexity here.

hadley: `vec_assert(values, logical())` adds little complexity and I don't think that a broad interpretation of `values` is useful here.

R/list_of.R
krlmlr: Added, but do we really need it? How do we remove elements in an integer vector?

krlmlr:
Addressed all comments.

R/c.R
```diff
@@ -39,8 +39,8 @@ vec_c <- function(..., .ptype = NULL) {
     return(NULL)
 
   ns <- map_int(args, vec_size)
-  out <- vec_na(ptype, sum(ns))
-  if (is.null(names(args))) {
+  out <- vec_na(unname(ptype), sum(ns))
```
hadley: This doesn't feel quite right to me. Should the names be stripped in either `vec_na()` or `vec_ptype()`? Otherwise, you need to explain why `vec_c()` needs special behaviour.

R/c.R
```diff
@@ -54,8 +54,9 @@ vec_c <- function(..., .ptype = NULL) {
 
     x <- vec_cast(args[[i]], to = ptype)
```
hadley: Doesn't this now violate the vctrs invariant that vectors are either all named or all unnamed?

R/list_of.R
hadley: We don't worry about it.

R/size.R
```diff
@@ -144,6 +145,8 @@ vec_slice <- function(x, i) {
       miss_args <- rep(list(missing_arg()), d - 1)
       eval_bare(expr(x[i, !!!miss_args] <- value))
     }
+
+    check_names_all_or_nothing(names(x))
```
hadley: This seems like a weird place to declare this invariant. I think you should remove it.

tests/testthat/test-vctr.R
```diff
@@ -92,6 +92,19 @@ test_that("equality functions remapped", {
   expect_true(anyDuplicated(x))
 })
 
+test_that("is.na<-() and new_vctr() commute", {
+  x <- new_vctr(1:4)
+  y <- 1:4
+
+  expect_error(is.na(x) <- c(FALSE, FALSE, TRUE, NA), NA)
```
hadley: Why is this an error?

R/list_of.R
krlmlr: > so there's still some way to remove elements.

So why do we worry about removing elements for list vectors?

I don't mind either way, just being picky here because I think consistency is important:

- Setting elements to `NA` should be supported (`NULL` for lists)
- Removing elements should be either supported consistently, or not supported at all

R/c.R
krlmlr: The names are restored afterwards: https://github.com/r-lib/vctrs/pull/176/files#diff-8659cf38b76d4f1d2bda41fdb89168d4R64.

Anyway, removing this change for now.

R/size.R
krlmlr: We have an issue for it: #171. Postponing, this doesn't affect _blob_ either.

R/c.R
krlmlr: Removing this change for now.

tests/testthat/test-vctr.R
krlmlr: It's not. Fixed here, added an issue to handle other instances of this anti-pattern.

tests/testthat/test-vctr.R
```diff
@@ -92,6 +92,19 @@ test_that("equality functions remapped", {
   expect_true(anyDuplicated(x))
 })
 
+test_that("is.na<-() and new_vctr() commute", {
+  x <- new_vctr(1:4)
+  y <- 1:4
+
+  expect_silent(is.na(x) <- c(FALSE, FALSE, TRUE, NA))
```
hadley: I still don’t understand what you are trying to test here. I think it would make more sense to set a specific value to NA with is.na<- and then check that the value at that position equals the correct NA.

krlmlr: Simplified the test, still testing the entire vector (because `is.na<-()` isn't supposed to touch other values).

R/list_of.R
hadley: I've changed my mind on this after talking to @lionel-. I think we should ensure that ``vec_size(`[[<-`(x, i, value)`` equals `vec_size(x)`, so that we don't need to support `zap()` here.

krlmlr: I've removed special-casing for `zap` here.

The `vec_size()` still can change when creating new elements past the existing length.

krlmlr:
Travis succeeds now. Anything else we want to address here?

hadley:
Thanks!

PR #184, opened 2019-02-10
vec_assert() throws classed error

krlmlr:
Closes #168.

PR #187, opened 2019-02-12
Prefix error classes with vctrs_

krlmlr:
Not sure about the existing `"error_unimplemented"` and `"error_unsupported"` classes, do they need a prefix too?

The class names should be part of the documented API, too.


krlmlr:
I like that approach. Should we start populating the _error_ package now?

hadley:
No

krlmlr:
Do we need to add/change anything here?

krlmlr:
Merging so that we can move forward with #184. Let's discuss there if any changes are needed.

hadley:
In the future, please don't merge with out approval from me.

PR #192, opened 2019-02-18
Adding difftime to date truncates the difftime to full dates

krlmlr:
For full _hms_ support. Do we need a lossy cast warning?


hadley:
Yes, we need a warning here.

R/type-date-time.R
```diff
@@ -427,24 +427,34 @@ vec_arith.difftime.POSIXct <- function(op, x, y) {
 #' @method vec_arith.Date difftime
 #' @export
 vec_arith.Date.difftime <- function(op, x, y) {
+  # Exit early to avoid warning
+  if (!(op %in% c("+", "-"))) {
```
hadley: I don't like moving this up front. Instead could we make a wrapper function, something like this:

```R
round_warn <- function(x, to) {
  x_floor <- floor(x)
  report_lossy_cast(x, to, x != x_floor)
  x_floor
}
```

Then could call instead `vec_arith_base()`?

krlmlr: Do you mind the fallthrough I've added?

hadley: Yeah, that's fine.

PR #194, opened 2019-02-18
Allow simple class names for one-class packages

krlmlr:
https://github.com/tidyverse/blob/pull/11#discussion_r257837961


hadley:
Thanks

PR #208, opened 2019-02-27
Optimize vec split

DavisVaughan:


DavisVaughan:
The only time I can make this fail is:

```r
# works fine
vctrs:::vec_self_split(1:49988)

# fails
vctrs:::vec_self_split(1:49989)
```

It fails with the error:

```
Error: protect(): protection stack overflow
```

And I tracked it down to the `growable_init()` function. In this scenario, where all values are distinct, it creates a growable for each individual value. So you get 49989 growables. During the initialization of the last one, at the `PROTECT_WITH_INDEX()` call, that error is thrown.

lionel-:
Thanks for looking into this @DavisVaughan. I think this might be getting a bit too complicated. Would the following strategy avoid the need for a growable of growables?

1. Record info: In a dict loop, find duplicate ids of `by`. Record keys as we see them, in order. Increase counts for each key. Return all of these.

2. Subset values: For each id, allocate a vector of type `x` with the unique count as size, then subset `x` to fill it in.

3. Wrap keys and vals in a tibble.

DavisVaughan:
The fact that I ended up with a growable of growables definitely made me think it was too complicated (but pretty cool that something like that can exist).

I have two main thoughts about your points:

* Do you really need the counts? Is your `2. Subset values` bullet happening at the C level or the R level? If it is at the R, we could just subset `x` as required for each key and it would have the right type and we could store each subset all together in a list.

* I think the part I am most stuck on is how to efficiently construct the indices to subset `x` with. I don't see how to do that with the information you suggest I collect. Can you finish out this example for me?

``` r
library(vctrs)

x  <- c("v", "w", "x", "y", "z")
by <- c("a", "b", "a", "c", "b")

# ////////////////////
# Record info

# Assume we can get these things at
# the C level

# can record keys as we see them easily
# with the dictionary+growable idea
keys <- c("a", "b", "c")

# this is easy enough, `vctrs_id()`
# does this already
dup_id <- c(1, 2, 1, 4, 2)

# this is slightly harder, I think
# but could be done if its actually required
counts <- c(2, 2, 1)

# ////////////////////
# Subset x

# Really what I want is this below,
# but I don't see how the pieces of 
# info above _efficiently_ get me
# here. The info is there in 
# `dup_id`, but how can it be 
# efficiently used?
x_a <- vec_slice(x, c(1L, 3L))
x_b <- vec_slice(x, c(2L, 5L))
x_c <- vec_slice(x, c(4L))

# Result
vals <- new_list_of(list(x_a, x_b, x_c), vec_type(x))
new_data_frame(list(key = keys, val = vals), n = vec_size(keys))
#>   key  val
#> 1   a v, x
#> 2   b w, z
#> 3   c    y
```

<sup>Created on 2019-02-28 by the [reprex package](https://reprex.tidyverse.org) (v0.2.1.9000)</sup>

lionel-:
We should probably implement it entirely in C, though I'm not sure we have the right primitives implemented at C level yet (I've just started to dive into vctrs).

I think the part we're missing is a way of associating ids to their key position, which is also the position in the output, see `which_key()` below. We might not be able to avoid allocating a second hash table. Though this one will only be large when there are many unique values in `by`, so that seems acceptable.

```r
keys <- c("a", "b", "c")
dup_id <- c(1, 2, 1, 4, 2)
counts <- c(2, 2, 1)


# This associates the duplicates id with their key id, i.e. their
# position in the output
which_key <- function(id) {
  switch(as.character(id),
    `1` = 1,  # "a"
    `2` = 2,  # "b"
    `4` = 3,  # "c"
    stop("Unknown `id`")
  )
}

# Allocate output and its components at the right size (wouldn't use
# seq_along() at C level)
out <- vector("list", length(counts))

for (i in seq_along(counts)) {
  out[[i]] <- x[0][seq_len(counts[[i]])]
}


# Now we loop over indices of `x`
counters <- integer(length(out)) + 1

for (i in seq_along(x)) {
  id <- dup_id[[i]]
  j <- which_key(id)
  k <- counters[[j]]

  out[[j]][[k]] <- x[[i]]

  counters[[j]] <- k + 1L
}

out

vals <- new_list_of(out, vec_type(x))
new_data_frame(list(key = keys, val = vals), n = vec_size(keys))
```

DavisVaughan:
@lionel- these were helpful comments, thanks! I don't think the infrastructure is currently there for doing exactly what you propose, _but_ I was able to use your approach the collect the index information that `x` is split with, and rewrite without any growables. I've left both approaches in there for the moment for benchmark comparisons.

The new way uses just slightly more memory than the growables approach, but is still less than base R. 

I was mostly impressed with how fast it was in the "worst case" scenarios. This happens for the growables approach when all the values are unique. See the example that has the comment `MUCH FASTER THAN THE OLD WAY!` for the most impressive demo of this. The tl;dr is that it is about 3x faster than the old way.

<details>

``` r
library(vctrs)
library(rlang)
```

``` r

vctrs:::vec_duplicate_split_new(c(1L, 1L, 2L, 1L, 3L, 2L))
#> [[1]]
#> [1] 1 2 4
#> 
#> [[2]]
#> [1] 3 6
#> 
#> [[3]]
#> [1] 5
```

``` r

big_iris <- dplyr::bind_rows(!!!rep_along(rep(1, 50000), list(iris)))

nrow(big_iris)
#> [1] 7500000
```

``` r

bench::mark(
  x <- vctrs:::vec_split_new(big_iris, big_iris$Species),
  x <- vctrs:::vec_split_old(big_iris, big_iris$Species),
  x <- vec_split(big_iris, big_iris$Species)
)
#> disabled.
#> # A tibble: 3 x 10
#>   expression   min  mean median   max `itr/sec` mem_alloc  n_gc n_itr
#>   <chr>      <bch> <bch> <bch:> <bch>     <dbl> <bch:byt> <dbl> <int>
#> 1 x <- vctr… 525ms 525ms  525ms 525ms      1.90     564MB     2     1
#> 2 x <- vctr… 564ms 564ms  564ms 564ms      1.77     471MB     2     1
#> 3 x <- vec_… 866ms 866ms  866ms 866ms      1.15     850MB     2     1
```

``` r
# same, but calling vec_duplicate_split() directly to compare

bench::mark(
  x <- vctrs:::vec_duplicate_split_new(big_iris$Species),
  x <- vctrs:::vec_duplicate_split_old(big_iris$Species),
  x <- unname(split(vec_seq_along(big_iris$Species), vec_duplicate_id(big_iris$Species)))
)
#> # A tibble: 3 x 10
#>   expression   min  mean median   max `itr/sec` mem_alloc  n_gc n_itr
#>   <chr>      <bch> <bch> <bch:> <bch>     <dbl> <bch:byt> <dbl> <int>
#> 1 x <- vctr… 297ms 299ms  299ms 300ms      3.35     306MB     0     2
#> 2 x <- vctr… 237ms 237ms  237ms 237ms      4.21     214MB     1     1
#> 3 x <- unna… 494ms 494ms  494ms 494ms      2.03     414MB     1     1
```

``` r
# what happens when there are a ton of unique values?

x <- 1:5000

mrk <- bench::mark(
  vctrs:::vec_duplicate_split_new(x),
  vctrs:::vec_duplicate_split_old(x),
  unname(split(vec_seq_along(x), vec_duplicate_id(x)))
)

mrk
#> # A tibble: 3 x 10
#>   expression      min     mean   median      max `itr/sec` mem_alloc  n_gc
#>   <chr>      <bch:tm> <bch:tm> <bch:tm> <bch:tm>     <dbl> <bch:byt> <dbl>
#> 1 vctrs:::v… 280.94µs 335.02µs  314.1µs 556.19µs     2985.     233KB    11
#> 2 vctrs:::v… 738.38µs 835.29µs 795.44µs   2.42ms     1197.     142KB     6
#> 3 unname(sp…   1.13ms   1.29ms   1.26ms   1.58ms      775.     616KB     5
#> # … with 2 more variables: n_itr <int>, total_time <bch:tm>
```

``` r
# what happens when there are REALLY REALLY BIG amount of values?
# (Worst case - all unique)

# MUCH FASTER THAN THE OLD WAY!

# many many unique values
x <- 1:500000

mrk2 <- bench::mark(
  y <- vctrs:::vec_duplicate_split_new(x),
  y <- vctrs:::vec_duplicate_split_old(x),
  y <- unname(split(vec_seq_along(x), vec_duplicate_id(x)))
)

mrk2
#> # A tibble: 3 x 10
#>   expression     min    mean  median     max `itr/sec` mem_alloc  n_gc
#>   <chr>      <bch:t> <bch:t> <bch:t> <bch:t>     <dbl> <bch:byt> <dbl>
#> 1 y <- vctr…  49.6ms  66.4ms  73.7ms  74.4ms     15.1     25.4MB     2
#> 2 y <- vctr… 196.5ms 196.5ms 196.5ms 196.5ms      5.09    15.6MB     2
#> 3 y <- unna… 209.3ms 211.8ms 212.9ms   213ms      4.72    57.1MB     0
#> # … with 2 more variables: n_itr <int>, total_time <bch:tm>
```

``` r
# what happens when there are REALLY REALLY BIG amount of values?
# (Best case - all same)

# many many unique values
x <- rep(1, times = 500000)

mrk2b <- bench::mark(
  y <- vctrs:::vec_duplicate_split_new(x),
  y <- vctrs:::vec_duplicate_split_old(x),
  y <- unname(split(vec_seq_along(x), vec_duplicate_id(x)))
)

mrk2b
#> # A tibble: 3 x 10
#> 1 y <- vctr… 23.4ms 24.6ms 24.4ms 27.8ms      40.6    23.4MB     1    18
#> 2 y <- vctr… 19.9ms 21.6ms 21.6ms 22.9ms      46.4    17.5MB     0    24
#> 3 y <- unna…   34ms 36.8ms 37.2ms   38ms      27.2    30.9MB     1    12
```

``` r
# What about a much smaller amount? - Worst case

x <- 1:10

mrk3 <- bench::mark(
  y <- vctrs:::vec_duplicate_split_new(x),
  y <- vctrs:::vec_duplicate_split_old(x),
  y <- unname(split(vec_seq_along(x), vec_duplicate_id(x)))
)

mrk3
#> # A tibble: 3 x 10
#>   expression    min   mean median     max `itr/sec` mem_alloc  n_gc n_itr
#>   <chr>      <bch:> <bch:> <bch:> <bch:t>     <dbl> <bch:byt> <dbl> <int>
#> 1 y <- vctr… 10.4µs 12.9µs 12.9µs  71.6µs    77360.        0B     1  9999
#> 2 y <- vctr…   12µs 13.7µs   13µs  58.9µs    73185.        0B     1  9999
#> 3 y <- unna… 42.5µs 49.4µs 46.3µs 140.6µs    20233.        0B     4  8333
```

``` r
# What about a much smaller amount? - Best case

x <- rep(1, times = 100)

mrk3b <- bench::mark(
  y <- vctrs:::vec_duplicate_split_new(x),
  y <- vctrs:::vec_duplicate_split_old(x),
  y <- unname(split(vec_seq_along(x), vec_duplicate_id(x)))
)

mrk3b
#> # A tibble: 3 x 10
#>   expression    min   mean median   max `itr/sec` mem_alloc  n_gc n_itr
#>   <chr>      <bch:> <bch:> <bch:> <bch>     <dbl> <bch:byt> <dbl> <int>
#> 1 y <- vctr… 13.8µs   19µs 18.7µs 132µs    52599.    5.69KB     1  9999
#> 2 y <- vctr… 14.3µs 18.4µs 18.2µs 137µs    54309.     4.2KB     0 10000
#> 3 y <- unna… 47.9µs 58.7µs 58.4µs 234µs    17026.    7.26KB     2  7969
```

``` r
# Even less? - Worst case

x <- 1:5

mrk4 <- bench::mark(
  y <- vctrs:::vec_duplicate_split_new(x),
  y <- vctrs:::vec_duplicate_split_old(x),
  y <- unname(split(vec_seq_along(x), vec_duplicate_id(x)))
)

mrk4
#> # A tibble: 3 x 10
#>   expression    min   mean median     max `itr/sec` mem_alloc  n_gc n_itr
#>   <chr>      <bch:> <bch:> <bch:> <bch:t>     <dbl> <bch:byt> <dbl> <int>
#> 1 y <- vctr… 10.5µs 12.6µs 12.8µs  64.4µs    79148.        0B     1  9999
#> 2 y <- vctr… 11.4µs 13.4µs 13.4µs    46µs    74412.        0B     1  9999
#> 3 y <- unna… 42.4µs 48.9µs 47.7µs 302.1µs    20437.        0B     3  9906
```

``` r
# Even less? - Best case

x <- rep(1, times = 5)

mrk4b <- bench::mark(
  y <- vctrs:::vec_duplicate_split_new(x),
  y <- vctrs:::vec_duplicate_split_old(x),
  y <- unname(split(vec_seq_along(x), vec_duplicate_id(x)))
)

mrk4b
#> # A tibble: 3 x 10
#>   expression    min   mean median     max `itr/sec` mem_alloc  n_gc n_itr
#>   <chr>      <bch:> <bch:> <bch:> <bch:t>     <dbl> <bch:byt> <dbl> <int>
#> 1 y <- vctr… 10.6µs 12.5µs 12.6µs  64.9µs    80006.        0B     0 10000
#> 2 y <- vctr… 10.9µs 13.2µs 13.4µs  62.2µs    75615.        0B     1  9999
#> 3 y <- unna…   42µs 47.7µs 45.7µs 185.7µs    20963.        0B     3  9997
```

<sup>Created on 2019-03-01 by the [reprex package](https://reprex.tidyverse.org) (v0.2.1.9000)</sup>

</details>

DavisVaughan:
It currently still does not return the unique keys themselves. It could do this, but I kind of think that this `vec_duplicate_split()` function would be useful on its own. That's the only reason I left them out and collect them on the R side.

lionel-:
Looks great!

I tried the following patch to avoid `INTEGER()` within the loop but it only makes a difference for very large numbers of unique values. The gain is pretty slight in that case, and the performance is actually hurt by the additional allocation in the other cases. Your new implementation seems optimal to me.

```diff
diff --git a/src/dictionary.c b/src/dictionary.c
index 3116a34..bd58fd9 100644
--- a/src/dictionary.c
+++ b/src/dictionary.c
@@ -380,10 +380,10 @@ SEXP vctrs_duplicate_split_new(SEXP x) {
   R_len_t n = vec_size(x);
 
   // Tells us which element of the final list x[i] goes in
-  SEXP out_pos = PROTECT(Rf_allocVector(INTSXP, n));
-  int* p_out_pos = INTEGER(out_pos);
+  SEXP j = PROTECT(Rf_allocVector(INTSXP, n));
+  int* p_out_pos = INTEGER(j);
 
-  // Fill dictionary, out_pos, and count
+  // Fill dictionary, j, and count
   for (int i = 0; i < n; ++i) {
     uint32_t k = dict_find(&d, x, i);
 
@@ -403,13 +403,21 @@ SEXP vctrs_duplicate_split_new(SEXP x) {
   int* p_counters = INTEGER(counters);
   memset(p_counters, 0, d.used * sizeof(int));
 
+  // Addresses of integer arrays ordered by key appearance
+  int** out_ptrs = (int**) RAW(PROTECT(Rf_allocVector(RAWSXP, sizeof(int*) * d.used)));
+
   // Set up empty container
   for (int k = 0; k < d.size; ++k) {
     if (d.key[k] == EMPTY) {
       continue;
     }
 
-    SET_VECTOR_ELT(out, p_tracker[k], Rf_allocVector(INTSXP, p_count[k]));
+    int j = p_tracker[k];
+
+    SEXP elt = Rf_allocVector(INTSXP, p_count[k]);
+    SET_VECTOR_ELT(out, j, elt);
+    out_ptrs[j] = INTEGER(elt);
   }
 
   // Fill container
@@ -417,11 +425,11 @@ SEXP vctrs_duplicate_split_new(SEXP x) {
     int j = p_out_pos[i];
     int k = p_counters[j];
 
-    INTEGER(VECTOR_ELT(out, j))[k] = i + 1;
+    out_ptrs[j][k] = i + 1;
     p_counters[j] = k + 1;
   }
 
-  UNPROTECT(5);
+  UNPROTECT(6);
   dict_free(&d);
   return out;
 }
```

DavisVaughan:
I basically tried this too and came to the same conclusion!
https://github.com/DavisVaughan/vctrs/commit/3c1e25c01cecbd609c49b2290fe97787cd12fed0

R/dictionary.R
```diff
@@ -280,3 +292,35 @@ vec_split <- function(x, by) {
...
+#' @export
+vec_split_new <- function(x, by) {
+
+  if (vec_size(x) != vec_size(by)) {
+    abort("`x` and `by` must have same size")
+  }
+
+  idx <- vec_duplicate_split_new(by)
+
+  keys <- vec_slice(by, map_int(idx, function(x) {x[1]}))
```
hadley: Could you return some thing from the C function to avoid having to do this in R?

R/dictionary.R
```diff
@@ -280,3 +292,35 @@ vec_split <- function(x, by) {
...
+
+  vals <- new_list_of(map(idx, vec_slice, x = x), vec_type(x))
```
hadley: I'd pull the map out on to it's own line for clarity

hadley:
Can you please also use `devtools::test_coverage()` to make sure that all your C code is covered by unit tests?

R/dictionary.R
DavisVaughan: ✅

R/dictionary.R
DavisVaughan: ✅

DavisVaughan:
@hadley done and ready for review. `test_coverage()` shows that all lines of `vec_duplicate_split()` and `vec_split()` are hit by existing unit tests.

R/dictionary.R
```diff
@@ -269,14 +269,25 @@ vec_in <- function(needles, haystack) {
 #'   as_tibble(vec_split(mtcars, mtcars[c("vs", "am")]))
 #' }
 vec_split <- function(x, by) {
-  # TODO: optimise to avoid creating the dictionary twice and to avoid split()
-  keys <- vec_unique(by)
+
```
lionel-: Extra space.

R/dictionary.R
```diff
@@ -269,14 +269,25 @@ vec_in <- function(needles, haystack) {
...
   if (vec_size(x) != vec_size(by)) {
     abort("`x` and `by` must have same size")
   }
 
-  idx <- unname(split(vec_seq_along(x), vec_duplicate_id(by)))
-  vals <- new_list_of(map(idx, vec_slice, x = x), vec_type(x))
+  ki <- vec_duplicate_split(by)
+
```
lionel-: Can you group these three lines together please? Each line as a paragraph is a bit too much whitespace.

R/dictionary.R
DavisVaughan: ✅

R/dictionary.R
DavisVaughan: ✅

DavisVaughan:
Done!

lionel-:
Thanks!

PR #212, opened 2019-03-04
Fix missing value cases in `vec_equal()`

### LINKED ISSUE #204: vec_equal() doesn't handle NAs correctly
``` r
library(vctrs)

vec_equal(1:3, NA)
#> [1] FALSE FALSE FALSE
vec_equal(NA, 1:3)
#> [1] NA NA NA
```

<sup>Created on 2019-02-26 by the [reprex package](https://reprex.tidyverse.org) (v0.2.1.9000)</sup>

Once that's fixed, also need to check for handling of data frames, because I don't think that's correct either.

lionel-:
Closes #204

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,7 @@
 # vctrs 0.1.0.9000
 
+* `vec_equal()` now handles missing values symmetrically (#204).
+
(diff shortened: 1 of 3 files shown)


hadley:
Thanks!

PR #224, opened 2019-03-05
Improve `vec_equal()` performance

lionel-:
This pulls the dereference of atomic SEXP arrays out of the `vec_equal()` loop to improve performance. A new `vctrs_type` enum supports the genericity over types.

I have tried several ways:

- `no_deref`: Pass a reference object to `equal_scalar()` that contains an array pointer if applicable:

    ```c
    struct vctrs_ref {
      SEXP object;
      enum vctrs_type type;
      void* pointer;
      R_len_t i;
    };
    ```

  This approach improved performance but not enough.

- `no_branch`: Same as above, but pulls the `if (na_equal)` branch out of the loop. This had no impact.

- `template`: switch over vctrs types with the same loop template. This is the approach that is implemented in this PR.

- `template_full`: Like `template` but with no `na_equal` branch. Again, no impact.

faster-loop-desktop

The `dotcall` case directly `.Call()`s into `vctrs_equal()` for better comparison with base R's `==`.

### DIFF
--- a/src/equal.c
+++ b/src/equal.c
@@ -1,27 +1,46 @@
+#include <math.h>
 #include "vctrs.h"
 
 int lgl_equal_scalar(int* x, int* y, bool na_equal) {
-  if (*x == NA_LOGICAL) return na_equal ? *y == NA_LOGICAL : NA_LOGICAL;
-  if (*y == NA_LOGICAL) return na_equal ? *x == NA_LOGICAL : NA_LOGICAL;
-  return *x == *y;
+  int xi = *x;
+  int yj = *y;
+  if (na_equal) {
+    return xi == yj;
+  } else {
+    return (xi == NA_LOGICAL || yj == NA_LOGICAL) ? NA_LOGICAL : xi == yj;
+  }
 }
 int int_equal_scalar(int* x, int* y, bool na_equal) {
-  if (*x == NA_INTEGER) return na_equal ? *y == NA_INTEGER : NA_LOGICAL;
-  if (*y == NA_INTEGER) return na_equal ? *x == NA_INTEGER : NA_LOGICAL;
-  return *x == *y;
+  int xi = *x;
+  int yj = *y;
+  if (na_equal) {
+    return xi == yj;
+  } else {
+    return (xi == NA_INTEGER || yj == NA_INTEGER) ? NA_LOGICAL : xi == yj;
+  }
 }
 int dbl_equal_scalar(double* x, double* y, bool na_equal) {
-  if (R_IsNA(*x)) return na_equal ? R_IsNA(*y) : NA_LOGICAL;
-  if (R_IsNaN(*x)) return na_equal ? R_IsNaN(*y) : NA_LOGICAL;
-  if (R_IsNA(*y)) return na_equal ? R_IsNA(*x) : NA_LOGICAL;
-  if (R_IsNaN(*y)) return na_equal ? R_IsNaN(*x) : NA_LOGICAL;
-  return *x == *y;
+  double xi = *x;
+  double yj = *y;
+  if (na_equal) {
+    if (R_IsNA(xi)) return R_IsNA(yj);
+    if (R_IsNaN(xi)) return R_IsNaN(yj);
+    if (R_IsNA(yj)) return R_IsNA(xi);
+    if (R_IsNaN(yj)) return R_IsNaN(xi);
+  } else {
+    if (isnan(xi) || isnan(yj)) return NA_LOGICAL;
+  }
+  return xi == yj;
 }
 int chr_equal_scalar(SEXP* x, SEXP* y, bool na_equal) {
-  if (*x == NA_STRING) return na_equal ? *y == NA_STRING : NA_LOGICAL;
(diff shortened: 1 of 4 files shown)


src/equal.c
```diff
@@ -1,68 +1,72 @@
 #include "vctrs.h"
 
-int equal_scalar(SEXP x, int i, SEXP y, int j, bool na_equal) {
-  if (TYPEOF(x) != TYPEOF(y))
-    return false;
+int lgl_equal_scalar(int* x, int* y, bool na_equal) {
+  if (*x == NA_LOGICAL) return na_equal ? *y == NA_LOGICAL : NA_LOGICAL;
```
hadley: I wonder if it's worth rewriting to something like:

```R
if (na_equal) {
  return *x == *y 
} else {
  return (*x == NA_LOGICAL || *y == NA_LOGICAL) ? NA_LOGICAL : *x == *y
}
```

I think that's easier to understand, and you're in a good position to see if it has any performance implications.

hadley: I have a vague feeling it might be worth re-checking if there's any difference in performance pulling the `na_equal` comparison after this change..... Now I've read further down, that would clearly introduce a bunch of duplication, so it's probably not worth trying for now.

src/types.c
```diff
@@ -11,3 +11,22 @@ bool is_record(SEXP x) {
 bool is_scalar(SEXP x) {
   return Rf_inherits(x, "vctrs_sclr");
 }
+
+enum vctrs_type vec_typeof(SEXP x) {
```
hadley: Looks good, and this will be straight forward to extend to other base vector types (e.g. factor, Date, ...)

src/vctrs.h
```diff
@@ -17,6 +35,8 @@ bool is_scalar(SEXP x);
 int equal_object(SEXP x, SEXP y, bool na_equal);
 bool equal_names(SEXP x, SEXP y);
 
+// Caller must ensure proper types and sizes. These functions are meant
```
hadley: It's not clear to me what "proper types and sizes" are. Can you be more specific?

src/equal.c
lionel-: I tried this in the `template_full` case (now `template_no_branch` in the new graphic) and it made no difference. I expected to see some difference as well, especially with a high proportion of missing values.

src/vctrs.h
lionel-: This is now:

```c
// Caller must ensure proper SEXPTYPEs and lengths for `x` and `y`.
// These functions are meant to be used in loops. These must check
// bounds.
```

lionel-:
I realised I didn't check performance when the vector contains missing values 🤦.

Interestingly base R was faster with a propertion of missing values close to 50% in the `NA` propagation case. I fixed this gap by recording the element values on the stack: https://github.com/r-lib/vctrs/pull/224/commits/ec10c7be9754ec3e87f660e7a650a9a734c0765a. I'm not sure why the compiler doesn't do this as an optimisation. Assembler output confirms that the pointer are derefed repeatedly when the values are not explicitly saved on the stack. Perhaps because of aliasing? But there is no operation that could change the pointed values...

Here is a more complete comparison (all of these directly `.Call()` into their implementation):

- `template2` uses your suggested control flow
- `template2_no_alias` is the same but with values saved on the stack

faster-loop-desktop

src/equal.c
```diff
@@ -1,68 +1,92 @@
...
-  if (TYPEOF(x) != TYPEOF(y))
-    return false;
-
-  switch(TYPEOF(x)) {
-  case LGLSXP: {
-    int xi = LOGICAL(x)[i], yj = LOGICAL(y)[j];
-    if (xi == NA_LOGICAL) return na_equal ? yj == NA_LOGICAL : NA_LOGICAL;
-    if (yj == NA_LOGICAL) return na_equal ? xi == NA_LOGICAL : NA_LOGICAL;
+int lgl_equal_scalar(int* x, int* y, bool na_equal) {
+  int xi = *x;
```
hadley: Can you add a comment here noting that this is a performance optimisation?

src/vctrs.h
```diff
@@ -17,6 +35,9 @@ bool is_scalar(SEXP x);
 int equal_object(SEXP x, SEXP y, bool na_equal);
 bool equal_names(SEXP x, SEXP y);
 
+// Caller must ensure proper SEXPTYPEs and lengths for `x` and `y`.
```
hadley: I still don't quite get what you mean by this. Do you mean that `x` and `y` must have the same `SEXPTYPE`? And that `i` and `j` must be between 0 and `length(x) - 1`/`length(y) - 1`?

lionel-: Exactly. How about:

```c
// Callers must ensure identical SEXPTYPEs for `x` and `y` and lengths
// greater than `i` and `j` respectively. These functions are meant to
// be used in loops which must check bounds.
```

hadley: I like that! I'd reword slightly:

These functions are meant to be used in loops so it is the callers responsibility to ensure that:
* `x` and `y` have identical SEXTYPES
* `i` is a valid index into `x`, and `j` is a valid index into `y`.
The behaviour is undefined if these conditions are not true.

lionel-: Thanks, done!

PR #226, opened 2019-03-06
Add `vec_is()`

### LINKED ISSUE #79: vec_is and vec_assert_is
```R
vec_is(1:10, integer())
# TRUE

vec_is(1:10, integer(2))
# FALSE

vec_assert_it(1:10, integer(2), name = "x")
# Error: `x` must be an integer vector of length 2, not an integer vector of length 10.
```

Requires thinking about the intersection of shape and type. Might be possible to delay general resolution of this issue by adding an `n` argument, or special-casing the use of length here.

lionel-:
And an `arg` argument for `vec_assert()` to pass argument names more easily.

Closes #79.


R/assert.R
```diff
@@ -1,31 +1,59 @@
...
 
     if (!identical(ptype, x_type)) {
-      msg <- paste0("`", x_name, "` must be <", vec_ptype_abbr(ptype), ">, not <", vec_ptype_abbr(x_type), ">.")
-      abort(
-        msg,
-        .subclass = c("vctrs_error_assert_ptype", "vctrs_error_assert"),
-        required = ptype,
-        actual = x_type
-      )
+      if (assert) {
```
hadley: Maybe refactor out `has_size()` and `has_type()` functions so that `vec_is()` can just be `has_size(x, size) && has_type(x,  ptype)` and `vec_assert()` can focus on throwing errors?

lionel-: This would require `vec_assert()` to dispatch twice for type and twice for size (when it becomes generic), making it more expensive than really needed. This might discourage users from using vector assertions.

hadley: Given that you'll generally check once at the start of a function, I don't think performance should be a dominant concern here.

lionel-: Won't functions created with vctrs be used in loops, e.g. in a grouped mutate? I thought the whole package needed to be performant.

hadley: This just feels like over eager performance optimisation — I think creating a type-unstable function like this (even though it's internal) dominates potential future concerns about performance.

lionel-: Done: https://github.com/r-lib/vctrs/pull/226/commits/05dd79df7274173fe7df4bd925bf240e122e4be4

This approach also suffers from code duplication. I think the original approach was indeed type unstable, but in a structured way. Just like gotos can be used in a structured, predictable, and readable way.

lionel-: hmm I guess with this duplication it doesn't make sense to call `has_ptype()` and dispatch twice, we should just inline it. Or am I missing an obvious way? Are we overthinking this little bit of code? ;)

lionel-: Does this make more sense? https://github.com/r-lib/vctrs/pull/226/commits/810aed54de492aa25b4b770425caf32ce47899d2

R/assert.R
```diff
@@ -46,3 +63,24 @@ vec_assert <- function(x, ptype = NULL, size = NULL) {
 
   invisible(x)
 }
+#' @rdname vec_assert
+#' @export
+vec_is <- function(x, ptype = NULL, size = NULL) {
+  if (!is_null(ptype)) {
+    ptype <- vec_type(ptype)
```
hadley: I was imagining one more step:

```R
vec_is_size <- function(x, size) {
  if (is.null(size)) {
    return(TRUE
  }

  size <- vec_recycle(vec_cast(size, integer()), 1L)
  x_size <- vec_size(x)
  identical(x_size, size)
}
```

Then `vec_is()` is just `vec_is_ptype(x, ptype) && vec_is_size(x, size)` and same checks are used in `vec_assert()` to determine if an error needs to be thrown.

lionel-: This is what I did in https://github.com/r-lib/vctrs/pull/226/commits/05dd79df7274173fe7df4bd925bf240e122e4be4 but this also implies code duplication. I think the code duplication in the last commit makes a bit more sense?

hadley: Oooh, I see now - you need the same calculations for the error messages.

Issue #114, opened 2018-10-09
R 3.1 - double dispatch requires explicit exports in packages that build on vctrs


DavisVaughan:
I noticed that in `probably`, I was failing travis on R 3.1 every time.

I've created a small repo that demonstrates the problem:
https://github.com/DavisVaughan/testtravis

The issue is this:
- Use `@importFrom vctrs vec_cast`
- Create a function that calls `vec_cast(thing, integer())`
- This should call `vec_cast.integer`
- Then more double dispatch...

This works everywhere except on R 3.1 where I get the error:

```
Error in vec_cast.integer(1, integer()) : 
  could not find function "vec_cast.integer"
```

See travis:
https://travis-ci.org/DavisVaughan/testtravis/builds/439111560

The failure line:
https://travis-ci.org/DavisVaughan/testtravis/jobs/439111566#L1868

If I explicitly do `@importFrom vctrs vec_cast.integer`, things work fine again.

My guess is that something must have changed from R 3.1 -> 3.2 where in 3.1 if you are dispatching to something that is another generic, R wouldn't recognize that as being a method of the first generic which it should be able to dispatch to without needing the method to be exported.

krlmlr:
Do we need to maintain support for R 3.1?

hadley:
Our policy is to support 5 versions of R, so when 3.6 comes out, we can drop 3.1 support.

DavisVaughan:
I have painfully also learned that this isn't enough for R 3.1 to always work. If you call `vec_cast.data.frame.data.frame` at some point in a dependent package, then at this line:

https://github.com/r-lib/vctrs/blob/36dd5592c2bca27e426d84c8e043b43323cf9163/R/type-data-frame.R#L161

the `vec_cast()` fails and can't find the first level dispatch function `vec_cast.<type>`. I have determined that this is because `vec_cast()` isn't wrapped in an anonymous function, so the dispatch can't occur correctly. See the [Note section of `lapply()`](https://www.rdocumentation.org/packages/base/versions/3.5.2/topics/lapply). (Strangely this is the second time in the past week I've seen this weird feature of `lapply()` and friends appear.)

If you change it to this everything works again:

https://github.com/DavisVaughan/vctrs/blob/e74fcf17fd9abcb7d529ad507aee3f9d3c519f62/R/type-data-frame.R#L161

(You'll still need the `@importFrom` tags in the downstream package too)

[...]

PR #228, opened 2019-03-07
R 3.1 double dispatch bugs

DavisVaughan:
This PR attempts to guard against all potential R 3.1 `lapply()` + dispatching bugs as mentioned in [this comment](https://github.com/r-lib/vctrs/issues/114#issuecomment-470294051).

### DIFF
--- a/R/cast.R
+++ b/R/cast.R
@@ -107,7 +107,7 @@ vec_cast <- function(x, to) {
 vec_cast_common <- function(..., .to = NULL) {
   args <- list2(...)
   type <- vec_type_common(!!!args, .ptype = .to)
-  map(args, vec_cast, to = type)
+  map(args, function(.x) vec_cast(.x, to = type))
 }
 
 #' @export
--- a/R/list_of.R
+++ b/R/list_of.R
@@ -26,7 +26,7 @@ list_of <- function(..., .ptype = NULL) {
     abort("Could not find common type for elements of `x`.")
   }
 
-  x <- map(args, vec_cast, to = ptype)
+  x <- map(args, function(.x) vec_cast(.x, to = ptype))
   new_list_of(x, ptype)
 }
 
--- a/R/type-data-frame.R
+++ b/R/type-data-frame.R
@@ -134,7 +134,10 @@ df_col_type2 <- function(x, y) {
   # Find types
   names <- set_partition(names(x), names(y))
   if (length(names$both) > 0) {
-    common_types <- map2(x_raw[names$both], y_raw[names$both], vec_type2)
+    common_types <- map2(
+      x_raw[names$both], y_raw[names$both],
+      function(.x, .y) vec_type2(.x, .y)
+    )
   } else {
     common_types <- list()
   }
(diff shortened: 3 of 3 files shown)


hadley:
Just to confirm, after these changes:

* what do packages that use vctrs need for R 3.1? Nothing?
* what do packages that extend vctrs need for R 3.1? Still `@importFrom vctrs vec_cast.integer`?

DavisVaughan:
tl;dr) Even after this PR:

- On 3.1, interactive use of vctrs should always be done with a call to `library(vctrs)`.

- When building packages, for 3.1 you must `@importFrom vctrs vec_cast.<type>` for all types you might cast to.

---

Testing on rstudio.cloud with the current dev version of vctrs and R 3.1:

- Interactive use of calling `vctrs::vec_cast(iris$Species, iris$Species)` without `library(vctrs)` currently fails with the following error. This PR will _not_ fix this problem.

```r
Error in vec_cast.factor(iris$Species, iris$Species) : 
  could not find function "vec_cast.factor"
```

- If building a package and exporting a function like `wrap_cast()` below, then you get the same error as above when calling `wrap_cast(iris$Species, iris$Species)`. To fix this, you _must_ use `@importFrom vctrs vec_cast.factor`. Yes, this means that you must `@importFrom` _all_ of the exported `vec_cast.<type>` functions [as I do here](https://github.com/DavisVaughan/hardhat/blob/master/R/aaa.R#L16). This PR does _not_ fix this issue.

```r
wrap_cast <- function(x, to) {
  vctrs::vec_cast(x, to)
}
```

- In the scenario above with `wrap_cast()`, say you have done the imports below, and call `wrap_cast(iris, iris)`. You will still get the error below about not being able to find `vec_cast.double`. This PR _does_ solve this issue with the anonymous function fix.

```r
#' @importFrom vctrs vec_cast.data.frame
#' @importFrom vctrs vec_cast.double
```

```r
 Error in vec_cast.double(dots[[1L]][[1L]], dots[[2L]][[1L]]) : 
  could not find function "vec_cast.double" 
```

lionel-:
Perhaps we should "fix" this by using internal genericity for base types. vctrs won't be extensible on R 3.1 but at least should work with base types.

hadley:
@lionel- ah, I think that's a good point since we were intending to do that anyway for performance reasons (by internal genericity, I assume you're meaning switchspatch on `vctrs_type`)

DavisVaughan:
So can this be closed because of this bit now?

https://github.com/r-lib/vctrs/blob/7da46b006134c81a0a3061d51846fd433c1abf26/R/zzz.R#L17

lionel-:
> So can this be closed because of this bit now?

This was meant to fix another bug, but maybe it works for yours as well?

DavisVaughan:
Yea it seems to have fixed it. I just reran my hardhat 3.1 travis build that had a Remote on vctrs and it is passing now. I think it's because `vec_cast()` itself no longer does the R level dispatch, and `vec_cast_dispatch()` does. Because `vec_cast_dispatch()` isn't called directly in the `lapply()`, it seems to work.

hadley:
Ok, great. I'm going to close this and the related issue.

PR #230, opened 2019-03-07
Add `vec_is_vector()`

### LINKED ISSUE #190: One argument vec_assert() should check that input is a vector
i.e. `vec_assert(mean)` should fail.

lionel-:
Closes #190.

* Atomic vectors and data frames are hardcoded as vectors.
* Non-vector base types are hardcoded as scalars.
* Bare lists are vectors.
* S3 lists are scalars by default but can be overriden to be vectors. Records are vectors.


R/assert.R
```diff
@@ -84,3 +93,85 @@ vec_is <- function(x, ptype = NULL, size = NULL) {
...
   TRUE
 }
+
+#' Is an object a vector?
+#'
+#' @description
+#'
+#' A vector is a collection of values. In practical terms, a vector is
+#' something that can be a column of a data frame. The opposite of a
+#' vector is a scalar. Such an object cannot be [sliced][vec_slice],
```
hadley: For example, a function is not a vector, and neither is a linear model.

R/assert.R
```diff
@@ -84,3 +93,85 @@ vec_is <- function(x, ptype = NULL, size = NULL) {
 
   TRUE
 }
+
+#' Is an object a vector?
+#'
+#' @description
```
hadley: I don't think you need `@description` here?

R/assert.R
```diff
@@ -84,3 +93,85 @@ vec_is <- function(x, ptype = NULL, size = NULL) {
...
+#'
+#' A vector is a collection of values. In practical terms, a vector is
```
hadley: Like a logical vector, a date, or a factor.

R/assert.R
```diff
@@ -84,3 +93,85 @@ vec_is <- function(x, ptype = NULL, size = NULL) {
...
+#' something that can be a column of a data frame. The opposite of a
```
hadley: I think we need to say something about how this definition relates to the usual definition of a vector in R (i.e. an atomic vector or a list)

R/assert.R
```diff
@@ -84,3 +93,85 @@ vec_is <- function(x, ptype = NULL, size = NULL) {
...
+#' fit <- stats::lm(disp ~ drat, mtcars)
+#' vec_is_vector(fit)
+#'
+#' # Other base types are never vectors:
+#' vec_is_vector(quote(foo(bar)))
+#' vec_is_vector(~foo)
+#' vec_is_vector(function() NULL)
+#' @export
+vec_is_vector <- function(x) {
+  switch(vec_typeof(x),
```
hadley: I think I'd like to see this function written in C. Not because the performance is going to be different, but we're going to have a bunch of our own internal generics in the future, and this seems like a good testing ground for the basic pattern.

lionel-:
Ready for another look @hadley

R/assert.R
```diff
@@ -31,6 +31,11 @@
 vec_assert <- function(x, ptype = NULL, size = NULL, arg = NULL) {
   arg <- arg %||% as_label(substitute(x))
 
+  if (!vec_is_vector(x)) {
+    msg <- paste0("`", arg, "` must be a vector, not a scalar")
```
hadley: Can we replace scalar with something more specific? Maybe, "not a `class(x)[[1]]`"?

hadley: This is an error that users will see, and I think "scalar" will be too confusing because they won't be familiar with our precise definition.

R/assert.R
```diff
@@ -84,3 +93,71 @@ vec_is <- function(x, ptype = NULL, size = NULL) {
 
   TRUE
 }
+
+#' Is an object a vector?
+#'
+#' @description
+#'
+#' A vector is a collection of values like a logical vector, a factor,
```
hadley: What if we called this a "data vector"?

R/assert.R
```diff
@@ -84,3 +93,71 @@ vec_is <- function(x, ptype = NULL, size = NULL) {
...
+#' is a vector, even a model fit object ([base::is.vector()] returns
+#' `TRUE` for these objects). For the most part, functions calls
+#' implement the vector interface and can generally be treated as
+#' vectors. In vctrs, all of these objects are scalars.
+#'
+#' @section Dispatch:
+#'
+#' This function is internally generic:
+#'
+#' * Atomic vectors are always vectors.
```
hadley:
```suggestion
#' * Atomic vectors are always vectors, regardless of their class.
```

R/assert.R
```diff
@@ -84,3 +93,71 @@ vec_is <- function(x, ptype = NULL, size = NULL) {
...
+#' * Bare lists are vectors.
+#' * Data frames are always vectors.
+#' * Non-vector base types are always scalars.
```
hadley: Here you're using "vector" to mean the base R definition, not our definition.

R/assert.R
```diff
@@ -84,3 +93,71 @@ vec_is <- function(x, ptype = NULL, size = NULL) {
...
+#'
+#' Only lists can be overriden by implementing an S3 method for
```
hadley: The behaviour of `vec_is_vector()` can be overridden for S3 objects built on top of lists by providing a method.

src/types.c
```diff
@@ -21,13 +21,16 @@ enum vctrs_type vec_typeof(SEXP x) {
   case STRSXP: return vctrs_type_character;
   case RAWSXP: return vctrs_type_raw;
   case VECSXP:
+    if (!OBJECT(x)) {
+      return vctrs_type_list;
+    }
```
hadley: I feel like this should be an `else if`?

R/assert.R
hadley: And recommend that you build on top of `vctrs_vctr` for new classes.

src/types.c
```diff
@@ -42,12 +45,55 @@ const char* vec_type_as_str(enum vctrs_type type) {
...
+  case vctrs_type_integer:
+  case vctrs_type_double:
+  case vctrs_type_complex:
+  case vctrs_type_character:
+  case vctrs_type_raw:
+  case vctrs_type_list:
+  case vctrs_type_dataframe:
+    return true;
+  case vctrs_type_s3:
+    if (Rf_inherits(x, "vctrs_vctr")) {
```
hadley: Should this be another case in the enum?

R/assert.R
lionel-: I included `friendly_type_of()` which we already use in rlang and purrr. I'll export it from rlang at the next release: https://github.com/r-lib/rlang/issues/458

R/assert.R
lionel-: This is now "Other objects like functions and calls"

R/assert.R
lionel-: I like it. Though what is the opposite of a data vector? An implementation vector?

It seems that in vctrs we'll completely hide the implementation, e.g. there'll be no way to slice a linear model without unclassing it first. From that point of view, it doesn't seem useful to make this distinction.

Another distinction would be data vector versus object vectors like lists and data frames. The latter are recursive because they contain objects rather than data.

src/types.c
lionel-: I thought the early return makes clearer code with less nestedness:

```c
    if (OBJECT(x)) {
      if (is_data_frame(x)) {
        return vctrs_type_dataframe;
      } else {
        return vctrs_type_s3;
      }
    } else {
      return vctrs_type_list;
    }
```

versus

```c
    if (!OBJECT(x)) {
      return vctrs_type_list;
    }
    if (is_data_frame(x)) {
      return vctrs_type_dataframe;
    } else {
      return vctrs_type_s3;
    }
```

The S3 branch will get more complicated in the future so it seems better to unnest it.

src/types.c
lionel-: I think we'll generally need to dispatch for `vctrs_vctr` objects so I kept it as S3 for now.

R/assert.R
hadley: I like object vectors. A data frame is both a data vector and an object vector, depending which way you slice it.

src/types.c
hadley: I just meant like this:

```c
if (!OBJECT(x)) {
  return vctrs_type_list;
} else if (is_data_frame(x)) {
  return vctrs_type_dataframe;
} else {
  return vctrs_type_s3;
}
```

hadley: But it's not a big deal.

src/types.c
hadley: Yeah, we'll need to think that through.

src/types.c
```diff
@@ -42,12 +46,60 @@ const char* vec_type_as_str(enum vctrs_type type) {
...
+      return true;
+    } else {
+      SEXP dispatch_call = PROTECT(Rf_lang2(vec_is_vector_dispatch_fn, x));
+      SEXP out = Rf_eval(dispatch_call, R_GlobalEnv);
```
hadley: Think about this more — we'll eventually need to write something about our "internal" generics, and how in return for performance we sacrifice a small amount of correctness, so that methods locally defined methods don't work. That tradeoff is worth making because the only place we routinely use local methods is in testing, and we have work arounds for that.

src/types.c
lionel-: Ah yes, done.

PR #233, opened 2019-03-11
Rewrite `vec_type2()` in C

lionel-:
* Add a `vctrs_dispatch` enumeration to support double dispatch.
* Move `vec_type2()` dispatch to C side.


hadley:
I'm confused as to why I don't see a bunch of R functions being removed.

Also it would be easier to understand if the general mechanism (e.g. `vec_dispatch_typeof`) was in a different file. It took me a while to discover `vctrs_type2` at the bottom of that file, so I was confused about how everything actually worked.

hadley:
I wonder if it's worth generating the double dispatch file with some R code - otherwise that file is going to get pretty tedious to maintain when we add the main base classes.

lionel-:
> Also it would be easier to understand if the general mechanism (e.g. vec_dispatch_typeof) was in a different file.

It's now in type-dispatch.c.

> I'm confused as to why I don't see a bunch of R functions being removed.

Ah that's because we still dispatch when there's a shape. I have added a comment in the source.

> I wonder if it's worth generating the double dispatch file with some R code - otherwise that file is going to get pretty tedious to maintain when we add the main base classes.

I think it will be ok, the tests check the tables are consistent so we don't need to worry about copy-pasting mistakes. It's only a few lines to add for each type, and adding a type will be a rare occurrence.

lionel-:
Yup:

```r
data <- rep(list(rlang::set_names(letters)), 10)

bench::mark(
  Before = vctrs::vec_rbind(!!!data),
  `No stopifnot` = vctrs2::vec_rbind(!!!data),
  `C vec_type2` = vctrs3::vec_rbind(!!!data),
  dplyr = dplyr::bind_rows(!!!data),
  check = FALSE,
  min_iterations = 100
)
#>   expression       min    mean  median     max `itr/sec` mem_alloc  n_gc
#>   <bch:expr>   <bch:t> <bch:t> <bch:t> <bch:t>     <dbl> <bch:byt> <dbl>
#> 1 Before        21.3ms  23.4ms  23.3ms  25.6ms      42.7  273.36KB    71
#> 2 No stopifnot  17.7ms  18.7ms  18.3ms  24.5ms      53.6    8.19MB    54
#> 3 C vec_type2   13.5ms  14.1ms  13.9ms  15.8ms      71.0    8.19MB    40
#> 4 dplyr          130µs 152.7µs 145.8µs 501.6µs    6547.     7.63KB    22
```

I will eventually post a detailed breakdown of all performance improvements.

PR #234, opened 2019-03-12
Rewrite `vec_cast()` in C

lionel-:
Still dispatches when inputs have a shape or when there is a lossy coercion detected.


R/cast.R
```diff
@@ -96,9 +96,10 @@
 #' vec_cast_common(factor("a"), factor(c("a", "b")))
 #' vec_cast_common(factor("a"), Sys.Date(), .to = list())
 vec_cast <- function(x, to) {
-  if (is.null(x) || is.null(to)) {
-    return(x)
-  }
+  return(.Call(vctrs_cast, x, to))
+  UseMethod("vec_cast", to)
```
hadley: I assume you leave the `UseMethod()` to trick roxygen?

R/type-date-time.R
```diff
@@ -268,7 +268,7 @@ vec_cast.POSIXlt.POSIXlt <- function(x, to) {
   as.POSIXlt(x, tz = tzone(to))
 }
 #' @export
-#' @method vec_cast.POSIXlt POSIXlt
+#' @method vec_cast.POSIXlt POSIXct
```
hadley: Oops!

src/cast.c
```diff
@@ -0,0 +1,271 @@
+#include "vctrs.h"
+
+// Initialised at load time
+static SEXP vec_cast_dispatch_fn = NULL;
+
+
+static SEXP int_as_logical(SEXP x, bool* lossy) {
```
hadley: Why is this static?

src/cast.c
```diff
@@ -0,0 +1,271 @@
...
+  SEXP out = PROTECT(Rf_allocVector(LGLSXP, n));
+  int* out_data = LOGICAL(out);
+
+  for (R_len_t i = 0; i < n; ++i, ++data) {
+    int elt = *data;
+
+    if (elt != 0 && elt != 1) {
+      *lossy = true;
+      UNPROTECT(1);
+      return R_NilValue;
```
hadley: Why do you have an early return here?

src/cast.c
```diff
@@ -0,0 +1,271 @@
...
+static SEXP lgl_as_double(SEXP x, bool* lossy) {
+  int* data = LOGICAL(x);
+  R_len_t n = Rf_length(x);
+
+  SEXP out = PROTECT(Rf_allocVector(REALSXP, n));
+  double* out_data = REAL(out);
+
+  for (R_len_t i = 0; i < n; ++i, ++data, ++out_data) {
+    int elt = *data;
+    *out_data = elt == NA_LOGICAL ? NA_REAL : elt;
```
hadley: I think this is a bit easier to read:

```suggestion
    *out_data = (elt == NA_LOGICAL) ? NA_REAL : elt;
```

src/cast.c
```diff
@@ -0,0 +1,271 @@
...
+    return x;
+  }
+  if (has_dim(x) || has_dim(to)) {
+    goto dispatch;
+  }
+
+  bool lossy = false;
+  SEXP out = R_NilValue;
+
+  switch (vec_typeof(to)) {
```
hadley: Why can't you use the same double dispatch enum as in `vec_type2`?

src/cast.c
hadley: Oooooh because it falls back to the regular method for lossy casts. I wonder if there's some other way to implement this that doesn't require so many contortions of the C code.

src/cast.c
```diff
@@ -0,0 +1,271 @@
...
+    case vctrs_type_logical:
+      out = lgl_as_integer(x, &lossy);
+      break;
+    case vctrs_type_integer:
+      return x;
+    case vctrs_type_double:
+      out = dbl_as_integer(x, &lossy);
+      break;
+    case vctrs_type_character:
+      // TODO: Implement with `R_strtod()` from R_ext/utils.h
```
hadley: Might also be worth looking at what readr/vroom does here.

R/cast.R
lionel-: Yes, or for other static analysis purposes. That's going to be a common trick in the package so I didn't add a comment as I did for `vec_type2()`.

src/cast.c
lionel-: To indicate these are support functions for other functions in this file. We can remove the static if we add these functions in vctrs.h.

src/cast.c
lionel-: We could wrap the dispatch function in a function and call it from the coercers. It seems cleaner to me to do the early return to make it clear that the output is produced by another mechanism though (R implementation).

This is all temporary, until we're able to signal the typed conditions from C. It requires a bit of effort because of the details on lossy casts included in the condition. I just realised the growable stuff might be helpful here.

src/cast.c
lionel-: Because the dispatch is not symmetric for casting.

src/cast.c
hadley: Ah got it.

src/cast.c
hadley: Oooh got it.

src/cast.c
```diff
@@ -146,7 +146,7 @@ static SEXP lgl_as_double(SEXP x, bool* lossy) {
 
   for (R_len_t i = 0; i < n; ++i, ++data, ++out_data) {
     int elt = *data;
-    *out_data = elt == NA_LOGICAL ? NA_REAL : elt;
+    *out_data = (elt == NA_LOGICAL) ? NA_REAL : elt;
```
hadley: BTW this needs to be applied to the other functions too

src/cast.c
hadley: Ok makes sense.

PR #235, opened 2019-03-15
Implement `vec_slice()` in C

lionel-:
New `vec_as_index()` function that transforms an index vector (missing, character, logical or integer) to an integer vector of positions. The semantics are identical to those of base R subsetting:

* Negative indices are inverted.
 * Zeros are filtered out.
* Logical vectors are recycled, including `NA`.
* Bounds are checked. OOB is an error.

`vec_as_index()` is currently unexported but should eventually make it easier to implement subsetting for custom classes because only one type of indices needs to be dealt with. Internally, it will avoid the recomputation of the index for each component when a composite structure such as a data frame or a record type is sliced.

The C implementations of slicing have these features:

* Missing values are propagated on slice and ignored on slice-assign.
* Unlike base R, attributes of unclassed structures are reproduced.


src/vctrs.h
```diff
@@ -165,3 +171,15 @@ SEXP growable_values(growable* g);
 // Shape --------------------------------------------------------
 
 bool has_dim(SEXP x);
+
+
+// Compatibility ------------------------------------------------
+
+#if (R_VERSION < R_Version(3, 5, 0))
+# define LOGICAL_RO(x) ((const int*) LOGICAL(x))
```
hadley: What does the RO stand for? read only?

R/size.R
```diff
@@ -99,43 +99,49 @@ vec_size2 <- function(nx, ny) {
...
 vec_slice <- function(x, i) {
-  i <- get_slice_index(i, x)
+  .Call(vctrs_slice, x, maybe_missing(i))
+}
 
-  if (is.null(x)) {
-    NULL
-  } else if (is.data.frame(x)) {
+vec_slice_dispatch <- function(x, i) {
+  if (is.data.frame(x)) {
```
hadley: Will you do data frames in the next PR? It seems like you already have most of the pieces in place.

R/size.R
```diff
@@ -99,43 +99,49 @@ vec_size2 <- function(nx, ny) {
 #'
 #' vec_slice(mtcars, 1:3)
 vec_slice <- function(x, i) {
-  i <- get_slice_index(i, x)
+  .Call(vctrs_slice, x, maybe_missing(i))
```
hadley: Do you want to make add `UseMethod("vec_slice")` now to make it more consistent with the other internal generics?

src/size.c
```diff
@@ -0,0 +1,346 @@
...
+  SLICE(STRSXP, SEXP, STRING_PTR, STRING_PTR_RO, NA_STRING);
+}
+static SEXP raw_slice(SEXP x, SEXP index) {
+  SLICE(RAWSXP, Rbyte, RAW, RAW_RO, 0);
+}
+
+#undef SLICE
+
+
+#define SLICE_BARRIER(RTYPE, GET, SET, NA_VALUE)                \
```
hadley: What does barrier stand for?

hadley: And is it really worth generalising for a single case?

src/size.c
```diff
@@ -0,0 +1,346 @@
...
+}
+
+SEXP vec_as_index(SEXP i, SEXP x) {
+  switch (TYPEOF(i)) {
+  case INTSXP: return int_as_index(i, x);
+  case LGLSXP: return lgl_as_index(i, x);
+  case STRSXP: return chr_as_index(i, x);
+  case SYMSXP: if (i == R_MissingArg) return i;
+
+  // Do we really want to forbid numeric indices here (> 2^31)?
```
hadley: Not really

src/size.c
```diff
@@ -0,0 +1,346 @@
+#include "vctrs.h"
```
hadley: This should probably be `slice.c` and we should have a new `.R` equivalent

src/size.c
```diff
@@ -0,0 +1,346 @@
...
+  SEXP out = vec_slice(x, index);
+
+  UNPROTECT(1);
+  return out;
+}
+
+static void slice_copy_attributes(SEXP to, SEXP from, SEXP index) {
+  SEXP attrib = PROTECT(Rf_shallow_duplicate(ATTRIB(from)));
+
+  for (SEXP node = attrib; node != R_NilValue; node = CDR(node)) {
```
hadley: Needs a comment: subsetting names.

src/size.c
```diff
@@ -0,0 +1,346 @@
...
+    break;
+  }
+  case vctrs_type_list: {
+    out = list_slice(x, index);
+    break;
+  }
+
+  default:
+  dispatch: {
+    SEXP dispatch_call = PROTECT(Rf_lang3(vec_slice_dispatch_fn, x, index));
```
hadley: Should we wrap this internal generic pattern up with a helper C function?

src/size.c
```diff
@@ -0,0 +1,346 @@
...
+    case NILSXP:
+      break;
+    case STRSXP:
+      nms = PROTECT(chr_slice(nms, index));
+      SETCAR(node, nms);
+      UNPROTECT(1);
+      break;
+    default:
+      Rf_error("Internal error: Expected character names in `vec_slice()`.");
+    }
```
lionel-: Could break out of the loop here.

src/vctrs.h
lionel-: Yup

R/size.R
lionel-: Yes, data frames and maybe `vec_restore()`

R/size.R
lionel-: I think we would need a `dispatch` argument so the default method can call back `vec_slice()`with `dispatch = FALSE`. This will be also useful for implementing `[` with `vec_slice()` semantics.

In general it seems a good principle to allow the caller to turn off dispatch. We seem to rely quite a lot on `.subset()` and `.subset2()` to manipulate the underlying R type, this would be useful in vctrs as well. Best implemented in a future PR I think.

src/size.c
lionel-: It stands for memory barrier, that's for type for which we don't have access to the C array.

This will be useful for altrep vectors so we can index them without materialising them. This would also allow us to reimplement character slicing with the memory barrier accessors in case they deprecate `STRING_PTR`. In the mean time, using `STRING_PTR` is better for performance (and is safe because we only assign CHARSXPs to a fresh vector that belongs to the youngest GC generation).

src/size.c
lionel-: This was a comment from @krlmlr in the R code. I think we should cast the doubles to ints for sure, but if we want to support long vectors this will make the code more complicated. We'll need to macro-template in many places. The hashtab implementations will need to be duplicated because the long vector indices take twice the memory.

R/size.R
hadley: Hmmmm, I think I would rather have a separate function for that. Otherwise the methods will also need to have `dispatch = FALSE` argument, which seems weird/confusing/wrong.

src/size.c
hadley: Ok, that makes sense. We'll tackle long vectors at some future point.

R/size.R
lionel-: Makes sense. The recursion is a bit tricky so best left for another PR.

PR #237, opened 2019-03-19
Make `vec_slice()` generic

lionel-:
* `vec_slice()` is now a generic. The default method dispatches to `[`.

* A new dispatch implementation allows (a) masking its arguments and (b) forwarding dots if they exist in the generic environment. Dispatch now occurs in the generic environment instead of the global env to make it easier to mask and to forward dots.

* New `vec_slice_bare()` function that is not a generic. This can be used to call `vec_slice()` from `[` methods without inflooping dispatch. `vec_typeof()` gains a `vec_typeof_impl()` variant as well to support this. It has a `dispatch` argument that disables `vctrs_type_s3`.

* Call `vec_slice_bare()` and `vec_slice()` from the `[` methods for `vctr` and `rcrd` types. Because `vec_slice()` preserves attributes, the constructor for `vctr`, called from `vec_restore()`, now allows (but ignores) attributes.


lionel-:
Oh I forgot, since `vec_slice()` preserves attributes, `vec_assert()` now treats bare prototypes as partial types. Only the SEXPTYPE is checked. This way `vec_assert(x, double())` works as expected in the vector restoration methods etc.

R/slice.R
```diff
@@ -18,14 +22,18 @@
 #' x
 #'
 #' vec_slice(mtcars, 1:3)
-vec_slice <- function(x, i) {
-  .Call(vctrs_slice, x, maybe_missing(i))
+vec_slice <- function(x, i, ...) {
+  return(.Call(vctrs_slice, x, maybe_missing(i), environment()))
```
hadley: `ellipsis::check_dots_used()`?

src/type2.c
```diff
@@ -37,11 +38,15 @@ SEXP vctrs_type2(SEXP x, SEXP y) {
 
   default:
   dispatch:
-    return vctrs_dispatch3(vec_type2_dispatch_fn, x, y);
+    return vctrs_dispatch2(syms_vec_type2_dispatch, fns_vec_type2_dispatch,
+                           syms_x, x,
+                           syms_y, y,
+                           R_GlobalEnv);
```
hadley: Why is this environment different to `vec_slice()` dispatch?

src/utils.c
```diff
@@ -7,11 +8,50 @@ bool is_bool(SEXP x) {
...
+ *   assigned.  The assignment occurs in `env` and the dispatch call
+ *   refers to these symbols.
+ * @param env The environment in which to dispatch. Should be the
+ *   global environment or inherit from it so methods defined there
+ *   are picked up. If the global environment, a child is created so
+ *   the call components can be masked.
+ *
+ *   If `env` contains dots, the dispatch call forwards dots.
+ */
+SEXP vctrs_dispatch2(SEXP fn_sym, SEXP fn,
```
hadley: This is to generate a better traceback?

tests/testthat/test-slice.R
```diff
@@ -213,6 +213,35 @@ test_that("vec_as_index() checks type", {
   expect_error(vec_as_index(quote(foo), 1), "must be an integer, character, or logical vector, not a symbol")
 })
 
+test_that("can `vec_slice()` S3 objects without infloop dispatch", {
```
hadley:
```suggestion
test_that("can `vec_slice()` S3 objects without dispatch infloop", {
```

src/type2.c
lionel-: Because we don't forward dots so we don't need to dispatch in the exec env of `vec_type2()`. This dispatches in a child of the global env without needing to add parameters and pass an environment across the C functions.

Maybe this should be `R_NilValue` so it doesn't suggest dispatch occurs exactly in the global env rather than a mask inheriting from it.

src/utils.c
lionel-: And to forward dots if the environment has dots.

R/slice.R
lionel-: Should the UI be `on.exit(check_dots_unused())`? I think hiding the `on.exit()` makes it harder to read. Also harder to experiment with. Might also lead to bugs by writing things like:

```r
check_dots_unused()

on.exit(cleanup())
```

src/type2.c
lionel-: Dispatch in the global environment is now signalled by `R_NilValue`, otherwise dispatch occurs in the environment supplied.

hadley: Isn't it better to be consistent across all our internal generics?

R/assert.R
```diff
@@ -94,6 +97,15 @@ vec_is <- function(x, ptype = NULL, size = NULL) {
   TRUE
 }
 
+# Bare prototypes act as partial types. Only the SEXPTYPE is checked.
+is_same_type <- function(x, y) {
```
hadley: I think this probably shouldn't have symmetric argument names, but instead be `x, ptype`, because this function will eventually check if `ptype` is a partial type, and do something special in that case.

R/slice.R
hadley: Actually, are you sure we need `...` here?

lionel-: This way we can use `vec_slice()` in the `[` methods of composite structures like data frames and records, and ensure the `...` are properly passed to `[` from `vec_slice.default()` on recursion.

src/type2.c
lionel-: I don't think it really matters in this case. Ideally we wouldn't dispatch in the vctrs namespace at all, we'd create a child of the global env and forward dots there. The PR implementation makes things a bit simpler. Would you like to forward dots in a child of global instead?

hadley: It just worries me that we have 3 (?) internal generics and two dispatch methods.

R/slice.R
hadley: Something seems faulty about that reasoning but I can't figure out what. Maybe it's that calling `[` is just a mostly a stop gap until we have our own implementations? Maybe that I see `vec_slice()` as considerably more restrictive than `[` so we shouldn't accept `...` without deliberation.

lionel-: Calling `[` from `vec_slice.default()` seems needed to be compatible with foreign types. And calling `vec_slice()` from `[.vctrs_xxx` is needed to get strict semantics with the subset syntax for our own types. I agree the proposed dispatch path is a bit convoluted but it seems like a reasonable compromise.

Would you prefer to call `check_dots_empty()` from `[.vctrs_xxx`?

hadley: Yeah, but why would they need extra arguments? `vec_slice()` is more restrictive.

lionel-: For the recursive case so we can dispatch to `[` from `vec_slice.default` and forward the dots passed by the user to `[.vctrs_xxx`.

lionel-: Here is the dispatch path for a record that contains a vector of class `foo`:

1. `[.vctrs_rcrd()`
2. `vec_slice.vctrs_rcrd()`
3. For each element: `vec_slice()`
4. Because `foo` doesn't implement vec_slice: `vec_slice.default()`
5. `[.foo`

The `...` are forwarded through all these levels.

src/type2.c
lionel-: I will change it to always dispatch in a child of the global environment.

R/slice.R
hadley: Yes, but I still don't understand why we need to forward `...` — `[.vctrs_rcrd` implements the `vec_slice()` interface not the more general `[` interface.

lionel-: Because we use `[` has a fallback for compatibility with foreign classes. And we have recursive types like records and data frames which may contain any class of vector.

The alternative is to check for empty dots in `[.vctrs_xxx` (this would be a change from current master which forwards dots).

lionel-:
@hadley `vec_slice()` no longer takes `...` and dispatch always occurs in a child of the global environment.

lionel-:
And `vec_slice()` no longer restores attributes, this will be handled in the default `vec_restore()` method.

R/utils.R
```diff
@@ -148,3 +148,16 @@ as_friendly_type <- function(type) {
 paste_classes <- function(x) {
   paste(class(x), collapse = "/")
 }
+
+paste_line <- function (...) {
+  paste(chr(...), collapse = "\n")
+}
+
+check_dots_empty_s3_consistency <- function(...) {
```
hadley: This will move to ellipsis?

lionel-: For ellipsis the message should probably be:

```r
      "`...` is not empty.",
      "These dots only exist to allow future extensions and should be empty."
```

Here the message is a bit different because it's not about extensions. Probably doesn't matter much though.

PR #238, opened 2019-03-22
Refactor: replace goto by helper functions

krlmlr:
- Separate repetitive `switch` code from more complex logic.
- Hide details of double-dispatch function invocation

This might fix a possible protection error in `vec_slice_impl()` too.


lionel-:
> This might fix a possible protection error in vec_slice_impl() too.

Where is that?

lionel-:
Merging so I can send another PR rebased on this.

PR #239, opened 2019-03-22
Move `vec_restore()` to C side

lionel-:
`vec_restore()` will be called from `vec_slice()` in a further PR.


src/cast.c
```diff
@@ -261,7 +263,111 @@ SEXP vec_cast(SEXP x, SEXP to) {
...
+  // SET_ATTRIB() does not set object bit when attributes include class
+  if (OBJECT(to)) {
+    SET_OBJECT(x, true);
+  }
+
+  UNPROTECT(n_protect);
+  return x;
+}
+
+static R_len_t df_raw_length(SEXP x) {
```
hadley: Isn't there substantial code duplication with `df_obs`?

src/cast.c
```diff
@@ -261,7 +263,111 @@ SEXP vec_cast(SEXP x, SEXP to) {
...
+    Rf_errorcall(R_NilValue, "Internal error: Attempt to restore data frame from a %s.",
+                 Rf_type2char(TYPEOF(x)));
+  }
+
+  int n_protect = 0;
+
+  // Compute size before changing attributes
+  R_len_t size = df_raw_length(x);
+
+  if (MAYBE_REFERENCED(x)) {
```
hadley: This is an optimisation to avoid duplicating unnamed inputs?

lionel-: When `vctrs_restore()` is called from R we need to duplicate to avoid changing attributes by reference. But we will also call this method from C with a fresh list that we own, and then it doesn't need to be duplicated.

src/cast.c
lionel-: oops

PR #242, opened 2019-03-22
Call `vec_restore()` from `vec_slice()`

lionel-:
* `[` methods can now just call `vec_slice()` and get a fully formed object.

* Restoration after slicing requires index information (at least the vector size) because zero-cols data frames do not carry size information in the raw data (an empty list).

  For this reason, `vec_restore()` gains an `i` argument that is non-null after slicing and contains the standardised index vector. This argument can be safely ignored most of the time.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,8 @@
 # vctrs 0.1.0.9000
 
+* `vec_slice()` now calls `vec_restore()` automatically. Attributes of
+  unclassed structures are preserved by default.
+
 * `vec_slice()` is now a generic. Its default method calls `[`.
 
 * `vec_size()` now works with positive short row names. This fixes
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,7 +1,7 @@
 # vctrs 0.1.0.9000
 
-* `vec_slice()` now calls `vec_restore()` automatically. Attributes of
-  unclassed structures are preserved by default.
+* `vec_slice()` now calls `vec_restore()` automatically. Unlike the
+  default `[` method from base R, attributes are preserved by default.
 
 * `vec_slice()` is now a generic. Its default method calls `[`.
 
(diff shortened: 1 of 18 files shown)


lionel-:
Last commit finishes implementing slicing of data frames in C as we now have all the pieces.

NEWS.md
```diff
@@ -1,5 +1,8 @@
 # vctrs 0.1.0.9000
 
+* `vec_slice()` now calls `vec_restore()` automatically. Attributes of
+  unclassed structures are preserved by default.
```
hadley: (unlike base R)

R/slice.R
```diff
@@ -4,8 +4,11 @@
 #' for all vector types, regardless of dimensionality. It is an analog to `[`
 #' that matches [vec_size()] instead of `length()`.
```
hadley: Maybe worth adding a new section that lists all differences with `[`? i.e. only slices along one dimension, preserves attributes of unclassed types, ...

src/slice.c
```diff
@@ -76,51 +77,82 @@ static SEXP list_slice(SEXP x, SEXP index) {
...
-  default:
-    return R_NilValue;
-  }
-}
+    out = list_slice(x, index);
+    break;
 
-static SEXP vec_slice_impl(SEXP x, SEXP index, bool dispatch) {
-  SEXP out = R_NilValue;
+  case vctrs_type_dataframe:
```
hadley: Slice records too?

hadley:
@yutannihilation you were right - `vec_restore()` does need an additional parameter.

src/slice.c
lionel-: They are not in the C types yet, I'll add them later on. I also like keeping a type fully implemented in R during the refactorings.

yutannihilation:
> `vec_restore()` does need an additional parameter.

Oh, I missed this. Thanks for the notice!

PR #247, opened 2019-03-24
warn_lossy_cast() is a no-op if the location argument is empty

krlmlr:
This makes it easier for the callers, especially in other packages. Need to think about exporting `report_lossy_cast()` or an equivalent.

### DIFF
--- a/R/conditions.R
+++ b/R/conditions.R
@@ -8,7 +8,8 @@
 #' @param details Any additional human readable details
 #' @param subclass Use if you want to further customise the class
 #' @param locations For `warn_lossy_cast()`, an optional vector giving the
-#'   locations where `x` lost information.
+#'   locations where `x` lost information.  If the vector is given and empty,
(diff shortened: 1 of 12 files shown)


R/conditions.R
```diff
@@ -8,7 +8,8 @@
 #' @param details Any additional human readable details
 #' @param subclass Use if you want to further customise the class
 #' @param locations For `warn_lossy_cast()`, an optional vector giving the
-#'   locations where `x` lost information.
+#'   locations where `x` lost information.  If the vector is given and empty,
+#'   no warning is issued.
```
hadley: Probably needs a `@value` section where it states it's called for its side-effect, and it only warns if there is some loss

lionel-:
We now have `maybe_lossy_cast()` which throws errors with a restart. It takes a `lossy` argument from which locations are picked up. If `any(lossy)`, a lossy cast error is thrown.

PR #248, opened 2019-03-25
Fix array restoration

### LINKED ISSUE #153: vec_restore.default() alters size and type
According to the docs:

> A `vec_restore()` method can assume that `x` has the correct type (although the length may be different) but all attributes have been lost and need to be restored.

This would lead me to believe that `x` can have a different size than `to`, which is the case when using `vec_restore()` with 1D vectors, but with matrices/arrays this isn't the case because all attributes of `to` are copied to `x`, including the `dim` (which embeds the size).

``` r
library(vctrs)

x <- matrix(1:5)

# dummy function to demonstrate losing attributes
# - type is the same (but attributes are needed)
# - size is now different
c_level_manip <- function(x) {
  vec_data(x)[1:2]
}

# I think this should work
vec_restore(c_level_manip(x), x)
#> Error in attributes(x) <- attributes(to): dims [product 5] do not match the length of object [2]

# special case the `dim` attribute
# pull size from x
# pull shape from to
vec_restore.default <- function(x, to) {
  restore_attr <- attributes(to)
  
  if ("dim" %in% names(restore_attr)) {
    restore_attr$dim <- c(vec_size(x), vctrs:::shape(to))
  }
  
  attributes(x) <- restore_attr
  x
}

vec_restore(vec_data(x)[1:2], x)
#>      [,1]
#> [1,]    1
#> [2,]    2
```

<sup>Created on 2019-01-07 by the [reprex package](https://reprex.tidyverse.org) (v0.2.1.9000)</sup>

Size is implicitly ignored with vectors:

``` r
library(vctrs)

ex <- factor(c("a", "a", "b"))

vec_restore(vec_data(ex[1]), ex)
#> [1] a
#> Levels: a b
```

<sup>Created on 2019-01-07 by the [reprex package](https://reprex.tidyverse.org) (v0.2.1.9000)</sup>

Update) This also would affect dimnames and general names. This is more difficult as the order or the rows would matter here when "restoring" names, but generally I think it isn't vctrs's job to restore names/dimnames so maybe they could just be removed.

``` r
library(vctrs)
x <- 1:5
names(x) <- letters[1:5]
vec_restore(vec_data(x[1:2]), x)
#> Error in attributes(x) <- attributes(to): 'names' attribute [5] must be the same length as the vector [2]
```

<sup>Created on 2019-01-07 by the [reprex package](https://reprex.tidyverse.org) (v0.2.1.9000)</sup>

lionel-:
* `vec_restore()` is called after `vec_slice()` fallbacks to `[`.
* The default restore method preserves dimnames.
* Review assumptions that can be made by restoration methods in `?vec_restore`

Closes #153.

Does this look good @DavisVaughan?

### DIFF
--- a/R/cast.R
+++ b/R/cast.R
@@ -46,14 +46,25 @@
 #' [vec_unspecified_cast()]).
 #'
 #' See `vignette("s3-vector")` for full details.
+#'
+#'
 #' @section Restoring attributes:
 #'
-#' A restore is a specialised type of cast, primarily used in conjunction
-#' with `NextMethod()` or a C-level function that works on the underlying
-#' data structure. A `vec_restore()` method can assume that `x` has the
-#' correct type (although the length may be different) but all attributes
-#' have been lost and need to be restored. In other words,
-#' `vec_restore(vec_data(x), x)` should yield `x`.
+#' A restore is a specialised type of cast, primarily used in
+#' conjunction with `NextMethod()` or a C-level function that works on
+#' the underlying data structure. A `vec_restore()` method can make
+#' the following assumptions about `x`:
+#'
+#' * It has the correct type.
+#' * It has the correct names.
+#' * It has the correct `dim` and `dimnames` attributes.
+#' * It is unclassed. This way you can call vctrs generics with `x`
+#'   without triggering an infinite loop of restoration.
+#'
+#' The length may be different (for example after [vec_slice()] has
+#' been called), and all other attributes may have been lost. The
+#' method should restore all attributes so that after restoration,
+#' `vec_restore(vec_data(x), x)` yields `x`.
 #'
 #' To understand the difference between `vec_cast()` and `vec_restore()`
 #' think about factors: it doesn't make sense to cast an integer to a factor,
--- a/R/slice.R
+++ b/R/slice.R
@@ -43,16 +43,18 @@ vec_slice_fallback <- function(x, i) {
   d <- vec_dims(x)
   if (d == 1) {
     if (is.object(x)) {
-      x[i]
+      out <- x[i]
     } else {
-      x[i, drop = FALSE]
+      out <- x[i, drop = FALSE]
     }
   } else if (d == 2) {
-    x[i, , drop = FALSE]
+    out <- x[i, , drop = FALSE]
   } else {
     miss_args <- rep(list(missing_arg()), d - 1)
-    eval_bare(expr(x[i, !!!miss_args, drop = FALSE]))
+    out <- eval_bare(expr(x[i, !!!miss_args, drop = FALSE]))
   }
+
+  vec_restore(out, x)
 }
 
 vec_slice_bare <- function(x, i) {
--- a/R/slice.R
+++ b/R/slice.R
@@ -37,21 +37,20 @@
 vec_slice <- function(x, i) {
   .Call(vctrs_slice, x, maybe_missing(i), TRUE)
 }
-vec_slice_fallback <- function(x, i) {
-  vec_assert(x)
 
-  d <- vec_dims(x)
+# Only called when `x` has dimensions
+vec_slice_fallback <- function(x, i) {
+  out <- unclass(vec_proxy(x))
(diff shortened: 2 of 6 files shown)


R/cast.R
```diff
@@ -46,14 +46,25 @@
...
-#' have been lost and need to be restored. In other words,
-#' `vec_restore(vec_data(x), x)` should yield `x`.
+#' A restore is a specialised type of cast, primarily used in
+#' conjunction with `NextMethod()` or a C-level function that works on
+#' the underlying data structure. A `vec_restore()` method can make
+#' the following assumptions about `x`:
+#'
+#' * It has the correct type.
+#' * It has the correct names.
+#' * It has the correct `dim` and `dimnames` attributes.
```
hadley: How can this be true if " The length may be different" ?

R/slice.R
```diff
@@ -48,16 +48,18 @@ vec_slice.default <- function(x, i) {
...
   } else if (d == 2) {
-    x[i, , drop = FALSE]
+    out <- x[i, , drop = FALSE]
   } else {
     miss_args <- rep(list(missing_arg()), d - 1)
-    eval_bare(expr(x[i, !!!miss_args, drop = FALSE]))
+    out <- eval_bare(expr(x[i, !!!miss_args, drop = FALSE]))
   }
+
+  vec_restore(out, x)
```
hadley: It seems weird to restore if we've called `[` on something that's an S3 object?

tests/testthat/test-cast.R
```diff
@@ -195,3 +195,15 @@ test_that("vec_restore() passes `i` argument to methods", {
...
+test_that("dimensions are preserved by default restore method", {
+  x <- foobar(1:4)
+  dim(x) <- c(2, 2)
+  dimnames(x) <- list(a = c("foo", "bar"), b = c("quux", "hunoz"))
+
+  exp <- foobar(c(1L, 3L))
+  dim(exp) <- c(1, 2)
+  dimnames(exp) <- list(a = "foo", b = c("quux", "hunoz"))
+
+  expect_identical(vec_slice(x, 1), exp)
```
hadley: Shouldn't this call `vec_restore()` since we're testing it?

R/cast.R
lionel-: Because we restore dim and dimnames so that they reflect the new size.

tests/testthat/test-cast.R
lionel-: `vec_slice()` calls `vec_restore()`.

R/slice.R
lionel-: Good point, I'll change to only call it on non-OO objects with dims.

lionel-: Actually, calling `vec_restore()` here is congruent with how `vec_slice()` automatically restores the class and attributes. The `[.foo` methods should call `vec_slice()` but not `vec_restore()`. Then the call path for a foo object with dims is `[` -> `vec_slice()` -> `unclass()` -> `[` -> `vec_restore()`.

I have simplified `vec_slice_fallback()` (which is only called with shaped objects and should be removed in the future when we support those natively) and added a missing `unclass()`.

Still need to figure out what to do when additional arguments are passed to `[` methods. Currently higher dimensions are ignored.

lionel-: I guess the current situation is confusing because of contradictory use cases:

- Using `vec_slice()` from our `[` methods for consistency of behaviour.
- Calling `[` from `vec_slice()` to handle shaped objects because we don't have native slicing for these in vctrs.
- Calling `[` from `vec_slice()` to handle S3 objects as a fallback.

With `[.vctrs_vctr` we'd like to restore after slicing shaped objects with `[`. With foreign classes we don't want to restore.

This will be less confusing once we slice shape natively. I'll try to find an interim solution.

hadley: You can leave it for now and tackle in a future PR if you want.

DavisVaughan:
@lionel- this PR allowed me to remove all of my hacks in rray! I'm happy with it.

PR #249, opened 2019-03-25
Fix slice dispatch

lionel-:
This untangles dispatch to `[` and dispatch from `[`.

* When `vec_slice()` is called with an S3 object, we fall back to `[` instead of slicing the `vec_proxy()` so we're immediately compatible with foreign types.

* Our own types implement `[` methods which call `vec_slice_native()`. The latter is a variant that turns off the `[` fallback. In this case, the `vec_proxy()` is sliced.

We'll probably want to special-case the vctrs classes in the S3 branch of `vec_slice()` to avoid the `vec_slice()` -> `[` -> `vec_slice_native()` -> `vec_proxy()` dance. Perhaps the C extension mechanism would be a good way to do so. We'll also need to export `vec_slice_native()` so third party classes can call `vec_slice()` from their `[` methods. This helper will probably need to handle dimensionality.


lionel-:
@hadley false alarm, this isn't right yet

lionel-:
@hadley Now fixed, and `vec_restore()` is called on data frames after slicin. This was necessary to transform it back to a record, and will also allow data frame subclasses to restore themselves.

R/cast.R
```diff
@@ -146,6 +146,10 @@ vec_restore_dispatch <- function(x, to, ..., i = NULL) {
 vec_restore.default <- function(x, to, ...) {
```
hadley: Doesn't this (and the method below) also need the `i` argument?

src/slice.c
```diff
@@ -77,16 +78,18 @@ static SEXP list_slice(SEXP x, SEXP index) {
...
+static SEXP df_slice(SEXP x, SEXP index) {
   R_len_t n = Rf_length(x);
   SEXP out = PROTECT(Rf_allocVector(VECSXP, n));
 
   SEXP nms = Rf_getAttrib(x, R_NamesSymbol);
   Rf_setAttrib(out, R_NamesSymbol, nms);
 
   for (R_len_t i = 0; i < n; ++i) {
-    SEXP sliced = vec_slice_impl(VECTOR_ELT(x, i), index, true);
+    SEXP elt = PROTECT(VECTOR_ELT(x, i));
```
hadley: Why does this need protecting?

src/slice.c
```diff
@@ -100,98 +103,136 @@ static SEXP vec_slice_fallback(SEXP x, SEXP index) {
...
+  SEXP out;
+  switch (type) {
+  case vctrs_type_logical:   out = lgl_slice(x, index); break;
+  case vctrs_type_integer:   out = int_slice(x, index); break;
+  case vctrs_type_double:    out = dbl_slice(x, index); break;
+  case vctrs_type_complex:   out = cpl_slice(x, index); break;
+  case vctrs_type_character: out = chr_slice(x, index); break;
+  case vctrs_type_raw:       out = raw_slice(x, index); break;
+  case vctrs_type_list:      out = list_slice(x, index); break;
+  default: Rf_error("Internal error: Non-base vector type `%s` in `vec_slice_base()`",
```
hadley:
```suggestion
  default: Rf_error("Internal error: Non-vector base type `%s` in `vec_slice_base()`",
```

src/slice.c
```diff
@@ -100,98 +103,136 @@ static SEXP vec_slice_fallback(SEXP x, SEXP index) {
...
+SEXP vctrs_slice(SEXP x, SEXP index, SEXP native) {
   if (x == R_NilValue || index == R_MissingArg) {
     return x;
   }
 
   index = PROTECT(vec_as_index(index, x));
-  SEXP out = vec_slice_impl(x, index, *LOGICAL(dispatch));
+
+  SEXP out;
+  if (*LOGICAL(native)) {
```
hadley: I think `LOGICAL(native)[0]` is a idiom that's slightly easier to understand (and more common)

src/slice.c
```diff
@@ -100,98 +103,136 @@ static SEXP vec_slice_fallback(SEXP x, SEXP index) {
...
+    return;
+  }
+
+  nms = PROTECT(chr_slice(nms, index));
+  Rf_setAttrib(x, R_NamesSymbol, nms);
+
+  UNPROTECT(2);
+}
+
+/**
```
hadley: Inconsistent commenting style

src/slice.c
```diff
@@ -8,7 +8,8 @@ SEXP fns_vec_slice_fallback = NULL;
 // Defined below
 SEXP vec_as_index(SEXP i, SEXP x);
 static void slice_names(SEXP x, SEXP to, SEXP index);
-static SEXP vec_slice_impl(SEXP x, SEXP index, bool dispatch);
+static SEXP vec_slice_impl(SEXP x, SEXP index, SEXP to, bool dispatch);
```
hadley: I think it might be worth including `vec_slice_bare()` here as well (even though it's not technically needed) and including a longer comment that describes the differences between the variants (and when you would use them)

src/slice.c
lionel-: I think we should use that style for docstrings with parameter lists, it's supported in vim and emacs.

src/slice.c
lionel-: In case it is generated by an altrep vector. Though I don't know anything about protection standards in that case.

hadley: I think ALTREP will have to handle protection, otherwise it would break a lot of existing code. @jimhester can you confirm? (or at least speculate given that I don't think we have ALTREP lists yet)

jimhester: We don't have altrep lists, and I think the semantics would have to be the same as normal lists, so you shouldn't need an additional PROTECT here.

lionel-:
@hadley All fixed.

Issue #115, opened 2018-10-09
Document NULL handling


hadley:
i.e. `NULL` is consistently treated as equivalent to a missing/absent _argument_

i.e. `vec_size_common(x, y)`  is equivalent to `vec_size_common(x, y, )` is equivalent to `vec_size_common(x, y, NULL)`

hadley:
Also related to what `vec_size_common()` and `vec_type_common()` return.

PR #250, opened 2019-03-25
i argument to vec_slice() gains NULL default

### LINKED ISSUE #246: Do we need/have an equivalent for x[] <- value?
Required when we start forwarding `[<-()` to `vec_slice()` .

``` r
library(vctrs)

x <- 1:3
x[] <- 4L

x <- 1:3
vec_slice(x, ) <- 4L
#> Error in vec_size(i): argument "i" is missing, with no default
```

<sup>Created on 2019-03-24 by the [reprex package](https://reprex.tidyverse.org) (v0.2.1.9000)</sup>

krlmlr:
I'm sure this can also be fixed using missing arguments, but explicit `NULL`s seem to be much easier to reason about.

Conflicts with #249, but easy to redo on top of that.

Closes #246.

### DIFF
--- a/src/slice.c
+++ b/src/slice.c
@@ -202,7 +202,7 @@ static SEXP vec_slice_impl(SEXP x, SEXP index, SEXP to, bool dispatch) {
 
 // [[export]]
 SEXP vctrs_slice(SEXP x, SEXP index, SEXP native) {
-  if (x == R_NilValue || index == R_MissingArg) {
+  if (x == R_NilValue || index == R_NilValue) {
     return x;
   }
 
@@ -402,11 +402,11 @@ static SEXP chr_as_index(SEXP i, SEXP x) {
 
 SEXP vec_as_index(SEXP i, SEXP x) {
   switch (TYPEOF(i)) {
+  case NILSXP: return lgl_as_index(Rf_ScalarLogical(1), x);
   case INTSXP: return int_as_index(i, x);
   case REALSXP: return dbl_as_index(i, x);
   case LGLSXP: return lgl_as_index(i, x);
   case STRSXP: return chr_as_index(i, x);
-  case SYMSXP: if (i == R_MissingArg) return i;
 
   default: Rf_errorcall(R_NilValue, "`i` must be an integer, character, or logical vector, not a %s.",
                         Rf_type2char(TYPEOF(i)));
--- a/src/slice.c
+++ b/src/slice.c
@@ -402,7 +402,7 @@ static SEXP chr_as_index(SEXP i, SEXP x) {
 
 SEXP vec_as_index(SEXP i, SEXP x) {
   switch (TYPEOF(i)) {
-  case NILSXP: return lgl_as_index(Rf_ScalarLogical(1), x);
+  case NILSXP: return lgl_as_index(vctrs_shared_true, x);
   case INTSXP: return int_as_index(i, x);
   case REALSXP: return dbl_as_index(i, x);
   case LGLSXP: return lgl_as_index(i, x);
--- a/src/slice.c
+++ b/src/slice.c
@@ -202,7 +202,7 @@ static SEXP vec_slice_impl(SEXP x, SEXP index, SEXP to, bool dispatch) {
 
 // [[export]]
 SEXP vctrs_slice(SEXP x, SEXP index, SEXP native) {
-  if (x == R_NilValue || index == R_NilValue) {
+  if (x == R_NilValue) {
     return x;
   }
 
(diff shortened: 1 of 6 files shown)


src/slice.c
```diff
@@ -352,11 +352,11 @@ static SEXP chr_as_index(SEXP i, SEXP x) {
 
 SEXP vec_as_index(SEXP i, SEXP x) {
   switch (TYPEOF(i)) {
+  case NILSXP: return lgl_as_index(Rf_ScalarLogical(1), x);
```
lionel-: This needs protection. Perhaps in `nil_as_index()` so the protection/unprotection doesn't mess up the switch.

lionel-: (I don't want to rely on the fact thet scalarlogical returns a shared constant)

Oh this can just be `vctrs_shared_true`.

lionel-:
@krlmlr Can you make `i` default to `TRUE` please?

krlmlr:
Do we also want to replicate R's behavior for `is.null(i)` ?

lionel-:
> Do we also want to replicate R's behavior for is.null(i) ?

Sorry I don't understand the question.

krlmlr:
Do we need this for `vec_slice()`:

``` r
iris[NULL, ]
#> [1] Sepal.Length Sepal.Width  Petal.Length Petal.Width  Species     
letters[NULL]
#> character(0)
letters[NULL] <- "A"
letters
#>  [1] "a" "b" "c" "d" "e" "f" "g" "h" "i" "j" "k" "l" "m" "n" "o" "p" "q"
#> [18] "r" "s" "t" "u" "v" "w" "x" "y" "z"
```

<sup>Created on 2019-03-26 by the [reprex package](https://reprex.tidyverse.org) (v0.2.1.9000)</sup>

hadley:
No

krlmlr:
Then I'd rather leave the default as `NULL`, that's the desired semantics (#115).

lionel-:
Your question is more general. Do we need this for vctrs:

```r
letters[integer()] <- "A"
letters
#>  [1] "a" "b" "c" "d" "e" "f" "g" "h" "i" "j" "k" "l" "m" "n" "o" "p" "q" "r"
#> [19] "s" "t" "u" "v" "w" "x" "y" "z"
```

Again this has compositional implications that should be carefully thought out.

I strongly recommend against `foo[c(idx1, idx2)]` having unstable semantics.

lionel-:
Maybe it'd make sense to have `NULL` as default with the current semantics. That would be the most consistent with #115 because `foo[bar]` would keep obs in `bar` and `foo[]` wouldn't keep any obs. But that'd make `[` methods powered by `vec_slice()` different from base R's. You'd need to use `TRUE` instead of a missing index.

I think base R's behaviour for missing indices makes sense in the context of multi-dim subsetting, where `[, ]` means subset all rows and all columns. So we can't really change that.

krlmlr:
I agree that the following invariant should hold also for `NULL` values in `i1` and `i2`:

```r
identical(vec_c(x[i1], x[i2]), x[vec_c(i1, i2)])
```

But it feels even more important that `x[]` should be the same as `vec_slice(x, )` . This means that we need to deviate from the "`NULL` means missing" idiom in this particular case. I think it's fine if we document it in #115.

lionel-:
Here is what I suggest:

- Avoid missing arg in `vec_slice()` and `vec_slice<-` because they're for programmatic use.

- Keep missing arg in `[`. Interface with `vec_slice()` with `maybe_missing(i, default = )`.

- For `[`, the default missing arg is translated to `TRUE`. For `[<-`, it is translated to `NULL`.

- The defaults for `vec_slice()` and `vec_slice<-` are both `NULL`, which is equivalent to an empty index.

This way `vec_slice(x)` and `vec_slice(x, int())`, as well as `vec_slice(x) <- foo` and `vec_slice(x, int()) <- foo`, have equivalent semantics. Not supplying an index and supplying an empty index are the same. As for our `[` methods, their behaviour is consistent in base R and with multi-dimensional subsetting.

lionel-:
> it feels even more important that x[] should be the same as vec_slice(x, )

It doesn't seem important to me.

krlmlr:
What do we gain by making `vec_slice(i, )` behave differently to `x[]` ?

lionel-:
Consistency. An absence of index and an empty index are the same.

`vec_slice()` is for programmatic use and one-dimensional subsetting. It doesn't have the same interface constraints as `[`.

In fact maybe a missing index should just be an error for `vec_slice()` and `vec_slice<-`. The caller should be forced to use the appropriate sentinel (e.g. `TRUE` for all obs, and `NULL` / `int()` for no obs).

krlmlr:
Fully agree, implemented this way.

hadley:
Should we make `vec_slice(x, NULL)` our standard approach to generate a size-0 slice?

lionel-:
> Should we make vec_slice(x, NULL) our standard approach to generate a size-0 slice?

Perhaps `vec_slice(x, int())` is a bit easier to understand for people not familiar with the syntax.

PR #253, opened 2019-03-27
Set `dim` attribute before any names in default `vec_restore()` method

DavisVaughan:
I hit a weird bug with the default `vec_restore()` method that has to do with how R treats names of 1D arrays. Here's the minimal example:

``` r
library(vctrs)

x <- array(c(1, 2), dimnames = list(c("nm1", "nm2")))
x
#> nm1 nm2 
#>   1   2

# `names` attribute is actually set?
vec_restore(x, matrix(1))
#> nm1 nm2 
#>   1   2 
#> attr(,"names")
#> [1] "nm1" "nm2"
```

A bit of exploring shows that R treats `names` special for 1D arrays

``` r
library(vctrs)

x <- array(c(1, 2), dimnames = list(c("nm1", "nm2")))
x
#> nm1 nm2 
#>   1   2

# only dimnames
attributes(x)
#> $dim
#> [1] 2
#> 
#> $dimnames
#> $dimnames[[1]]
#> [1] "nm1" "nm2"

# but this registers names
names(x)
#> [1] "nm1" "nm2"

# so does this
attr(x, "names", exact = TRUE)
#> [1] "nm1" "nm2"

# no problems here
vec_restore(x, 1L)
#> nm1 nm2 
#>   1   2

# `names` attribute is actually set?
vec_restore(x, matrix(1))
#> nm1 nm2 
#>   1   2 
#> attr(,"names")
#> [1] "nm1" "nm2"

x_2D <- as.matrix(x)
x_2D
#>     [,1]
#> nm1    1
#> nm2    2

# no problem here
vec_restore(x_2D, matrix(1))
#>     [,1]
#> nm1    1
#> nm2    2
```

I think what is happening on the vctrs side is that the default `vec_restore()`:

- Collects the `R_NamesSymbol` from `x` as `nms`, which is actually a thing for 1D arrays, but otherwise returns NULL for 2D+ structures.
- Calls `SET_ATTRIB(x, attrib)` which sets `dim = c(1,1)`, so its now 2D
- Calls `Rf_setAttrib(x, R_NamesSymbol, nms);`, which actually sets the names attribute with `nms` for a matrix. Great. If this was a 1D array still, I think it would skip over it and wouldn't set the `"names"` attribute directly.
- Then `Rf_setAttrib(x, R_DimSymbol, dim);` restores the 1D dimensionality.

I think that _ideally_ `SET_ATTRIB(x, attrib)` call wouldn't set incorrect attributes to begin with (meaning it shouldn't have used the names, dim, and dimnames of `to`, which immediately get overwritten anyways).

The easiest alternative solution is implemented here. I just set the `R_DimSymbol` first so that the setting of the `R_NamesSymbol` is skipped over.

### DIFF
--- a/src/cast.c
+++ b/src/cast.c
@@ -280,6 +280,29 @@ SEXP vctrs_restore_default(SEXP x, SEXP to) {
     ++n_protect;
   }
 
+  // Remove vectorised attributes which might be incongruent after reshaping.
+  // Shouldn't matter for GNU R but other R implementations might have checks.
+  {
+    SEXP node = attrib;
+    SEXP prev = R_NilValue;
+
+    while (node != R_NilValue) {
+      SEXP tag = TAG(node);
+      if (tag == R_NamesSymbol || tag == R_DimSymbol || tag == R_DimNamesSymbol) {
+        if (prev == R_NilValue) {
+          attrib = CDR(attrib);
+          node = CDR(node);
+          continue;
+        }
+
+        SETCDR(prev, CDR(node));
+      }
+
+      prev = node;
+      node = CDR(node);
+    }
+  }
+
   // Copy attributes but keep names and dims
   SEXP nms = PROTECT(Rf_getAttrib(x, R_NamesSymbol));
   ++n_protect;
--- a/src/cast.c
+++ b/src/cast.c
@@ -282,13 +282,22 @@ SEXP vctrs_restore_default(SEXP x, SEXP to) {
 
   // Remove vectorised attributes which might be incongruent after reshaping.
   // Shouldn't matter for GNU R but other R implementations might have checks.
+  // Also record class to set it later with `Rf_setAttrib()`. This restores
+  // the OBJECT bit and is likely more compatible with other implementations.
+  SEXP class = R_NilValue;
+
   {
     SEXP node = attrib;
     SEXP prev = R_NilValue;
 
     while (node != R_NilValue) {
       SEXP tag = TAG(node);
-      if (tag == R_NamesSymbol || tag == R_DimSymbol || tag == R_DimNamesSymbol) {
+
+      if (tag == R_NamesSymbol || tag == R_DimSymbol ||
+          tag == R_DimNamesSymbol || tag == R_ClassSymbol) {
+        if (tag == R_ClassSymbol) {
+          class = CAR(node);
+        }
         if (prev == R_NilValue) {
           attrib = CDR(attrib);
           node = CDR(node);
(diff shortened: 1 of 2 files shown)


src/cast.c
```diff
@@ -291,8 +291,8 @@ SEXP vctrs_restore_default(SEXP x, SEXP to) {
   ++n_protect;
 
   SET_ATTRIB(x, attrib);
-  Rf_setAttrib(x, R_NamesSymbol, nms);
   Rf_setAttrib(x, R_DimSymbol, dim);
+  Rf_setAttrib(x, R_NamesSymbol, nms);
```
lionel-: Can you add a comment please?

lionel-:
> I think that ideally SET_ATTRIB(x, attrib) call wouldn't set incorrect attributes to begin with

Agreed, this could pose issues on other implementations of R which might have additional checks. I have pushed a few commits to avoid bypassing the public API.

src/cast.c
```diff
@@ -280,24 +280,58 @@ SEXP vctrs_restore_default(SEXP x, SEXP to) {
...
-  // Copy attributes but keep names and dims
-  SEXP nms = PROTECT(Rf_getAttrib(x, R_NamesSymbol));
-  ++n_protect;
+  // Remove vectorised attributes which might be incongruent after reshaping.
+  // Shouldn't matter for GNU R but other R implementations might have checks.
+  // Also record class to set it later with `Rf_setAttrib()`. This restores
+  // the OBJECT bit and is likely more compatible with other implementations.
+  SEXP class = R_NilValue;
+
+  {
```
DavisVaughan: Extra bracket?

src/cast.c
```diff
@@ -280,24 +280,58 @@ SEXP vctrs_restore_default(SEXP x, SEXP to) {
...
 
-  SET_ATTRIB(x, attrib);
-  Rf_setAttrib(x, R_NamesSymbol, nms);
-  Rf_setAttrib(x, R_DimSymbol, dim);
-  Rf_setAttrib(x, R_DimNamesSymbol, dimnames);
+  if (dim == R_NilValue) {
+    SEXP nms = PROTECT(Rf_getAttrib(x, R_NamesSymbol));
+    SET_ATTRIB(x, attrib);
+    Rf_setAttrib(x, R_NamesSymbol, nms);
+    UNPROTECT(1);
```
DavisVaughan: Not sure if you are trying to be consistent with `++n_protect` here or not

src/cast.c
lionel-: For scope.

src/cast.c
lionel-: There is no need for counted protection here.

DavisVaughan:
LGTM

lionel-:
Thanks!

PR #257, opened 2019-03-28
Fix assertion of partial frames

### LINKED ISSUE #198: vec_assert needs to work with partial specs
``` r
library(vctrs)

df <- data.frame(x = 1, col_name = 2)
vec_assert(
  df,  
  partial_frame(
    col_name = character(), 
    measure = character()
  )
)
#> Error: `df` must be <<partial_frame>>, not <df[,2]>.
```

<sup>Created on 2019-02-21 by the [reprex package](https://reprex.tidyverse.org) (v0.2.1.9000)</sup>

lionel-:
Closes #198.

* Forget unlearned columns when partial frame is finalised.
* Finalise partial types in `vec_assert()`.

I'm not sure this is the right approach as I'm just learning about partial types.

### DIFF
--- a/tests/testthat/test-partial-frame.R
+++ b/tests/testthat/test-partial-frame.R
@@ -18,9 +18,10 @@ test_that("order of variables comes from data", {
   expect_named(vec_type_common(df, pf), c("x", "y"))
 })
 
-test_that("partial variables added to end if not in data", {
+test_that("partial variables are forgotten if not in data", {
   pf <- partial_frame(y = 1)
   df <- data.frame(x = 1)
-  expect_named(vec_type_common(pf, df), c("x", "y"))
-  expect_named(vec_type_common(df, pf), c("x", "y"))
+  expect_named(vec_type_common(pf, df), "x")
+  expect_named(vec_type_common(df, pf), "x")
+})
 })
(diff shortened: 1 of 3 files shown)


tests/testthat/test-partial-frame.R
```diff
@@ -18,9 +18,15 @@ test_that("order of variables comes from data", {
...
   df <- data.frame(x = 1)
-  expect_named(vec_type_common(pf, df), c("x", "y"))
-  expect_named(vec_type_common(df, pf), c("x", "y"))
+  expect_named(vec_type_common(pf, df), "x")
+  expect_named(vec_type_common(df, pf), "x")
+})
+
+test_that("can assert partial frames", {
+  partial <- partial_frame(col_name = character(), measure = character())
+  expect_true(vec_is(data.frame(x = 1, col_name = 2), partial))
```
hadley: The logic doesn't seem quite right here — I think this should assert that `df` contains at least the columns in the partial frame.

hadley:
I'm not entirely sure my thinking about partial types is clear, but they are about making minimal assertions - i.e. this is a data frame (without asserting what columns it contains), or this is a data frame containing at least a column called `id`, or a column called `group` that is a factor (without specifying levels).

lionel-:
That makes sense. Let's start over.

PR #259, opened 2019-03-28
Throw lossy errors with restarts

lionel-:
The errors can be selectively suppressed by calling `suppress_errors_lossy_cast()` with prototypes for `x` and `to`.

Also take `to` instead of `y` in `stop_lossy_cast()`, as this makes it easier to map the arguments from the cast. We still store `to` as `y` in the condition (but also as `to`).


R/conditions.R
```diff
@@ -79,23 +87,53 @@ stop_incompatible_op <- function(op, x, y, details = NULL, ..., message = NULL,
...
+      ...,
+      .subclass = c(.subclass, "vctrs_error_cast_lossy")
+    )
+  )
+}
+#' @rdname vctrs-conditions
+#' @param x_ptype,to_ptype Suppress only the casting errors where `x`
+#'   or `to` match these [prototypes][vec_type].
+#' @export
+suppress_errors_lossy_cast <- function(expr, x_ptype = NULL, to_ptype = NULL) {
```
hadley: My only question is the name of this function — would it be better as `allow_lossy_cast()`?

lionel-: I like it.

lionel-:
`suppress_errors_lossy_cast()` is now `allow_lossy_cast()` and `stop_lossy_cast()` is unexported in favour of `maybe_lossy_cast()`.

R/conditions.R
```diff
@@ -1,15 +1,23 @@
...
 #'
+#' @description
+#'
 #' These errors and warnings have custom classes and structures to make
 #' testing easier.
 #'
+#' By default, lossy casts are an error. Use
+#' `allow_lossy_cast()` to silence these errors and continue
+#' with the partial results. The lost values may be set to `NA` or to
+#' a lower value resolution, depending on the type of cast.
```
hadley: I think you need a paragraph or two here about `maybe_lossy_cast()` — it either throws an error or returns a vector, if `allow_lossy_cast()` is used

PR #264, opened 2019-04-03
Rewrite `vec_type()` in C

### LINKED ISSUE #263: Lazily generate label in vec_assert
Currently `arg %||% as_label(substitute(x))` appears to take about 25% of the run time when the assertion passes

``` r
library(vctrs)
library(rlang)

label1 <- function(x, arg = NULL) {
  arg %||% as_label(substitute(x))
}
label2 <- function(x, arg = NULL) {
  delayedAssign("label", arg %||% as_label(substitute(x)))
  10
}

x <- 1:10

bench::mark(
  vctrs::vec_assert(x, integer(), size = 10),
  label1(x),
  label2(x),
  check = FALSE
)[1:6]
#> # A tibble: 3 x 6
#>   expression                         min    mean  median      max `itr/sec`
#>   <chr>                         <bch:tm> <bch:t> <bch:t> <bch:tm>     <dbl>
#> 1 vctrs::vec_assert(x, integer…   43.3µs 54.51µs 48.03µs 228.47µs    18345.
#> 2 label1(x)                       11.7µs 15.75µs 13.53µs  19.72ms    63489.
#> 3 label2(x)                        1.4µs  1.99µs  1.62µs   2.15ms   502699.
```

<sup>Created on 2019-03-31 by the [reprex package](https://reprex.tidyverse.org) (v0.2.1.9000)</sup>

It's probably worth worrying about the performance so we can use this as a standard tool for making assertions about argument types across the tidyverse.

lionel-:
- Genericity achieved through `vec_proxy()` + `vec_restore()`. `vec_type()` is no longer an S3 generic.

- `vec_type()` is now always a no-op with scalars. This way `vctrs_sclr` does not need to be special-cased. Ideally `vctrs_sclr` should only serve to implement scalars with base R methods, not with vctrs methods.

- `vec_type2()` on the other hand always requires vectors and that is now enforced.

This improves the performance of vctrs primitives like `vec_assert()` and closes #263.


R/conditions.R
```diff
@@ -243,6 +246,15 @@ stop_unimplemented <- function(x, method) {
   )
 }
 
+stop_scalar_type <- function(x, arg = NULL) {
+  if (is_null(arg)) {
```
hadley: Are you sure we shouldn't require this argument?

src/conditions.c
```diff
@@ -0,0 +1,18 @@
+#include "vctrs.h"
+#include "utils.h"
+
+
+void stop_scalar_type(SEXP x, const char* arg_str) {
+  SEXP arg;
+  if (!strlen(arg_str)) {
+    arg = PROTECT(Rf_mkString(arg_str));
+  } else {
+    arg = PROTECT(R_NilValue);
```
hadley: That doesn't actually need protecting, right?

lionel-: Right, this is just to avoid counted protection.

R/conditions.R
lionel-: I think it would be difficult to make it a requirement. I have two other PRs pending locally and IIRC I call `stop_scalar_type()` without `arg` because passing the arg would require adding it as argument to many vctrs primitives. I'll think about it though.

src/unspecified.c
```diff
@@ -1,5 +1,35 @@
 #include "vctrs.h"
+#include "utils.h"
 
+// Initialised at load time
+static SEXP unspecified_class = NULL;
```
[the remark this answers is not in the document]
lionel-: For classes with no variable attributes I think it's better to compare the whole attribute pairlist as in https://github.com/r-lib/rlang/blob/76f692b6cdff0f07c594283709893a8af645a1a0/src/internal/dots.c#L22

I'll do a quick benchmark with and without the pointer comparison.

src/type.c
```diff
@@ -5,6 +5,59 @@
...
+  SEXP proxy = PROTECT(vec_proxy(x));
+  SEXP out = PROTECT(vec_type_rec(proxy, false));
+  out = vec_restore(out, x, vctrs_shared_empty_int);
+  UNPROTECT(2);
+  return out;
+}
+
+static SEXP vec_type_rec(SEXP x, bool dispatch) {
+  switch (vec_typeof_impl(x, dispatch)) {
+  case vctrs_type_scalar:    return x;
```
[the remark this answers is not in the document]
lionel-: This is currently required because partial frames are not vectors. Maybe they should become vectors though, since you're planning for a partial vector type.

src/conditions.c
hadley: It's just a bit weird since you don't unprotect (because of the error)

src/type.c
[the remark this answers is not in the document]
hadley: Ok, that makes sense

src/unspecified.c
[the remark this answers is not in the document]
hadley: Oh yeah, even better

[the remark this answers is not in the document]
lionel-: 30 ns gain compared to `Rf_inherits()` with few classes / few attributes so it only matters for very common operations in very low level functions.

src/conditions.c
lionel-: ah yes, not protecting would cause a protection balance warning in rchk. And I didn't unprotect partly because it's weird to add unreachable code, and partly because I wanted to test whether rchk understands the `error` call.

lionel-:
```r
bench::mark(
  vec_type_before(NA),
  vec_type(NA)
)[1:6]
#>   expression               min     mean   median      max `itr/sec`
#>   <chr>               <bch:tm> <bch:tm> <bch:tm> <bch:tm>     <dbl>
#> 1 vec_type_before(NA)   1.78µs   2.14µs   2.02µs     52µs   466920.
#> 2 vec_type(NA)           373ns 450.99ns    409ns   16.3µs  2217345.

bench::mark(
  vec_type_before(letters),
  vec_type(letters)
)[1:6]
#>   expression                    min     mean   median      max `itr/sec`
#>   <chr>                    <bch:tm> <bch:tm> <bch:tm> <bch:tm>     <dbl>
#> 1 vec_type_before(letters)   2.22µs   2.67µs    2.5µs    166µs   373978.
#> 2 vec_type(letters)           385ns 494.87ns    455ns   32.2µs  2020729.

df <- data.frame(x = 1:2, y = c(NA, NA))
bench::mark(
  vec_type_before(df),
  vec_type(df)
)[1:6]
#>   expression               min     mean   median      max `itr/sec`
#>   <chr>               <bch:tm> <bch:tm> <bch:tm> <bch:tm>     <dbl>
#> 1 vec_type_before(df)   12.6µs   15.6µs   14.5µs   2.08ms    64205.
#> 2 vec_type(df)           821ns  957.9ns    890ns  47.33µs  1044000.
```

PR #265, opened 2019-04-03
Only call `vec_proxy_equal()` once in `vec_unique()`

DavisVaughan:
`vec_unique_loc()` already calls `vec_proxy_equal()` so I don't think there is need to do it in `vec_unique()` as well.


lionel-:
Thanks! Note that `vec_proxy_equal()` will be replaced by `vec_proxy()` which is idempotent and a no-op when called multiple times. Still seems a good idea to take the proxy only when needed.

PR #276, opened 2019-04-04
Explain type differences when assertion fails

### LINKED ISSUE #260: Need a better error message for `vec_assert()` with data frames
This isn't too descriptive of what is happening

``` r
library(vctrs)
iris_ptype <- iris[0,]
iris_ptype$Species <- as.character(iris_ptype$Species)
vec_assert(iris, iris_ptype)
#> `iris` must be <df[,5]>, not <df[,5]>.
```

<sup>Created on 2019-03-29 by the [reprex package](https://reprex.tidyverse.org) (v0.2.1.9000)</sup>

lionel-:
Fixes #260

I wonder if we should lay out factors in the same way as data frames and only hash the labels when there are too many?

### DIFF
--- a/R/assert.R
+++ b/R/assert.R
@@ -41,7 +41,7 @@ vec_assert <- function(x, ptype = NULL, size = NULL, arg = as_label(substitute(x
     ptype <- vec_type(ptype)
     x_type <- ununspecify(vec_type(x))
     if (!is_same_type(x_type, ptype)) {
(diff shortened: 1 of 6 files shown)


hadley:
Yeah, that'd be great. The primary difference with data frames though, is that the individual levels can also be very long, so giving a nice abbreviation is a reasonable amount of work. (Showing the hash is just a hack until we do that)

PR #277, opened 2019-04-04
Fail early when partial type doesn't match

lionel-:
* `vec_type2(df, pf)` is now an error when the type don't match

* `vec_is()` wraps `vec_type_common()` in a `maybe()` to catch incompatible types errors.

### DIFF
--- a/R/assert.R
+++ b/R/assert.R
@@ -95,12 +95,11 @@ vec_is <- function(x, ptype = NULL, size = NULL) {
 
 is_same_type <- function(x, ptype) {
   if (is_partial(ptype)) {
-    ptype <- maybe(vec_type_common(x, ptype), "vctrs_error_incompatible_type")
-
-    if (!is_null(ptype$error)) {
-      return(FALSE)
-    }
-    ptype <- ptype$value
+    env <- environment()
+    ptype <- tryCatch(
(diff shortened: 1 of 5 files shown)


R/assert.R
```diff
@@ -95,8 +95,14 @@ vec_is <- function(x, ptype = NULL, size = NULL) {
 
 is_same_type <- function(x, ptype) {
   if (is_partial(ptype)) {
-    ptype <- vec_type_common(x, ptype)
+    ptype <- maybe(vec_type_common(x, ptype), "vctrs_error_incompatible_type")
```
hadley: Given that we're not fully committed to this pattern, I think it would be better to inline the `tryCatch()` here and avoid any abstraction to hide the ugliness of the approach.

lionel-:
@hadley There's an issue with the binary dispatch in case of inheritance. To make sure dispatch between partial frames and tibbles is symmetric, I changed tibble's default method to call the data frame method: https://github.com/r-lib/vctrs/pull/277/commits/a533dd3246d1350984d6f06d30f9c27356833795

That's one more constraint for the double dispatch, the parent methods must be called explicitly from default methods.

hadley:
Ah yes, that makes sense. Can you file an issue to document that somewhere? I think we might need to start collecting stuff for a vignette about double dispatch problems.

PR #281, opened 2019-04-07
`vec_slice()` fails when slicing shaped objects by name

DavisVaughan:
The following currently fails:

``` r
library(vctrs)

x <- matrix(1:2)
dimnames(x) <- list(c("foo", "bar"))

vec_slice(x, "foo")
#> Error: Can't use character to index an unnamed vector.
```

I know eventually shaped object slicing will be incorporated fully, but this might be a general solution for `chr_as_index()`.

### DIFF
--- a/src/slice.c
+++ b/src/slice.c
@@ -392,7 +392,7 @@ static SEXP chr_as_index(SEXP i, SEXP x) {
 
   if (has_dim(x)) {
     SEXP dim_nms = Rf_getAttrib(x, R_DimNamesSymbol);
-    if (dim_nms != R_NilValue) {
+    if (dim_nms != R_NilValue && Rf_length(dim_nms) >= 1) {
       nms = VECTOR_ELT(dim_nms, 0);
     }
   } else {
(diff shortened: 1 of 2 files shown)


src/slice.c
```diff
@@ -387,7 +387,17 @@ static SEXP lgl_as_index(SEXP i, SEXP x) {
 }
 
 static SEXP chr_as_index(SEXP i, SEXP x) {
-  SEXP nms = Rf_getAttrib(x, R_NamesSymbol);
+
+  SEXP nms = R_NilValue;
+
+  if (has_dim(x)) {
+    SEXP dim_nms = Rf_getAttrib(x, R_DimNamesSymbol);
+    if (dim_nms != R_NilValue) {
```
hadley: I think you also need to check length is at least 1 here

DavisVaughan: Gotcha. I've never actually seen that case before.

``` r
x <- matrix(1)
dimnames(x) <- list()
dimnames(x)
#> NULL
```

lionel-:
Thanks!

PR #283, opened 2019-04-08
Allow downgrading lossy cast errors to deprecation warnings

### LINKED ISSUE #280: Consider bringing back warn_lossy_cast()
In {hms} I'd like to introduce a warning when characters can't be parsed as `hms()`. I don't like to make this an error because I'd like to give downstream maintainers a chance to adapt seamlessly. I think this scenario will occur very frequently as vector classes are lifted to use {vctrs}.

This can be as simple as adding an optional argument to `maybe_lossy_cast()`. Then, `warn_lossy_cast()` could forward to `maybe_lossy_cast()` and be marked soft-deprecated.

Also, there are quite a few instances of `warn_lossy_cast` floating around on GitHub: https://github.com/search?q=warn_lossy_cast&type=Code.

Currently I'm rolling my own version of `warn_lossy_cast()`.

lionel-:
Closes #280

To detect that the user has wrapped their code in `allow_lossy_cast()`, we signal a lossy cast error with a fake restart point. If the latter is invoked, it means the user has correctly called `allow_lossy_cast()` with the right prototypes.

### DIFF
--- a/R/conditions.R
+++ b/R/conditions.R
@@ -114,6 +114,10 @@ stop_incompatible_op <- function(op, x, y, details = NULL, ..., message = NULL, 
 #'   own location vector, possibly empty.
 #' @param locations An optional integer vector giving the
 #'   locations where `x` lost information.
+#' @param .deprecation If `TRUE`, the error is downgraded to a
+#'   deprecation warning. This is useful for transitioning your class
+#'   to a stricter conversion scheme. The warning advises your users
+#'   to wrap their code with `allow_lossy_cast()`.
 #' @export
 maybe_lossy_cast <- function(result, x, to,
                              lossy = NULL,
@@ -121,8 +125,13 @@ maybe_lossy_cast <- function(result, x, to,
                              details = NULL,
                              ...,
                              message = NULL,
-                             .subclass = NULL) {
+                             .subclass = NULL,
+                             .deprecation = FALSE) {
   if (!any(lossy)) {
+    return(result)
+  }
+  if (.deprecation) {
+    maybe_warn_deprecated_lossy_cast(x, to)
     return(result)
   }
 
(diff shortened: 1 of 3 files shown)


R/conditions.R
```diff
@@ -180,6 +189,38 @@ allow_lossy_cast <- function(expr, x_ptype = NULL, to_ptype = NULL) {
...
+    warn_deprecated(paste_line(
+      glue::glue("We detected a lossy transformation from `{ from }` to `{ to }`."),
+      "The result will contain lower-resolution values or missing values.",
+      "To suppress this warning, wrap your code with `allow_lossy_cast()`:",
+      "",
+      "  # Allow all lossy transformations:",
+      "  vctrs::allow_lossy_cast(mycode())",
+      "",
+      "  # Allow only a specific transformation:",
+      "  vctrs::allow_lossy_cast(mycode(), x_ptype = from, to_ptype = to)",
```
krlmlr:
```suggestion
      "  vctrs::allow_lossy_cast(mycode(), x_ptype = from(), to_ptype = to())",
```

Not entirely sure about that, maybe skip entirely because it's hard to be accurate here. The warning is already long enough.

R/conditions.R
```diff
@@ -180,6 +189,38 @@ allow_lossy_cast <- function(expr, x_ptype = NULL, to_ptype = NULL) {
...
+      "",
+      "Consult `?vctrs::allow_lossy_cast` for more information."
```
krlmlr:
```suggestion
      "Consult `?vctrs::allow_lossy_cast` for more information, in particular about allowing only specific lossy coercions."
```

lionel-:
I'm not sure it's worth it to soft-deprecate the `warn_` variant given it's only used in one package and we can coordinate with Davis for the update. What do you think @DavisVaughan? Would you prefer a soft-deprecated `warn_lossy_cast()` for the probably package?

R/conditions.R
lionel-: You can just pass the full objects, we take the prototype automatically.

lionel-: I was wondering if we should rename `ptype` to `type` for that reason.

DavisVaughan:
I'm happy to submit a quick release of probably after a vctrs release rather than support backwards compat here. There will be a momentary breakage of CRAN vctrs with CRAN probably, but I don't think too many people use it yet so I'm not that worried about it.

R/conditions.R
```diff
@@ -180,6 +189,38 @@ allow_lossy_cast <- function(expr, x_ptype = NULL, to_ptype = NULL) {
...
+  # to handle the condition
+  handled <- withRestarts(
+    vctrs_restart_error_cast_lossy = function() TRUE,
+    {
+      cnd_signal(cnd("vctrs_error_cast_lossy", x = x, to = to))
+      FALSE
+    }
+  )
+
+  if (!handled) {
```
hadley: Would be a bit simpler to early return here.

R/conditions.R
hadley: I think `ptype` is fine; an object is by definition its own prototype.

R/conditions.R
hadley: I think existing wording is fine

PR #284, opened 2019-04-08
Support additional suscripts in index methods

### LINKED ISSUE #269: Is `[.vctrs_vctr` too restrictive for objects with shape?
If you create a new vctr class that acts like a matrix and try and subset it, you immediately hit problems with `[.vctrs_vctr` only allowing `i` through.

``` r
library(vctrs)

new_2d <- function(.data, dim) {
  vctrs::new_vctr(.data, dim = dim, class = "vctrs_2d")
}

x <- new_2d(c(1, 2), dim = c(2L, 1L))

# works
x[1]
#> <vctrs_2d[1]>
#>      [,1]
#> [1,] 1

# nope
x[, 1]
#> `...` is not empty.
#> These dots only exist for technical reasons and should be empty.

# nope
x[1, 1]
#> `...` is not empty.
#> These dots only exist for technical reasons and should be empty.


# What _should_ it return?
# should it just pass on to 
# the matrix method and drop dimensions?
# (i.e. we say that the underlying class
# controls whether dims are dropped or not)
y <- matrix(1:2)
y[, 1]
#> [1] 1 2

# or should it return this?
y[, 1, drop = FALSE]
#>      [,1]
#> [1,]    1
#> [2,]    2

# same question here I guess
y[1, 1]
#> [1] 1
```

If you do pass everything on to the underlying class, I think it might be good to think through what these return:

```r
x[1] # might call `vec_slice_native()` and drop dims
x[1,] # might call underlying method and not drop dims
```

On a similar note, `[<-.vctrs_vctr` should allow more than just `i` through as well. But I don't think the fix is as simple as just replacing `i` with `...` there.

lionel-:
Closes #269

Eventually `vec_index()` should be exported to make it easy to create `[` methods with vctrs semantics.

How does this look @DavisVaughan?

### DIFF
--- a/R/vctr.R
+++ b/R/vctr.R
@@ -157,8 +157,7 @@ format.vctrs_vctr <- function(x, ...) {
 
 #' @export
 `[.vctrs_vctr` <- function(x, i, ...) {
-  check_dots_empty_s3_consistency(...)
-  vec_slice_native(x, maybe_missing(i, TRUE))
+  vec_index(x, i, ...)
 }
 
 #' @export
--- a/R/rcrd.R
+++ b/R/rcrd.R
@@ -113,8 +113,7 @@ vec_cast.vctrs_rcrd.default <- function(x, to) {
 
 #' @export
 `[.vctrs_rcrd` <-  function(x, i, ...) {
-  check_dots_empty_s3_consistency(...)
-  vec_slice_native(x, i)
+  vec_index(x, i, ...)
 }
 
 #' @export
--- a/R/slice.R
+++ b/R/slice.R
@@ -94,6 +94,20 @@ vec_as_index <- function(i, x) {
   .Call(vctrs_as_index, i, x)
 }
 
+vec_index <- function(x, i, ...) {
+  i <- maybe_missing(i, TRUE)
+
+  if (!dots_n(...)) {
+    return(vec_slice_native(x, i))
+  }
+
+  out <- unclass(vec_proxy(x))
+  vec_assert(out)
+
+  i <- vec_as_index(i, out)
+  vec_restore(out[i, ..., drop = FALSE], x, i = i)
+}
+
 #' Create a missing vector
 #'
 #' @param x Template of missing vector
--- a/tests/testthat/test-rcrd.R
+++ b/tests/testthat/test-rcrd.R
@@ -210,8 +210,8 @@ test_that("dangerous methods marked as unimplemented", {
 
 # slicing -----------------------------------------------------------------
 
-test_that("dots must be empty", {
-  expect_error(new_rcrd(list(foo = "foo"))[1, 2], "is not empty")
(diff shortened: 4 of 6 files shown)


DavisVaughan:
Just confirming, you want to enforce `drop = FALSE` for all subscripts? Not saying it is necessarily a bad thing, just making sure it was intended.

``` r
library(vctrs)

new_shaped <- function(.data, dim) {
  vctrs::new_vctr(.data, dim = dim, class = "vctrs_shaped")
}

x <- new_shaped(c(1, 2), dim = c(1L, 2L))
x[, 1]
#> <vctrs_shaped[1]>
#>      [,1]
#> [1,] 1
```

DavisVaughan:
Would it be possible to export `vec_as_index()` as well? I have a wrapper written around it in rray that calls it on each index dimension, not just the first. The error messages aren't perfect (it doesn't tell you which dimension failed), but I like the features.

lionel-:
Yes I think it's best to hardcode `drop`. And `vec_as_index()` is intended for export, just didn't get around to it yet. I'll send a PR.

> The error messages aren't perfect (it doesn't tell you which dimension failed)

I wonder if we could/should throw errors with "enrichment restarts". You'd invoke the restart to return a derived error from a calling handler.

lionel-:
Do you think we should apply `vec_as_index()` on all unnamed arguments in `...`?

DavisVaughan:
> Yes I think it's best to hardcode drop

Sounds good.

> as the simpler catching and rethrowing pattern.

Would this be using `rlang::with_handlers()`? `vec_as_index()` doesn't currently throw typed errors, but I guess that's fine.

> And vec_as_index() is intended for export

This would be great!

> Do you think we should apply vec_as_index() on all unnamed arguments in ...?

I do this in rray, so yea probably. One thing I have to work around for this is `vec_as_index(i, x)` takes `x` which should be the right size for `i` (and possibly have the right names). So for 2D for example, for the columns I kind of have to create a fake object with size = number of columns, and if `i` is a character I also attach the column names to the fake object.

lionel-:
We might want to add `dim = 1` argument to `vec_as_index()`.

R/vctr.R
```diff
@@ -157,8 +157,7 @@ format.vctrs_vctr <- function(x, ...) {
 
 #' @export
 `[.vctrs_vctr` <- function(x, i, ...) {
-  check_dots_empty_s3_consistency(...)
-  vec_slice_native(x, maybe_missing(i, TRUE))
+  vec_index(x, i, ...)
```
hadley: Interestingly this makes it clear that `[` is not type-stable. I think that's correct, so nothing needs to change, but it's worth noting.

DavisVaughan: I tend to think of it as dimensionality-stable, but not dimension-stable (which makes it not type-stable).

PR #287, opened 2019-04-13
Optimization) `shape_broadcast()`

DavisVaughan:
I've noticed that sometimes `shape_broadcast()` can be an expensive function to call when it often doesn't need to be. With the following:

```
library(vctrs)

x <- matrix(1:10000) + 0L
y <- matrix(1:10000) + 0L

vec_cast(x, y)
```

- 3 copies are made on 3.5.2 (1 from `dim(x)<- dim_x`, and two from the final `[` call)
- 2 copies are made on 3.6.0 (the `dim(x)<- dim_x` copy is ALTREPed away)

For this example, I really think it should be much cheaper to call `vec_cast()`, so this PR makes two changes:

- If the shape of `x` and `y` are the same, return `x` early
- If the dimensionality of `x` does not change, do not set the dim of `x`

The first change helps on 3.5.2 and 3.6, the second change probably only helps on 3.5.2 but I don't think it hurts.

### DIFF
--- a/R/shape.R
+++ b/R/shape.R
@@ -32,10 +32,6 @@ shape_broadcast <- function(x, to) {
   if (length(dim_x) == 1L && length(dim_to) == 1L)
     return(x)
 
-  # Don't broadcast if nothing changes
-  if (identical(dim_x[-1], dim_to[-1]))
-    return(x)
-
   if (length(dim_x) > length(dim_to))
     stop_incompatible_cast(x, to, details = "Can not decrease dimensions")
 
@@ -45,13 +41,19 @@ shape_broadcast <- function(x, to) {
   if (any(!ok))
     stop_incompatible_cast(x, to, details = "Non-recyclable dimensions")
 
+  # Increase dimensionality if required
+  if (vec_dims(x) != length(dim_x))
+    dim(x) <- dim_x
+
   recycle <- dim_x != dim_to
+
+  # Avoid expensive subset
+  if (all(!recycle)) {
+    return(x)
+  }
 
   indices <- rep(list(missing_arg()), length(dim_to))
   indices[recycle] <- map(dim_to[recycle], rep_len, x = 1L)
-
-  if (vec_dims(x) != length(dim_x))
-    dim(x) <- dim_x
 
   eval_bare(expr(x[!!!indices, drop = FALSE]))
 }
--- a/R/shape.R
+++ b/R/shape.R
@@ -22,28 +22,33 @@ shape_common <- function(x, y) {
 }
 
 shape_broadcast <- function(x, to) {
-  if (is.null(x) || is.null(to))
+  if (is.null(x) || is.null(to)) {
     return(x)
+  }
 
   dim_x <- vec_dim(x)
   dim_to <- vec_dim(to)
 
   # Don't set dimensions for vectors
-  if (length(dim_x) == 1L && length(dim_to) == 1L)
+  if (length(dim_x) == 1L && length(dim_to) == 1L) {
     return(x)
+  }
 
-  if (length(dim_x) > length(dim_to))
+  if (length(dim_x) > length(dim_to)) {
     stop_incompatible_cast(x, to, details = "Can not decrease dimensions")
+  }
 
   dim_x <- dim2(dim_x, dim_to)$x
   dim_to[[1]] <- dim_x[[1]] # don't change number of observations
   ok <- dim_x == dim_to | dim_x == 1 | dim_to == 0
-  if (any(!ok))
+  if (any(!ok)) {
     stop_incompatible_cast(x, to, details = "Non-recyclable dimensions")
+  }
 
   # Increase dimensionality if required
-  if (vec_dims(x) != length(dim_x))
+  if (vec_dims(x) != length(dim_x)) {
     dim(x) <- dim_x
+  }
 
   recycle <- dim_x != dim_to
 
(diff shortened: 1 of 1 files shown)


DavisVaughan:
Actually, I think what is happening for `[` is that it makes 2 allocations to generate the indices (rows and columns), and 1 allocation to generate the container for the final result. The above example was bad because the size of the matrix happened to be the same size of the indices it makes.

These examples are much clearer. 

- In the first you see an allocation for the row indices and the final container (the column indices are so small they don't show up).
- In the second you see an allocation for the row and column indices, and the final container.

``` r
# (10000, 1)
x <- matrix(1:10000 + 0) # make it a double
trash <- profmem::profmem(x[,,drop=F])
profmem::profmem(x[,,drop=F])
#> x[, , drop = F]
#> 
#> Memory allocations:
#>        what  bytes      calls
#> 1     alloc  40048 <internal>
#> 2     alloc  80048 <internal>
#> total       120096

# (10000, 10000)
x <- matrix(1:1e8 + 0, ncol = 10000)
trash <- profmem::profmem(x[,,drop=F])
profmem::profmem(x[,,drop=F])
#> x[, , drop = F]
#> 
#> Memory allocations:
#>        what     bytes      calls
#> 1     alloc     40048 <internal>
#> 2     alloc     40048 <internal>
#> 3     alloc 800000048 <internal>
#> total       800080144

object.size(1:10000)
#> 40048 bytes
```

DavisVaughan:
I made one more tweak that optimizes the case where dimensionality is the only thing that changes. This comes up for me a decent bit in rray when I do subassignment.

```
library(vctrs)

x <- 1:10000
y <- matrix(1:10000)

vec_cast(x, y)
```

R/shape.R
```diff
@@ -41,12 +41,20 @@ shape_broadcast <- function(x, to) {
   if (any(!ok))
     stop_incompatible_cast(x, to, details = "Non-recyclable dimensions")
 
+  # Increase dimensionality if required
+  if (vec_dims(x) != length(dim_x))
```
hadley: Since you touched this, can you please update the style too? (To use `{}`)

DavisVaughan:
Same shape bench mark:

``` r
library(vctrs)
library(bench)

x <- matrix(1:10000 + 0)

# so memory allocations aren't polluted
trash <- vec_cast(x, x)

# 3.5.2 - Before:
mark(vec_cast(x, x))
#> # A tibble: 1 x 10
#>   expression   min   mean median   max `itr/sec` mem_alloc  n_gc n_itr
#>   <chr>      <bch> <bch:> <bch:> <bch>     <dbl> <bch:byt> <dbl> <int>
#> 1 vec_cast(…  46µs 77.3µs   61µs 214µs    12939.     195KB    18  5121

# 3.5.2 - After:
#> # A tibble: 1 x 10
#> 1 vec_cast(… 10.8µs 12.7µs 12.4µs 66.9µs    78722.        0B    14  9986

# 3.6.0 - Before:
#> # A tibble: 1 x 10
#>   expression    min   mean median   max `itr/sec` mem_alloc  n_gc n_itr
#>   <chr>      <bch:> <bch:> <bch:> <bch>     <dbl> <bch:byt> <dbl> <int>
#> 1 vec_cast(… 57.4µs 77.3µs   67µs 236µs    12929.     117KB    12  5524

# 3.6.0 - After:
#> # A tibble: 1 x 10
#>   expression   min   mean median    max `itr/sec` mem_alloc  n_gc n_itr
#>   <chr>      <bch> <bch:> <bch:> <bch:>     <dbl> <bch:byt> <dbl> <int>
#> 1 vec_cast(…  11µs 13.3µs 13.1µs 66.3µs    75127.        0B    13  9987
```

Dimensionality change only bench mark:

``` r
library(vctrs)
library(bench)

x <- matrix(1:10000 + 0)
y <- 1:10000 + 0

# so memory allocations aren't polluted
trash <- vec_cast(x, x)
trash <- vec_cast(y, x)

# 3.5.2 - Before:
mark(vec_cast(y, x))
#> # A tibble: 1 x 10
#>   expression    min   mean median   max `itr/sec` mem_alloc  n_gc n_itr
#>   <chr>      <bch:> <bch:> <bch:> <bch>     <dbl> <bch:byt> <dbl> <int>
#> 1 vec_cast(… 46.9µs 69.3µs 60.2µs 245µs    14420.     195KB    20  5815

# 3.5.2 - After:
#> # A tibble: 1 x 10
#>   expression    min   mean median   max `itr/sec` mem_alloc  n_gc n_itr
#>   <chr>      <bch:> <bch:> <bch:> <bch>     <dbl> <bch:byt> <dbl> <int>
#> 1 vec_cast(… 16.3µs 32.7µs 23.3µs 434µs    30591.    78.2KB    14  9986

# 3.6.0 - Before:
#> # A tibble: 1 x 10
#>   expression    min   mean median   max `itr/sec` mem_alloc  n_gc n_itr
#>   <chr>      <bch:> <bch:> <bch:> <bch>     <dbl> <bch:byt> <dbl> <int>
#> 1 vec_cast(… 46.8µs 59.4µs 55.3µs 234µs    16838.     117KB    15  7407

# 3.6.0 - After:
#> # A tibble: 1 x 10
#>   expression    min   mean median   max `itr/sec` mem_alloc  n_gc n_itr
#>   <chr>      <bch:> <bch:> <bch:> <bch>     <dbl> <bch:byt> <dbl> <int>
#> 1 vec_cast(… 10.1µs 14.2µs 12.8µs  78µs    70193.        0B     3  9997
```

hadley:
For future reference, you should only fix the style of the lines you changed, not the whole file.

DavisVaughan:
Ah, that was a misunderstanding on my part. I read your original comment wrong even though your request was clear.

PR #290, opened 2019-04-16
Implement `vec_type_finalise()` in C

lionel-:
Also extracts `vec_recurse()` and `df_map()` as useful C utils for common vctrs patterns.

Partial frames are kept in the R side until we solidify the partial type API.

```r
x <- vctrs::vec_type(data.frame(a = 1, b = NA))

before <- vctrs::vec_type_finalise
after <- vctrs2::vec_type_finalise

bench::mark(
  before(x),
  after(x)
)[1:6]
#> 1 before(x)    23.6µs  29.56µs   27.9µs  150.7µs    33830.
#> 2 after(x)      785ns   1.04µs    950ns   39.3µs   960796.
```


src/type.c
```diff
@@ -72,6 +48,31 @@ SEXP vec_type(SEXP x) {
   return vec_type_rec(x, true);
 }
 
+
+SEXP vec_type_finalise_rec(SEXP x, bool dispatch) {
+  if (OBJECT(x)) {
```
hadley: I don't quite understand the logic between how you've broken up the cases here. I would expect to see `unspecified`, `data.frame`, and `partial` classes more prominently

src/utils.c
```diff
@@ -92,6 +92,33 @@ SEXP vctrs_dispatch3(SEXP fn_sym, SEXP fn,
...
+  SEXP nms = PROTECT(Rf_getAttrib(df, R_NamesSymbol));
+  Rf_setAttrib(out, R_NamesSymbol, nms);
+
+  out = df_restore(out, df, vctrs_shared_empty_int);
+
+  UNPROTECT(2);
+  return out;
+}
+
+SEXP vec_recurse(SEXP x, SEXP (*rec)(SEXP, bool), SEXP i) {
```
hadley: Why is this called `vec_recurse()`?

lionel-: Because it recurses into `rec` with `dispatch = false`.

src/type.c
lionel-: `unspecified` and `partial` are not part of the vctrs enum so are/will be handled in the `if (OBJECT(x))` branch. Partial types are kept in R for now but will be included here in the future.

src/utils.c
hadley: What do you mean by recurses into? Where is the recursion?

lionel-: I mean `vec_method(dispatch = true) => vec_recurse() => vec_method(dispatch = false)`

hadley: Yeah, I just don't find it to be a particularly evocative name

lionel-: maybe `vec_dispatch()`?

PR #293, opened 2019-04-18
Pass dots to `format()`, not `vec_data()`

### LINKED ISSUE #292: Dots in format.vctrs_vctr() should not be passed to vec_data()
https://github.com/r-lib/vctrs/blob/8251896d6db132ada6e7702b753de24dee66dc56/R/vctr.R#L153

I assume they should be passed to `format()`

DavisVaughan:
Closes #292 

`...` should probably be passed for `format()` not `vec_data()`


lionel-:
Thanks!

hadley:
Woops!

PR #294, opened 2019-04-18
Refactor dispatch helpers

lionel-:


src/utils.c
```diff
@@ -20,84 +20,56 @@ bool is_bool(SEXP x) {
  * Dispatch with two arguments
  *
  * @param fn The method to call.
- * @param x,y,z Arguments passed to the method.
- * @param fn_sym,x_sym,y_sym,z_sym Symbols to which `x`, `y` and `z`
- *   should be assigned.  The assignment occurs in `env` and the
- *   dispatch call refers to these symbols.
+ * @param syms Symbols to which arguments are assigned.  The
```
hadley: What is `syms`? A null-terminated array?

src/utils.c
```diff
@@ -299,6 +271,43 @@ SEXP r_peek_option(const char* option) {
   return Rf_GetOption1(Rf_install(option));
 }
 
+/**
+ * Create a call or pairlist
+ *
+ * @param tags Optional. If not `NULL`, an array of symbols.
```
hadley: a null-terminated array of symbols

src/utils.c
lionel-: ah good point, I'll make this more precise.

PR #295, opened 2019-04-18
Build argument tags recursively in data frame methods of `vec_type2()`

lionel-:
```r
tibble <- tibble::tibble

vec_type2(
  tibble(x = tibble(y = tibble(z = 1))),
  tibble(x = tibble(y = tibble(z = "a")))
)
#> Error: No common type for `x$y$z` <double> and `x$y$z` <character>.
```

It is a bit disruptive to add new arguments to generics when `...` hasn't been part of the signature from the beginning. Maybe we should add dots to all generics to make them extensible?

I don't love the added boilerplate, but I'm not sure how to make it better.


DESCRIPTION
```diff
@@ -35,7 +36,8 @@ Suggests:
     testthat,
     tibble
 Remotes:
-    r-lib/rlang
+    r-lib/rlang,
+    r-lib/ellipsis
```
hadley: Do you still need the dev version here?

R/conditions.R
```diff
@@ -51,17 +51,38 @@ stop_incompatible <- function(x, y, details = NULL, ..., message = NULL, .subcla
...
   vec_assert(x)
   vec_assert(y)
 
-  message <- message %||% glue_lines(
-    "No common type for <{vec_ptype_full(x)}> and <{vec_ptype_full(y)}>.",
-    details
-  )
+  if (is_null(message)) {
+    if (nzchar(x_arg)) {
+      x_name <- paste0(" `", x_arg, "` ")
```
hadley: Where do you use `x_name` and `y_name`?

R/type2.R
```diff
@@ -32,20 +32,23 @@
 #' @keywords internal
 #' @param x,y Either vector types; i.e.
 #' @export
-vec_type2 <- function(x, y) {
-  return(.Call(vctrs_type2, x, y))
+vec_type2 <- function(x, y, ..., x_arg = "", y_arg = "") {
```
hadley: Are you sure we should use `...` here? It saves a bunch of typing, but is it the right thing to do?

R/type2.R
```diff
@@ -32,20 +32,23 @@
 #' @keywords internal
```
hadley: Need to document `x_arg` and `y_arg`

R/zzz.R
```diff
@@ -8,6 +8,14 @@
   s3_register("generics::as.ordered", "vctrs_vctr")
   s3_register("generics::as.difftime", "vctrs_vctr")
 
+  ns <- ns_env("vctrs")
+
+  if (getRversion() < "3.5.0") {
```
hadley: Should we use backports here, just for consistency?

tests/testthat/test-type2.R
```diff
@@ -62,3 +62,30 @@ test_that("stop_incompatible_type() checks for scalars", {
...
+  expect_args(factor(), new_hidden(), x_arg = "foo", y_arg = "bar")
+  expect_args(ordered(""), new_hidden(), x_arg = "foo", y_arg = "bar")
+  expect_args(ordered(""), factor(), x_arg = "foo", y_arg = "bar")
+  expect_args(bit64::as.integer64(1), new_hidden(), x_arg = "foo", y_arg = "bar")
+})
+
+test_that("vec_type2() data frame methods builds argument tags", {
+  df1 <- tibble(x = tibble(y = tibble(z = 1)))
+  df2 <- tibble(x = tibble(y = tibble(z = "a")))
+  expect_error(
```
hadley: I think you should consider using `expect_known_output()` here

hadley:
I think the amount of boilerplate is to be expected — a lot would go away if we moved more of the dispatch into C.

R/type2.R
lionel-: If we don't add `...` to the generic, it'll be hard to add new arguments once vctrs is more widely used. Shouldn't we add `...` to all generics?

DESCRIPTION
lionel-: I don't see `check_dots_empty()` in the CRAN version.

R/type2.R
hadley: I'm not sure. Maybe it's ok now that we have ellipsis?

Issue #57, opened 2018-08-21
Better error messages


hadley:
`vec_c()`, `vec_rbind()`, and `vec_cbind()` need to create error messages that make it easier to determine the source of the error, even when nested.

e.g. how can we do better here?

``` r
library(vctrs)

vec_c(1, 2, vec_c(3, vec_c(4, "x")))
#> Error: No common type for double and character
```

njtierney:
My first (and probably not very good) thought: Perhaps it could print back some kind of AST of the vector:

```
> lobstr::ast(c(1,2, c(1, c(4, "x"))))
█─c 
├─1 
├─2 
└─█─c 
  ├─1 
  └─█─c 
    ├─4 
    └─"x" {"character provided in a majority numeric vector"}
```

And then provide an error message / cross at the location where there is an error?

That said, this print / error warning wouldn't scale.

hadley:
Hmmm, interesting idea. I think the idea of some tree like display might be useful.

Although I can't quite imagine how the tree would look in cases like this:

```R
df1 <- data.frame(a = 1)
df1$b <- data.frame(b = 2)
df1$b$c <- data.frame(c = 3)
df1$b$c$d <- data.frame(d = 3)

df2 <- data.frame(a = 1)
df2$b <- data.frame(b = 2)
df2$b$c <- data.frame(c = 3)
df2$b$c$d <- data.frame(d = "x")

vctrs::vec_rbind(df1, df2)
#>  Error: No common type for double and factor<5a425> 
```

njtierney:
I think that the trees idea is kinda neat, but I'm not sure how useful it would be in implementation, it would need a pretty special print method, I think. Maybe there's some kind of scope for a reduced print method for `ast` similar to how `tibble` reduces the printing of rows for `data.frame`s.

I'm not sure what the tree would look like in that case either, although I must say that the `vctr` error message is better than base `rbind` and `dplyr::bind_rows`

``` r
df1 <- data.frame(a = 1)
df1$b <- data.frame(b = 2)
df1$b$c <- data.frame(c = 3)
df1$b$c$d <- data.frame(d = 3)

df2 <- data.frame(a = 1)
df2$b <- data.frame(b = 2)
df2$b$c <- data.frame(c = 3)
df2$b$c$d <- data.frame(d = "x")

rbind(df1,df2)
#> Warning: non-unique value when setting 'row.names': '1'
[...]

PR #297, opened 2019-04-23
Include `vec_type_common()` index in argument tags

lionel-:
```r
vec_c(df1, df1, df2)
#> Error: No common type for `list(...)[[2]]$x$y$z` <double> and `list(...)[[3]]$x$y$z` <character>.
```

Laziness is introduced to avoid an expensive generation of tags / manipulation of global buffers (the index would have to be formatted at each iteration). The laziness also makes it easy to enlabel user inputs without cost in the future, if we want to.

Methods that might fail no longer take a `const char*` as `arg` parameters. Instead, they take an object with `fill()` method which will generate the string when needed.

Memory management of the string buffer is done by allocating a RAWSXP with R. If it is too small, it is reallocated and the tag generation is repeated until success (or R throwing if the stack is full). I use low tech `memcpy()` idioms to manipulate the C strings, as suggested [here](https://stackoverflow.com/a/4689766/1725177) and other places. The caller always manages memory, and the callee get a pointer and a remaining size. A negative value should be returned if the buffer is too small. This way there is no truncation and no buffer overflow. This should be safe and I think the code is readable.


hadley:
Could you please provide a bit more of a guide to reviewing this PR?

lionel-:
vctrs.h

- `struct vctrs_arg{}`: Object with `fill()` method. When called, the method should write the argument tag that the object represents to a buffer passed as argument. If there is no more room in the buffer, a negative value must be returned.

  These objects have a parent pointer so they can form a linked list. For instance `list(...)[[1]]$foo$bar` would be represented with a chain of 3 objects.


conditions.c

- `vctrs_arg()`: This takes a `struct vctrs_arg{}` linked list and calls the recursive function `fill_arg_buffer()`. It allocates a buffer in a RAWSXP of size 100, which is grown by a factor of 1.5 each time the `fill()` methods return a negative value. Returns a character vector of size 1 containing the materialised argument tag.

- `fill_arg_buffer()`: Takes a `struct vctrs_arg{}` linked list and calls the `fill()` method on each of those, recursively. Unless an error occurred, it returns the current size written to the buffer so we can track the remaining memory available in the buffer after recursion.


conditions.c

- `new_vctrs_arg()`: Creates a `vctrs_arg` object which is a simple wrapper around a `const char*` argument tag. The `fill()` method just writes that tag to the buffer. This is the most common way to create an argument tag and replaces the previous `const char* arg` parameters.

- `arg_fill()`: `vctrs_arg::fill()` method for objects created with `new_vctrs_arg()`.


type.c

- `new_common_arg()`: Creates a `vctrs_arg` object for `vec_type_common()`. It wraps a pointer to the current index of reduce loop that finds the common type.

- `common_arg_fill()`: `vctrs_arg::fill()` method for objects created with `new_common_arg()`. Writes a tag like `list(...)[[i]]` to the buffer. The index is formatted into a string by R through `Rf_coerceVector()`.

hadley:
Does this close #57?

lionel-:
Not really:

```
vec_c(1, 2, vec_c(3, vec_c(4, "x")))
#> Error: No common type for `list(...)[[1]]` <double> and `list(...)[[2]]` <character>.
```

However I'm not sure it's possible to make this particular case better.

hadley:
I'd say it closes it as much as it possible can be closed 😄

jimhester:
Why can't you use `0` if the buffer is not big enough to fill? Using a typedef of `r_ssize_t` with a signed integral is kind of misleading, as `size_t` is always an unsigned integral.

lionel-:
I'm not sure I understand, `r_ssize_t` is signed.

lionel-:
It's possible an argument tag is `""`, in which case the number of bytes written is 0, this would then infloop-alloc.

lionel-:
`ssize_t` is the standard denomination for a signed size, you'll find it in the POSIX and CPython API for instance (e.g. https://github.com/python/cpython/commit/18e165558b24d29e7e0ca501842b9236589b012a).

jimhester:
Oh right of course, but then why not just use `ssize_t` directly then?

lionel-:
ah it's not in the C99 standard, only in the POSIX API. I think it makes sense to reuse R's definition for long signed sizes.

lionel-:
The `vctrs_arg` objects are now implemented with C-style polymorphism, as described in https://www.python.org/dev/peps/pep-3123/

Two things are not quite right in this PR. First, the error message is confusing when one argument was coerced:

```r
vec_type_common(1, TRUE, "chr")
#> Error: No common type for `list(...)[[2]]` <double> and `list(...)[[3]]` <character>.
```

Second, the argument names are not used when they exist:

```r
vec_type_common(foo = 1, bar = "")
#> Error: No common type for `list(...)[[1]]` <double> and `list(...)[[2]]` <character>.
```

An upcoming PR solves both issues.

lionel-:
Thanks for the reviews!

PR #303, opened 2019-04-25
Fix row-binding of unspecified vectors

lionel-:
Only unnamed unspecified vectors are treated as unspecified.

Introduces `vec_default_type2()` which can be used to decrease method boilerplate.

### DIFF
--- a/R/type-data-frame.R
+++ b/R/type-data-frame.R
@@ -83,7 +83,7 @@ vec_type2.data.frame.data.frame <- function(x, y, ...) df_col_type2(x, y, ...)
 #' @method vec_type2.data.frame default
 #' @export
 vec_type2.data.frame.default <- function(x, y, ..., x_arg = "", y_arg = "") {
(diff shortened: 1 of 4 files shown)


R/type-data-frame.R
```diff
@@ -83,7 +83,7 @@ vec_type2.data.frame.data.frame <- function(x, y, ...) df_col_type2(x, y, ...)
 #' @method vec_type2.data.frame default
 #' @export
 vec_type2.data.frame.default <- function(x, y, ..., x_arg = "", y_arg = "") {
-  stop_incompatible_type(x, y, x_arg = x_arg, y_arg = y_arg)
+  vec_default_type2(x, y, x_arg = x_arg, y_arg = y_arg)
```
hadley: You should make a note to deprecate the `stop_` functions

lionel-: I thought they might still be useful in case unspecified needs to be treated specially by the method (I'm not sure if that will be ever needed though).

PR #305, opened 2019-04-29
Handle names and coercion in argument tags

lionel-:
This PR fixes support for names and coercions in `vec_type_common()` error messages. Ideally these two issues would be implemented in different PRs, unfortunately it would be hard to untangle the development history.

#### Coercion

The error message is currently confusing when an argument was coerced along the way:

```r
vec_type_common(1, TRUE, "chr")
#> Error: No common type for `list(...)[[2]]` <double> and `list(...)[[3]]` <character>.
```

One way to fix this would be to describe the coercion pattern:

```r
#> Error: No common type for `list(...)[[2]]` <double> coerced from <logical> and `list(...)[[3]]` <character>.
```

However this would be hard to implement with recursive data structures like data frames. A simpler solution is to keep track of the argument corresponding to the current type:

```r
#> Error: No common type for `list(...)[[1]]` <double> and `list(...)[[3]]` <character>.
```

To support this, `vec_type2()` now takes an output parameter `int* left` that indicates if the common type comes from the LHS or the RHS. It returns 1 for the LHS, 0 for the RHS, and -1 when both types are the same. The last value means that the LHS is taken as the default when `left` is coerced to a boolean.

#### Names

This PR also adds support for names:

```r
vec_type_common(foo = 1, bar = "")
Error: No common type for `foo` <double> and `bar` <character>.

vec_type_common(foo = df1, bar = df2)
#> Error: No common type for `foo$x$y$z` <double> and `bar$x$y$z` <character>.
```

The implementation is a bit more complicated than I expected because of the need to keep track of counters within splice boxes. When I first implemented `vec_type_common()` with `rlang::dots_values()` to avoid copies on splicing, I did not realise this would introduce such complexity later on. Not sure it was worth it in the end:

```r
many <- rep(list(1.5, 1:10), 1000)
bench::mark(
  dots_values = vec_type_common(!!!many),
  dots_list = vec_type_common_splice(!!!many)
)[1:6]
#>   expression       min     mean   median      max `itr/sec`
#>   <chr>       <bch:tm> <bch:tm> <bch:tm> <bch:tm>     <dbl>
#> 1 dots_values    224µs    280µs    229µs 732.52µs     3572.
#> 2 dots_list      276µs    352µs    299µs   1.45ms     2842.
```

### DIFF
--- a/src/arg.c
+++ b/src/arg.c
@@ -113,7 +113,6 @@ static r_ssize_t counter_arg_fill(struct vctrs_arg* self, char* buf, r_ssize_t r
 
 struct vctrs_arg_counter new_counter_arg(struct vctrs_arg* parent,
                                          R_len_t* i,
-                                         int offset,
                                          SEXP* names,
                                          R_len_t* names_i) {
   struct vctrs_arg_counter counter = {
@@ -122,7 +121,6 @@ struct vctrs_arg_counter new_counter_arg(struct vctrs_arg* parent,
       .fill = &counter_arg_fill
     },
     .i = i,
-    .offset = offset,
     .names = names,
     .names_i = names_i
   };
@@ -132,10 +130,10 @@ struct vctrs_arg_counter new_counter_arg(struct vctrs_arg* parent,
 
 static r_ssize_t counter_arg_fill(struct vctrs_arg* self, char* buf, r_ssize_t remaining) {
   struct vctrs_arg_counter* counter = (struct vctrs_arg_counter*) self;
-  R_len_t i = *counter->i + counter->offset;
+  R_len_t i = *counter->i;
 
   SEXP names = *counter->names;
-  R_len_t names_i = *counter->names_i + counter->offset;
+  R_len_t names_i = *counter->names_i;
 
   int len;
   if (r_has_name_at(names, names_i)) {
--- a/src/arg.c
+++ b/src/arg.c
@@ -113,13 +113,15 @@ static r_ssize_t counter_arg_fill(struct vctrs_arg* self, char* buf, r_ssize_t r
 
 struct vctrs_arg_counter new_counter_arg(struct vctrs_arg* parent,
                                          R_len_t* i,
+                                         int offset,
                                          SEXP* names) {
   struct vctrs_arg_counter counter = {
     .iface = {
       .parent = parent,
       .fill = &counter_arg_fill
     },
-    .i = (void*) i,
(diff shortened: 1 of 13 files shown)


lionel-:
I'm not sure I understand what you call a new memory management system. The counter objects maintained on the stack rather than on the heap? The C-style dynamic polymorphism of `vctrs_arg` structs?

lionel-:
We could talk about this at the work week @jimhester?

jimhester:
If the purpose of this is just for error messages could we just use a static character buffer? R already has a hard limit on the total error message size, so we could do the same thing, which would avoid the need to dynamically allocate memory.

lionel-:
I'm planning to use auto-labelling of arguments for e.g. `foo[i] <- value`. If we don't have laziness, this will be too slow because of the deparsing cost. Also users tend to bind together very large lists. This tag generation will be used everywhere, in our variants of `rbind()`, `c()`, [<-`, etc, so I'd prefer avoiding writing the tag buffer unless we're in the error path.

I don't think the `vctrs_arg` are particularly complex to develop or maintain. They are just C objects following a standard idiom of polymorphism. It's very similar to C++ runtime polymorphism, only with a more cumbersome syntax.

The main source of complexity in this PR is not due to the `vctrs_arg` or the character buffers, it's because we're reducing over a list and the potential splice boxes it might contain, which makes it harder to keep track of the relevant counters. If it becomes a maintenance issue, we can splice the arguments into a flat list to remove this complexity. For the moment, I recommend keeping the approach as is because it's already implemented and makes the fundamental vctrs primitive `vec_type_common()` a little faster when large lists of objects are spliced. Maybe we don't have to touch it again, and if we do, the counters might not get in the way.

lionel-:
After thinking about it this weekend, I think the polymorphism of `vctrs_arg` might be overkill. When I have time I'll change it to a simple struct of a `fill` method taking a `void*` data argument. This will simplify the call sites (the struct can be passed by value, and if passed by pointer there's no longer need to cast to a pointer of the parent type). Also a function pointer taking a data argument should be a familiar idiom to any C programmer.

As for the complexity in this PR, it is caused by a combination of two things:

- The reduction is ragged by the splice boxes.
- We need to keep track of the state of counters inside splice boxes because of the laziness of argument tag.

We can remove the complexity either by writing the tags to a buffer eagerly, or by removing the raggedness. I think the latter should be the first solution to consider.

PR #309, opened 2019-05-01
Move names repair from tibble to vctrs

lionel-:
* New `vec_names()` function with `repair` argument. It takes names on the observation dimension rather than from `base::names()`.

* `vec_as_names()` takes a vector of names and repairs it. This should be useful for tibble which repairs column names rather than observation names: `vec_as_names(names(df), repair = repair)`. For simplicity, it doesn't take a `n` parameter and expects a fully formed character vector.

* `vec_repair_names()` is a shortcut for `set_names(x, vec_names(x))`.

* The original `vec_names()` function has been renamed to `vec_bare_names()`.


lionel-:
When we add support for row names of data frames, the default names repair should probably be upgraded from minimal to unique because data frames can't have duplicated row names. This would make the UI slightly strange, but it's probably better than taking `NULL` and renouncing to self-documentation?

lionel-:
You're probably right that `vec_names()` should be a straight getter. But it should always return minimal names right?

About the repair output, I find it hard to read. Should we add some whitespace?

```r
vec_names(letters, repair = "unique")
#> New names:
#> * `` -> ...1
#> * `` -> ...2
#> * `` -> ...3
#> * `` -> ...4
#> * `` -> ...5
#> * ...
#>  [1] "...1"  "...2"  "...3"  "...4"  "...5"  "...6"  "...7"  "...8"  "...9" 
#> [10] "...10" "...11" "...12" "...13" "...14" "...15" "...16" "...17" "...18"
#> [19] "...19" "...20" "...21" "...22" "...23" "...24" "...25" "...26"

=>

#> New names:
#>
#> * `` -> ...1
#> * `` -> ...2
#> * `` -> ...3
#> * `` -> ...4
#> * `` -> ...5
#> * ...
#>
#>  [1] "...1"  "...2"  "...3"  "...4"  "...5"  "...6"  "...7"  "...8"  "...9"
#> [10] "...10" "...11" "...12" "...13" "...14" "...15" "...16" "...17" "...18"
#> [19] "...19" "...20" "...21" "...22" "...23" "...24" "...25" "...26"
```

krlmlr:
Yes, always minimal names.

I don't have a strong opinion about the empty line.

NAMESPACE
```diff
@@ -431,6 +431,7 @@ export(vec_arith.difftime)
 export(vec_arith.logical)
 export(vec_arith.numeric)
 export(vec_arith_base)
+export(vec_as_names)
```
jennybc: Another possible name for this is `vector_make_names()`, to make the connection to `base::make.names()` more clear.

I'm not saying I like that better, just raising it. No matter what, it's a connection we should make in docs.

NAMESPACE
```diff
@@ -470,6 +471,7 @@ export(vec_math)
 export(vec_math_base)
 export(vec_na)
 export(vec_na_along)
+export(vec_names)
```
jennybc: So this is the heir to `rlang::names2()`, right?

R/names.R
```diff
@@ -0,0 +1,405 @@
...
+#' vector. It has the following differences:
+#'
+#' * It always returns __minimal names__. Whereas `names()` returns
+#'   `NULL` when a vector doesn't have names, `vec_names()` returns a
+#'   character vector of empty strings by default. `NA` names are also
+#'   converted to `""`. You can control how names are repaired with
+#'   the `repair` argument.
+#'
+#' * With arrays it returns the row names. With data frames it returns
+#'   an empty character vector as long as the number of rows.
```
jennybc: This data frame behaviour is very unexpected to me.

R/names.R
```diff
@@ -0,0 +1,405 @@
...
+#'     names to make them unique.
+#'
+#'   * Universal names are unique and syntactic, meaning that you can
+#'     safely use the names as variables without causing a syntax
+#'     error.
+#'
+#'   See [vec_repair_names()] for a complete overview of names repair.
+#' @param quiet By default, the user is informed of any renaming
+#'   caused by names repairs. This only concerns unique and universal
+#'   repairing. Set `quiet` to `TRUE` to silence the messages.
```
jennybc: Propose:

By default, the user is informed of any renaming caused by making names unique or universal.

R/names.R
```diff
@@ -0,0 +1,405 @@
...
+#'
+#' * `minimal` names exist. The `names` attribute is not `NULL`. The
+#'   name of an unnamed element is `""` and never `NA`. For instance,
+#'   [vec_names()] always returns minimal names and data frames
+#'   created by the tibble package have names that are, at least,
+#'   `minimal`.
+#'
+#' * `unique` names are `minimal`, have no duplicates, and can be used
+#'   where a variable name is expected.  Empty names, and `...` or
+#'   `..` followed by a sequence of digits are banned.  - All columns
```
jennybc: Reflowing problem here goofed up the bullet list.

R/names.R
```diff
@@ -0,0 +1,405 @@
...
+#'   can be accessed by name via `df[["name"]]` and `` df$`name` ``
+#'   and ``with(df, `name`)``.
+#'
+#' * `universal` names are `unique` and syntactic (see Details for
+#'   more).  - Names work everywhere, without quoting: `df$name` and
```
jennybc: Ditto

NAMESPACE
lionel-: Correct, though `names2()` is lower level and works with base R definition of vector (e.g. `colnames()` of data frames).

R/names.R
lionel-: We discussed with @hadley about returning the row names instead.

jennybc:
To elaborate a bit more on my surprise that `vec_names()` and `vec_repair_names()` address the row names of a data frame ...

If you digest all this with a base R mentality, this is surprising because a data frame is a list and you'd expect to be dealing with the names of that list.

If you filter it through a tidyverse mentality, you're used to living as if data frame *don't even have* row names.

lionel-:
If we want to be consistent with, on the one hand, base R data frames which have row names, and on the other hand, the vctrs definition of a vector, returning the row names makes sense.

Also record types are data frames that hide their internals, just like factors are integers that hide their internals. It'd make sense for record types to have names just like any other vector, and then implementing with row names is natural. (Well almost natural, it'd require using a different class than data.frame because the latter can't have duplicated row names.)

jennybc:
I'm not necessarily resisting, but rather giving a preview of the "consumer education" that will be necessary for people to connect this to their existing mental models and feel like it makes sense.

lionel-:
I understand and I think you're right, we went through the same thinking process. The tension between data frames as rectangular lists and data frames as vectors of tuples hasn't been resolved yet, and perhaps never will be.

lionel-:
Perhaps we need a different name than "names" to make the distinction easier to learn and remember, just like we have "size" versus "length". I feel like this would make vctrs more complicated though, because of the added vocabulary.

jennybc:
> The tension between data frames as rectangular lists and data frames as vectors of tuples

This statement really nails what I'm getting at ... how does this tension get resolved in the name domain? 🤔

lionel-:
I've been wondering whether df-as-rectangular-lists are in fact transposed df-as-tuple-vectors.

jennybc:
> I've been wondering whether df-as-rectangular-lists are in fact transposed df-as-tuple-vectors.

You mean, like, "In the beginning, there were tuple-vectors. And on the second day we transposed them and called them data frames". Are you talking about which data structure is somehow the more basic?

lionel-:
I don't think one is more basic than the other, but instead they are two views of the same data structure. Unlike matrices which have their actual values transposed, the transposability of data frames is a logical view, a flippable boolean that changes how the structure is indexed but not how it is stored. Since `t()` is a generic, we could actually implement a data frame class this way.

lionel-:
Row-major data frames are not recursive, but column-major data frames are?

R/names.R
hadley: I wonder if we need to (also) expose the name repair as a function that takes a character vector.

hadley: @jennybc I think the invariant here is that `vec_size(vec_names(x))` needs to equal `vec_size(x)`. But we need to figure out how to resolve that with the naming we need for data frames etc.

hadley:
`vec_obs_names()` could be a possible name option.

lionel-:
Then should `vec_size()` be `vec_obs_length()`?

Does having a qualifier implies other types of length and names are possible in vctrs?

hadley:
No, we can't consider a general renaming at this point.

lionel-:
In this case I think `vec_names()` is still the best option, with the understanding that all vctrs functions operato along the observation dimension. Since they are developer facing, `vec_names()` and `vec_size()` won't be reexported in funs right?

hadley:
Agreed.

krlmlr:
Regarding the formula syntax for the repair strategy: is there a reason for not supporting it here?

lionel-:
> But we still need some way to apply this renaming to the columns of a data frame, particularly since this is the motivating case.

This should be covered by `vec_as_names()` which takes a character vector and a `repair` parameter.

> Regarding the formula syntax for the repair strategy: is there a reason for not supporting it here?

The timing/place of application of that function doesn't seem so clear to me, so I'd rather wait before including this. I think it's better to wait and see how custom name repair is used in practice by tibble users.

jennybc:
> Regarding the formula syntax for the repair strategy: is there a reason for not supporting it here?
>> The timing/place of application of that function doesn't seem so clear to me, so I'd rather wait before including this. I think it's better to wait and see how custom name repair is used in practice by tibble users.

Two observations:

  * Re: the flexibility of `.name_repair` in tibble and its applicability here. There is an intermediate option which is to accept a function as the `repair` parameter here, but not the anonymous formula syntax.
  * It's hard to predict, but I think the need to enact some sort of custom name repair might be just as likely to happen internally, in packages dealing in tibbles, as it is for tibble end-users.

lionel-:
Do you recommend supporting repair functions right away?

hadley:
I think we need to match the tibble interface.

lionel-:
Is it correct that the `transform` parameter of the compat file and the support for functions in the tibble interface have different semantics? IIUC the former is applied after stripping the names from repair suffixes, while the latter applies the function on the original unaltered vector of names. Is the former only an implementation detail for transforming to universal names or was it meant to be public?

I guess the reason of my confusion is that you were talking about repair functions from the tibble UI, which is not implemented in the compat file, and I thought this was about the `transform` parameter, which is implemented in the compat file?

jennybc:
I don't totally follow your question @lionel-. But will try to move the conversation along.

`transform` here:

https://github.com/tidyverse/tibble/blob/815c9027b82ceab58b41aa45c015315a3cac778b/R/compat-name-repair.R#L35

is only relevant to names that are getting *at least* the "unique" level of repair (but it's still the identity). So it's only really really relevant to names that are getting the "universal" level of repair, where `transform = `make_syntactic`.

> the support for functions in the tibble interface

If a user specifies a function to `.name_repair`, they are completely on their own. All we do is pass "minimal" names to their function and make sure whatever they return is still "minimal".

lionel-:
Got it, thanks. `transform` is an internal implementation thing, and support for `repair` functions is separate.

NAMESPACE
hadley: `names2()` still has it's place - it has the invariant that `length(names2(x))` equals `length(x)`, where `vec_size(vec_names(x))` equals `vec_size(x)` (which is very different for data frames)

R/names.R
```diff
@@ -0,0 +1,405 @@
+#' Retrieve and repair names
```
hadley: Now that we've implemented this here, it doesn't feel quite right to me — name repair is most important for data frames, and this, the primary function, doesn't help at all. Maybe it would be better to call this `vec_rec_names()` or `vec_col_names()` or similar?

R/names.R
```diff
@@ -0,0 +1,405 @@
...
+#' # Universal repairing fixes any non-syntactic name:
+#' x <- c("_foo" = 1, "+" = 2)
+#' vec_names(x, repair = "universal")
+#'
+#' @export
+vec_names <- function(x,
+                      ...,
+                      repair = c("minimal", "unique", "universal"),
+                      quiet = FALSE) {
+  if (...length()) {
```
hadley: Doesn't this require 3.5? Or do you have a backport elsewhere?

R/names.R
```diff
@@ -0,0 +1,405 @@
...
+#'   `NULL` when a vector doesn't have names, `vec_names()` returns a
+#'   character vector of empty strings by default. `NA` names are also
+#'   converted to `""`. You can control how names are repaired with
+#'   the `repair` argument.
+#'
+#' * With arrays it returns the row names. With data frames it returns
+#'   an empty character vector as long as the number of rows.
+#'
+#' @param x A vector.
+#' @param ... These dots are for future extensions and must be empty.
```
hadley: Why are we using `...` here?

NAMESPACE
jennybc: 💡 is it called `names2()` just because it's a second take on `names()`? Probably.

But I just realized you can say that, for a data frame, vec_names() gives names along dimension 1 and `names2()` gives names along dimension 2.

R/names.R
lionel-: Maybe we should just have `vec_as_names()` for the time being?

R/names.R
lionel-: There's a backport but I've been standardising towards `!missing(...)` lately.

R/names.R
lionel-: So that `repair` must be named.

R/names.R
lionel-: "caused by repairing the names"? Since the `repair` function might be supplied by the user.

NAMESPACE
lionel-: I like `vec_as_names()` because it is similar to `vec_as_index()`. Good point about the connection, I have added a mention of `make.names()` in the doc.

NAMESPACE
lionel-: Though that doesn't work with matrices and simple vectors don't really have 2 dimensions (and if taken as a column vector, `names2()` returns the names of dim 1 rather than 2).

R/names.R
lionel-: I have unexported `vec_names()`.

lionel-:
`vec_as_names()` now supports user-supplied functions.

R/names.R
```diff
@@ -0,0 +1,429 @@
...
+  )
+}
+vec_repair_names <- function(x,
+                             repair = c("minimal", "unique", "universal"),
+                             ...,
+                             quiet = FALSE) {
+  set_bare_names(x, vec_names(x, ..., repair = repair, quiet = quiet))
+}
+
+minimal_names <- function(x) {
```
hadley: Should we make these `names_minimal` etc?

R/c.R
```diff
@@ -54,13 +54,14 @@ vec_c <- function(..., .ptype = NULL) {
 
     x <- vec_cast(args[[i]], to = ptype)
 
-    names[pos:(pos + n - 1)] <- outer_names(names(args)[[i]], vec_names(args[[i]]), length(x))
+    names[pos:(pos + n - 1)] <- outer_names(names(args)[[i]], vec_bare_names(args[[i]]), length(x))
```
hadley: I think you should leave these as `vec_names()`

R/names.R
lionel-: Then it would be strange to have `as_minimal_names()`, and `as_names_minimal()` looks strange as well? These are only used internally and in tests though, so perhaps not worth worrying about.

lionel-:
Thanks for your inputs!

krlmlr:
Is there a specific reason why `"check_unique"` isn't a possible value for the `repair` argument?

lionel-:
That it wasn't part of the compat file in the tibble repo would be a specific reason.

krlmlr:
Would you review a PR?

lionel-:
A PR would be great.

PR #310, opened 2019-05-01
Implement unique names repair in C

lionel-:
Expects #309.

* Export dictionary declarations to `dictionary.h` so we can detect duplicates in other files.

* New `sexp_cow` structure that wraps a `SEXP` and a `PROTECT_INDEX`. With this, an object protected from up high can be duplicated only if necessary. Used in names repair to avoid duplicating names vector if no repairs are needed.

* `repair = "unique"` is now implemented in C.

```r
old <- tibble:::unique_names
new <- vctrs::as_unique_names

bench::mark(
  old = old(letters),
  new = new(letters)
)[1:6]
#> 1 old        155.94µs 183.04µs 174.42µs  405.1µs     5463.
#> 2 new          4.71µs   5.57µs   5.14µs   95.8µs   179546.

dups <- rep_along(letters, "")
bench::mark(
  old = old(dups, quiet = TRUE),
  new = new(dups, quiet = TRUE)
)[1:6]
#> 1 old        120.42µs 141.11µs 134.38µs    371µs     7087.
#> 2 new          6.64µs   8.16µs   7.24µs     51µs   122471.


long <- as.character(seq(1:1000))

bench::mark(
  old = old(long),
  new = new(long)
)[1:6]
#> 1 old          1.03ms   1.25ms   1.19ms   1.99ms      802.
#> 2 new         95.19µs 106.35µs  98.41µs  291.5µs     9403.


long_dups <- rep_len("", 1000)

bench::mark(
  old = old(long_dups, quiet = TRUE),
  new = new(long_dups, quiet = TRUE)
)[1:6]
#> 1 old          1.22ms   1.34ms   1.31ms   2.11ms      748.
#> 2 new        219.35µs 234.38µs 220.91µs 510.17µs     4267.
```

cc @jennybc @krlmlr

### DIFF
--- a/R/names.R
+++ b/R/names.R
@@ -229,13 +229,7 @@ minimal_names <- function(x) {
 }
 
 vec_names <- function(x) {
-  if (vec_dims(x) == 1) {
-    names(x)
-  } else if (is.data.frame(x)) {
-    NULL
-  } else {
-    rownames(x)
-  }
+  .Call(vctrs_names, x)
 }
 `vec_names<-` <- function(x, value) {
   if (vec_dims(x) == 1) {
--- a/R/names.R
+++ b/R/names.R
@@ -228,7 +228,7 @@ minimal_names <- function(x) {
   .Call(vctrs_minimal_names, x)
 }
 unique_names <- function(x, quiet = FALSE) {
-  .Call(vctrs_unique_names, x)
+  .Call(vctrs_unique_names, x, quiet)
 }
 
 vec_names <- function(x) {
@@ -250,13 +250,7 @@ as_minimal_names <- function(names) {
   .Call(vctrs_as_minimal_names, names)
 }
 as_unique_names <- function(names, quiet = FALSE) {
-  out <- .Call(vctrs_as_unique_names, names)
-
-  if (!quiet) {
-    describe_repair(names, out)
-  }
-
-  out
+  .Call(vctrs_as_unique_names, names, quiet)
 }
 as_universal_names <- function(names, quiet = FALSE) {
   as_unique_names_impl(names, quiet, TRUE)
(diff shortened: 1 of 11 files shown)


src/names.c
```diff
@@ -0,0 +1,379 @@
...
+      elt = strings_empty;
+      SET_STRING_ELT(names, i, elt);
+      SET_STRING_ELT(d.x, i, elt);
+    }
+
+    // Strip `...n` suffixes
+    const char* nm = CHAR(elt);
+    int pos = suffix_pos(nm);
+    if (pos >= 0) {
+      R_CheckStack2(pos + 1);
```
hadley: What does this do?

lionel-: This checks there is sufficient space to allocate on the stack.

src/names.c
```diff
@@ -0,0 +1,379 @@
...
+    UNPROTECT(1);
+    return R_NilValue;
+  }
+
+  SEXP out = VECTOR_ELT(dimnames, 0);
+  UNPROTECT(1);
+  return out;
+}
+
+static struct sexp_cow as_minimal_names(struct sexp_cow cow_names) {
```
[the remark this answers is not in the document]
lionel-: I thought we should avoid allocs and copies if we can, this will be called once per inputs, which can be many when assembling data frames from vectors with e.g. purrr.

The structure is always the same: first pass to check if a modif is needed, and then second pass with the SEXP copied.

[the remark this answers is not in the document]
lionel-: Sorry, on second inspection the cow structure was completely silly. I have removed it. Is this now better?

src/names.c
hadley: Oh `buf[pos + 1]` is doing the allocation. Isn't that a C99 idiom? I thought we couldn't use that in R code.

R/names.R
```diff
@@ -213,7 +213,7 @@ vec_names <- function(x,
 
   switch(arg_match(repair),
     minimal = minimal_names(x),
-    unique = as_unique_names(minimal_names(x), quiet = quiet),
+    unique = unique_names(x, quiet = quiet),
```
hadley: Shouldn't some corresponding R code now disappear?

lionel-:
> Isn't that a C99 idiom? I thought we couldn't use that in R code.

R depends on C99, so we can use all the C99 stuff. We also use the new struct initialisation syntax for instance.

R/names.R
lionel-: This code is used for `repair = "universal"`, which remains implemented in R.

Issue #308, opened 2019-04-29
`vec_compare(<type>, NULL)` throws an error


DavisVaughan:
Should this error?

``` r
vctrs::vec_compare(1:5, NULL)
#> Error: `x` and `y` must have same types and lengths
```

I think `integer(0)` might be an appropriate result, since that is what this does:

``` r
vctrs::vec_compare(NULL, NULL)
#> integer(0)
```

This error means that comparison operations between vctrs and `NULL` error, unlike base R, which I think gives an intuitive result:

``` r
vctrs:::new_hidden(1) > NULL
#> Error: `x` and `y` must have same types and lengths

1 > NULL
#> logical(0)
```

If it returned `integer(0)`, then the comparison operator would work since this operation powers `>.vctrs_vctr`:

``` r
integer(0) > 0
#> logical(0)
```

However, returning `integer(0)` would break an invariant of `vec_size_common(x, y) == vec_size(vec_compare(x, y))` so maybe it should be an error after all.

The current error is here at the C level. Since `vec_compare()` calls `vec_recycle_common()` and `vec_cast_common()` before this, I don't see how this error could ever be reached except in this 1 case with `NULL`. Maybe it could be improved some if that was the case?
https://github.com/r-lib/vctrs/blob/master/src/compare.c#L125

lionel-:
> However, returning integer(0) would break an invariant of vec_size_common(x, y) == vec_size(vec_compare(x, y)) so maybe it should be an error after all.

I think the common size with a `NULL` should be zero. However I'm not familiar with this part of vctrs yet so that might not be practical.

DavisVaughan:
I agree! Here are my thoughts:

``` r
library(vctrs)

# i think this should be 0
vec_size_common(NULL)
#> NULL

# this is NULL
vec_size_common()
#> NULL

# i think this is 0
vec_size_common(1, NULL)
#> [1] 1
```

I think all you'd have to do is remove the `compact()` from `vec_size_common()`

hadley:
NULL should be treated as the absence of an input, definitely not as a zero length vector. I think we need a fallback for when all inputs are NULL, not to change existing behaviour otherwise.

DavisVaughan:
Should `vec_size_common(NULL) == vec_size(NULL)`?

I imagine those two cases should be consistent?

In `vec_size()` there is this note, and from that I got that both should then equal `0`:

```
#' vec_size(NULL) == 0
[...]

PR #311, opened 2019-05-02
`vec_size()` identity and `NULL` handling

DavisVaughan:
This PR is in regard to the comments about `vec_size()` and `vec_size_common()` in #308.

In particular, the following are now all equivalent and equal to `0` in light of the concept that `NULL` should be treated as the absence of input, ~and that the identity of `vec_size()` should be `0` since that is the value you can add to other sizes without anything changing.~

* ~`vec_size()`~ updated to now still be an error
* `vec_size(NULL)`
* `vec_size_common()`
* `vec_size_common(NULL)`
* `vec_size_common(NULL, NULL)`

_The last 3 are new behavior that previously returned `NULL`_

With `NULL` being the absence of input, the following still holds:

* `vec_size_common(1:5, NULL) == 5`

The empty-punning behavior of recycling to size `0` still holds as well:

* `vec_size_common(1:5, integer()) == 0`


lionel-:
I don't think we need to allow `vec_size()` without input because it's not reducible, especially if it means having to deal with the missing arg.

lionel-:
>  the identity of vec_size() should be 0 since that is the value you can add to other sizes without anything changing.

0 is the identity of addition, and sizes can be summed, but I don't think this means 0 is the identity of sizes in general. Quantities don't have identities by themselves, it's in relation to an operation. And `vec_size_common()` is not a monoid and doesn't have an identity.

DavisVaughan:
That makes sense, thanks for clarifying. I think I was trying to use the language without fully understanding the concepts.

lionel-:
It's possible that I am the source of your confusion because I was confused myself this morning, see the edits in my reply to your post: https://github.com/r-lib/vctrs/issues/308#issuecomment-488580439

Well, actually I think I was still confused. `vec_size_common()` _could_ be a reducing operation. We could define a `vec_size2()` that returns either the LHS or the RHS depending on its length:

* `vec_size2(1:2, 1:3)` is an error
* `vec_size2(1, 1:3)` is `1:3`
* `vec_size2(1:2, int())` is `int()`

Then we just need an identity, which could be `NULL`. But this `vec_size2()` operation is very different from `vec_size()`, and may even need another name to avoid confusion. I think what threw us off is that `vec_type()`, `vec_type2()`, `vec_type_common()` all return the same thing, which isn't the case for size.

Actually there's a difference in return value between `vec_type_common()` and `vec_type()` on the one hand, and `vec_type2()` on the other hand: the latter doesn't return a finalised type. So we really have:

```
vec_type_common(x) == vec_type_finalise(reduce(x, vec_type2))
```

Similarly we could have:

```
vec_size_common(x) == vec_size(reduce(x, vec_size2))
```

DavisVaughan:
>  I think what threw us off is that vec_type(), vec_type2(), vec_type_common() all return the same thing, which isn't the case for size.

This in particular is something I have come back to a number of times in rray, where I also have `rray_dim_common()` which has the same issue as `vec_size_common()`. But yes, for `vec_size()` and `vec_size_common()`, the inputs are multiple objects to take the size of. For `vec_size2()` the inputs are the actual sizes of 2 objects, and the output is the common size. That is the only way for that operation to be reducible.

lionel-:
> For vec_size2() the inputs are the actual sizes of 2 objects, and the output is the common size. That is the only way for that operation to be reducible.

For this operation to be reducible, I think the inputs should be actual vectors, and the output one of the two inputs. If you return a size, the operation is not reducible because the domain and the range of the operation are not the same. Do the `vec_size2()` examples above make sense?

lionel-:
oops I see what you mean, you'd do `reduce(map(x, vec_size), vec_size2)`, then the inputs and outputs are sizes?

DavisVaughan:
That is what happens right now in `vec_size_common()`, right?

lionel-:
Maybe `vec_size_common()` needs a `.empty` or `.absent` parameter that defines what to return on absence of inputs. If not supplied, this is an error. `vec_size(NULL)` would always be an error too.

DavisVaughan:
Note that `vec_size_common()` is also used in `vec_cbind()` and `vec_recycle_common()` so that might need to become arguments to those as well. Something about that doesn't feel quite right.

lionel-:
> Note that vec_size_common() is also used in vec_cbind() and vec_recycle_common() so that might need to become arguments to those as well.

They would just specify the behaviour they need (the current behaviour) without taking a new argument.

hadley:
Shouldn't it return 1 since that's the identity under the vctrs recycling rules?

Or should it just return an error, forcing the user to deal with the problem of no inputs?

DavisVaughan:
Does it matter that that would break this invariant in `?vec_size`?

```
vec_size(vec_c(x, y)) == vec_size(x) + vec_size(y)
```

```
vec_size(vec_c(NULL, 2)) == vec_size(NULL) + vec_size(2) == 1
```

hadley:
Due to its nature, I don't think `NULL` participates in size related invariants in that way.

DavisVaughan:
I made the change locally to let `vec_size(NULL) == 1` and `vec_size_common() == vec_size_common(NULL) == 1`. It broke a few things. Specifically, any time you do `map_int(args, vec_size)` to get a list of sizes that can be used to construct an output container, the output size will be wrong. So the following break after the change:

- `vec_rbind(data_frame(x = 1), NULL)` errors
- `vec_c(NULL, 1L)` errors
- `vec_cbind(NULL)` gives a 0 col, 1 row empty data frame

I suppose they can be somewhat solved by wrapping the `list2()` calls in each of those in `compact()` but I still find it strange that `vec_cbind()` by itself would return a 1 row df because of this.

lionel-:
I think the caller of `vec_size_common()` has to decide the return value on absence of inputs, unless we find a compelling invariant for the size of `NULL`, which it appears we can't. `vec_size(NULL)` should be an error.

DavisVaughan:
@lionel- I updated this PR based on how I'm interpreting your assumptions. To keep previous code working:

- `vec_cbind()` and `vec_recycle_common()` both state that `.empty = 0L`
- `vec_rbind()` and `vec_c()` both now remove `NULL` values after the `ptype` is checked, but before the sizes are computed

The one test that still fails is this, which I'm not sure how to handle.

```
> vec_compare(NULL, NULL)
Error: The size of `NULL` is not defined.
```

R/size.R
```diff
@@ -48,14 +46,20 @@ vec_size <- function(x) {
 
 #' @export
 #' @rdname vec_size
-vec_size_common <- function(..., .size = NULL) {
+vec_size_common <- function(..., .size = NULL, .empty) {
```
hadley: This either violates https://principles.tidyverse.org/args-required.html or https://principles.tidyverse.org/args-data-details.html — I think the easiest fix would be `.empty = stop_empty()` 

```R
stop_empty <- function() {
  abort("`...` is empty, and no `.empty` value was supplied.")
}
```

DavisVaughan: This brings up another point. Should `stop_empty()` be exported? I've always felt that if a function is the default of an argument, a user should be able to call it. Which feels like part of https://principles.tidyverse.org/args-magical-defaults.html

hadley: FWIW I don't think `.empty = NULL` is really a valid value here since it violates type stability.

I have a very rough start on exporting default helpers at https://principles.tidyverse.org/args-defaults-short.html#helper-function. I'll have to think about how to crosslink different problems that have the same solutions.

lionel-: Perhaps it doesn't violate type stability because the type can be inferred from the arguments statically. In statical typing terms, the type of the function is parameterised in its return value (the return value becomes an option type with the option depending on the declared type parameter).

So I think it is type-stable as long as we pass a constant.

lionel-: (or an expression whose type is constant, e.g. `new_foo()` or `dbl()`)

DavisVaughan: For now I have:
- Set `.empty = NULL` as the default, which errors if no `...` are provided
- If provided, validate that `.empty` is a length 1 integer with `vec_assert()`

lionel-: I like that. Despite my notes above, there's no reason to allow the return type to be more complex here, since the caller can set a `na_int` sentinel if needed.

My only concern is the name, it seems `.empty` might suggest this is the size used for empty vectors. Should it be `.absent` instead?

DavisVaughan: Updated to `.absent`

DavisVaughan:
This conversation also makes me wonder what `vec_dim(NULL)` and `vec_dims(NULL)` should return.

My guess is a redefinition as:

```r
vec_dim <- function(x) {
  dim(x) %||% vec_size(x)
}

vec_dims <- function(x) {
  d <- vec_dim(x)
  vec_size(d)
}
```

So:

```r
> vec_dim(NULL)
Error: The size of `NULL` is not defined.
> vec_dims(NULL)
Error: The size of `NULL` is not defined.
```

lionel-:
@DavisVaughan  `vec_compare(NULL, NULL)` fails because the inputs are passed to `vec_size_common()`. Since the common size is then passed to `vec_recycle()`, that's a case where it makes sense to pass the recycling identity which is 1. This seems to confirm that parameterising the return value on absence of inputs is the right approach.

I'm going to finish the PR and merge it before I rewrite `vec_size_common()` in C, as part of the rewriting of `vec_cbind()`. This way we'll avoid painful merge conflicts.

lionel-:
I fixed the `vec_compare()` issue by making it an error to supply scalar objects. This is consistent with absence of inputs which is also an error.

@hadley Is this now good to go?

src/compare.c
```diff
@@ -122,11 +122,13 @@ int compare_scalar(SEXP x, R_len_t i, SEXP y, R_len_t j, bool na_equal) {
 // R interface -----------------------------------------------------------------
 
 SEXP vctrs_compare(SEXP x, SEXP y, SEXP na_compare_) {
-  if (TYPEOF(x) != TYPEOF(y) || vec_size(x) != vec_size(y))
+  R_len_t n = vec_size(x);
+
+  if (TYPEOF(x) != TYPEOF(y) || n != vec_size(y)) {
     Rf_errorcall(R_NilValue, "`x` and `y` must have same types and lengths");
```
DavisVaughan: Since we recycle to common size and cast to common type at the R level, is this error even reachable? Maybe it's good practice to keep it in, just something I noticed.

lionel-: I was thinking we should probably remove these checks.

PR #312, opened 2019-05-02
Introduce coercible-cast for assigning into vectors

### LINKED ISSUE #140: Type stable [<- for base classes
i.e. ensures that the type of `x` does not change. 

Needed for https://github.com/tidyverse/purrr/pull/579

lionel-:
Makes `[<-`, `[[<-`, and `vec_slice<-` type-stable and closes #140.

Type-stability is ensured by casting after checking that the LHS and RHS types are coercible. If we just cast the RHS to the LHS, we have:

```r
x <- 1:2
vec_slice(x, 1) <- "10"
x
#> [1] 10  2

x <- c("foo", "bar")
vec_slice(x, 1) <- 10
x
#> [1] "10"  "bar"
```

With coercible-cast, we have:

```r
x <- 1:2
vec_slice(x, 1) <- "10"
#> Error: No common type for `x` <integer> and `value` <character>.

x <- c("foo", "bar")
vec_slice(x, 1) <- 10
#> Error: No common type for `x` <character> and `value` <double>.
```

The following used to work but is now an error:

```r
l <- list_of(1, 2, .ptype = int())
l[[1]] <- list(TRUE)
#> Error: No common type for `x` <double> and `value` <list>.
```

It should be one of these instead:

```r
l[1] <- list(TRUE)
l[[1]] <- TRUE
```


R/cast.R
```diff
@@ -133,6 +133,15 @@ vec_cast.default <- function(x, to) {
   stop_incompatible_cast(x, to)
 }
 
+# Cast `x` to `to` but only if they are coercible
+vec_coercible_cast <- function(x, to, ..., x_arg = "", to_arg = "") {
+  if (!missing(...)) {
+    ellipsis::check_dots_empty()
+  }
+  vec_type2(to, x, x_arg = to_arg, y_arg = x_arg)
```
hadley: This is being called purely for its side effect of generating an error if there's no common type?

R/slice.R
```diff
@@ -42,6 +48,34 @@
...
 #'
 #' vec_slice(mtcars, 1:3)
+#'
+#'
+#' # Type stability --------------------------------------------------
+#'
+#' # The assign variant is type stable. It always returns the same
+#' # type as the input.
+#' x <- 1:5
+#' vec_slice(x, 2) <- 20.0
```
hadley: What happens if you do `vec_slice(x, 2) <- 20.5`? You get a warning from the cast?

R/slice.R
```diff
@@ -42,6 +48,34 @@
...
+#'
+#' # `x` is still an integer vector because the RHS was cast to the
+#' # type of the LHS:
+#' typeof(x)
```
hadley: `vec_type()` ?

R/cast.R
lionel-: Correct, I'll add a comment.

R/slice.R
lionel-: This is a lossy cast error that you can lift with `allow_lossy_cast()`.

lionel-: This will surely trip people in general. Maybe we can introduce a custom lossy cast error to suggest creating sequences with `dbl(1:5)` to allow fractional doubles?

lionel-: Maybe we should have used `real` instead of `dbl`...

hadley: That behaviour seems fine to me

R/cast.R
```diff
@@ -133,6 +133,19 @@ vec_cast.default <- function(x, to) {
   stop_incompatible_cast(x, to)
 }
 
+# Cast `x` to `to` but only if they are coercible
+vec_coercible_cast <- function(x, to, ..., x_arg = "", to_arg = "") {
+  if (!missing(...)) {
```
krlmlr: I'm curious: is the check here really required?

lionel-: It's not required, that's just a micro-optimisation. I don't recommend doing this except in very low level packages.

PR #314, opened 2019-05-03
Break cyclic dependencies in s3_register()

krlmlr:
This is required so that I can `load_all()` in {pillar}.


hadley:
@jimhester did this come up in another package?

krlmlr:
Note that this only affects a package X if it calls `s3_register()` for a method defined in Y and Y imports X. Here, X = {vctrs}, Y = {pillar}. For this reason I can't declare imports from {vctrs} in my `NAMESPACE`.

Do we have a similar case somewhere?

jimhester:
Not that I am aware of

R/register-s3.R
```diff
@@ -67,4 +67,16 @@ s3_register <- function(generic, class, method = NULL) {
...
+can_s3_register_now <- function(generic, class, method, package) {
+  if (!(package %in% loadedNamespaces())) {
+    return(FALSE)
+  }
+
+  envir <- asNamespace(package)
+
+  # Avoid registration failures during loading (pkgload or regular),
+  # only register if generic can be accessed
+  return(exists(generic, envir))
```
hadley:
```suggestion
  exists(generic, envir)
```

lionel-:
Thanks!

This might be similar to this issue reported by @wch https://github.com/klutometis/roxygen/issues/623

lionel-:
This causes deploy to fail: https://travis-ci.org/r-lib/vctrs/jobs/531646750#L1274

Might be because of dangling pointers after a load-all, though not sure of the exact mechanism.

lionel-:
@jimhester To solve the issue of dangling pointers once and for all, how about updating the old namespace to make it a mirror environment of the new namespace? All that's needed is to change the CAR (for non-hashtab envs) and TAG (hashtab envs) to those of the new environment. They will share the same bindings.

```r
foo <- env(a = 1)
bar <- env(a = "")

node_poke_tag(foo, node_tag(bar))

env_print(foo)
#> <environment: 0x7fd265e01658>
#> parent: <environment: global>
#> bindings:
#>  * a: <chr>

bar$quux <- TRUE
env_print(foo)
#> <environment: 0x7fd265e01658>
#> parent: <environment: global>
#> bindings:
#>  * a: <chr>
#>  * quux: <lgl>
```

This would also be much faster (load-all'ing rlang is kind of slow because of the binding updates).

lionel-:
hmm the current builds seem to pass successfully!

lionel-:
Actually the deploy step still regularly fails, for instance: https://travis-ci.org/r-lib/vctrs/jobs/533339856

@krlmlr Should we roll this back to see if that fixes it?

krlmlr:
Sure, go ahead. I'm using the old `register_s3_method()` in {pillar} for now. Maybe dynamic registration should live in {rlang}?

lionel-:
@krlmlr Something strange is going on https://community.rstudio.com/t/multiple-error-package-or-namespace-load-failed/30768

lionel-:
It seems pillar is running into dangling pointers after load-alling vctrs.

```r
#> Error in .Call(vctrs_is_vector, x) (from assert.R#108) : Incorrect number of arguments (1), expecting 18 for ''
#> Call `rlang::last_error()` to see a backtrace

rlang::last_error()
#> <error>
#> class:   `rlang_error`
#> backtrace:
#>   1. (function (x, ...) ...
#>   2. tibble:::print.tbl(x)
#>   7. tibble:::format.tbl(x, ..., n = n, width = width, n_extra = n_extra)
#>   9. tibble:::format.trunc_mat(mat)
#>  10. pillar::squeeze(x$mcf, width = width)
#>  11. pillar:::colonnade_get_width(x, width, rowid_width)
#>  12. pillar:::map_named(x[seq_len(init_cols)], pillar_capital)
#>  13. pillar:::map2(.x, names, .f)
#>  14. base::Map(.f, .x, .y, ...)
#>  15. base::mapply(FUN = f, ..., SIMPLIFY = FALSE)
#>  17. pillar::new_pillar_type(x, ...)
#>  18. pillar:::get_pillar_type(x)
#>  20. pillar:::type_sum.default(x)
#>  21. vctrs::vec_is(x)
#>  22. vctrs::vec_is_vector(x) R/assert.R:72:2
#> Call `rlang::last_trace()` to see the full backtrace
```

@krlmlr I'm not sure what to think about pillar depending on vctrs.
0.1.0 is very experimental... And the ptype-abbr API might still need a bit of thinking, in relation to the rlang API, e.g. https://github.com/r-lib/rlang/issues/458.

krlmlr:
Is this also true with #338?

lionel-:
Yes I'm still seeing dangling pointer errors after load-all. This time they seem random and I get a crash.

```
 *** caught bus error ***
address 0x7fe08b01ec00, cause 'non-existent physical address'

Traceback:
 1: .Call(vctrs_is_vector, x)
 2: vec_is_vector(x)
 3: vec_is(x)
 4: type_sum.default(x)
 5: type_sum(x)
 6: get_pillar_type(x)
 7: new_pillar_type(x, ...)
 8: (function (x, title = NULL, ...) {  ...
 9: mapply(FUN = f, ..., SIMPLIFY = FALSE)
10: Map(.f, .x, .y, ...)
11: map2(.x, names, .f)
12: map_named(x[seq_len(init_cols)], pillar_capital)
13: colonnade_get_width(x, width, rowid_width)
14: pillar::squeeze(x$mcf, width = width)
15: format.trunc_mat(mat)
16: format(mat)
17: format.tbl(x, ..., n = n, width = width, n_extra = n_extra)
18: format(x, ..., n = n, width = width, n_extra = n_extra)
19: paste0(..., "\n")
20: cat(paste0(..., "\n"), sep = "")
21: cat_line(format(x, ..., n = n, width = width, n_extra = n_extra))
22: print.tbl(x)
23: function (x, ...) UseMethod("print")(x)
```

lionel-:
oh that's probably because of https://github.com/r-lib/pillar/blob/113845e016b1bc3c5db6da357b5827eb6aa31b78/R/zzz.R#L23

IIRC && IIUC pkgload updates the bindings in imports envs but here you're assigning into the namespace. I think this kind of things would be solved if pkgload created a mirror environment as I suggested above.

Can you please wrap this up in a function so `vec_is()` is not assigned in your namespace, but resolved dynamically instead?

lionel-:
I'm a bit puzzled as to the decision of making CRAN pillar use new functions from dev vctrs if available. What if they change?

lionel-:
@krlmlr Would it make sense to send pillar back to CRAN without a dependency on the experimental vctrs 0.1.0 and without the conditional dependency on dev vctrs?

krlmlr:
I'm only using `vec_is()`, `vec_ptype_abbr()` and `vec_dim()` for now. Would you consider marking these as "maturing" instead?

PR #315, opened 2019-05-03
Fix fallback to `[`

lionel-:
Currently the call graph looks like this when slicing S3 objects:

`vec_slice()` -> `[` -> `vec_slice_native()` -> `vec_proxy()` -> `vec_slice_impl()`

This PR adds a function to get S3 methods in the vctrs namespace or the global env. This is used to determine wether an object implements `vec_proxy()`, and avoid fallback to `[` in that case. The call graph then looks like:

`vec_slice()` -> `vec_proxy()` -> `vec_slice_impl()`

### DIFF
--- a/src/utils.c
+++ b/src/utils.c
@@ -480,7 +480,14 @@ bool r_has_name_at(SEXP names, R_len_t i) {
 }
 
 SEXP r_env_get(SEXP env, SEXP sym) {
-  return Rf_findVarInFrame3(env, sym, FALSE);
+  SEXP obj = Rf_findVarInFrame3(env, sym, FALSE);
+
+  // Force lazy loaded bindings
+  if (TYPEOF(obj) == PROMSXP) {
+    obj = Rf_eval(obj, R_BaseEnv);
+  }
+
+  return obj;
 }
 
 bool r_is_function(SEXP x) {
@@ -512,7 +519,7 @@ SEXP fns_quote = NULL;
 
 void vctrs_init_utils(SEXP ns) {
   vctrs_ns_env = ns;
-  vctrs_method_table = Rf_findVar(Rf_install(".__S3MethodsTable__."), ns);
+  vctrs_method_table = r_env_get(ns, Rf_install(".__S3MethodsTable__."));
 
   vctrs_shared_empty_str = Rf_mkString("");
   R_PreserveObject(vctrs_shared_empty_str);
--- a/src/utils.c
+++ b/src/utils.c
@@ -197,7 +197,7 @@ inline void never_reached(const char* fn) {
 
 static char s3_buf[200];
 
-static SEXP s3_sym(const char* generic, const char* class) {
+static SEXP s3_method_sym(const char* generic, const char* class) {
   int gen_len = strlen(generic);
   int class_len = strlen(class);
   int dot_len = 1;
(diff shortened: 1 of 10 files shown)


lionel-:
Added benefit is that we won't need to export `vec_slice_native()` to avoid inflooping when called from `[` methods. Users can just call `vec_slice()` from `[`, as long as they have implemented `vec_proxy()` (even with just the identity function).

I'll update the documentation if you agree with this approach.

lionel-:
We'll also be able to use this to distinguish between scalar S3 lists and vector S3 lists. Currently they require a proxy that unclasses its input. Using method inspection, an identity proxy would be enough.

The advantage of method inspection over adding a `has_proxy()` generic is that it requires only one dispatch and doesn't complicate the implementation of a new class.

Alternatively, we could add a method getter generic. In addition to implementing proxy, restore, etc, users would have to implement an interface generic:

```r
vec_iface.myclass <- function(...) {
  vec_methods(
    proxy = vec_proxy.myclass,
    restore = vec_restore.myclass
  )
}
```

This would be one more generic to implement. The individual generics would still have to be implemented so that R CMD check can check the signature consistency, otherwise we'd need to check it ourselves. One advantage is that the interface as a whole can then be documented. Like the approach implemented in the PR, inspection and invokation can be done using a single dispatch.

src/utils.c
```diff
@@ -113,6 +114,65 @@ inline void never_reached(const char* fn) {
...
+
+  method = r_env_get(vctrs_method_table, sym);
+  if (r_is_function(method)) {
+    return method;
+  }
+
+  return R_NilValue;
+}
+
+SEXP s3_method(SEXP x, const char* generic) {
```
hadley: Need comment noting that doesn't look for default.  Maybe call it `s3_find_method()`? Also feels like `generic` should be the first arg, since methods are properties of generics.

src/utils.c
```diff
@@ -113,6 +114,65 @@ inline void never_reached(const char* fn) {
...
+  memcpy(buf, generic, gen_len); buf += gen_len;
+  *buf = '.'; ++buf;
+  memcpy(buf, class, class_len); buf += class_len;
+  *buf = '\0';
+
+  return Rf_install(s3_buf);
+}
+
+// First check in global env, then in method table
+static SEXP get_s3_method(const char* generic, const char* class) {
```
hadley: `s3_get_method()`?

src/utils.c
```diff
@@ -113,6 +114,65 @@ inline void never_reached(const char* fn) {
 }
 
 
+static char s3_buf[200];
+
+static SEXP s3_sym(const char* generic, const char* class) {
```
hadley: `s3_method_sym()`?

src/utils.c
```diff
@@ -113,6 +114,65 @@ inline void never_reached(const char* fn) {
...
+// First check in global env, then in method table
+static SEXP get_s3_method(const char* generic, const char* class) {
+  SEXP sym = s3_sym(generic, class);
+
+  SEXP method = r_env_get(R_GlobalEnv, sym);
+  if (r_is_function(method)) {
+    return method;
+  }
+
+  method = r_env_get(vctrs_method_table, sym);
```
hadley: Will methods defined in other packages get registered here?

lionel-: Yes the methods get registered in the table of the namespace of the generic.

lionel-:
:white_check_mark:

PR #319, opened 2019-05-05
Test dev versions of downstream packages

krlmlr:
as a separate element in the build matrix.


krlmlr:
Solved with daily CRON jobs in the respective repos -- {vctrs} is experimental.

PR #320, opened 2019-05-05
Tweaks to ?vec_type2()

krlmlr:
I'm using `@inheritParams` in {hms} and {blob}.

I have no idea why the checks currently don't complain about the missing `...` in the docs.

### DIFF
--- a/R/type2.R
+++ b/R/type2.R
@@ -24,13 +24,14 @@
 #' but should be tested.
 #'
 #' Whenever you implemenet a `vec_type2.new_class()` generic/method,
-#' make sure to always provide `vec_type2.new_class.default()` (
-#' which should call [stop_incompatible_cast()]) and
+#' make sure to always provide `vec_type2.new_class.default()`
+#' (which should call [stop_incompatible_type()]) and
 #' `vec_type2.new_class.vctrs_unspecified()` (which should return `x`).
 #'
 #' See `vignette("s3-vector")` for full details.
 #' @keywords internal
-#' @param x,y Either vector types; i.e.
+#' @inheritParams ellipsis::dots_empty
+#' @param x,y Vector types.
 #' @param x_arg,y_arg Argument names for `x` and `y`. These are used
 #'   in error messages to inform the user about the locations of
 #'   incompatible types (see [stop_incompatible_type()]).
--- a/man/vec_type2.Rd
+++ b/man/vec_type2.Rd
@@ -28,7 +28,9 @@ vec_type2(x, y, ..., x_arg = "", y_arg = "")
 \method{vec_type2}{list}(x, y, ...)
 }
 \arguments{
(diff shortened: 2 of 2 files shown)


R/type-tibble.R
```diff
@@ -1,5 +1,6 @@
 # Coercion ----------------------------------------------------------------
 
+#' @include type2.R
```
lionel-: Why is the `@include` needed?

R/type2.R
```diff
@@ -24,13 +24,14 @@
...
-#' make sure to always provide `vec_type2.new_class.default()` (
-#' which should call [stop_incompatible_cast()]) and
+#' make sure to always provide `vec_type2.new_class.default()`
+#' (which should call [stop_incompatible_type()]) and
 #' `vec_type2.new_class.vctrs_unspecified()` (which should return `x`).
 #'
 #' See `vignette("s3-vector")` for full details.
 #' @keywords internal
-#' @param x,y Either vector types; i.e.
+#' @param x,y Vector types
```
lionel-: This should end with a full stop.

R/type2.R
```diff
@@ -24,13 +24,14 @@
...
+#' @param ... Unused
```
lionel-: This should be:

```
These dots are for future extensions and must be empty.
```

I have added an issue in ellipsis about documentation inheritance: https://github.com/r-lib/ellipsis/issues/22

R/type-tibble.R
krlmlr: It's necessary, because otherwise `type-tibble.R` will be processed before `type2.R` . Could achieve the same with renaming, but this is brittler.

hadley: I don't think the cost is worth the payoff here.

krlmlr:
Merged files instead. I think the order in the "Usage" section is important.

lionel-:
I would like to keep the files separate so the methods are easy to find.

lionel-:
I would like `type-tibble.R` to retain that name instead of being renamed to `type2tibble.R`, so it's consistent with the others.

lionel-:
Isn't this a roxygen issue? It seems in the case of generics and methods, roxygen should keep the generic at the top, then lay out base methods, then S3 methods.

hadley:
Yes, don't worry about the precise order of the functions here.

krlmlr:
Disregarding order for now, opened klutometis/roxygen#863 to deal with it later.

lionel-:
@krlmlr That's weird, I don't think I got a notification for your review request...

krlmlr:
Thanks. I re-requested from Hadley only, initially I did request a review from both of you.

lionel-:
oh I'm seeing this in my github frontend:

> krlmlr requested a review from lionel- 16 hours ago

PR #324, opened 2019-05-08
Move back to fct and cpl

### LINKED ISSUE #323: Reconsider vec_ptype_abbr() for factor and complex
After switching to `vec_ptype_abbr()` in {pillar}, two downstream packages now fail their checks. This is mostly due to the change from `"fct"` to `"fctr"` and `"cpl"` to `"cplx"` .

`type_sum()` moved *away* from `"fctr"` and `"cplx"` just over a year ago: https://github.com/r-lib/pillar/pull/80, suggested in https://github.com/r-lib/pillar/issues/71. It feels inconsistent that now `vec_ptype_abbr()` returns `"fctr"` and `"cplx"` .

@lionel- @hadley: Can we please move to `"fct"` and `"cpl"` here to save quite a few headaches downstream?

Thanks to @nacnudus for pointing this out: https://github.com/nacnudus/unpivotr/issues/24#issuecomment-490627895.

krlmlr:
for consistency with `type_sum()`.

Closes #323.

### DIFF
--- a/R/ptype-abbr-full.R
+++ b/R/ptype-abbr-full.R
@@ -57,7 +57,7 @@ vec_ptype_abbr.default <- function(x) {
       integer = "int",
       double = "dbl",
       character = "chr",
-      complex = "cplx",
+      complex = "cpl",
       list = "list",
       expression = "expr"
     )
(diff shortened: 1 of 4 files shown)


lionel-:
Thanks!

PR #326, opened 2019-05-09
Implement `vec_type2()` data frame methods in C

lionel-:
```r
df1 <- tibble(x = tibble(y = tibble(z = 1)))
df2 <- tibble(x = tibble(y = tibble(Z = "a")))

bench::mark(
  before = vec_type_common_before(df1, df2),
  after = vec_type_common(df1, df2)
)[1:6]
#> 1 before        379µs    412µs     2342.    88.9KB     26.9
#> 2 after        11.1µs   13.3µs    72223.        0B     21.7

bench::mark(
  before = vec_type_common_before(df1, df2, df1),
  after = vec_type_common(df1, df2, df1)
)[1:6]
#> 1 before        877µs    971µs      980.        0B     13.3
#> 2 after        15.2µs   18.4µs    50660.        0B     15.2
```

type2.c

* Implements `df_type2()` in C. Partitions the columns of the two data frames by calling `vec_match()` twice.

* Tibbles are special-cased so we don't need to add them to our type enums.

arg.c:

* New `struct vctrs_arg` derived from `struct vctrs_arg_wrapper`. It adds a `$` prefix to the argument tag, but only if there's no parent tag so it looks like `x$y$z` instead of `$x$y$z` (only a problem for `vec_type2()` since `vec_type_common()` leads with the position/name of dots).

utils.c:

* Predicates for data frames and tibbles that compare the class elements to cached strings.
* Constructors `init_data_frame()` and `init_tibble()`.

### DIFF
--- a/src/utils.c
+++ b/src/utils.c
@@ -148,7 +148,7 @@ static SEXP new_compact_rownames(R_len_t n) {
 static bool is_tibble_class(SEXP class);
 static bool is_data_frame_class(SEXP class);
 
-bool is_tibble(SEXP x) {
+bool is_bare_tibble(SEXP x) {
   SEXP class = PROTECT(Rf_getAttrib(x, R_ClassSymbol));
   bool out = is_tibble_class(class);
   UNPROTECT(1);
--- a/src/utils.c
+++ b/src/utils.c
@@ -9,6 +9,13 @@ SEXP (*rlang_unbox)(SEXP) = NULL;
 SEXP (*rlang_env_dots_values)(SEXP) = NULL;
 SEXP (*rlang_env_dots_list)(SEXP) = NULL;
 
+SEXP strings_tbl = NULL;
+SEXP strings_tbl_df = NULL;
+SEXP strings_data_frame = NULL;
+
+SEXP classes_data_frame = NULL;
+SEXP classes_tibble = NULL;
+
 
 bool is_bool(SEXP x) {
   return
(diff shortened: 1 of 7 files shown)


src/utils.c
```diff
@@ -108,6 +115,80 @@ R_len_t compact_rownames_length(SEXP x) {
...
+  ++class_ptr;
+  if (*class_ptr != strings_tbl) {
+    return false;
+  }
+  ++class_ptr;
+  if (*class_ptr != strings_data_frame) {
+    return false;
+  }
+
+  return true;
```
hadley: Why doesn't this need to fall back to `Rf_inherits()`?

lionel-: This is used for optimisating objects that are actually tibbles. Objects inheriting from it should call `vec_restore()`. The dispatch path for tibbles is still not entirely clear to me though.

I'll add a `dispatch` parameter so the semantics at the call site are clearer.

lionel-: This is now `is_bare_tibble()`.

hadley: Actually I was confused by the flow, because you're not first comparing `class` to a cached character vector. Or did we decide that optimisation was not worth it?

lionel-: This would require tibble to use a constructor exported in vctrs. Comparing the length and elements of the class should be fast as well.

lionel-: Also this method works with unserialised objects.

hadley: Right, but I think that's irrelevant compared to the cost of having a subtly different way of checking for different classes. It's better to standardise on a single approach and use it everywhere.

PR #329, opened 2019-05-10
Fix `min()` and `max()` methods

DavisVaughan:
`min.vctrs_vctr()` and `max.vctrs_vctr()` have small bugs.

- Both break with missing values when `na.rm = FALSE`
- `min()` is using `which.max()` when `na.rm = TRUE`

``` r
library(vctrs)

x <- c(NA, 1, 2)
x_vctr <- new_vctr(x)

# not working 
min(x)
#> [1] NA
min(x_vctr)
#> Error in idx[[1]]: subscript out of bounds

max(x)
#> [1] NA
max(x_vctr)
#> Error in idx[[1]]: subscript out of bounds

# typo in min.vctrs_vctr()! woops!
min(x, na.rm = TRUE)
#> [1] 1
min(x_vctr, na.rm = TRUE)
#> <vctrs_vctr[1]>
#> [1] 2

# all good here
max(x, na.rm = TRUE)
#> [1] 2
max(x_vctr, na.rm = TRUE)
#> <vctrs_vctr[1]>
#> [1] 2
```

<sup>Created on 2019-05-10 by the [reprex package](https://reprex.tidyverse.org) (v0.2.1.9000)</sup>

I fixed the error by using `vec_equal(na_equal = TRUE)`. It went past the 80 chr line limit so I took the chance to refactor a bit.

### DIFF
--- a/R/type-vctr.R
+++ b/R/type-vctr.R
@@ -422,7 +422,7 @@ min.vctrs_vctr <- function(x, ..., na.rm = FALSE) {
   rank <- xtfrm(x)
 
   idx <- if (isTRUE(na.rm)) {
-    which.max(rank)
+    which.min(rank)
   } else {
     which(vec_equal(rank, min(rank), na_equal = TRUE))
   }
--- a/R/type-vctr.R
+++ b/R/type-vctr.R
@@ -421,10 +421,10 @@ min.vctrs_vctr <- function(x, ..., na.rm = FALSE) {
   # TODO: implement to do vec_arg_min()
   rank <- xtfrm(x)
 
-  idx <- if (isTRUE(na.rm)) {
-    which.min(rank)
+  if (isTRUE(na.rm)) {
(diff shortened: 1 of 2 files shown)


R/type-vctr.R
```diff
@@ -420,15 +420,27 @@ quantile.vctrs_vctr <- function(x, ..., type = 1, na.rm = FALSE) {
 min.vctrs_vctr <- function(x, ..., na.rm = FALSE) {
   # TODO: implement to do vec_arg_min()
   rank <- xtfrm(x)
-  idx <- if (isTRUE(na.rm)) which.max(rank) else which(rank == min(rank))
+
+  idx <- if (isTRUE(na.rm)) {
```
hadley: You should move `idx <- ` inside the if clauses (we don't have a principle for this, but should)

lionel-:
Thanks!

PR #330, opened 2019-05-13
Use `vec_proxy_method()` to detect vectors

lionel-:


src/type.c
```diff
@@ -318,38 +317,35 @@ const char* vec_type_as_str(enum vctrs_type type) {
...
-
-  case vctrs_type_s3: {
-    SEXP proxy = PROTECT(vec_proxy(x));
-    bool out = vec_is_vector_rec(proxy, false);
-    UNPROTECT(1);
-    return out;
-  }
-
+  case VECSXP:
+    if (!OBJECT(x) || is_data_frame(x)) {
```
hadley: Do we still need `is_data_frame()` here? It vaguely confused me for a minute.

lionel-: I have reviewed the design of proxy/restore in the internals and this code is now gone, PR incoming.

PR #335, opened 2019-05-16
Fix restoration of data frames

lionel-:
* Rewrite class detection as a state machine.
* Use this in `vec_restore()` to avoid dispatch with native data frames.
* Fix a couple of bugs when restoring to subclasses of data frames.

New enum to support this:

```c
enum vctrs_class_type {
  vctrs_class_data_frame,
  vctrs_class_bare_data_frame,
  vctrs_class_bare_tibble,
  vctrs_class_rcrd,
  vctrs_class_posixlt,
  vctrs_class_unknown,
  vctrs_class_none
};
```

It's still unclear how much of it should be in `enum vctrs_type`.


src/utils-dispatch.c
```diff
@@ -0,0 +1,105 @@
...
+  UNPROTECT(1);
+  return type;
+}
+
+static enum vctrs_class_type class_type_impl(SEXP class) {
+  int n = Rf_length(class);
+  SEXP const* class_ptr = STRING_PTR(class);
+  SEXP const* p = class_ptr;
+
+  switch (n) {
```
hadley: I think this needs a comment like "First check for data frames"

lionel-: This checks for bare classes, i.e. those for which we know exactly how many strings compose their class. Good point that this needs comments.

PR #343, opened 2019-05-20
Factor counted reduce out of common type finder

lionel-:


lionel-:
This should make it easier to reuse the current reduction mechanism in other primitives like finding the common size.

PR #346, opened 2019-05-20
Vectorise hashing

lionel-:
This pulls the dereference of arrays out of loops, and computes the hashes of nested data frames columnwise rather than rowwise.

```r
df <- tibble(x = 1:10, y = rep(list(iris), 10))

bench::mark(
  before = obj_hash_before(df)
  after = obj_hash(df)
)[1:6]
#> 1 before        333µs    443µs    355µs   1.94ms     2258.
#> 2 after         152µs    181µs    154µs    700µs     5515.

bench::mark(
  before = vec_unique_before(df)
  after = vec_unique(df)
)[1:6]
#> 1 before        337µs    446µs    363µs   1.56ms     2243.
#> 2 after         161µs    201µs    168µs    969µs     4981.
```

### DIFF
--- a/src/hash.c
+++ b/src/hash.c
@@ -41,97 +41,39 @@ int32_t hash_double(double x) {
 }
 
 
+// Hashing scalars -----------------------------------------------------
+
+static int32_t lgl_hash_scalar(const int* x);
+static int32_t int_hash_scalar(const int* x);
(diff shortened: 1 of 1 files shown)


lionel-:
hash.c should probably use unsigned types for hashes

PR #349, opened 2019-05-20
Add `vec_structure()` for use in `new_vctr()`

### LINKED ISSUE #341: Performance) `new_vctr()` overhead
Currently, with R < 3.6.0, `new_vctr()` makes 4 copies every time it is called. Since this should be a really cheap function, I was hoping we could do something about this. As is, it's really killing my rray performance.

There are two main issues, both stem from the fact that `attributes(x) <- list(...)` creates two copies when there are attributes to set. See my examples below:

``` r
x <- 1:50000 + 0L
garbarge <- capture.output(profmem::profmem(attributes(x) <- NULL))

# set to nothing (same as setting to NULL)
x <- 1:50000 + 0L
profmem::profmem(attributes(x) <- list())
#> attributes(x) <- list()
#> 
#> Memory allocations:
#>        what  bytes      calls
#> 1     alloc 200048 <internal>
#> total       200048

# set attributes. Agh! Two copies!
x <- 1:50000 + 0L
profmem::profmem(attributes(x) <- list(stuff = "hi"))
#> attributes(x) <- list(stuff = "hi")
#> 
#> Memory allocations:
#>        what  bytes      calls
#> 1     alloc 200048 <internal>
#> 2     alloc 200048 <internal>
#> total       400096

# Clever way to call it, which only copies once
x <- 1:50000 + 0L
profmem::profmem(x <- `attributes<-`(x, value = list(stuff = "hi")))
#> x <- `attributes<-`(x, value = list(stuff = "hi"))
#> 
#> Memory allocations:
#> Number of 'new page' entries not displayed: 1
#>        what  bytes      calls
#> 2     alloc 200048 <internal>
#> total       200048
```

Based on the above results, I have two suggestions.

1) `validate_attr()` should validate and return names, but not set them. It will simply pass them on to `structure()`, ~which always removes all attributes from `.Data` before adding new ones.~ (not true, see the Update below)

2) We should wrap `structure()` so we have a version that does the exact same thing, except it sets attributes using `attributes<-(.Data, value = c(attributes(.Data), attrib))` since that only makes 1 copy, as seen above. Maybe call it `vec_structure_shim()`.

To be as forward compatible as possible, I think it would make sense to create this `vec_structure_shim()` in `.onLoad()`. If the version is >= 3.6.0, we can just use base R's provided `structure()`, which has fixed the double copy issue, and with ALTREP often makes 0 copies until needed. Otherwise, we use our shim.

PS: For the interested, I believe this `attributes<-()` double copy behavior might have been some kind of bug that was just fixed. There was a time between 3.5.3 and 3.6.0 where this was what `structure()` did, which leads me to believe that even after they had ALTREP parts in place, it was still making a copy because of some compiler bug. 
https://github.com/wch/r-source/commit/8d332d420f83c4343cf2e1a5694faeffc9de599e#diff-7eda0a571a9f91eb3be4d1627b8d4c2aL37

Here is a before/after performance comparison:

``` r
x <- 1:500000 + 0L
dim = c(5000L, 100L)

bench::mark(
  new_vctr(x, dim = dim),
  new_vctr2(x, dim = dim)
)
#> # A tibble: 2 x 10
#>   expression   min     mean   median    max `itr/sec` mem_alloc  n_gc n_itr
#>   <chr>      <bch> <bch:tm> <bch:tm> <bch:>     <dbl> <bch:byt> <dbl> <int>
#> 1 new_vctr(… 706µs   1.77ms   1.11ms 5.56ms      565.    7.65MB    43   152
#> 2 new_vctr2… 135µs 484.16µs 279.96µs 1.22ms     2065.    1.96MB    39   801
```

Update) `structure()` does not actually remove existing attributes from `.Data`, but maybe this is ok? Otherwise I don't see a way to add _only_ the attributes in `...` without either rewriting `structure()` or making a second copy with `attributes(x) <- NULL`

DavisVaughan:
Closes #341 

- New `vctrs_set_attributes()`, inspired heavily from `do_attributesgets()`, which removes all attributes from `x` and then adds the `attrib`.
   - All attributes must be named.
   - `dim` is always set first, consistent with `do_attributesget()` to avoid an issue of attempting to set dim names on a non-array object if the user does `vec_structure(x, dimnames = list(...), dim = c(...))`
   - Exposed as `vec_set_attributes()`

- New `vec_structure()` that tries to do the same thing no matter what R version you are on. Both versions try to first remove attributes of `.data`, then add the new ones.

- `new_vctr()` now uses `vec_structure()` and just passes `names` through rather than calling `attributes(.data) <- list(names = nms)` separately.

- `vec_data()` now uses `vec_set_attributes()`.

Questions)

- I'm setting `r_version` in the `.onLoad()` because it's actually quite slow to call `getRVersion()`. Is this right?
- Does `vctrs_set_attributes()` do what you were intending?
- Is there a better way to avoid having to switch off the r version? Maybe register these in `.onLoad()` rather than storing the R version?
- I can add tests, I guess specifically for `vec_set_attributes()`?

### DIFF
--- a/R/type-vctr.R
+++ b/R/type-vctr.R
@@ -56,27 +56,11 @@ new_vctr <- function(.data, ..., class = character()) {
   if (!is_vector(.data)) {
     abort("`.data` must be a vector type.")
   }
+
   nms <- validate_names(.data)
+  attrib <- list(names = nms, ..., class = c(class, "vctrs_vctr"))
 
-  restructure(.data, names = nms, ..., class = c(class, "vctrs_vctr"))
-}
-
-restructure <- function(.data, ...) {
-  if (r_version_at_least_3.6.0) {
-    attributes(.data) <- NULL
-    structure(.data, ...)
-  } else {
-    vec_set_attributes(.data, list(...))
-  }
-}
-
-vec_set_attributes <- function(x, attrib) {
-  if (r_version_at_least_3.6.0) {
-    attributes(x) <- attrib
-    x
-  } else {
-    .Call(vctrs_set_attributes, x, attrib)
-  }
+  vec_set_attributes(.data, attrib)
 }
 
 validate_names <- function(.data) {
--- a/R/type-vctr.R
+++ b/R/type-vctr.R
@@ -62,7 +62,7 @@ new_vctr <- function(.data, ..., class = character()) {
 }
 
 vec_structure <- function(.data, ...) {
-  if (r_version >= '3.6.0') {
+  if (r_version_at_least_3.6.0) {
     attributes(.data) <- NULL
     structure(.data, ...)
   } else {
@@ -71,7 +71,7 @@ vec_structure <- function(.data, ...) {
 }
 
 vec_set_attributes <- function(x, attrib) {
-  if (r_version >= '3.6.0') {
+  if (r_version_at_least_3.6.0) {
     attributes(x) <- attrib
     x
   } else {
--- a/R/type-vctr.R
+++ b/R/type-vctr.R
@@ -65,8 +65,7 @@ vec_structure <- function(.data, ...) {
   if (r_version >= '3.6.0') {
     attributes(.data) <- NULL
     structure(.data, ...)
-  }
-  else {
+  } else {
     vec_set_attributes(.data, list(...))
   }
 }
--- a/R/type-vctr.R
+++ b/R/type-vctr.R
@@ -67,12 +67,17 @@ vec_structure <- function(.data, ...) {
     structure(.data, ...)
   }
   else {
-    vec_set_attributes(.data, ...)
+    vec_set_attributes(.data, list(...))
   }
 }
 
-vec_set_attributes <- function(.x, ...) {
-  .Call(vctrs_set_attributes, .x, list(...))
+vec_set_attributes <- function(x, attrib) {
+  if (r_version >= '3.6.0') {
+    attributes(x) <- attrib
+    x
+  } else {
+    .Call(vctrs_set_attributes, x, attrib)
+  }
 }
 
 validate_names <- function(.data) {
--- a/R/type-vctr.R
+++ b/R/type-vctr.R
@@ -56,20 +56,33 @@ new_vctr <- function(.data, ..., class = character()) {
   if (!is_vector(.data)) {
     abort("`.data` must be a vector type.")
   }
-  .data <- validate_attr(.data)
+  nms <- validate_names(.data)
 
-  structure(.data, ..., class = c(class, "vctrs_vctr"))
+  vec_structure(.data, names = nms, ..., class = c(class, "vctrs_vctr"))
 }
 
-validate_attr <- function(.data) {
+vec_structure <- function(.data, ...) {
+  if (r_version >= '3.6.0') {
+    attributes(.data) <- NULL
+    structure(.data, ...)
+  }
+  else {
+    vec_set_attributes(.data, ...)
+  }
+}
+
+vec_set_attributes <- function(.x, ...) {
+  .Call(vctrs_set_attributes, .x, list(...))
+}
+
+validate_names <- function(.data) {
   nms <- names(.data)
 
   if (!names_all_or_nothing(nms)) {
     stop("If any elements of `.data` are named, all must be named", call. = FALSE)
   }
 
-  attributes(.data) <- list(names = nms)
-  .data
+  nms
 }
 names_all_or_nothing <- function(names) {
   if (is.null(names)) {
--- a/R/type-vctr.R
+++ b/R/type-vctr.R
@@ -58,10 +58,10 @@ new_vctr <- function(.data, ..., class = character()) {
   }
   nms <- validate_names(.data)
 
-  vec_structure(.data, names = nms, ..., class = c(class, "vctrs_vctr"))
+  restructure(.data, names = nms, ..., class = c(class, "vctrs_vctr"))
 }
 
-vec_structure <- function(.data, ...) {
+restructure <- function(.data, ...) {
   if (r_version_at_least_3.6.0) {
     attributes(.data) <- NULL
     structure(.data, ...)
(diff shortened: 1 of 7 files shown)


R/type-vctr.R
```diff
@@ -56,20 +56,37 @@ new_vctr <- function(.data, ..., class = character()) {
...
   }
-  .data <- validate_attr(.data)
+  nms <- validate_names(.data)
 
-  structure(.data, ..., class = c(class, "vctrs_vctr"))
+  vec_structure(.data, names = nms, ..., class = c(class, "vctrs_vctr"))
 }
 
-validate_attr <- function(.data) {
+vec_structure <- function(.data, ...) {
```
lionel-: Should this be called `restructure()`? This way the name hints at the removal of existing attributes.

R/type-vctr.R
```diff
@@ -56,20 +56,37 @@ new_vctr <- function(.data, ..., class = character()) {
...
+  if (r_version >= '3.6.0') {
+    attributes(.data) <- NULL
+    structure(.data, ...)
+  } else {
+    vec_set_attributes(.data, list(...))
+  }
+}
+
+vec_set_attributes <- function(x, attrib) {
+  if (r_version >= '3.6.0') {
```
lionel-: I would cache that comparison from the onLoad hook.

src/type.c
```diff
@@ -7,6 +7,65 @@ static SEXP syms_vec_type_finalise_dispatch = NULL;
 static SEXP fns_vec_is_vector_dispatch = NULL;
 static SEXP fns_vec_type_finalise_dispatch = NULL;
 
+// [[ register() ]]
```
lionel-: The file doesn't seem appropriate. Maybe utils.c?

R/type-vctr.R
```diff
@@ -56,20 +56,37 @@ new_vctr <- function(.data, ..., class = character()) {
   if (!is_vector(.data)) {
     abort("`.data` must be a vector type.")
   }
-  .data <- validate_attr(.data)
+  nms <- validate_names(.data)
 
-  structure(.data, ..., class = c(class, "vctrs_vctr"))
+  restructure(.data, names = nms, ..., class = c(class, "vctrs_vctr"))
```
hadley: Why can't we use `vec_set_attributes()` here?

src/utils.c
```diff
@@ -85,6 +85,66 @@ SEXP vctrs_dispatch3(SEXP fn_sym, SEXP fn,
   return vctrs_dispatch_n(fn_sym, fn, syms, args);
 }
 
+// [[ register() ]]
+SEXP vctrs_set_attributes(SEXP x, SEXP attrib) {
```
hadley: Need a comment explaining why we have this function.

R/zzz.R
```diff
@@ -1,11 +1,15 @@
...
+r_version_at_least_3.6.0 <- NULL
+
 .onLoad <- function(libname, pkgname) {
   backports::import(pkgname, "strrep")
 
   s3_register("generics::as.factor", "vctrs_vctr")
   s3_register("generics::as.ordered", "vctrs_vctr")
   s3_register("generics::as.difftime", "vctrs_vctr")
 
+  r_version_at_least_3.6.0 <<- getRversion() >= '3.6.0'
```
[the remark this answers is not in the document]
DavisVaughan: I did that first, but lionel suggested https://github.com/r-lib/vctrs/pull/349#discussion_r285744185

R/type-vctr.R
DavisVaughan: So you would no longer call `structure()` when >= 3.6? Maybe that would be fine, actually, since `structure()` is just `attributes<-()` with some extra stuff on top.

R/zzz.R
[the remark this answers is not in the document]
hadley: Ooooh, maybe it would be better to define the function on load to avoid the weird variable name (and that would make it ever faster).

DavisVaughan:
Benchmarks:

``` r
library(vctrs)

vec_set_attributes <- vctrs:::vec_set_attributes

base_set_attributes <- function(x, attrib) {
  attributes(x) <- attrib
  x
}

# Stripping attributes
x <- structure(1, foo = "bar")

bench::mark(
  base_set_attributes(x, NULL),
  vec_set_attributes(x, NULL),
  iterations = 100000
)
#> # A tibble: 2 x 10
#>   expression   min  mean median    max `itr/sec` mem_alloc  n_gc n_itr
#>   <chr>      <bch> <bch> <bch:> <bch:>     <dbl> <bch:byt> <dbl> <int>
#> 1 base_set_… 393ns 535ns  469ns 23.8µs  1869515.        0B    10 99989
#> 2 vec_set_a… 304ns 451ns  402ns 49.4µs  2216623.        0B     8 99992

# Adding attributes
x <- structure(1, foo = "bar")
attrib <- list(a = "a", b = "b")

bench::mark(
  base_set_attributes(x, attrib),
  vec_set_attributes(x, attrib),
  iterations = 100000
)
#> # A tibble: 2 x 10
#>   expression   min  mean median    max `itr/sec` mem_alloc  n_gc n_itr
#>   <chr>      <bch> <bch> <bch:> <bch:>     <dbl> <bch:byt> <dbl> <int>
#> 1 base_set_… 440ns 654ns  595ns 23.9µs  1528826.        0B     8 99992
#> 2 vec_set_a… 400ns 500ns  470ns 14.5µs  1999424.        0B     8 99992

# Large object + setting attributes = expose the double copy issue
x <- 1:5000000 + 0L
attrib <- list(a = "a", b = "b")

bench::mark(
  base_set_attributes(x, attrib),
  vec_set_attributes(x, attrib)
)
#> # A tibble: 2 x 10
#> 1 base_set_… 5.58ms 6.56ms 6.73ms 8.77ms      152.    38.1MB    29    18
#> 2 vec_set_a… 2.19ms 3.06ms 3.28ms 3.44ms      327.    19.1MB    25    98
```

<sup>Created on 2019-05-21 by the [reprex package](https://reprex.tidyverse.org) (v0.2.1.9000)</sup>

lionel-:
Thanks!

PR #350, opened 2019-05-21
Use `vctrs_error_scalar_type` class in `expect_error()`

DavisVaughan:
This was throwing warnings for me with the latest version of testthat


lionel-:
Can you leave the testing of message contents please?

hadley:
Is that important for this case?

lionel-:
I guess if the message doesn't vary as a function of inputs it doesn't matter.

lionel-:
Thanks!

PR #352, opened 2019-05-24
Fix hash type

lionel-:
* Use unsigned `uint32_t` in hash.c, consistently with the type used in dictionary.c

* Return a raw vector instead of an integer vector from `vec_hash()` and `obj_hash()` to avoid implementation defined behaviours (unsigned to signed conversion, and size of int).


lionel-:
The invariant is `vec_size(vec_hash(x))` equals `vec_size(x)` * 4. Do you plan to export this function? I thought it was just for tests.

On some systems the conversion saturates to `INT_MAX`, on others it wraps around. I guess this is already tested with vctrs 0.1.0 on CRAN's Solaris machine, so we can get away with the signed integers.

However if `vec_hash()` is meant to be exported, I'd rather avoid implementation defined behaviour as it looks a bit icky, and create a little wrapper class on top of the raw vector. Such a class could be generally useful, if parameterised to specify the byte length of atomic elements.

hadley:
Oh sure, I thought we exported it.  Can you just add a (non-roxygen) comment describing the behavior?

PR #355, opened 2019-05-28
Implement `vec_rbind()` in C

lionel-:
Performance is now closer too dplyr for small vectors, and scales better than dplyr.

```r
library(vctrs)
library(dplyr)
library(data.table)

bench <- function(data) {
  bench::mark(
    dplyr = bind_rows(!!!data),
    vctrs = vec_rbind(!!!data)
  )[1:6]
}

# For lists of data frames, includes data.table
bench2 <- function(data) {
  bench::mark(
    dplyr = bind_rows(!!!data),
    vctrs = vec_rbind(!!!data),
    data.table = rbindlist(data),
    check = FALSE
  )[1:6]
}

# Run manually with r-lib/vctrs@d7f2a72f0f
bench_before <- function(data) {
  bench::mark(
    before = vec_rbind(!!!data)
  )[1:6]
}
```

### Small vectors

```r
data <- rep(list(rlang::set_names(letters)), 10)
bench(data)
#> 1 dplyr         125µs    142µs     6637.     7.2KB     65.2
#> 2 vctrs         180µs    205µs     4784.      27KB     16.2

bench_before(data)
#> 1 before       8.96ms   10.5ms      91.6    1.64MB     9.16
```

### Large vectors

Much faster than dplyr.

```r
data <- rep(list(rlang::set_names(1:1e4)), 2)
bench(data)
#> 1 dplyr         4.31s    4.31s     0.232  382.19MB    36.2 
#> 2 vctrs        8.82ms   9.66ms    97.4      1.95MB     3.98

bench_before(data)
#> 1 before         2.2s     2.2s     0.454    17.4MB     11.3
```

### Few small data frames

```r
data <- rep(list(mtcars), 10)
bench2(data)
#> 1 dplyr        55.8µs   69.7µs    13509.      28KB    17.6
#> 2 vctrs        94.2µs  105.4µs     8859.    29.3KB     8.89
#> 3 data.table   72.5µs   83.2µs    11003.    44.8KB    14.8

bench_before(data)
#> 1 before       13.2ms     14ms      71.2     590KB     17.1
```

### Many small data frames

```r
data <- rep(list(mtcars), 1000)
bench2(data)
#> 1 dplyr        4.31ms   5.82ms     175.     2.69MB    14.6
#> 2 vctrs        9.05ms  10.21ms      97.5    2.82MB     6.65
#> 3 data.table    1.6ms   2.58ms     389.     2.75MB    11.7

bench_before(data)
#> 1 before         4.6s     4.6s     0.218    5.25GB     3.92
```

### Few large data frames

Faster than dplyr.

```r
large <- vec_rbind(!!!rep(list(mtcars), 1000))
data <- rep(list(large), 10)
bench2(data)
#> 1 dplyr       14.47ms  14.98ms      60.7    26.9MB    152.
#> 2 vctrs        7.01ms   8.76ms      91.9    28.1MB     91.9
#> 3 data.table   2.85ms   3.53ms     175.     26.9MB    175.

bench_before(data)
#> 1 before        348ms    349ms      2.86     580MB        0
```

### Many large data frames

Faster than dplyr and data.table.

```r
data <- rep(list(large), 1000)
bench2(data)
#> 1 dplyr         3.17s    3.17s     0.316    2.62GB    0.316
#> 2 vctrs         2.34s    2.34s     0.427    2.74GB    0.853
#> 3 data.table    2.87s    2.87s     0.349    2.62GB    0.349

bench_before(data)
# *too slow, had to kill*
```

### Factors

Still slow because of type2/cast implemented in R.

```r
data <- rep(list(iris), 10)
bench2(data)
#> 1 dplyr       57.53µs  64.19µs    14415.      53KB     4.37
#> 2 vctrs        1.09ms   1.24ms      797.   223.3KB     7.52
#> 3 data.table  72.07µs   82.8µs    11036.    69.4KB     6.72

bench_before(data)
#> 1 vctrs        8.73ms   9.19ms      108.    1.32MB     9.17
```


src/type-data-frame.c
```diff
@@ -0,0 +1,57 @@
...
+  init_compact_rownames(x, n);
+}
+
+static void init_compact_rownames(SEXP x, R_len_t n) {
+  SEXP rn = PROTECT(new_compact_rownames(n));
+  Rf_setAttrib(x, R_RowNamesSymbol, rn);
+  UNPROTECT(1);
+}
+static SEXP new_compact_rownames(R_len_t n) {
+  if (n <= 0) {
```
hadley: This is mildly surprising — would it be better to always use the same representation of row names?

src/type-data-frame.c
```diff
@@ -0,0 +1,57 @@
...
+
+
+// [[ include("utils.h") ]]
+SEXP new_data_frame(SEXP x, R_len_t n) {
+  x = PROTECT(r_maybe_duplicate(x));
+
+  init_data_frame(x, n);
+
+  if (Rf_length(x) == 0) {
+    Rf_setAttrib(x, R_NamesSymbol, vctrs_shared_empty_chr);
```
hadley: Why isn't this done in `init_data_frame()`?

src/type-data-frame.c
```diff
@@ -0,0 +1,57 @@
...
+  }
+
+  UNPROTECT(1);
+  return x;
+}
+
+// [[ include("utils.h") ]]
+bool is_compact_rownames(SEXP x) {
```
hadley: I feel like these probably should have some `dataframe_` prefix, but it's not that important (and orthogonal to this PR)

src/utils.c
```diff
@@ -265,6 +234,27 @@ SEXP s3_find_method(const char* generic, SEXP x) {
 }
 
 
+// Initialised at load time
+SEXP compact_seq_attrib = NULL;
+
+// Returns a compact sequence that `vec_slice()` understands
+SEXP compact_seq(R_len_t from, R_len_t to) {
```
hadley: Not sure where this is used, but should it check that `to` > `from`? (or return empty vector)

src/utils.c
```diff
@@ -345,6 +335,16 @@ void r_int_fill_seq(SEXP x, int start) {
   }
 }
 
+SEXP r_seq(R_len_t from, R_len_t to) {
+  R_len_t n = to - from;
```
hadley: This definitely needs a check

src/bind.c
```diff
@@ -0,0 +1,113 @@
...
+
+  if (ptype == R_NilValue) {
+    UNPROTECT(1);
+    return new_data_frame(vctrs_shared_empty_list, 0);
+  }
+  if (!is_data_frame(ptype)) {
+    Rf_errorcall(R_NilValue, "Can't bind objects that are not coercible to a data frame.");
+  }
+
+  R_len_t nrow = 0;
```
hadley: `// Find total size of output`

src/bind.c
```diff
@@ -0,0 +1,113 @@
...
+  // The common type holds information about common column names,
+  // types, etc. Each element of `xs` needs to be cast to that type
+  // before assignment.
+  ptype = PROTECT(vctrs_type_common_impl(xs, ptype));
+
+  if (ptype == R_NilValue) {
+    UNPROTECT(1);
+    return new_data_frame(vctrs_shared_empty_list, 0);
+  }
+  if (!is_data_frame(ptype)) {
```
hadley: What should `vec_rbind(NA, NA)` return? It's currently an error, which doesn't seem right to me.

src/bind.c
```diff
@@ -0,0 +1,113 @@
...
+    R_len_t size = ns[i];
+    if (!size) {
+      continue;
+    }
+
+    SEXP tbl = PROTECT(vec_cast(VECTOR_ELT(xs, i), ptype));
+    idx_ptr[0] = counter;
+    idx_ptr[1] = counter + size;
+
+    df_assign(out, idx, tbl, false);
```
hadley: What's the advantage of a single `idx` here instead of two integers?

src/slice-assign.c
```diff
@@ -139,19 +175,54 @@ static SEXP raw_assign(SEXP x, SEXP index, SEXP value) {
...
 #undef ASSIGN_BARRIER
+#undef ASSIGN_BARRIER_INDEX
+#undef ASSIGN_BARRIER_COMPACT
 
 
-static SEXP df_assign(SEXP x, SEXP index, SEXP value) {
-  R_len_t n = Rf_length(x);
-  SEXP out = PROTECT(Rf_shallow_duplicate(x));
+/**
+ * `out` and `value` must be rectangular lists. `value` must have the
```
hadley: I'd suggest putting each precondition on its own line

src/bind.c
lionel-: Good catch, I think it should return `tibble(...1 = c(NA, NA)`.

src/bind.c
lionel-: `df_assign()` is part of `vec_slice<-` and thus expects a full index vector. Eventually we'll have ALTREP sequences and repetitions that `vec_slice()` will treat specially. The compact sequence stuff I added here is an experiment in that direction. It didn't make much difference in terms of performance but I decided to keep it since the altrep seqs and reps are still on the todo list.

src/type-data-frame.c
lionel-: Good point.

src/type-data-frame.c
lionel-: This is from `.set_row_names()`'s implementation. It's probably safer to reproduce its behaviour exactly.

src/utils.c
lionel-: `Rf_allocVector()` checks for negative inputs, but I added another check.

lionel-:
> Performance looks great! I assume adding C implementations for factors, Dates etc is on your near term todo list?

Yes, right after `vec_cbind()` gets a C implementation.

PR #358, opened 2019-05-29
Fix implementation of `vec_dim()` and add `vec_dims()`

DavisVaughan:
There are two things that are somewhat confusing here:

- It looks like the original C `vec_dim()` was really trying to be `vec_dims()`.
- If it was trying to be `vec_dims()`, I don't think it was doing the right thing with arrays. See this diff line in particular, which I don't think is correct. https://github.com/r-lib/vctrs/compare/master...DavisVaughan:fix-dim?expand=1#diff-9eb7d88ef60b7c3fcaa4c07f6fd75c90L9

I've fixed this by implementing `vec_dims()` and `vec_dim()` to be consistent with their R level equivalents.

`vec_names()` now works as I'd expect

``` r
library(rray)
library(vctrs)

x <- matrix(1, dimnames = list("row", "col"))

vctrs:::vec_names(x)
#> [1] "row"

x_rray <- as_rray(x)

vctrs:::vec_names(x_rray)
#> [1] "row"
```


lionel-:
Is `vec_dims()` a confusing name? It's plural yet may return a smaller vector than `vec_dim()`. If I had to guess what these functions were doing, I would have swapped them.

Perhaps `vec_dims()` should be `vec_dim_n()` or `vec_dimensionality()`?

tests/testthat/test-names.R
```diff
@@ -8,6 +8,8 @@ test_that("vec_names() retrieves names", {
   expect_identical(vec_names(set_names(letters)), letters)
   expect_null(vec_names(mtcars))
   expect_identical(vec_names(Titanic), dimnames(Titanic)[[1]])
+  x <- matrix(1L, dimnames = list("row", "col"))
+  expect_identical(vec_names(x), dimnames(x)[[1]])
```
lionel-: Perhaps also a test with multiple rows?

src/size.c
```diff
@@ -3,13 +3,19 @@
...
 
+SEXP vec_dim(SEXP x) {
+  SEXP dim = PROTECT(Rf_getAttrib(x, R_DimSymbol));
 
-R_len_t vec_dim(SEXP x) {
-  if (has_dim(x)) {
-    return Rf_length(x);
-  } else {
-    return 1;
+  if (r_is_null(dim)) {
```
lionel-: We use `dim == R_NilValue`.

src/utils.c
```diff
@@ -439,6 +439,10 @@ bool r_is_string(SEXP x) {
     STRING_ELT(x, 0) != NA_STRING;
 }
 
+bool r_is_null(SEXP x) {
```
lionel-: Please remove in favour of pointer comparison.

tests/testthat/test-names.R
DavisVaughan: The Titanic example is a 3D table with multiple rows, so I thought that sufficed. It happened to pass with the previous implementation of `vec_dim()`. This length 1 case was just to expose the issue.

DavisVaughan:
I think I would call is `vec_n_dim()` as that matches how I would say it. "The number of dimensions" = dimensionality.

But note that `vec_dims()` is exported and was included in the first version of vctrs, so if the change was made at the R level this would be backwards incompatible.

FWIW I use `dims` in rray to mean dimensionality, as is currently done here. I'm not tied to it, I suppose, and this does seem clearer.

lionel-:
I think it's ok to change the R API of an experimental package, if it avoids confusion.

DavisVaughan:
Can we do it in a separate PR? Feels a bit different than this one since this was really about fixing that `vec_names()` issue

hadley:
Let's do it in a separate PR, and lets think about names a bit more broadly.

lionel-:
Thanks!

PR #365, opened 2019-05-30
Fix array hashing

### LINKED ISSUE #327: Bug with `hash_scalar()` affecting `vec_unique()` with arrays
https://github.com/r-lib/vctrs/blob/79adc0f63552a87705fd326599419671b83b94a1/src/hash.c#L44

`hash_scalar()` is currently only looking at the _first_ value in each row for matrices / arrays. So you get strange behavior where sometimes matrices are seen as equal, when they clearly are not.

``` r
library(vctrs)

x <- matrix(c(1, 2, 3, 4), c(2, 2))
y <- matrix(c(1, 2, 3, 5), c(2, 2))

# these should clearly be different!
vec_unique(list(x, y))
#> [[1]]
#>      [,1] [,2]
#> [1,]    1    3
#> [2,]    2    4

# the problem is that `has_scalar()`
# doesn't look at the entire row, just the first value
# in the row. So these are seen as different:
x <- matrix(c(2, 2, 3, 4), c(2, 2))
y <- matrix(c(1, 2, 3, 4), c(2, 2))

vec_unique(list(x, y))
#> [[1]]
#>      [,1] [,2]
#> [1,]    2    3
#> [2,]    2    4
#> 
#> [[2]]
#>      [,1] [,2]
#> [1,]    1    3
#> [2,]    2    4
```

<sup>Created on 2019-05-09 by the [reprex package](https://reprex.tidyverse.org) (v0.2.1.9000)</sup>

This also affects the rest of the `vec_unique_*()` family, and maybe the duplicate functions too.

Maybe it makes sense to `vec_slice()` the i-th row, and then iterate over it, `hash_combine()`-ing the values?

lionel-:
Closes #327.

`vec_hash()` and `vec_proxy_compare()` now transform matrices and arrays to data frames. This allows dictionary functions to behave properly.

cc @DavisVaughan

### DIFF
--- a/R/compare.R
+++ b/R/compare.R
@@ -23,7 +23,7 @@ vec_proxy_compare <- function(x, relax = FALSE) {
 
 #' @export
 vec_proxy_compare.data.frame <- function(x, relax = FALSE) {
-  x[] <- lapply(x[], vec_proxy_compare, relax = TRUE)
+  x[] <- lapply(x[], vec_proxy_compare_default, relax = TRUE)
   x
 }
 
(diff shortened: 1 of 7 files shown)


lionel-:
This is just a stopgap, I think in the long term we should treat shaped vectors natively.

DavisVaughan:
LGTM for a temp hack

PR #369, opened 2019-05-31
Replace `vec_dims()` with `vec_n_dim()`

DavisVaughan:
This PR makes 3 main changes:

- `vec_dims()` is replaced with `vec_n_dim()`
- The R level `vec_dim()` and `vec_n_dim()` functions now call their C level equivalents.
- Calling the C level `vec_dim()` function at the R level exposed a bug with data frames, which have their own `dim.data.frame()` method. `df_dim()` was added to handle this special case.

My only worry with this PR is the possibility that someone has a `dim()` method for their custom class that does something non-trivial. In that case, the C level `vec_dim()` might not do the right thing. Besides calling the R `dim()` function from C, I'm not sure how to resolve this.


lionel-:
> My only worry with this PR is the possibility that someone has a dim() method for their custom class that does something non-trivial.

The problem with R level `dim()` dispatch is that it doesn't correspond to our notion of shaped vector. A shaped vector is one with `dim` attributes, and the `dim()` generic might return dimensions for non-shaped vectors such as data frame. For this reason, I don't think it's appropriate to dispatch with `base::dim()` since it seems to belong to another paradigm.

DavisVaughan:
You don't expect this?

```r
> dim(mtcars)
[1] 32 11
```

lionel-:
It's useful for extracting with `[`, but not useful as a vctrs primitive.

DavisVaughan:
I guess some tests need to change then:
https://github.com/r-lib/vctrs/blob/c6c9c85f0152fb3275a1788c5174895b43b62dde/tests/testthat/test-bind.R#L164

lionel-:
I think this test is fine. On the other hand `has_dim()`, used internally, should never dispatch.

DavisVaughan:
Hmm this also means we get this result because `vec_ptype_abbr()` -> `vec_ptype_shape()` -> `vec_dim()` is called.

```r
> vec_ptype_abbr(mtcars)
[1] "df"
> "df[,11]"
[1] "df[,11]"
```

lionel-:
Good point, I guess that anything leaking into the user interface should reflect the two-dimensional subsetting with `[`. Maybe we should have both `vec_dim()` and `dim2()`, the latter defined as `dim(x) %||% length(x)`? This would be unexported but could be exported in funs.

DavisVaughan:
- `dim2()` is now more appropriately named `n_dim2()` to reflect that it extends to a common dimensionality
- A new `dim2()` was added to respect that `dim()` is a generic and sometimes we want data frames to look like 2d objects (i.e. `vec_ptype_abbr()`)

DavisVaughan:
@hadley any thoughts on this PR? If you think the name change is appropriate I'll resolve the merge conflicts today

hadley:
I don't really like `vec_n_dim()` as a name (shouldn't it be `vec_dim_n()`?), but I can't think of anything better.

lionel-:
Thanks!

PR #377, opened 2019-06-04
Implement `vec_cbind()` in C

lionel-:
Old marks are with current master.

```r
library("dplyr")

bench <- function(..., base = cbind) {
  bench::mark(
    vctrs = vec_cbind(...),
    base = base(...),
    dplyr = bind_cols(...),
    before = vec_cbind_old(...),
    check = FALSE
  )[1:6]
}
bench2 <- function(..., base = cbind) {
  bench::mark(
    vctrs = vec_cbind(...),
    dplyr = bind_cols(...),
    before = vec_cbind_old(...),
    check = FALSE
  )[1:6]
}


## Fast when names are unique

bench(mtcars, set_names(mtcars, ~ paste("foo_", .x)))
#>   <bch:expr> <bch:tm> <bch:tm>     <dbl> <bch:byt>    <dbl> <int> <dbl>
#> 1 vctrs        22.5µs   29.8µs    30016.      632B     6.00  9998     2
#> 2 base         52.3µs   59.9µs    15480.    1.55KB     4.20  7372     2
#> 3 dplyr       108.3µs  125.8µs     7759.   11.08KB     2.08  3727     1
#> 4 before      620.2µs    693µs     1380.  244.99KB     6.49   638     3

bench(mtcars, x = 1:32, y = 32:1)
#>   <bch:expr> <bch:tm> <bch:tm>     <dbl> <bch:byt>    <dbl> <int> <dbl>
#> 1 vctrs          22µs   26.9µs    34347.        0B     3.44  9999     1
#> 2 base          127µs    145µs     6430.      912B     4.24  3030     2
#> 3 dplyr         109µs  120.1µs     7870.    3.56KB     4.21  3739     2
#> 4 before        677µs  761.3µs     1238.   41.84KB     4.21   588     2


## Describing names repairs is slow

bench(mtcars, mtcars)
#> New names:
#> * mpg -> mpg...1
#> * cyl -> cyl...2
#> * disp -> disp...3
#> * hp -> hp...4
#> * drat -> drat...5
#> * ...
#>
#>   <bch:expr> <bch:tm> <bch:tm>     <dbl> <bch:byt>    <dbl> <int> <dbl>
#> 1 vctrs        1.32ms   1.44ms      662.   15.77KB     4.20   315     2
#> 2 base        51.91µs  58.88µs    15564.    1.55KB     4.16  7490     2
#> 3 dplyr      111.26µs  127.2µs     7566.   11.08KB     2.07  3654     1
#> 4 before     648.43µs 719.35µs     1331.   11.81KB     6.51   613     3


## Assembling many uniquely named data frames is almost as fast

args <- rep(list(mtcars), 1000)
args <- map2(args, 1:1000, function(x, i) set_names(x, ~ paste0(i, "_", .)))
bench2(!!!args)
#>   <bch:expr> <bch:tm> <bch:tm>     <dbl> <bch:byt>    <dbl> <int> <dbl>
#> 1 vctrs        8.57ms   9.74ms    94.0    401.83KB     5.76    49     3
#> 2 dplyr        7.52ms   8.01ms   119.       2.19MB     1.98    60     1
#> 3 before        1.29s    1.29s     0.773    1.36GB    11.6      1    15


## Assembling many column vectors is 5 times slower, not sure why

args <- set_names(rep(list(1:32), 1000), 1:1000)
bench2(mtcars, !!!args)
#>   <bch:expr> <bch:tm> <bch:tm>     <dbl> <bch:byt>    <dbl> <int> <dbl>
#> 1 vctrs        3.11ms   3.82ms    256.      71.5KB     9.32   110     4
#> 2 dplyr      775.26µs 800.25µs   1213.     207.1KB     2.06   588     1
#> 3 before     116.72ms 118.58ms      8.43    12.1MB     8.43     2     2
```


R/conditions.R
```diff
@@ -127,6 +127,49 @@ stop_incompatible_op <- function(op, x, y, details = NULL, ..., message = NULL,
...
+#' @export
+stop_incompatible_size <- function(x, y,
+                                   x_size, y_size,
+                                   x_arg = "", y_arg = "",
+                                   details = NULL,
+                                   ...,
+                                   message = NULL,
+                                   .subclass = NULL) {
+  vec_assert(x)
+  vec_assert(y)
```
hadley: Plus `vec_asssert(x_size, numeric(), 1)` etc ?

R/conditions.R
```diff
@@ -127,6 +127,49 @@ stop_incompatible_op <- function(op, x, y, details = NULL, ..., message = NULL,
...
+      x_name <- "vector of size"
+    }
+    if (nzchar(y_arg)) {
+      y_name <- paste0("`", y_arg, "` of size")
+    } else {
+      y_name <- "vector of size"
+    }
+
+    message <- glue_lines(
+      "No common size for {x_name} [{x_size}] and {y_name} [{y_size}].",
```
hadley: Would it be more with `stop_incompatible_type()` to generate something like ``No common size for `x`[3] and `y`[4]`` ? Otherwise I don't think you'd want the square brackets, and generate ``No common size for `x`, size 3, and `y`, size 4.``

src/bind.c
```diff
@@ -122,3 +122,175 @@ static SEXP as_df_row_impl(SEXP x, bool quiet) {
...
+  } else if (!is_data_frame(type)) {
+    type = r_as_data_frame(type);
+  }
+  UNPROTECT(1);
+  PROTECT(type);
+
+
+  R_len_t nrow;
+  if (size == R_NilValue) {
+    nrow = vec_size_common(xs);
```
hadley: Where do we check for (e.g.) `vec_cbind(NULL, NULL)` ?

src/bind.c
```diff
@@ -122,3 +122,175 @@ static SEXP as_df_row_impl(SEXP x, bool quiet) {
...
+  SEXP containers = PROTECT(map(xs, &cbind_container_type));
+  ptype = PROTECT(cbind_container_type(ptype));
+
+  SEXP type = PROTECT(vctrs_type_common_impl(containers, ptype));
+  if (type == R_NilValue) {
+    type = new_data_frame(vctrs_shared_empty_list, 0);
+  } else if (!is_data_frame(type)) {
+    type = r_as_data_frame(type);
+  }
+  UNPROTECT(1);
```
hadley: Why are you unprotecting here?

src/bind.c
```diff
@@ -122,3 +122,175 @@ static SEXP as_df_row_impl(SEXP x, bool quiet) {
...
+
+  R_len_t nrow;
+  if (size == R_NilValue) {
+    nrow = vec_size_common(xs);
+  } else {
+    nrow = size_validate(size, ".size");
+  }
+
+
+  // Validate input data frames and collect total number of columns
```
hadley:
```suggestion
  // Convert input to data frames, validate, and collect total number of columns
```

src/bind.c
```diff
@@ -122,3 +122,175 @@ static SEXP as_df_row_impl(SEXP x, bool quiet) {
...
+
+  for (R_len_t i = 0; i < n; ++i) {
+    SEXP x = VECTOR_ELT(xs, i);
+
+    if (x == R_NilValue) {
+      continue;
+    }
+
+    R_len_t xn = Rf_length(x);
+    idx_ptr[0] = counter;
```
hadley: Every time I read this idiom, I have to think a bit. I think it would be better to just have start and end parameters.

hadley: If you don't want to get rid of it, I think you should group the modification of the index with the call to `list_assign()`

src/bind.c
```diff
@@ -122,3 +122,175 @@ static SEXP as_df_row_impl(SEXP x, bool quiet) {
...
+static SEXP vec_as_df_col(SEXP x, SEXP outer);
+
+// [[ register() ]]
+SEXP vctrs_as_df_col(SEXP x, SEXP outer) {
+  return as_df_col(x, r_chr_get(outer, 0));
+}
+static SEXP as_df_col(SEXP x, SEXP outer) {
+  if (is_data_frame(x)) {
+    return df_as_df_col(x, outer);
+  }
```
hadley: Feels weird to use early-return style here. I think it would be clearer as a single if block.

NAMESPACE
```diff
@@ -426,6 +424,7 @@ export(partial_frame)
 export(s3_register)
 export(stop_incompatible_cast)
 export(stop_incompatible_op)
+export(stop_incompatible_size)
```
hadley: Needs a news bullet

R/bind.R
```diff
@@ -96,12 +96,13 @@ vec_rbind <- function(..., .ptype = NULL) {
...
 #' @rdname vec_bind
-#' @param .size If, `NULL`, the default, will determing the number of
+#' @param .size If, `NULL`, the default, will determine the number of
 #'   rows in `vec_cbind()` output by using the standard recycling rules.
 #'
 #'   Alternatively, specify the desired number of rows, and any inputs
 #'   of length 1 will be recycled appropriately.
 vec_cbind <- function(..., .ptype = NULL, .size = NULL) {
+  return(.External2(vctrs_cbind, .ptype, .size))
   args <- list2(...)
```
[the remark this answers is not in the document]
lionel-: oops

R/conditions.R
lionel-: We currently don't have a numeric type. If we're being restrictive, we should probably opt for integer rather than double.

lionel-: Perhaps we need `vec_assert_coercible()`?

R/conditions.R
lionel-: I like the second. What should be the message when no arg is passed?

```
No common size for vector, size 3, and vector, size 4.
```

or

```
No common size for vector of size 3, and vector of size 4.
```

src/bind.c
lionel-: So I can reprotect `type`.

src/bind.c
lionel-: We skip `NULL` inputs.

src/bind.c
lionel-: How about

```c
init_compact_seq(idx_ptr, counter, counter + xn);
```

src/bind.c
lionel-: Ah that's because I find chained if-else very hard to read. With early return style, it's easier for me to check the last expression of each block and determine that the control flow pattern is homogeneous.

src/bind.c
lionel-: Implemented in https://github.com/r-lib/vctrs/pull/377/commits/1d144849f77ab2a9d43e1fba1223dbb3127cd829

src/bind.c
lionel-: Would you find `PROTECT_WITH_INDEX()` easier to read? It seems a bit heavy handed when there's no intervening protections, but might make the protection pattern clearer.

R/conditions.R
hadley: I think in that case you'd just call `vec_cast()`? (and make a separate assertion about the size) At some point we need to fully consider all of the combinations of asserting, casting to fixed type, and casting to common type, and then how to most elegantly express them.

src/bind.c
```diff
@@ -75,9 +75,7 @@ static SEXP vec_rbind(SEXP xs, SEXP ptype) {
     }
 
     SEXP tbl = PROTECT(vec_cast(VECTOR_ELT(xs, i), ptype));
-    idx_ptr[0] = counter;
-    idx_ptr[1] = counter + size;
-
+    init_compact_seq(idx_ptr, counter, counter + size);
```
hadley: I mean get rid of `idx_ptr` altogether and just pass two integers. I think this change actually makes the code harder to read because it's not immediately obvious what the connection between this line and the next is. 

Wouldn't `df_assign(out, counter, counter + size, tbl, false)` be more clear?

src/bind.c
hadley: Ah ok, I just missed that.

src/bind.c
hadley: But what happens if they're all `NULL` and `xs` is empty? I guess we already have a unit test for it so it must work, it's just not super clear how.

src/bind.c
lionel-: It can't be `df_assign()`, because this is the code that takes an R index. We could have a variant with the signature you suggest but then that variant would have to allocate a compact idx at each iteration, which can be many when assembling vectors with `!!!`. Alternatively we could duplicate all the slice/assign functions, but I think this wouldn't be desirable.

src/bind.c
lionel-: `nrow` will be 0 in that case. Then we cast an empty list to a data frame (the common prototype).

PR #378, opened 2019-06-04
Assert that `vec_assign()`'s `value` argument must be a vector

### LINKED ISSUE #285: What should `vec_slice(x, i) <- NULL` do?
In the spirit of having good error messages for edge cases, what should this do? Is this a no-op or an error?

``` r
library(vctrs)
x <- 1:5
vec_slice(x, 1L) <- NULL
#> Error in x[i] <- value: replacement has length zero
```

DavisVaughan:
Closes #285 

Rather than special casing `NULL`, this felt more appropriate. Since we already `vec_assert()` that `x` is a vector, it makes sense that `value` must be one too.

``` r
library(vctrs)
x <- 1
vec_slice(x, 1L) <- NULL
#> `value` must be a vector, not NULL
vec_slice(x, 1L) <- environment()
#> `value` must be a vector, not an environment
```

<sup>Created on 2019-06-04 by the [reprex package](https://reprex.tidyverse.org) (v0.2.1.9000)</sup>

### DIFF
--- a/src/slice-assign.c
+++ b/src/slice-assign.c
@@ -29,6 +29,7 @@ SEXP vec_assign(SEXP x, SEXP index, SEXP value) {
   struct vctrs_arg x_arg = new_wrapper_arg(NULL, "x");
   struct vctrs_arg value_arg = new_wrapper_arg(NULL, "value");
   vec_assert(x, &x_arg);
+  vec_assert(value, &value_arg);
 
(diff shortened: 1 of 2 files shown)


hadley:
(@DavisVaughan you should feel free to merge this yourself, and in general, you should feel free to merge your PRs once Lionel or I have approved)

PR #386, opened 2019-06-06
DRY repair values

krlmlr:
We want to put the possible values for `repair` into the signature of `vec_as_names()`, but we want to check in `validate_repair()`.

### DIFF
--- a/tests/testthat/test-names.R
+++ b/tests/testthat/test-names.R
@@ -1,6 +1,13 @@
 context("test-names")
 
 
+# API -----------------------------------------------------------------
+
+test_that("Consistent values for repair argument", {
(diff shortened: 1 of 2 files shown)


tests/testthat/test-names.R
```diff
@@ -1,6 +1,13 @@
 context("test-names")
 
 
+# API -----------------------------------------------------------------
+
+test_that("Consistent values for repair argument", {
+  expect_identical(formals(validate_repair)$repair, formals(vec_as_names)$repair)
+  expect_identical(formals(validate_repair)$repair, formals(vec_names2)$repair)
+})
+
```
lionel-:
```suggestion


```

lionel-: > You do not have permission to push to this repository

:(

PR #388, opened 2019-06-06
Rename to avoid confusion in RStudio IDE

krlmlr:
Ideally, the IDE would know...

### DIFF
--- a/src/names.c
+++ b/src/names.c
@@ -5,7 +5,7 @@
 
 #include <ctype.h>
 
-static void describe_repair(SEXP old, SEXP new);
+static void describe_repair(SEXP old_names, SEXP new_names);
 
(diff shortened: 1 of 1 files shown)


lionel-:
What confusion is this fixing?

lionel-:
I'm a bit reticent to introduce ad hoc style changes that can't be implemented consistently and intuitively by code authors.

krlmlr:
RStudio IDE thinks `new` is a keyword and messes up the syntax highlighting and highlights errors. It's not, and it shouldn't.

lionel-:
Then I'd like proper names instead of an underscore suffix, but I don't think we should bother about this.

krlmlr:
It's the only instance of `new` here.

PR #389, opened 2019-06-06
Use isdigit()

krlmlr:
instead of own `switch()`, but beware signed-unsigned character trap.

### DIFF
--- a/src/names.c
+++ b/src/names.c
@@ -1,3 +1,5 @@
+#include <ctype.h>
+
 #include "vctrs.h"
 #include "utils.h"
 
@@ -213,11 +215,6 @@ static bool is_dotdotint(const char* name) {
   return (bool) strtol(name, NULL, 10);
 }
 
-// Wrap for casting to unsigned char
-static bool is_digit(const unsigned char c) {
-  return isdigit(c);
-}
-
 static ptrdiff_t suffix_pos(const char* name) {
   int n = strlen(name);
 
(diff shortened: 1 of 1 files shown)


src/names.c
```diff
@@ -213,22 +213,9 @@ static bool is_dotdotint(const char* name) {
...
-  case '6':
-  case '7':
-  case '8':
-  case '9':
-    return true;
-  default:
-    return false;
-  }
+// Wrap for casting to unsigned char
+static bool is_digit(const unsigned char c) {
```
lionel-: The signedness of `char` is implementation defined. `isdigit()` takes an `int` (presumably so that `EOF` can be passed). It seems this is doing an unnecessary roundtrip of signedness casting here. As long as we pass `char`, this should be fine.

krlmlr: It's convoluted, but https://stackoverflow.com/a/45007070/946850 suggests it's important.

lionel-: Interesting, the important bit seems to be:

> The problem is not just a theoretical one: several widespread C library implementations use the provided value straight as an index into a lookup table, so negative values will read into unallocated memory and segfault your program.

Should we rather use `unsigned char` everywhere we need bit data rather than integers?

krlmlr: `isdigit()` is for character data, they come as `char*` from `CHAR()`. Other than that, perhaps `typedef unsigned char byte` would be useful.

lionel-: We could define `UCHAR()` and `UCHAR_RO()` as well.

PR #390, opened 2019-06-06
Fix two minor glitches in name repair

krlmlr:
- Need duplication
- Need to copy encoding information

### DIFF
--- a/src/names.c
+++ b/src/names.c
@@ -6,7 +6,7 @@
 static void describe_repair(SEXP old, SEXP new);
 
 // 3 leading '.' + 1 trailing '\0' + 24 characters
-static const int max_iota_size = 28;
+static const int MAX_IOTA_SIZE = 28;
 
 // [[ register(); include("vctrs.h") ]]
 SEXP vec_names(SEXP x) {
@@ -161,7 +161,7 @@ SEXP as_unique_names_impl(SEXP names, bool quiet) {
     const char* name = CHAR(elt);
 
     int size = strlen(name);
-    int buf_size = size + max_iota_size;
+    int buf_size = size + MAX_IOTA_SIZE;
 
     R_CheckStack2(buf_size);
     char buf[buf_size];
@@ -328,8 +328,8 @@ SEXP vctrs_unique_names(SEXP x, SEXP quiet) {
 
 
 static SEXP names_iota(R_len_t n) {
-  char buf[max_iota_size];
-  SEXP nms = r_chr_iota(n, buf, max_iota_size, "...");
+  char buf[MAX_IOTA_SIZE];
+  SEXP nms = r_chr_iota(n, buf, MAX_IOTA_SIZE, "...");
 
   if (nms == R_NilValue) {
     Rf_errorcall(R_NilValue, "Too many names to repair.");
--- a/src/names.c
+++ b/src/names.c
@@ -123,7 +123,7 @@ SEXP as_unique_names_impl(SEXP names, bool quiet) {
   const R_len_t n = Rf_length(names);
 
   SEXP new_names = PROTECT(Rf_shallow_duplicate(names));
-  SEXP* const new_names_ptr = STRING_PTR(new_names);
+  const SEXP* new_names_ptr = STRING_PTR_RO(new_names);
 
   for (R_len_t i = 0; i < n; ++i) {
     SEXP elt = new_names_ptr[i];
--- a/src/names.c
+++ b/src/names.c
@@ -1,8 +1,12 @@
 #include "vctrs.h"
 #include "utils.h"
 
+#include <ctype.h>
+
 static void describe_repair(SEXP old, SEXP new);
 
+// 3 leading '.' + 1 trailing '\0' + 24 characters
+static const int max_iota_size = 28;
 
 // [[ register(); include("vctrs.h") ]]
 SEXP vec_names(SEXP x) {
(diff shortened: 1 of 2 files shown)


src/names.c
```diff
@@ -90,102 +95,95 @@ SEXP as_unique_names(SEXP names, bool quiet) {
   }
 
   R_len_t i = 0;
-  R_len_t n = Rf_length(names);
-  SEXP* ptr = STRING_PTR(names);
+  const R_len_t n = Rf_length(names);
```
lionel-: Is there a practical reason to use `const` here, such as better assembly output? I'd rather not overdo const-qualifiers as a matter of style.

src/names.c
```diff
@@ -90,102 +95,95 @@ SEXP as_unique_names(SEXP names, bool quiet) {
...
+  SEXP* const names_ptr = STRING_PTR(names);
```
lionel-:
```suggestion
  const SEXP* names_ptr = STRING_PTR_RO(names);
```

src/names.c
```diff
@@ -90,102 +95,95 @@ SEXP as_unique_names(SEXP names, bool quiet) {
...
 
   SEXP dups = PROTECT(vctrs_duplicated(names));
-  int* dups_ptr = LOGICAL(dups);
+  int* const dups_ptr = LOGICAL(dups);
```
lionel-:
```suggestion
  const int* dups_ptr = LOGICAL_RO(dups);
```

src/names.c
```diff
@@ -90,102 +95,95 @@ SEXP as_unique_names(SEXP names, bool quiet) {
...
+}
 
-  ptr = STRING_PTR(names);
+SEXP as_unique_names_impl(SEXP names, bool quiet) {
+  const R_len_t n = Rf_length(names);
 
-  for (; i < n; ++i, ++ptr) {
-    SEXP elt = *ptr;
+  SEXP new_names = PROTECT(Rf_shallow_duplicate(names));
+  SEXP* const new_names_ptr = STRING_PTR(new_names);
```
lionel-:
```suggestion
  const SEXP* new_names_ptr = STRING_PTR_RO(new_names);
```

src/names.c
```diff
@@ -90,102 +95,95 @@ SEXP as_unique_names(SEXP names, bool quiet) {
...
   }
 
   // Append all duplicates with a suffix
-  char buf[100] = "";
-  ptr = STRING_PTR(names);
 
-  dups = PROTECT(vctrs_duplicated(names));
-  dups_ptr = LOGICAL(dups);
+  SEXP dups = PROTECT(vctrs_duplicated(new_names));
+  int* const dups_ptr = LOGICAL(dups);
```
lionel-:
```suggestion
  const int* dups_ptr = LOGICAL_RO(dups);
```

src/names.c
```diff
@@ -1,8 +1,12 @@
 #include "vctrs.h"
 #include "utils.h"
 
+#include <ctype.h>
+
 static void describe_repair(SEXP old, SEXP new);
 
+// 3 leading '.' + 1 trailing '\0' + 24 characters
+static const int max_iota_size = 28;
```
lionel-: I would like to keep global constants upper case, as it was before.

src/names.c
```diff
@@ -90,102 +95,95 @@ SEXP as_unique_names(SEXP names, bool quiet) {
...
     int pos = suffix_pos(nm);
     if (pos >= 0) {
-      R_CheckStack2(pos + 1);
-      char buf[pos + 1];
-      memcpy(buf, nm, pos);
-      buf[pos] = '\0';
-
-      elt = Rf_mkChar(buf);
-      SET_STRING_ELT(names, i, elt);
+      elt = Rf_mkCharLenCE(nm, pos, Rf_getCharCE(elt));
```
lionel-: Nice!

src/names.c
krlmlr: It does ease the burden on the optimizer, but this is not the point.

It's the guarantee that the value of `n` doesn't change, in line with pure functional programming style. Anyone reading the code immediately knows what the value of `n` must be at any occurrence. Occasionally this style will catch errors, when a typo would lead to updating the wrong variable.

It's too bad that the `const` adds so much noise. Maybe with typedefs, or written as `R_len_t const n = ...` ?

src/names.c
krlmlr: First mention of `STRING_PTR_RO` in wch/r-source@6ac7b5670cd (Nov 2017), but now I see we have compat code. Neat!

This is a constant pointer to (modifiable) `SEXP`, not a modifiable pointer to const `SEXP`: https://stackoverflow.com/a/1143272/946850. If we chose to switch to const qualifiers everywhere, we would write:

```suggestion
  SEXP const * const names_ptr = STRING_PTR_RO(names);
```

krlmlr:
Removed the `const` (we can discuss const-ness elsewhere if you want), all-caps static constant.

src/names.c
lionel-: I don't think this is important and that we should use `const` everywhere just because we can. This overall hinders reading code. I'd rather use it with data arrays, and where it actually helps the compiler.

src/names.c
lionel-: I don't want to mark the pointers as const in this codebase.

lionel-: What's important here is that the array can't change:

```c
const SEXP* data = STRING_PTR_RO(i);

data[3] = R_NilValue;

   slice.c:408:9: error: read-only variable is not assignable
     ip[3] = R_NilValue;
     ~~~~~ ^
   1 error generated.
```

krlmlr:
I removed all undesired instances of `const` here.

lionel-:
Thanks!

PR #391, opened 2019-06-06
Test invariants

### LINKED ISSUE #245: Provide implementation for vec_data(), preserving shape and vec_names
Invariants: 

```r
identical(vec_names(vec_data(x)), vec_names(x))
identical(vec_dim(vec_data(x)), vec_dim(x))
```

The documentation currently states that names are preserved. The names of a vector can also be row names (of a data frame or matrix).

Ideally, an ALTREP implementation would avoid copying.

``` r
library(vctrs)

x <- diag(2)
rownames(x) <- letters[1:2]
colnames(x) <- LETTERS[1:2]

vctrs:::vec_names(x)
#> [1] "a" "b"

x
#>   A B
#> a 1 0
#> b 0 1

vec_data(x)
#> [1] 1 0 0 1
```

<sup>Created on 2019-03-23 by the [reprex package](https://reprex.tidyverse.org) (v0.2.1.9000)</sup>

krlmlr:
Implementation already in place.

Closes #245.

### DIFF
--- a/tests/testthat/test-data.R
+++ b/tests/testthat/test-data.R
@@ -60,6 +60,7 @@ test_that("vec_data() preserves names (#245)", {
   rownames(x) <- letters[1:2]
   colnames(x) <- LETTERS[1:2]
   expect_identical(vec_names(x), vec_names(vec_data(x)))
+  expect_identical(dimnames(x), dimnames(vec_data(x)))
 })
 
 test_that("vec_data() preserves size (#245)", {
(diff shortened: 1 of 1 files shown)


DavisVaughan:
It looks like we already test that names are preserved here:
https://github.com/r-lib/vctrs/blob/331f79973fca5450dfca82a7edf31a1a2a0b078b/tests/testthat/test-data.R#L3

And test that dim / dimnames are preserved here:
https://github.com/r-lib/vctrs/blob/331f79973fca5450dfca82a7edf31a1a2a0b078b/tests/testthat/test-data.R#L11

But personally I like your tests more as the original tests are doing two things, testing that some attributes are stripped, and testing that some are kept. 

Maybe you could replace those two linked tests with your 3 new ones. And then if you add a test that is purely about stripping attributes (maybe one test on a 1D vctr and one on a 2D vctr created with `new_vctr()`) then that would be very clear and cover all the bases.

krlmlr:
Thanks, done.

lionel-:
Thanks!

PR #394, opened 2019-06-08
Alter `vec_as_index()` signature

### LINKED ISSUE #381: Export `vec_as_index()`
@lionel- I believe you mentioned you were planning on exporting this, correct? I use it in rray (although I use it over any dimension, not just the 1st dimension, which requires a bit of a hack).

DavisVaughan:
Closes #381 

- `vec_as_index()` has changed from `vec_as_index(i, x)` -> `vec_as_index(i, size, names = NULL)`. This more accurately captures the required information used in this function, and makes it generic enough to work over any dimension, rather than just the first dimension.

- `vctrs_as_index()` is now a required function at the C level (because of the `size` arg), and has become the new bridge function between C and R.

- Two new tests were added regarding input validation for `size` and `names`.

- Internal C level naming consistency was improved by this PR. Rather than using `vec_n` or `data_n` for `vec_size(x)`, now only `size` is used.

- I added a check to `chr_as_index()` to ensure that `names` is a character vector (if not already caught by the `NULL` check). Since `names` is now an argument the user can pass in, this felt like an appropriate check.

- I added a `vec_assert()` for `size` at the R level to ensure that it is a length 1 integer.

I have tested this in rray, and it solves the issue raised in #381.

It feels appropriate to export `vec_as_index()` in a separate PR, so I will do that after this gets merged.

Question: I currently use `size` as the new argument name, which is consistent with how it is used in vctrs, but for rray I technically pass in the length of any dimension here. Should it just be `n`? If so, it will require a bit of renaming at the C level since we currently use `n` in many places for the index length.

Question: Should `names` default to `NULL`? They are only required when `i` is character and it felt odd repeatedly specifying it as `NULL` in the tests when `i` was not a character.


DavisVaughan:
I also found an error in the ordering of the arguments in the error message for `lgl_as_index()`. Since I was already touching it, I fixed it in cd8e8ea

src/slice-assign.c
```diff
@@ -35,7 +35,7 @@ SEXP vec_assign(SEXP x, SEXP index, SEXP value) {
   value = PROTECT(vec_coercible_cast(value, x, &value_arg, &x_arg));
   SEXP value_proxy = PROTECT(vec_proxy(value));
 
-  index = PROTECT(vec_as_index(index, x));
+  index = PROTECT(vec_as_index(index, vec_size(x), vec_names(x)));
```
lionel-: `vec_names()` needs a `PROTECT()` because `Rf_getAttrib()` might return a fresh SEXP. Alternatively we could import the attribute getter implemented in rlang which doesn't allocate (but then you can't get  names vector for pairlists for instance).

DavisVaughan: I've just `PROTECT()`ed in two places for now.

DavisVaughan:
- Documented and exported `vec_as_index()` 
- Used `n` rather than `size` in `vec_as_index()`

R/slice.R
```diff
@@ -125,8 +125,45 @@ vec_assign_fallback <- function(x, i, value) {
...
+#' Create an index
+#'
+#' This provides a means of standardizing common indexing methods such as
+#' integer, character or logical indexing. The input is a vector of one
+#' of these three types, and the output is always an integer vector that can
+#' be used for subsetting.
+#'
+#' @inheritParams vec_slice
+#' @param n A single integer representing the total size of the
+#' object that `i` is meant to index into.
```
lionel-: The `@param` paragraphs need 2 spaces indent.

PR #400, opened 2019-06-12
Recursive reset_rownames()

krlmlr:
to support deeply nested data frames.

Probably harmless now, but perhaps this function will be used later in other contexts.


lionel-:
Though we'll move towards supporting row names in the near future.

PR #405, opened 2019-06-13
Extract is_unique_names() and duplicated_any()

krlmlr:
to further simplify name repair and to prepare for `"check_unique"`.

### DIFF
--- a/src/names.c
+++ b/src/names.c
@@ -92,33 +92,35 @@ static bool needs_suffix(SEXP str);
 
 // [[ include("vctrs.h") ]]
 SEXP as_unique_names(SEXP names, bool quiet) {
+  if (is_unique_names(names)) {
+    return names;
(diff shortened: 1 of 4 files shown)


src/names.c
```diff
@@ -92,33 +93,48 @@ static bool needs_suffix(SEXP str);
...
     Rf_errorcall(R_NilValue, "`names` must be a character vector");
   }
 
-  R_len_t i = 0;
   R_len_t n = Rf_length(names);
   const SEXP* names_ptr = STRING_PTR_RO(names);
 
-  SEXP dups = PROTECT(vctrs_duplicated(names));
-  const int* dups_ptr = LOGICAL_RO(dups);
+  if (duplicated_any(names)) return false;
```
lionel-: Return statement needs to be on its own line.

PR #407, opened 2019-06-13
Tweak minimal repair

krlmlr:
I can't show any failures with the original code, but I believe unconditional duplication is correct here.

### DIFF
--- a/src/names.c
+++ b/src/names.c
@@ -54,7 +54,7 @@ SEXP vctrs_as_minimal_names(SEXP names) {
     return names;
   }
 
-  names = PROTECT(r_maybe_duplicate(names));
+  names = PROTECT(Rf_shallow_duplicate(names));
 
   for (; i < n; ++i, ++ptr) {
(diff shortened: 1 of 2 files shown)


lionel-:
Not that it matters here since the closure wrapper will always bump namedness. But if it didn't, conditional duplication would be correct.

When called from C that's another thing. On the one hand, we'd like to avoid duplication when the caller owns the object. On the other hand, it's very easy from C code to get a foreign object and forget to mark it shared (get an attribute, extract a list component, etc).

krlmlr:
I vividly remember in-place modifications when calling `vec_as_names()` from R code before #390 (as in the tests I've added), but I can no longer replicate them.

Maybe we can design internal APIs to have `xxx_in_place()` versions for all `xxx()` that operate on a vector? The `_in_place()` version would assume exclusive ownership.

lionel-:
That would be an option, though we may not need it here. Making a first pass without copy should be enough.

PR #408, opened 2019-06-13
Support check_unique repair

krlmlr:
using a simple R implementation for now. Can move to C after #405.

This is for consistency with tibble's interface.


lionel-:
Thanks, I have rewritten this to follow existing conventions and file names. FYI, we generally construct conditions using the signaller functions.

I like the idea of adding `vctrs_error` class to all vctrs errors to enable package-wide filtering. I'll send another PR to use this in other conditions.

krlmlr:
Thanks, looks better. The tests in tibble still pass.

PR #412, opened 2019-06-14
Inline `can_s3_register_now()`

### LINKED ISSUE #371: s3_register() can't call internal functions
Because you should be able to copy and paste the source into another package.

DavisVaughan:
Closes #371 

- Inlines `can_s3_register_now()` so `s3_register()` stands alone
- Uses `isNamespaceLoaded()` which was added in R 3.2.0
- Stores `asNamespace(package)` as `envir` rather than calling it twice

I've reordered things a bit as well for maximal clarity. The hook is set first, and then we have a potential early exit if the namespace is not loaded. Then we only register if we can also find the generic in the namespace. I think this makes it a bit more readable, and the comments read in a nice order.

### DIFF
--- a/R/register-s3.R
+++ b/R/register-s3.R
@@ -64,7 +64,7 @@ s3_register <- function(generic, class, method = NULL) {
 
   # Avoid registration failures during loading (pkgload or regular)
   if (!isNamespaceLoaded(package)) {
-    return()
+    return(invisible())
   }
 
   envir <- asNamespace(package)
@@ -73,6 +73,8 @@ s3_register <- function(generic, class, method = NULL) {
   if (exists(generic, envir)) {
     registerS3method(generic, class, method, envir = envir)
   }
+
+  invisible()
 }
 
 # nocov end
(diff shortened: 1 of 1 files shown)


DavisVaughan:
Not that it matters much here, but the claim in the docs that `isNamespaceLoaded()` is faster than `pkg %in% loadedNamespaces()` seems true

```r
> bench::mark(isNamespaceLoaded("generics"), "generics" %in% loadedNamespaces())
# A tibble: 2 x 14
  expression                    min     mean   median      max `itr/sec` mem_alloc  n_gc n_itr total_time result   memory      time   gc          
  <chr>                    <bch:tm> <bch:tm> <bch:tm> <bch:tm>     <dbl> <bch:byt> <dbl> <int>   <bch:tm> <list>   <list>      <list> <list>      
1 "isNamespaceLoaded(\"ge…    290ns 328.22ns    315ns   34.2µs  3046773.        0B     0 10000     3.28ms <lgl [1… <Rprofmem … <bch:… <tibble [10…
2 "\"generics\" %in% load…   1.44µs   1.77µs   1.61µs  164.7µs   564951.      864B     0 10000     17.7ms <lgl [1… <Rprofmem … <bch:… <tibble [10…
```

hadley:
300x performance improvement 😛

PR #413, opened 2019-06-14
Construct `base::order()` call with `expr()`

### LINKED ISSUE #356: Can't sort empty data frame
``` r
library(vctrs)

vec_sort(data.frame())
#> Error in `$<-.data.frame`(`*tmp*`, "decreasing", value = FALSE): replacement has 1 row, data has 0
```

<sup>Created on 2019-05-28 by the [reprex package](https://reprex.tidyverse.org) (v0.2.1.9000)</sup>

DavisVaughan:
Closes #356 

This failed because we were trying to attach the `base::order()` arguments to the unnamed data frame, which had 0 rows.

```
> vec_sort(data.frame())
Error in `$<-.data.frame`(`*tmp*`, "decreasing", value = FALSE) : 
  replacement has 1 row, data has 0
```

Instead, I now construct the `base::order()` expression with `expr()`, using `!!!` for the unnamed data frame.

This results in the following behavior:

``` r
library(vctrs)

vec_order(data.frame())
#> NULL

# consistent with order() called with no `...` args
# because that is essentially what happens, but maybe we want something
# different?
order()
#> NULL

vec_sort(data.frame())

vec_order(data.frame(x = numeric()))
#> integer(0)

vec_sort(data.frame(x = numeric()))
#> [1] x
```

I feel like it might be appropriate for `vec_order(data.frame())` to return `integer(0)`, but I think the only way to do that is an early exit if `vec_size(x) == 0`

### DIFF
--- a/R/compare.R
+++ b/R/compare.R
@@ -144,7 +144,7 @@ order_proxy <- function(proxy, direction = "asc", na_value = "largest") {
   if (is.data.frame(proxy)) {
     args <- unname(proxy)
     order_expr <- expr(
-      base::order(!!! args, decreasing = decreasing, na.last = na.last)
+      base::order(!!!args, decreasing = decreasing, na.last = na.last)
     )
     eval_bare(order_expr)
   } else if (is_character(proxy) || is_logical(proxy) || is_integer(proxy) || is_double(proxy)) {
--- a/R/compare.R
+++ b/R/compare.R
@@ -142,6 +142,11 @@ order_proxy <- function(proxy, direction = "asc", na_value = "largest") {
   }
 
   if (is.data.frame(proxy)) {
+    # Avoid empty data frames getting a `NULL` order
+    if (vec_size(proxy) == 0L) {
+      return(integer(0L))
+    }
+
     args <- unname(proxy)
     order_expr <- expr(
       base::order(!!!args, decreasing = decreasing, na.last = na.last)
--- a/R/compare.R
+++ b/R/compare.R
@@ -143,10 +143,10 @@ order_proxy <- function(proxy, direction = "asc", na_value = "largest") {
 
   if (is.data.frame(proxy)) {
     args <- unname(proxy)
-    args$decreasing <- decreasing
-    args$na.last <- na.last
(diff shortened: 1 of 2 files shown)


R/compare.R
```diff
@@ -143,10 +143,10 @@ order_proxy <- function(proxy, direction = "asc", na_value = "largest") {
 
   if (is.data.frame(proxy)) {
     args <- unname(proxy)
-    args$decreasing <- decreasing
-    args$na.last <- na.last
-
-    do.call(base::order, args)
+    order_expr <- expr(
+      base::order(!!! args, decreasing = decreasing, na.last = na.last)
```
lionel-: No space after `!!!` or `!!`.

R/compare.R
```diff
@@ -142,11 +142,16 @@ order_proxy <- function(proxy, direction = "asc", na_value = "largest") {
...
-    args$decreasing <- decreasing
-    args$na.last <- na.last
+    # Avoid empty data frames getting a `NULL` order
+    if (vec_size(proxy) == 0L) {
+      return(integer(0L))
+    }
 
-    do.call(base::order, args)
+    args <- unname(proxy)
+    order_expr <- expr(
```
hadley: Wouldn't `exec()` be simpler here?

R/compare.R
```diff
@@ -142,11 +142,16 @@ order_proxy <- function(proxy, direction = "asc", na_value = "largest") {
   }
 
   if (is.data.frame(proxy)) {
-    args <- unname(proxy)
-    args$decreasing <- decreasing
-    args$na.last <- na.last
+    # Avoid empty data frames getting a `NULL` order
```
hadley: I think it's more like: Work around type-instability in base::order()

R/compare.R
DavisVaughan: Oh! Definitely. Great example of `exec()` being nicer than `do.call()`

R/compare.R
DavisVaughan: Makes sense, I'll change

PR #414, opened 2019-06-14
Followup tweaks to #413

DavisVaughan:
- Improved comment
- Use `exec()` rather than `expr()` + `eval_bare()`

### DIFF
--- a/R/compare.R
+++ b/R/compare.R
@@ -147,7 +147,7 @@ order_proxy <- function(proxy, direction = "asc", na_value = "largest") {
       return(integer(0L))
     }
     args <- unname(proxy)
-    exec(base::order, !!!args, decreasing = decreasing, na.last = na.last)
+    exec("order", !!!args, decreasing = decreasing, na.last = na.last)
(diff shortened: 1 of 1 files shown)


R/compare.R
```diff
@@ -142,16 +142,12 @@ order_proxy <- function(proxy, direction = "asc", na_value = "largest") {
...
     if (vec_size(proxy) == 0L) {
       return(integer(0L))
     }
-
     args <- unname(proxy)
-    order_expr <- expr(
-      base::order(!!!args, decreasing = decreasing, na.last = na.last)
-    )
-    eval_bare(order_expr)
+    exec(base::order, !!!args, decreasing = decreasing, na.last = na.last)
```
lionel-: Maybe `"order"` to improve backtraces in case `order()` fails?

DavisVaughan: Is the backtrace also going to be ugly since `!!!args` is being done?

lionel-: No because these are scalars. Also the function is most important because it helps to identify the source of a problem.

lionel-: (oops sorry these are not scalars)

PR #419, opened 2019-06-17
Fix documentation of binding invariants

### LINKED ISSUE #286: Inconsistent behavior of `vec_rbind()` with 0-length input
I don't think these first two are quite right. They break this invariant:

```r
vec_size(vec_rbind(x, y)) == vec_size(x) + vec_size(y)
```

``` r
library(vctrs)

# not right?
vec_rbind(integer(), 1)
#> New names:
#>  -> ..1
#>   ..1
#> 1  NA
#> 2   1

# not right?
vec_rbind(integer(), data.frame(x = 1))
#>    x
#> 1 NA
#> 2  1

# this seems right
vec_rbind(integer(), integer())
```

Should the result just have 1 row, with `integer()` just being used to help determine the output container type?

lionel-:
Closes #286

### DIFF
--- a/R/bind.R
+++ b/R/bind.R
@@ -6,8 +6,26 @@
 #' across all arguments.
 #'
 #' @section Invariants:
+#'
+#' All inputs are first converted to a data frame. The conversion for
+#' 1d vectors depends on the direction of binding:
+#'
+#' * For `vec_rbind()`, each element of the vector becomes a column in
+#'   a single row.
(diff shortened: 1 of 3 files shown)


R/bind.R
```diff
@@ -6,9 +6,27 @@
 #' across all arguments.
 #'
 #' @section Invariants:
+#'
+#' All inputs are first converted to a data frame. How simple vectors
```
hadley: The conversion for 1d vectors depends on the direction of binding:

PR #420, opened 2019-06-17
Add `vctrs_error` class to all structured vctrs errors

lionel-:
To make it easy to filter errors based on their provenance.

### DIFF
--- a/tests/testthat/test-conditions.R
+++ b/tests/testthat/test-conditions.R
@@ -1,0 +1,17 @@
+context("conditions")
+
+test_that("conditions inherit from `vctrs_error`", {
+  expect_error(stop_incompatible(NULL, NULL), class = "vctrs_error")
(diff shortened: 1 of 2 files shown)


lionel-:
In conditions.R, it was introduced by Kirill for names errors in https://github.com/r-lib/vctrs/pull/408.

tests/testthat/test-conditions.R
```diff
@@ -0,0 +1,17 @@
+context("conditions")
+
+test_that("conditions inherit from `vctrs_error`", {
+  expect_error(stop_incompatible(NULL, NULL), class = "vctrs_error")
```
krlmlr: Do you want to test the wording of the error messages here or elsewhere? I've found this useful in tibble, especially for complex messages.

lionel-: The wordings should already be tested elsewhere.

PR #423, opened 2019-06-17
Add name repair parameter to `vec_cbind()`, `vec_rbind()`, and `vec_c()`

### LINKED ISSUE #227: vec_cbind() needs .name_repair argument
which has sensible default

lionel-:
Closes #227.

The micro-optimisation of inlining `.name_repair` in the formals list buys 0.6us:

```r
bench::mark(
  before = vec_c_old(),
  after = vec_c()
)[1:8]
#>   <bch:expr> <bch:tm> <bch:tm>     <dbl> <bch:byt>    <dbl> <int> <dbl>
#> 1 before       2.48µs    3.1µs   303418.        0B      0   10000     0
#> 2 after         1.9µs   2.49µs   389138.        0B      0   10000     0
```

### DIFF
--- a/R/bind.R
+++ b/R/bind.R
@@ -36,6 +36,9 @@
 #'
 #'   `NULL` inputs are silently ignored. Empty (e.g. zero row) inputs
 #'   will not appear in the output, but will affect the derived `.ptype`.
+#' @param .name_repair One of `"unique"`, `"universal"`, or
+#'   `"check_unique"`. See [vec_as_names()] for the meaning of these
+#'   options.
 #' @inheritParams vec_c
 #' @return A data frame, or subclass of data frame.
 #'
@@ -108,9 +111,12 @@ NULL
 
 #' @export
 #' @rdname vec_bind
-vec_rbind <- function(..., .ptype = NULL) {
-  .External2(vctrs_rbind, .ptype)
+vec_rbind <- function(...,
+                      .ptype = NULL,
+                      .name_repair = c("unique", "universal", "check_unique")) {
+  .External2(vctrs_rbind, .ptype, .name_repair)
 }
+vec_rbind <- fn_inline_formals(vec_rbind, ".name_repair")
 
 #' @export
 #' @rdname vec_bind
@@ -119,9 +125,6 @@ vec_rbind <- function(..., .ptype = NULL) {
 #'
 #'   Alternatively, specify the desired number of rows, and any inputs
 #'   of length 1 will be recycled appropriately.
-#' @param .name_repair One of `"unique"`, `"universal"`, or
-#'   `"check_unique"`. See [vec_as_names()] for the meaning of these
-#'   options.
 vec_cbind <- function(...,
                       .ptype = NULL,
                       .size = NULL,
--- a/R/bind.R
+++ b/R/bind.R
@@ -128,7 +128,7 @@ vec_rbind <- fn_inline_formals(vec_rbind, ".name_repair")
 vec_cbind <- function(...,
                       .ptype = NULL,
                       .size = NULL,
-                      .name_repair = c("unique", "universal", "check_unique")) {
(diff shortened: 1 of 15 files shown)


R/bind.R
```diff
@@ -36,6 +36,9 @@
 #'
 #'   `NULL` inputs are silently ignored. Empty (e.g. zero row) inputs
 #'   will not appear in the output, but will affect the derived `.ptype`.
+#' @param .name_repair One of `"unique"`, `"universal"`, or
```
jennybc: Is there some reason the `"minimal"` level is not relevant here?

lionel-: I didn't add it because data frames must have unique names. Now I see `tibble::tibble()` allows duplicated names when `"minimal"` repair name is specified. This is surprising to me. What is the point of allowing the creation of malformed data frames in a high-level UI?

jennybc: From https://principles.tidyverse.org/names-attribute.html#unique-names 3.5.3 "When is minimal better than unique?"

> Why would you ever want to import a tibble and enforce only minimal names, instead of unique? Sometimes the first row of a data source – allegedly variable names – actually contains data and the resulting tibble will be reshaped with, e.g., tidyr::gather(). In this case, it is better to not munge the names at import. This is a common special case of the “data stored in names” phenomenon.

> In general, you may want to tolerate minimal names when the dysfunctional names are just an awkward phase that an object is passing through and a more definitive solution is applied downstream.

I definitely saw specific examples when we were working on the revdeps, too.

lionel-: hmm following this reasoning minimal repair could make sense for `vec_cbind()`. However I'm not sure about `vec_rbind()` where columns are actually matched together based on their names. If you have duplicates, you'd be matching to the first occurrence, which seems rather ad hoc and prone to errors. But then we'd need to document why cbind accepts minimal repairs but not rbind. Since we don't have the problem of backward compatibility, maybe it makes sense to omit minimal repairs until the need actually comes up?

Another question: Do we want to be consistent with tibble and use the stricter `"check_unique"` option as default? cc @hadley who wrote the first implementation.

jennybc: I think it may be reasonable to have different name repair default for `vec_cbind()` and `vec_rbind()`.

Question: is `.name_repair` applied to the inputs to the putative output or both? 🤔 I'm not sure we've really articulated that.

This dplyr code suggests that `.name_repair` is a check on inputs but doesn't dictate name repair on the output.

``` r
library(tidyverse)

df1 <- tibble(a = 1, b = 2, a = 3, .name_repair = "minimal")
df2 <- tibble(a = 1, c = 4)
df3 <- tibble(a = 1, d = 5)

left_join(df1, df2)
#> Error: Column `a` must have a unique name
left_join(df3, df2)
#> Joining, by = "a"
#> # A tibble: 1 x 3
#>       a     d     c
#>   <dbl> <dbl> <dbl>
#> 1     1     5     4

df4 <- tibble(e = 1, f = 6)
df5 <- tibble(e = 1, f = 7)

left_join(df4, df5, by = "e")
#> # A tibble: 1 x 3
#>       e   f.x   f.y
#>   <dbl> <dbl> <dbl>
#> 1     1     6     7
```

<sup>Created on 2019-06-18 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0.9000)</sup>

lionel-: I have added the minimal option to `vec_cbind()`.

`vec_rbind()` checks the inputs, while `vec_cbind()` checks the output. The difference makes sense given the way the data frames are constructed (standardisation then matching of columns for rbind, simple concatenation for cbind).

hadley: I think you need a little more documentation here along the lines of your paragraph above. I think you need to reinforce that `vec_rbind()` has to repair the input because if there are duplicated columns in the input, we don't know how to align columns based only on their name.

PR #425, opened 2019-06-18
Native array slicer

DavisVaughan:
A native `vec_slice()`r for arrays. I'll document in full tomorrow, but initial tests are promising! ~It doesn't quite work right with vctrs objects, but hopefully we can figure that out easily. I probably don't have something wired up correctly.~ (Update: falling back to `vec_slice_fallback()` when required)

I have changed `slice_names()` to return the sliced names rather than assign them directly. I need to use it to slice row names.

``` r
library(vctrs)
library(ggplot2)
library(magrittr)

# Short 1D array (select half of the rows)
x <- 1:10
dim(x) <- 10L

bench::mark(vec_slice(x, 1:5), x[1:5, drop = FALSE], iterations = 100000)
#> # A tibble: 2 x 10
#>   expression   min     mean   median    max `itr/sec` mem_alloc  n_gc n_itr
#>   <chr>      <bch> <bch:tm> <bch:tm> <bch:>     <dbl> <bch:byt> <dbl> <int>
#> 1 vec_slice… 893ns   1.29µs   1.25µs 45.8µs   773623.    6.45KB     8 99992
#> 2 x[1:5, dr… 371ns 517.39ns    499ns 41.8µs  1932768.        0B     8 99992

# Medium sized matrix (selecting all rows)
z <- matrix(1:1000, nrow = 500)
idx <- 1:500 + 0L

bench::mark(vec_slice(z, idx), z[idx, , drop = FALSE], iterations = 100000)
#> # A tibble: 2 x 10
#> 1 vec_slice… 2.28µs 2.99µs 2.68µs 6.52ms   334311.    3.95KB     9 99991
#> 2 z[idx, , … 2.15µs 2.78µs 2.67µs 5.25ms   360133.    3.95KB     9 99991

# (125000, 4, 2) double array
y <- 1:1000000 + 0
y <- array(y, c(length(y) / 8, 4, 2))

short_results <- bench::press(
  length = c(1, 10, 50, 100, 200, 500, 1000),
  {
    idx <- seq_len(length) + 0L
    bench::mark(
      vec_slice(y, idx),
      y[idx, , , drop = FALSE],
      iterations = 100000
    )
  }
)

short_results %>%
  ggplot(aes(x = length, y = median, color = expression)) +
  geom_line() +
  ggtitle("Short index test")
```

``` r

long_results <- bench::press(
  length = c(5000, 10000, 20000, 50000, 75000, 100000),
  {
    idx <- seq_len(length) + 0L
    bench::mark(
      vec_slice(y, idx),
      y[idx, , , drop = FALSE],
      iterations = 100
    )
  }
)

long_results %>%
  ggplot(aes(x = length, y = median, color = expression)) +
  geom_line() +
  ggtitle("Long index test")
```

<sup>Created on 2019-06-17 by the [reprex package](https://reprex.tidyverse.org) (v0.2.1)</sup>


src/slice.c
```diff
@@ -191,9 +191,25 @@ static SEXP vec_slice_impl(SEXP x, SEXP index) {
...
+        names = PROTECT_N(Rf_shallow_duplicate(names), &nprot);
+        SEXP row_names = VECTOR_ELT(names, 0);
+        SET_VECTOR_ELT(names, 0, slice_names(row_names, index));
+        Rf_setAttrib(out, R_DimNamesSymbol, names);
+      }
+    } else {
+      out = PROTECT_N(vec_slice_base(info.type, data, index), &nprot);
+
+      SEXP names = PROTECT_N(Rf_getAttrib(x, R_NamesSymbol), &nprot);
+      Rf_setAttrib(out, R_NamesSymbol, slice_names(names, index));
```
DavisVaughan: @lionel- this doesn't need a `PROTECT` right? Since we set the attrib immediately?

lionel-: oh yes it does, allocating the attribute pairlist nodes might trigger a gc.

PR #426, opened 2019-06-18
Add `...` to all vctrs generics

### LINKED ISSUE #399: Take `...` in all vctrs generics
This way their signature can be extended without causing downstream failures.

lionel-:
This makes it easier to extend the generics with new arguments without breaking backward compatibility. This is not to allow extensibility of methods, so we check the dots are empty before dispatch.

Closes #399


lionel-:
Now also in the S3 vignette: https://github.com/r-lib/vctrs/commit/1b97ed2d

R/arith.R
```diff
@@ -14,6 +14,8 @@
 #' @param op An arithmetic operator as a string
 #' @param x,y A pair of vectors. For `!`, unary `+` and unary `-`, `y` will be
 #'   a sentinel object of class `MISSING`, as created by `MISSING()`.
+#' @inheritParams ellipsis::dots_empty
```
hadley: Does this do anything? The argument of `check_dots_empty()` is `env`?

lionel-: This inherits from `dots_empty` which was written for this purpose: https://github.com/r-lib/ellipsis/blob/d8bf8a34eb3be04a52169bc69759ebccc7f247e2/R/dots.R#L6

hadley: Ah got it.

Issue #225, opened 2019-03-05
`vec_cast()` should use the nice `vec_type2()` error messaging


DavisVaughan:
I've been working on letting `hardhat` use `vctrs` for the checking of `new_data` you get at prediction time against the reference data used at fit time. The actual type casting implementation all seems correct, but the error messages are currently a bit frustrating to an end user. There are 3 main types of casting issues a user might see:

``` r
suppressPackageStartupMessages({
  library(vctrs)
  library(gapminder)
  library(rlang)
})

options(rlang__backtrace_on_error = "none")

gap_train <- gapminder[1:3,]
gap_test <- gapminder[4:6,]

# 0 row reference slice
gap_train_0 <- vec_slice(gap_train, 0L)


# Case 1)
# Lossy cast warnings - too many columns

gap_test2 <- gap_test
gap_test2$x <- 1

vec_cast(gap_test2, gap_train_0)
#> Warning: Lossy cast from <tbl_df<
#>   country  : factor<bf6dc>
#>   continent: factor<69262>
#>   year     : integer
#>   lifeExp  : double
#>   pop      : integer
#>   gdpPercap: double
#>   x        : double
#> >> to <tbl_df<
#>   country  : factor<bf6dc>
#>   continent: factor<69262>
#>   year     : integer
#>   lifeExp  : double
#>   pop      : integer
#>   gdpPercap: double
#> >>.
#> Locations: 
#> Dropped variables: `x`
#> # A tibble: 3 x 6
#>   country     continent  year lifeExp      pop gdpPercap
#>   <fct>       <fct>     <int>   <dbl>    <int>     <dbl>
#> 1 Afghanistan Asia       1967    34.0 11537966      836.
#> 2 Afghanistan Asia       1972    36.1 13079460      740.
#> 3 Afghanistan Asia       1977    38.4 14880372      786.


# Case 2)
# Lossy cast warnings - at the individual column level

gap_test3 <- gap_test
gap_test3$country <- as.character(gap_test3$country)
gap_test3$country[1] <- "jupyter"

vec_cast(gap_test3, gap_train_0)
#> Warning: Lossy cast from <character> to <factor<bf6dc>>.
#> Locations: 1
#> # A tibble: 3 x 6
#>   country     continent  year lifeExp      pop gdpPercap
#>   <fct>       <fct>     <int>   <dbl>    <int>     <dbl>
#> 1 <NA>        Asia       1967    34.0 11537966      836.
[...]

PR #428, opened 2019-06-18
Fix cast error messages

### LINKED ISSUE #57: Better error messages

lionel-:
We now have:

```r
x <- tibble(a = tibble(b = "foo"))
y <- tibble(a = tibble(b = 10))

vec_cast(x, y)
#> Error: Lossy cast from `x$a$b` <character> to `to$a$b` <double>.
#> Locations: 1

x <- tibble(a = tibble(b = factor("foo")))

vec_cast(x, y)
#> Error: Can't cast `x$a$b` <factor<bd40e>> to `to$a$b` <double>.
#> Call `rlang::last_error()` to see a backtrace

vec_cast_common(x, y)
#> Error: No common type for `..1$a$b` <factor<bd40e>> and `..2$a$b` <double>.
```

I'm not sure about the last commit which adds default argument tags to `vec_cast()`. It improves the messages above (`to$a$b` instead of just `a$b`), but this looks slightly suboptimal in the non-nested case because of the repetition of "to":

```r
vec_cast("foo", 10.5)
#> Error: Lossy cast from `x` <character> to `to` <double>.
```

Closes #57 and #225. 
Case 1 in #225 isn't solved, but I don't think we can make it better for this release.

### DIFF
--- a/R/cast-list.R
+++ b/R/cast-list.R
@@ -10,9 +10,14 @@
 #'
 #' @param x A list
 #' @param to Type to coerce to
+#' @inheritParams ellipsis::dots_empty
+#'
 #' @export
 #' @keywords internal
-vec_list_cast <- function(x, to) {
+vec_list_cast <- function(x, to, ..., x_arg = "", to_arg = "") {
+  if (!missing(...)) {
+    ellipsis::check_dots_empty()
+  }
   ns <- map_int(x, vec_size)
 
   n <- vec_size(x)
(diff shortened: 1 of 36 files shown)


DavisVaughan:
Case 1 isn't that bad, so that definitely doesn't bother me for this release

DavisVaughan:
I don't have a solution for the default arg tags. ``Lossy cast between `x` and `to` `` doesn't really have the directional aspect to it. I don't mind it that much though.

Does it matter that `vec_type2()` doesn't have defaults for `"x"` and `"y"`?

```
> vctrs::vec_type2("x", 1)
Error: No common type for <character> and <double>.

> vctrs::vec_type2(data.frame(x = "x"), data.frame(x = 1))
Error: No common type for `x` <factor<5a425>> and `x` <double>.
```

lionel-:
@DavisVaughan I didn't change type2's defaults because I wasn't sure it's an improvement, but in the end I think it does make it easier to understand. Now changed.

PR #430, opened 2019-06-19
Add `vec_as_names_legacy()`

### LINKED ISSUE #359: Provide legacy naming function
From https://github.com/tidyverse/tidyr/issues/641#issuecomment-496583453 (note that the default arguments are slightly different from those in readxl)

```R
legacy_repair <- function(nms, prefix = "V", sep = "") {
  if (length(nms) == 0) return(character())
  blank <- nms == ""
  nms[!blank] <- make.unique(nms[!blank], sep = sep)
  new_nms <- setdiff(paste(prefix, seq_along(nms), sep = sep), nms)
  nms[blank] <- new_nms[seq_len(sum(blank))]
  nms
}
```

lionel-:
Addresses tidyverse/readxl#546
Addresses tidyverse/tidyr#641
Closes #359


lionel-:
hmm on second thought maybe it's best to make it a "retired" function? This would mean it will not be deprecated and we'll continue exporting this tool undefinitely.

krlmlr:
Why would you want to export indefinitely?

lionel-:
Because it doesn't represent any maintenance cost. Similarly `purrr::invoke()` is retired in favour of `exec()` instead of being deprecated.

This would be a guarantee for the users in tidyverse/readxl#546 that their old scripts will keep working for the foreseeable future, if they use `vec_as_names_legacy()` (which is an easier fix than changing the names).

lionel-:
Thanks for the tests, I'll look into it.

> Do we want to support repair = "legacy"?

I don't think so, it's best to avoid exposing users to deprecated/retired stuff. This is why I marked the function with `@keywords internal`.

> What's the easiest way for users of packages to adapt their scripts, perhaps {pkgconfig}?

Passing the legacy repair function to the relevant data frame constructors seems reasonable.

PR #432, opened 2019-06-19
Return logical vectors from predicate methods

### LINKED ISSUE #251: Generic predicates should operate on proxy but not restore
They probably should always return an untyped bool.

```r
new_cached_sum <- function(x = double(), sum = 0L) {
  new_vctr(x, sum = sum, class = "vctrs_cached_sum")
}
vec_restore.vctrs_cached_sum <- function(x, to, ..., i = NULL) {
  new_cached_sum(x, sum(x))
}
x <- new_cached_sum(runif(10))

is.finite(x)
#> <vctrs_cached_sum[10]>
#>  [1] TRUE TRUE TRUE TRUE TRUE TRUE TRUE TRUE TRUE TRUE
```

lionel-:
Closes #251

This still goes through `vec_math()` because a `vec_logical()` might not make sense for `is.finite()` or `is.nan()` which are defined on numeric data. Since we need to special-case them in `vec_math()`, a new generic just for `any()` and `all()` seems a bit overkill.

### DIFF
--- a/R/numeric.R
+++ b/R/numeric.R
@@ -42,15 +42,22 @@ vec_math <- function(fun, x, ...) {
 
 #' @export
 vec_math.default <- function(fun, x, ...) {
-  if (is_double(x)) {
-    vec_restore(vec_math_base(fun, x, ...), x)
-  } else if (fun %in% c("any", "all") && is_logical(x)) {
-    vec_restore(vec_math_base(fun, x, ...), x)
-  } else {
-    # nocov start
+  if (!is_double(x) && !is_logical_dispatch(fun, x)) {
     stop_unimplemented(x, "vec_math")
-    # nocov end
   }
+
+  out <- vec_math_base(fun, x, ...)
+
+  # Don't restore output of logical predicates like `any()`,
+  # `is.finite()`, or `is.nan()`
+  if (is_double(out)) {
+    out <- vec_restore(out, x)
+  }
+
+  out
+}
+is_logical_dispatch <- function(fun, x) {
+  is_logical(x) && fun %in% c("any", "all")
 }
 
 #' @export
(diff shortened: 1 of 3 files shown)


R/numeric.R
```diff
@@ -42,15 +42,17 @@ vec_math <- function(fun, x, ...) {
...
-  } else {
-    # nocov start
+  if (!is_double(x) && !(is_logical(x) && fun %in% c("any", "all"))) {
     stop_unimplemented(x, "vec_math")
-    # nocov end
   }
+
+  out <- vec_math_base(fun, x, ...)
+
+  if (!fun %in% c("any", "all", "is.finite", "is.infinite", "is.nan")) {
```
hadley: What if we just restore if `out` is also double?

R/numeric.R
```diff
@@ -42,15 +42,17 @@ vec_math <- function(fun, x, ...) {
 
 #' @export
 vec_math.default <- function(fun, x, ...) {
-  if (is_double(x)) {
-    vec_restore(vec_math_base(fun, x, ...), x)
-  } else if (fun %in% c("any", "all") && is_logical(x)) {
-    vec_restore(vec_math_base(fun, x, ...), x)
-  } else {
-    # nocov start
+  if (!is_double(x) && !(is_logical(x) && fun %in% c("any", "all"))) {
```
hadley: If found this compound expression rather hard to parse.

tests/testthat/test-type-vctr.R
```diff
@@ -438,8 +438,8 @@ test_that("na.rm is forwarded to summary generics", {
 
   x <- new_vctr(lgl(TRUE, NA))
 
-  expect_identical(all(x, na.rm = FALSE), new_vctr(lgl(NA)))
-  expect_identical(all(x, na.rm = TRUE), new_vctr(TRUE))
+  expect_identical(all(x, na.rm = FALSE), lgl(NA))
```
hadley: This is a change, so it's worth giving it a news bullet.

R/numeric.R
lionel-: This also works, done.

PR #433, opened 2019-06-19
Rename `fun` to `fn` in `vec_math()`

lionel-:
For consistency with rlang.

Includes #432.

### DIFF
--- a/R/arith.R
+++ b/R/arith.R
@@ -91,8 +91,8 @@ vec_arith.numeric.numeric <- function(op, x, y, ...) vec_arith_base(op, x, y)
 vec_arith_base <- function(op, x, y) {
   c(x, y) %<-% vec_recycle_common(x, y)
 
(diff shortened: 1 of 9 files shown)


hadley:
Oh, this cause any new revdep failures?

DavisVaughan:
I think this one might be clean. I don't use it in probably and the evaluator pkg doesn't seem to either

lionel-:
Also we already cause failures because of signature consistency, since we have added `...` to generics, so it makes sense to bundle other breaking changes in this release.

hadley:
I'm happy with breaking changes; just wanted to make sure it didn't introduce any new ones (I couldn't remember if you'd already done a round of notifications or not)

lionel-:
oh it also needs to be `.fn` and `.x` to prevent matching with methods arguments

hadley:
Oh good point!

PR #435, opened 2019-06-19
Add vctrs_ prefix to all C API functions

krlmlr:
via:

```
sed -r -i 's/(vec_(dim|restore|restore_default|proxy|type|type_finalise|names|assign)[(,])/vctrs_\1/g' src/*
```

This makes these functions easier to search for in the C part of the codebase.

### DIFF
--- a/src/bind.c
+++ b/src/bind.c
@@ -230,7 +230,7 @@ static SEXP vec_cbind(SEXP xs, SEXP ptype, SEXP size, enum name_repair_arg name_
   names = PROTECT(vec_as_names(names, name_repair, false));
   Rf_setAttrib(out, R_NamesSymbol, names);
 
-  out = vec_restore(out, type, R_NilValue);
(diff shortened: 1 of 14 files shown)


DavisVaughan:
Hmm, I've been under the impression that many of these function:
- Have an R level `vec_fun()`
- Have a bridge between R and C called `vctrs_fun()`
- Have a pure C implementation called `vec_fun()`

This has had a really nice cognitive load reducing side effect of being able to reach for `vec_size()` from C without even thinking about it having a different name.

See `vec_dim_n()` and `vctrs_dim_n()` for an example:
https://github.com/r-lib/vctrs/blob/a62a37e90028b0b09f86c2ec363e08aae5700280/src/size.c#L19

lionel-:
I recommend closing this and not worry about perfect consistency between the different naming logics in the API.

PR #440, opened 2019-06-19
Implement `range.vctrs_vctr()`

### LINKED ISSUE #434: Implement range.vctrs_vctr()
Rather than going through `vec_math()`, as it currently does.

It should have the same signature as `min.vctrs_vctr`, taking `x` and ignoring the rest of the `...`

Should the `finite` argument be included? It is in the default method but not the generic.

Are there any optimizations we can do with `NA` values ahead of time if `na.rm = TRUE`? Can we discard them once and then pass on to `min()` and `max()` with `na.rm = FALSE`?

DavisVaughan:
Closes #434 

The only thing we could optimize away would be a call to `xtfrm()` by essentially inlining `min()` and `max()` in `range()`, but I didn't think that was worth it so this just calls `min()` and `max()` directly.

This also tweaks documentation by adding missing parenthesis in `?new_vctr` docs, and by consistently doing `fn<-`, not `fn<-()`.

---

I also tested replacing the following in `min.vctrs_vctr`:

```
idx <- which(vec_equal(rank, min(rank), na_equal = TRUE))
```

with:

```
idx <- vec_match(min(rank), rank))
```

as it would use less memory, but was surprised to see it was a good bit slower.


hadley:
Hmmm, I'm not sure it's worth it unless we do combine the implementation of `min()` and `max()` so we only have to do a single pass. Otherwise, `range.default` already just calls `min()` and `max()` (with a bit of messing around on top, but nothing too extreme). 

I prefer your initial implementation in 748cd87ba01 (which I think should be about 2x faster?).  What made you move away from it?

DavisVaughan:
Combining the implementations lets you call `xtfrm()` once, but you still have to call `which(vec_equal(...))` twice, which is time consuming.

When `x` is an integer/double, the call to `xtfrm()` doesn't have to do much, so the results are similar. When `xtfrm()` actually has to do some work, only calling it once is noticeably faster with large vectors, but not quite 2x faster.

Below, the results with `x` and `y` are about equal, with the fast method pulling a little ahead.
The results with `x_lgl` and `y_lgl` show `xtfrm()` having to call `order(order_proxy())`, and the fast method is the clear winner.

``` r
library(vctrs)

range_slow <- vctrs:::range.vctrs_vctr

range_fast <- function(x, ..., na.rm = FALSE) {
  if (vec_is_empty(x)) {
    return(vec_cast(c(Inf, -Inf), x))
  }
  
  rank <- xtfrm(x)
  
  if (isTRUE(na.rm)) {
    idx_min <- which.min(rank)
    idx_max <- which.max(rank)
  } else {
    idx_min <- which(vec_equal(rank, min(rank), na_equal = TRUE))
    idx_max <- which(vec_equal(rank, max(rank), na_equal = TRUE))
  }
  
  c(x[[idx_min[[1]]]], x[[idx_max[[1]]]])
}

x <- new_vctr(sample(1000) + 0L)
x_lgl <- new_vctr(rep(c(TRUE, FALSE), 500))

bench::mark(
  slow = range_slow(x),
  fast = range_fast(x),
  iterations = 100000
)
#> # A tibble: 2 x 10
#>   expression   min  mean median    max `itr/sec` mem_alloc  n_gc n_itr
#>   <chr>      <bch> <bch> <bch:> <bch:>     <dbl> <bch:byt> <dbl> <int>
#> 1 slow       185µs 236µs  230µs 2.24ms     4246.    31.6KB   638 99362
#> 2 fast       169µs 205µs  200µs 1.15ms     4886.    31.6KB   581 99419

bench::mark(
  slow = range_slow(x_lgl),
  fast = range_fast(x_lgl),
  iterations = 10000
)
#> # A tibble: 2 x 10
#>   expression   min  mean median   max `itr/sec` mem_alloc  n_gc n_itr
#>   <chr>      <bch> <bch> <bch:> <bch>     <dbl> <bch:byt> <dbl> <int>
#> 1 slow       296µs 352µs  348µs 781µs     2840.    92.8KB    90  9910
#> 2 fast       225µs 272µs  265µs 650µs     3678.    43.5KB    72  9928

y <- new_vctr(sample(100000) + 0L)
y_lgl <- new_vctr(rep(c(TRUE, FALSE), 50000))

bench::mark(
  slow = range_slow(y),
  fast = range_fast(y),
  iterations = 100000
)
#> # A tibble: 2 x 10
#>   expression   min   mean median    max `itr/sec` mem_alloc  n_gc n_itr
#>   <chr>      <bch> <bch:> <bch:> <bch:>     <dbl> <bch:byt> <dbl> <int>
#> 1 slow       949µs 1.23ms 1.18ms 4.32ms      815.    3.05MB  7143 92857
#> 2 fast       902µs 1.19ms 1.15ms 2.82ms      837.    3.05MB  7678 92322

bench::mark(
  slow = range_slow(y_lgl),
  fast = range_fast(y_lgl),
  iterations = 10000
)
#> # A tibble: 2 x 10
#> 1 slow       2.38ms 3.15ms 3.06ms 5.28ms      317.    5.34MB  1236  8764
#> 2 fast        1.6ms 2.15ms 2.07ms 3.61ms      466.     4.2MB   953  9047
```

lionel-:
I see we have both `vec_math()` and `vec_math_base()` exported. This makes me wonder if it'd make sense to dispatch to our own primitives from `vec_math()`, in case we come up with implementations that are more generic, simpler, or more efficient?

Issue #446, opened 2019-06-24
`vec_cbind()` should pack named inputs


lionel-:
`vec_cbind(foo = mtcars, mtcars)` should return `tibble::tibble(foo = mtcars, !!!mtcars)`.

PR #448, opened 2019-06-24
Add `.name_spec` support in `vec_c()`

lionel-:
I didn't modify `vec_cbind()` as it might interfere with #446.

From our Slack discussion:

By default, outer names can only be combined with scalars or empty vectors:

```r
vec_c(foo = 1)
#>foo
#>  1

vec_c(foo = integer())
#> integer(0)
```

It is an error to provide both an outer name and a named vector:

```r
vec_c(foo = c(a = 1))
#> Error: Can't merge the outer name `foo` with a named vector.
```

And an error to provide an outer name and a length > 1 vector:

```r
vec_c(foo = 1:3)
#> Error: Can't merge the outer name `foo` with a vector of length > 1.
```

To get out of these errors, you can supply a glue specification:

```r
vec_c(foo = c(a = 1), bar = 1:3, .name_spec = "{outer}__{inner}")
#> foo__a bar__1 bar__2 bar__3
#>      1      1      2      3
```

You can also supply a function, if the inner vector is unnamed `inner` is character, otherwise it is integer:

```r
my_merge <- function(outer, inner) {
  if (is_character(inner)) {
    paste0(outer, "__", inner)
  } else {
    paste0(outer, inner * 10)
  }
}

vec_c(foo = c(a = 1), bar = 1:3, .name_spec = my_merge)
#> foo__a  bar10  bar20  bar30
#>      1      1      2      3
```

Lambda formulas are supported:

```r
vec_c(foo = 1:3, .name_spec = ~ .x)
#> foo foo foo 
#>   1   2   3 
```

Another application of this will be the vctrs-powered flatten functions:

```r
flatten(list(foo = list(a = 1, b = 2)), .name_spec = "{outer}_{inner}")
#> $foo_a
#> [1] 1
#>
#> $foo_b
#> [1] 2
```


NEWS.md
```diff
@@ -1,5 +1,30 @@
 # vctrs 0.1.0.9000
 
+* `vec_c()` now fails when an input is supplied with a name but has
+  internal names or is length > 1:
+
+  ```
+  vec_c(foo = c(a = 1))
+  #> Error: Can't merge the outer name `foo` with a named vector.
+  #> Please supply a `.name_merge` specification.
```
hadley: `.name_spec`?

NEWS.md
```diff
@@ -1,5 +1,30 @@
...
+
+  You can supply a name specification that describes how to combine
+  the external name of the input with its internal names or positions:
+
+  ```
+  # Name spec as a function:
+  vec_c(foo = c(a = 1), .name_spec = function(outer, inner) paste(outer, inner, sep = "_"))
+  vec_c(foo = c(a = 1), .name_spec = ~ paste(.x, .y, sep = "_"))
+
+  # Name spec as glue string:
```
hadley: I'd show this first

R/names.R
```diff
@@ -509,3 +509,90 @@ vec_as_names_legacy <- function(names, prefix = "V", sep = "") {
 
   names
 }
+
+
+#' Name specifications
+#'
+#' @description
+#'
+#' A name specification describes how to combine an external name with
```
hadley: Inner and outer names?

R/names.R
```diff
@@ -509,3 +509,90 @@ vec_as_names_legacy <- function(names, prefix = "V", sep = "") {
...
+#'   vec_c(external = c(internal1 = 1, internal2 = 2))
+#'   ```
+#'
+#' * Unnamed vector:
+#'
+#'   ```
+#'   vec_c(external = 1:2)
+#'   ```
+#'
+#' In r-lib and tidyverse packages, these cases are errors by default.
```
hadley: the cases are errors by default, because there's no behaviour that works well for every case. Instead, you can provide ...

R/names.R
```diff
@@ -509,3 +509,90 @@ vec_as_names_legacy <- function(names, prefix = "V", sep = "") {
...
+#' @param name_spec,.name_spec A name specification for combining
+#'   outer and inner names. This is relevant for inputs passed with a
+#'   name, when these inputs are themselves named, like `outer =
+#'   c(inner = 1)`, or when they have length greater than 1: `outer =
+#'   1:2`. By default, these cases trigger an error. You can resolve
+#'   the error by providing a specification that describes how to
+#'   combine the names or the indices of the inner vector with the
+#'   name of the input. This specification can be:
+#'
+#'   * A function of two arguments. The exterior name is passed as a
```
hadley:
```suggestion
#'   * A function of two arguments. The outer name is passed as a
```

src/names.c
```diff
@@ -441,6 +442,88 @@ SEXP outer_names(SEXP names, SEXP outer, R_len_t n) {
...
+  if (r_is_empty_names(inner)) {
+    if (n == 1) {
+      return r_str_as_character(outer);
+    }
+    inner = PROTECT(r_seq(1, n + 1));
+  } else {
+    inner = PROTECT(inner);
+  }
+
+  switch (TYPEOF(name_spec)) {
```
hadley: Would it make more sense to do this in R?

lionel-: The C switch doesn't seem bad.

hadley: Plus the 3 new helper functions 😉

lionel-: oh I think I'll export an `rlang_as_function()` C callable. Here I'm also experimenting with passing an `arg` param, maybe that should become standard in all dev-oriented tools.

PR #454, opened 2019-06-25
Add packing in `vec_cbind()`

lionel-:


DavisVaughan:
Extreme edge case) Should packing + `vec_cbind()` care if the data frame being packed has no column names?

``` r
library(vctrs)
library(tibble)

# 2 unnamed columns, 1 row
# should this care that there were no column names to begin with?
x <- vec_cbind(y = new_data_frame(list(1, 2), n = 1L))
str(x)
#> 'data.frame':    1 obs. of  1 variable:
#>  $ y:'data.frame':   1 obs. of  2 variables:
#>   ..$ : num 1
#>   ..$ : num 2

x
#>   y. y.
#> 1  1  2

# I ask because this seems to care
as_tibble(vec_cbind(x = array(numeric(), c(1, 2))))
#> New names:
#> * `` -> ...1
#> * `` -> ...2
#> # A tibble: 1 x 1
#>   x$...1 $...2
#>    <dbl> <dbl>
#> 1     NA    NA

# and so does this
vec_cbind(new_data_frame(list(1, 2), n = 1L))
#> New names:
#> * `` -> ...1
#> * `` -> ...2
#>   ...1 ...2
#> 1    1    2

# And by "not caring" this has some odd behavior downstream
# (this also feels like a tibble bug)
as_tibble(x)
#> # A tibble: 1 x 1
#>      y$    NA
#>   <dbl> <dbl>
#> 1     1     2
```

<sup>Created on 2019-06-25 by the [reprex package](https://reprex.tidyverse.org) (v0.2.1)</sup>

DavisVaughan:
Looks like packing with >2D arrays is a bit broken, regarding the names.

``` r
library(vctrs)
library(tibble)

# looks good
two_cols <- array(1, c(1, 2, 1))
as_tibble(vec_cbind(x = two_cols))
#> New names:
#> * `` -> ...1
#> * `` -> ...2
#> # A tibble: 1 x 1
#>   x$...1 $...2
#>    <dbl> <dbl>
#> 1      1     1

# hmm, not right
two_deep <- array(1, c(1, 1, 2))
as_tibble(vec_cbind(x = two_deep))
#> New names:
#> * `` -> ...1
#> # A tibble: 1 x 1
#>   x$...1   $NA
#>    <dbl> <dbl>
#> 1      1     1

# It might have to do with this:
# number of columns change
ncol(two_deep)
#> [1] 1
ncol(as.data.frame(two_deep))
#> [1] 2

# note that not packing works
as_tibble(vec_cbind(two_deep))
#> New names:
#> * `` -> ...1
#> New names:
#> * NA -> ...2
#> # A tibble: 1 x 2
#>    ...1  ...2
#>   <dbl> <dbl>
#> 1     1     1

# packing when there are names doesn't work either, but might offer more insight
two_deep_with_names <- array(1, c(1, 1, 2), dimnames = list("r1", "c1", c("d1", "d2")))
as_tibble(vec_cbind(x = two_deep_with_names))
#> # A tibble: 1 x 1
#>    x$c1   $NA
#>   <dbl> <dbl>
#> 1     1     1
```

My first thought was to wait until after it has been converted to a data frame to extract the names in `shaped_as_df_col()`. But you then inherit the ugly column names that `as.data.frame()` gives the collapsed array.

When `as.data.frame()` is called on a >2D array it seems to collapse the names of the 2nd dimension and up by separating them with a `.`

``` r
two_deep <- array(1, c(1, 1, 2))
as.data.frame(two_deep)
#>   V1 V2
#> 1  1  1

two_deep_with_names <- array(1, c(1, 1, 2), dimnames = list("r1", "c1", c("d1", "d2")))
as.data.frame(two_deep_with_names)
#>    c1.d1 c1.d2
#> r1     1     1

two_deep_with_no_colnames <- array(1, c(1, 1, 2), dimnames = list("r1", NULL, c("d1", "d2")))
as.data.frame(two_deep_with_no_colnames)
#>    1.d1 1.d2
#> r1    1    1
```

lionel-:
We now fully preserve the data structure of packed columns. Matrices and 1D arrays are preserved:

```r
m <- matrix(1:12, 3, 4)
as_tibble(vec_cbind(m))
#> # A tibble: 3 x 4
#>    ...1  ...2  ...3  ...4
#>   <int> <int> <int> <int>
#> 1     1     4     7    10
#> 2     2     5     8    11
#> 3     3     6     9    12
```

Regarding arrays of ndim > 2, I suggest we just make it an error.

PR #458, opened 2019-06-26
Slice row names

### LINKED ISSUE #397: Handle row names
For instance in `vec_slice()`:

```r
vec_slice(mtcars, 1)
#>   mpg cyl disp  hp drat   wt  qsec vs am gear carb
#> 1  21   6  160 110  3.9 2.62 16.46  0  1    4    4
```

@hadley Perhaps this should make it into 0.2.0? Supporting it later on might be disruptive if unit tests in downstream deps make incidental assumptions about row names.

lionel-:
Since row names can't be `NA` or duplicates, they are repaired after slicing:

```r
vec_slice(mtcars[1:3, ], NA)
#> New names:
#> * NA -> ...1
#> * NA -> ...2
#> * NA -> ...3
#>      mpg cyl disp hp drat wt qsec vs am gear carb
#> ...1  NA  NA   NA NA   NA NA   NA NA NA   NA   NA
#> ...2  NA  NA   NA NA   NA NA   NA NA NA   NA   NA
#> ...3  NA  NA   NA NA   NA NA   NA NA NA   NA   NA
```

```r
vec_slice(mtcars, c(NA, 2, NA))
#> New names:
#> * NA -> ...1
#> * NA -> ...3
#>               mpg cyl disp  hp drat    wt  qsec vs am gear carb
#> ...1           NA  NA   NA  NA   NA    NA    NA NA NA   NA   NA
#> Mazda RX4 Wag  21   6  160 110  3.9 2.875 17.02  0  1    4    4
#> ...3           NA  NA   NA  NA   NA    NA    NA NA NA   NA   NA
```

Closes #397


lionel-:
It doesn't. I assumed it's best to not special-case, and deal with rownames whenever they exist?

hadley:
If I understand correctly won't `vec_slice(tibble(x = 1:4), c(1,1,1,1))` now create a bunch of messages about renaming rownames that didn't exist before?

hadley:
I'm also a little worried that if we roll this out generally across the tidyverse it's going to create a bunch of new messages for cases that were previously silent. Maybe we're being a bit too informative about changing rownames?

lionel-:
This doesn't cause messages unless the row names exist:

```r
vec_slice(tibble(x = 1:4), c(1,1,1,1))
#> # A tibble: 4 x 1
#>       x
#>   <int>
#> 1     1
#> 2     1
#> 3     1
#> 4     1

vec_slice(structure(tibble(x = 1:4), row.names = letters[1:4]), c(1,1,1,1))
#> New names:
#> * a -> a...1
#> * a -> a...2
#> * a -> a...3
#> * a -> a...4
#> # A tibble: 4 x 1
#>       x
#> * <int>
#> 1     1
#> 2     1
#> 3     1
#> 4     1
```

In the latter case, it seems the verbosity is better than silent dropping or renaming?

hadley:
Oh ok, maybe it's not as bad as I thought then.

PR #462, opened 2019-06-27
Alter `r_as_data_frame()` to only extract column names of 2D objects

### LINKED ISSUE #461: Calling `colnames()` on a 1D array with `dimnames = list(NULL)` breaks
The latest changes to `r_as_data_frame()` cause:

``` r
x <- array(c(1, 1, 2, 2, 3), dimnames = list(NULL))
vctrs::vec_unique_loc(x)
#> Error in dn[[2L]]: subscript out of bounds
```

Following this back, the culprit is `r_as_data_frame()` calling `vec_unique_colnames()` here
https://github.com/r-lib/vctrs/blob/ee9185941ea9b8d337c44df0eb14aee85fdb506e/src/utils.c#L782

Which calls base R's `colnames()` here
https://github.com/r-lib/vctrs/blob/a8393eba66094e13d916e5bdd741ff0440ff0269/src/names.c#L374

Apparently these are different:

``` r
y <- array(c(1, 1, 2, 2, 3))
colnames(y)
#> NULL

x <- array(c(1, 1, 2, 2, 3), dimnames = list(NULL))
colnames(x)
#> Error in dn[[2L]]: subscript out of bounds
```

Maybe `vec_unique_colnames()` should immediately return `NULL` if `has_dim(x) && vec_dim_n(x) == 1` so we don't call `colnames()`?

Or our C `colnames()` could check this, that would probably be more general in case we use `colnames()` in other places.

DavisVaughan:
Closes #461 

Here is one potential solution to #461.

- `r_as_data_frame()` gains a `quiet` argument to solve the noisiness issue when calling `vec_unique()` on an array without column names. It is now called quietly from the hashing functions.
- `r_as_data_frame()` now only attempts to extract column names from 2D objects. A 1D / 3D+ object has default names repaired (`...1`, `...2`, etc). 

This solves the problem of trying to extract `colnames()` from a 1D array with `NULL` dimnames, which is an error.

This also solves the problem of trying to extract column names from a 3D array, and then trying to assign those column names to the newly created data frame. This is problematic because the 3D array might have dimensions `(1, 2, 2)` resulting in 2 column names, but the corresponding data frame will have 4 columns.

_One thing to note is that `vec_unique_names_impl()` is now exposed in `util.h`_

### DIFF
--- a/src/bind.c
+++ b/src/bind.c
@@ -147,7 +147,7 @@ static SEXP as_df_row_impl(SEXP x, enum name_repair_arg name_repair, bool quiet)
     Rf_errorcall(R_NilValue, "Can't bind arrays.");
   }
   if (ndim == 2) {
-    return r_as_data_frame(x);
+    return r_as_data_frame(x, false);
   }
 
   x = PROTECT(r_as_list(x));
@@ -202,7 +202,7 @@ static SEXP vec_cbind(SEXP xs, SEXP ptype, SEXP size, enum name_repair_arg name_
   if (type == R_NilValue) {
     type = new_data_frame(vctrs_shared_empty_list, 0);
   } else if (!is_data_frame(type)) {
-    type = r_as_data_frame(type);
+    type = r_as_data_frame(type, false);
   }
   UNPROTECT(1);
   PROTECT(type);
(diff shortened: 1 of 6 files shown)


lionel-:
oops sorry, I didn't see you posted a fix! Just pushed a commit.

DavisVaughan:
I like yours more as it simplifies `r_as_data_frame()` to be more in line with `as.data.frame2()`. I didn't like that I was doing extra things inside `r_as_data_frame()` that `as.data.frame2()` wasn't doing

PR #463, opened 2019-06-27
Rename `type` particle to `ptype`

lionel-:
Compats added to `lifecycle-retired.R`.


lionel-:
hmm since `vec_ptype_show()` is a straight pretty-printer like `env_print()` or `expr_print()`, it should actually be called `vec_ptype_print()` I think.

lionel-:
> hmm since vec_ptype_show() is a straight pretty-printer like env_print() or expr_print(), it should actually be called vec_ptype_print() I think.

oops now I see it first takes the common type of `...` inputs before printing, so `_show` is appropriate.

Issue #167, opened 2019-01-25
Specify behavior of as.data.frame.vctrs_sclr()


krlmlr:
https://github.com/r-lib/vctrs/pull/163#discussion_r250687447

lionel-:
Closing as this doesn't seem important at the moment.

PR #464, opened 2019-06-28
used `vec_size()` instead of `Rf_length()` in `vec_dim()` (pillar #167)

earowang:
This PR is related to my issue about `list_of()` size display in the list-columns (now transferred to https://github.com/r-lib/pillar/issues/167). After digging into the source code, I think this issue lies in the **vctrs**, not **pillar**. For robustness, `Rf_length()` is replaced by `vec_size()` in `vec_dim()`. It fixes the size issue.

``` r
library(vctrs)

new_rational <- function(n = integer(), d = integer()) {
  vec_assert(n, ptype = integer())
  vec_assert(d, ptype = integer())
  
  new_rcrd(list(n = n, d = d), class = "vctrs_rational")
}

rational <- function(n, d) {
  c(n, d) %<-% vec_cast_common(n, d, .to = integer())
  c(n, d) %<-% vec_recycle_common(n, d)
  
  new_rational(n, d)
}

format.vctrs_rational <- function(x, ...) {
  n <- field(x, "n")
  d <- field(x, "d")
  
  out <- paste0(n, "/", d)
  out[is.na(n) | is.na(d)] <- NA
  
  out
}

vec_ptype_abbr.vctrs_rational <- function(x, ...) "rtnl"
vec_ptype_full.vctrs_rational <- function(x, ...) "rational"

x <- rational(1, 1:10)
x
#> <rational[10]>
#>  [1] 1/1  1/2  1/3  1/4  1/5  1/6  1/7  1/8  1/9  1/10

iris %>% 
  group_by(Species) %>% 
  summarise(
    ratio = list_of(rational(as.integer(Sepal.Length), as.integer(Sepal.Width)))
  )
#> # A tibble: 3 x 2
#>   Species           ratio
#>   <fct>      <list<rtnl>>
#> 1 setosa             [50]
#> 2 versicolor         [50]
#> 3 virginica          [50]
```

<sup>Created on 2019-06-28 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

### DIFF
--- a/src/size.c
+++ b/src/size.c
@@ -8,7 +8,7 @@ SEXP vec_dim(SEXP x) {
   SEXP dim = PROTECT(vec_bare_dim(x));
 
   if (dim == R_NilValue) {
-    dim = r_int(Rf_length(x));
+    dim = r_int(vec_size(x));
   }
 
   UNPROTECT(1);


DavisVaughan:
The only thing I can think of for this is, what should `vec_dim()` of a data frame return? Because this alters that (potentially in a good way)

lionel-:
Since this is exposed to end users, we need to represent all dimensions available for subsetting with `[` (not `vec_slice()`). This looks like a case where we need to use `dim(x) %||% length(x)`.

Perhaps revert your changes in the C code and use a new R function with the above definition?

lionel-:
Ah this is actually an issue in the pillar package, it should not use `vec_dim()` but something like `vctrs:::dim2`.

cc @krlmlr: pillar should represent the dimensions available for subsetting with `[`, not `vec_slice()`, since it's user-oriented. This means it needs to dispatch with the `dim()` generic and can't use `vec_dim()`.

I am unsure whether `vec_dim()` should return `vec_size()` for 1d vectors. It would certainly feel more vctrs-y. cc @hadley

krlmlr:
If my understanding of the implementation is correct (see my PR), I think we should keep returning `length()`.

hadley:
I think it would oddly inconsistent for `vec_dim()[1]` to return `length()` not `vec_size()`. However, maybe we shouldn't export it at all? Or should we give it a prefix other than `vec`?

krlmlr:
Do we have a name for `c(vec_size(x), vec_shape(x))` ?

krlmlr:
If in doubt, I agree we should refrain from exporting for now.

lionel-:
I was thinking the same, best to not export until we have a clearer view of the interaction between shape, size, and length.

lionel-:
Thanks for looking into this @earowang! It allowed us to discover a part of the API that wasn't ready to be exported.

Issue #161, opened 2019-01-23
Rethink casting for factor and ordered


krlmlr:
- [x] Should be able to cast ordered to factor
- [ ] Common type of factor and ordered should be factor
- [ ] Common type of ordered and ordered should be factor (!), perhaps with warning

hadley:
Current behaviour:

``` r
library(vctrs)

f <- factor("a")
o <- ordered("b")

vec_c(f, o)
#> Error: No common type for `..1` <factor<127a2>> and `..2` <ordered<ddf10>>.
vec_c(o, o)
#> [1] b b
#> Levels: b
```

<sup>Created on 2020-04-03 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

DavisVaughan:
Not sure if the second example is a typo, but that one is an ordered factor, it just doesn't have >1 levels so you wouldn't know just from the print method. Helpful!

``` r
library(vctrs)

o <- ordered("b")

vec_c(o, o)
#> [1] b b
#> Levels: b

class(vec_c(o, o))
#> [1] "ordered" "factor"
```

<sup>Created on 2020-04-03 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

lionel-:
We have now settled on the current behaviour:

- factor and ordered are never coercible
- ordered are coercible if they have the exact same set of levels

PR #465, opened 2019-06-28
Implement `partial_factor()`

### LINKED ISSUE #106: Partial type for factors
Where levels are learned from the data

### LINKED ISSUE #261: How strict should `vec_assert()` be with factors?
I think this is a broader question around what exactly `vctrs:::is_same_type()` should do, but consider the case of validating that a user provided a factor as the `outcome` to a tree-based model (as they would be required to do).

``` r
library(vctrs)

# I think I want this to pass
vec_assert(iris$Species, factor())
#> `iris$Species` must be <fctr>, not <fctr>.

# Because there's no way i could know the levels ahead of time,
# just that its a factor
vec_assert(iris$Species, factor(levels = c("setosa", "versicolor", "virginica")))

# I suppose I could use this
outcome <- vec_cast(iris$Species, factor())

# But it doesn't give as good of an error message
# when there is an issue
vec_cast(iris$Sepal.Length, factor())
#> Can't cast <double> to <factor<>>


# (on a separate note)
# I see how is_same_type() lets me do this, but is it right?
vec_assert(iris$Species, integer())
```

<sup>Created on 2019-03-29 by the [reprex package](https://reprex.tidyverse.org) (v0.2.1.9000)</sup>

DavisVaughan:
Closes #106 
Closes #261 

Somewhat related to #161 

This PR implements a `partial_factor()` that learns levels from data.

Should the abbreviation be `prtl` like with partial frames?

``` r
library(vctrs)

partial_factor()
#> partial_factor<
#> 
#> >

partial_factor("x")
#> partial_factor<
#>   5a425 {partial}
#> >

learned <- vec_ptype2(partial_factor("x"), factor("y"))
learned
#> partial_factor<
#>   5a425 {partial}
#>   df698
#> >

# levels come from `learned` first
vec_type_finalise(learned)
#> factor(0)
#> Levels: y x

vec_ptype_common(partial_factor("x"), factor("y"))
#> factor(0)
#> Levels: y x

# mainly wanted it for this
vec_assert(factor("x"), partial_factor())

# but this is unfortunate
vec_assert(1, partial_factor())
#> `y` must be a vector, not a `vctrs_partial_factor/vctrs_partial/vctrs_sclr`
#> object

# same problem with partial frames
vec_assert(1, partial_frame())
#> `y` must be a vector, not a `vctrs_partial_frame/vctrs_partial/vctrs_sclr`
#> object


# inherits the type rules with factors/ordered factors
# which might be problematic
vec_ptype2(partial_factor(), ordered("x"))
#> No common type for `x` <factor<>> and `y` <ordered<5a425>>.
```

<sup>Created on 2019-06-28 by the [reprex package](https://reprex.tidyverse.org) (v0.2.1)</sup>


R/partial-factor.R
```diff
@@ -0,0 +1,102 @@
+#' Partially specify a factor
+#'
+#' This special class can be passed as a `ptype` in order to specify that the
+#' result should be a factor with at least some minimum set of levels.
```
hadley:
```suggestion
#' result should be a factor that contains at least the specified levels.
```

R/partial-factor.R
```diff
@@ -0,0 +1,102 @@
...
+#'
+#' @inheritParams new_factor
+#' @export
+#' @examples
+#' # Assert that `x` is a factor with any levels
```
hadley:
```suggestion
#' # Assert that `x` is a factor
```

Maybe also discuss what `vec_assert(factor("x"), factor())`  does?

R/partial-factor.R
```diff
@@ -0,0 +1,102 @@
...
+  paste0(
+    "partial_factor<\n",
+    paste0(details, collapse = "\n"),
+    "\n>"
+  )
+}
+
+#' @export
+vec_ptype_abbr.vctrs_partial_factor <- function(x, ...) {
+  "prtl"
```
hadley: I think this needs to include `fctr`

PR #467, opened 2019-06-29
Refine documentation of vec_dim() and export dim2()

krlmlr:
The documentation reflects my understanding of the implementation and the difference between `vec_dim()` and `dim2()`. 

I think a `dim2()` function can be useful, e.g. in pillar.

Closes #464.


lionel-:
If we unexport `vec_dim()`, does this mean that `vec_dim_n()` should be renamed to `dim_n()` to be consistent with `dim2()`? But then, is it confusing to have both `dim2()` and `dim_n()`, since the suffix doesn't have the same meaning? Should we just unexport all these helpers and keep copying implementations in our utils.R files?

krlmlr:
I'd say let's unexport everything for now, perhaps by substituting `@export` by `@noRd`. Then we can pick up later.

Is it fair to say that vctrs's notion of "dim" is the combination of size and shape?

lionel-:
> Is it fair to say that vctrs's notion of "dim" is the combination of size and shape?

I think so.

> I'd say let's unexport everything for now,

Sounds good, but will this affect pillar?

krlmlr:
I have already removed the calls to `vec_dim()`.

PR #469, opened 2019-07-01
Register tibble methods conditionally

### LINKED ISSUE #437: Only conditionally export tbl_df methods
So that these methods can be gracefully taken over by the tibble package, without having to do another vctrs release.

@krlmlr

lionel-:
Closes #437


lionel-:
Just adding these implementations should be good.

lionel-:
Without the last commit, vctrs crashes after the second `load_all()`.

R/register-s3.R
```diff
@@ -27,6 +27,12 @@
 #' @param method Optionally, the implementation of the method. By default,
 #'   this will be found by looking for a function called `generic.class`
 #'   in the package environment.
+#'
+#'   Note that supplying this method can be dangerous if you use
```
hadley:
```suggestion
#'   Note that providing `method` can be dangerous if you use
```

PR #474, opened 2019-07-02
Tweak vec_dim() documentation

krlmlr:
Is my understanding correct?

Follow-up to #467.

### DIFF
--- a/R/dim.R
+++ b/R/dim.R
@@ -1,11 +1,20 @@
-#' Vector dimensions
+#' Actual vector dimensions
 #'
+#' @description
 #' * `vec_dim_n()` gives the dimensionality (i.e. number of dimensions)
 #' * `vec_dim()` returns the size of each dimension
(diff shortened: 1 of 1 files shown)


lionel-:
I don't think we should document that `vec_dim()` returns length instead of size.

krlmlr:
This represents the current implementation, I feel it's valuable as such. Do we have an issue that outlines the desired state?

lionel-:
This is already documented as:

```
# FIXME: Should `vec_dim()` return the size instead of the length?
```

I think this represents better the status of the current implementation than making it look like the current semantics are well thought out.

I think the `FIXME` note is sufficient and I'd rather not have another open issue. `vec_dim()` is not used much in the package (but makes one test fail if we change it). We can figure it out later.

lionel-:
Merging as we don't need to spend time on this.

PR #494, opened 2019-07-21
Fix hash test results on big-endian systems.

### LINKED ISSUE #492: Hash test results are not endian-neutral
Building vctrs 0.2.0 across [various arches](https://koji.fedoraproject.org/koji/taskinfo?taskID=36317857) works everywhere except [s390x](https://koji.fedoraproject.org/koji/taskinfo?taskID=36317896), which is a big-endian architecture. The error occurs in a check of hashes from some stored known value:

>   1. Failure: hashes are consistent from run to run (@test-hash.R#61) 

I did not investigate too deeply where the endian-ness comes into play. But since hashes are supposed to be fast-ish, you may not want to implement byte-swapping. I could provide the big-endian hash results instead.

QuLogic:
The result is the same, but looking at the *raw* bytes, the results are
byte swapped. So comparing against a list of bytes from a little-endian
system won't work.

Just swap the byte order before comparing against the ground truth
values. This means there's no need to save two copies of the ground
truth.

Fixes #492.

### DIFF
--- a/tests/testthat/test-hash.R
+++ b/tests/testthat/test-hash.R
@@ -57,6 +57,17 @@ test_that("hashes are consistent from run to run", {
   )
   hash <- lapply(df, vec_hash)
 
+  # Big-endian results are byte-swapped, but otherwise equivalent.
+  # Swap results so that there's no need to save results twice.
+  if (.Platform$endian == "big") {
+    hash <- lapply(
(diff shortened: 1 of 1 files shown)


lionel-:
Thanks for the diagnostic and the fix!

PR #502, opened 2019-07-24
Convert `as_df_row()` inputs to lists with elements of size 1

### LINKED ISSUE #480: `vec_rbind()` internal error with a list of data frames
I'm not entirely sure what I expected as a result here, but something doesn't feel correct.

Error here?

``` r
vctrs::vec_rbind(list(mtcars))
#> New names:
#> * `` -> ...1
#> Error in vctrs::vec_rbind(list(mtcars)): Internal error in `vec_assign()`: `value` should have been recycled to fit `x`.
```

But not here:

``` r
vctrs::vec_rbind(list(x = 1, y = 2))
#>   x y
#> 1 1 2
```

My guess is that it has something to do with this mess:

``` r
vctrs:::as_df_row(list(mtcars))
#> New names:
#> * `` -> ...1
#> Warning in format.data.frame(if (omit) x[seq_len(n0), , drop = FALSE] else
#> x, : corrupt data frame: columns will be truncated or padded with NAs
#>   ...1.mpg ...1.cyl ...1.disp ...1.hp ...1.drat ...1.wt ...1.qsec ...1.vs
#> 1     21.0        6     160.0     110      3.90   2.620     16.46       0
#>   ...1.am ...1.gear ...1.carb
#> 1       1         4         4
```

DavisVaughan:
Closes #480 

This PR technically fixes #480, but I am hoping it can be improved, as I think it will be a bit slow in the long run.

It replaces the call to `r_as_list()` with an alternative, `as_list_row()`, that does something slightly different. It turns its input into a list such that each element of the result is of size 1. Meaning that `list(1:2, 1)` becomes `list(list(1:2), list(1))`. Another way to think of it is that it iterates through `x`, calling `vec_slice(x, i)` for each `i`, and places the result of the slice into the `i`th element of the output list. I think this is the most consistent way for `as_df_row()` to work.

I think `as_list_row()` is going to be somewhat slow though, as the repeated calls to `vec_slice()` might be a problem. Note that this isn't called for data frames (`as_df_row_impl()` early exits), but is called when the user does something like `vec_rbind(1:5)`, which might hurt performance. I'm open to other ideas!


DavisVaughan:
We could probably remove `r_as_list()` now because I think this is the only place it was used

DavisVaughan:
I've optimized it as much as I know how. Here are a few comparisons. We sacrifice a bit of speed when passing in vectors, but gain correct behavior with lists. Again, when in the most common case of working with data frames as input, we won't have any performance changes.

``` r
library(vctrs)
library(tibble)
library(rlang, warn.conflicts = FALSE)

# Will make a 2 row data frame
x <- set_names(1:5000)
y <- set_names(5001:10000, nm = names(x))
```

```r
# Can tell a performance difference between the two

# before
bench::mark(vec_rbind(x, y), iterations = 500)
#> # A tibble: 1 x 6
#> 1 vec_rbind(x, y)   4.13ms   4.91ms      195.     946KB     7.28

# after
bench::mark(vec_rbind(x, y), iterations = 500)
#> # A tibble: 1 x 6
#> 1 vec_rbind(x, y)   5.13ms   5.89ms      160.     985KB     8.40
```

```r
as_df_row <- vctrs:::as_df_row

# Performance loss is here :/

# before
bench::mark(as_df_row(x), iterations = 500)
#> # A tibble: 1 x 6
#> 1 as_df_row(x)    292µs    322µs     3076.    90.8KB     31.1

# after
bench::mark(as_df_row(x), iterations = 500)
#> # A tibble: 1 x 6
#> 1 as_df_row(x)    771µs    993µs      995.    90.8KB     18.2
```

```r
# But this works correctly now:

# before 
as_df_row(list(x = 1:2, y = 2))
#> Warning in format.data.frame(if (omit) x[seq_len(n0), , drop = FALSE] else
#> x, : corrupt data frame: columns will be truncated or padded with NAs
#>   x y
#> 1 1 2

# after
as_df_row(list(x = 1:2, y = 2))
#>      x y
#> 1 1, 2 2
```

```r
# And so does this:

# before
as_tibble(as_df_row(list(cars = mtcars)))
#> Tibble columns must have consistent lengths, only values of length one are recycled:
#> * Length 1: Requested with `.rows` argument
#> * Length 32: Column `cars`

# after
as_tibble(as_df_row(list(cars = mtcars)))
#> # A tibble: 1 x 1
#>   cars               
#>   <list>             
#> 1 <df[,11] [32 × 11]>
```

lionel-:
Maybe a generic `r_as_list()` using the slice internals would solve the performance problems and be generally useful?

DavisVaughan:
I think this works differently from `r_as_list()` though. If you pass `list(1, 2)` into `r_as_list()` I would expect it to return the results unchanged. But for `vec_rbind()` it would need to return `list(list(1), list(2))`. That's what I'm calling a "list row".

lionel-:
> I think this works differently from r_as_list() though.

Good point, then it could be called `vec_split_list()`? This seems to make sense to me as an operation name because if you have an S3 list `<mylist: 1, 2, 3>`, you get `<list: <mylist: 1>, <mylist: 2>, <mylist: 3>>`.

> Note that this does use vec_slice_impl() now to avoid repeated calls to vec_as_index()

There's still redudant work done when calling `vec_slice_impl()` repeatedly with the same vector.

DavisVaughan:
`vec_split_list()` makes sense to me. What you describe with the S3 object is exactly what should happen 👍 

> There's still redundant work done when calling vec_slice_impl() repeatedly with the same vector.

Yea I see that too. For how it will be used in `as_df_row()`, we could also make the assumption that it is not a shaped vector, and is not a data frame. I think one of the main speedups would be that we would be able to only call `vec_proxy_info()` once. And we wouldn't care about slicing names (`as_df_row()` takes care of that). I'll give it a shot.

DavisVaughan:
One more example that didn't work before because of an inconsistency with `as.list()` that works now:

``` r
library(vctrs)
library(rlang)
as_df_row <- vctrs:::as_df_row

# before
as_df_row(set_names(factor(c("a", "b"))))
#> New names:
#> * `` -> ...1
#> * `` -> ...2
#>   ...1 ...2
#> 1    a    b

# after
as_df_row(set_names(factor(c("a", "b"))))
#>   a b
#> 1 a b
```

DavisVaughan:
- `vec_split_list()` is fully implemented, and works in a generic enough way that it could be useful elsewhere. It is exported to the R side, but not to the user, and is fully tested.
- `vec_split_list()` basically eliminates the performance issues from before.

``` r
library(vctrs)
library(tibble)
library(rlang, warn.conflicts = FALSE)

# Will make a 2 row data frame
x <- set_names(1:5000)
y <- set_names(5001:10000, nm = names(x))
```

No more performance issues:

```r
# before
bench::mark(vec_rbind(x, y), iterations = 2500)
#> # A tibble: 1 x 6
#> 1 vec_rbind(x, y)   3.88ms   4.51ms      206.     946KB     7.96

# after
bench::mark(vec_rbind(x, y), iterations = 2500)
#> # A tibble: 1 x 6
#> 1 vec_rbind(x, y)   4.11ms   4.57ms      202.     985KB     7.81
```

It's still slightly slower than using `as.list()` like before, but it scales much better.

```r
as_df_row <- vctrs:::as_df_row

# before
bench::mark(as_df_row(x), iterations = 5000)
#> # A tibble: 1 x 6
#> 1 as_df_row(x)    237µs    300µs     3301.    90.8KB     29.3

# after
bench::mark(as_df_row(x), iterations = 5000)
#> # A tibble: 1 x 6
#> 1 as_df_row(x)    373µs    501µs     1957.    90.8KB     17.0
```

We can see what is taking time in `as_df_row()` by calling `vec_split_list()` by itself. Note that we need to unname `x` to get accurate timings, which is what `as_df_row()` does anyways, because that ends up removing all attributes from `x` so the repeated calls to `vec_restore()` run faster.

So `vec_split_list()` takes about 300us, and `as_df_row()` takes 500us. More investigation showed that `vec_unique_names()` was the other 200us.

```r
vec_split_list <- vctrs:::vec_split_list

bench::mark(vec_split_list(x), iterations = 10000)
#> # A tibble: 1 x 6
#> 1 vec_split_list(x)   1.05ms   1.49ms      657.    39.1KB     36.0

xx <- unname(x)
bench::mark(vec_split_list(xx), iterations = 10000)
#> # A tibble: 1 x 6
#> 1 vec_split_list(xx)    202µs    302µs     3335.    39.1KB     29.9
```

Creating a row data frame from a classed object like a factor takes a little longer because we have to fall back to `[`. This example does not work with the current version because of how `as.list()` with factors works (the names are dropped from the factor and name repair has to kick in). I was impressed that `vec_split_list()` ends up being as fast / faster than `as.list.factor()`

```
# falls back to `[`
fctr <- factor(rep(letters, 100))
names(fctr) <- paste0("nm-", as.character(seq_along(fctr)))

bench::mark(as_df_row(fctr), iterations = 100)
#> # A tibble: 1 x 6
#> 1 as_df_row(fctr)   8.69ms   10.7ms      93.8    57.3KB     55.1

bench::mark(vec_split_list(fctr), iterations = 100)
#> # A tibble: 1 x 6
#> 1 vec_split_list(fctr)   10.8ms   12.4ms      70.0    20.4KB     52.8

bench::mark(as.list(fctr))
#> # A tibble: 1 x 6
#> 1 as.list(fctr)   17.6ms   17.6ms      56.9    36.9KB    1536.
```

In my bench mark with `vec_split_list(x)`, I noted that it matters to unname `x` so it has no attributes. This is because we have to call `vec_restore()` on each sliced result from `x`, and this can be costly if there are any attributes. I don't think there is a good solution to this though.

```
z <- 1:5000
z_with_stuff <- structure(z, stuff = "stuff")

bench::mark(vec_split_list(z)) # vec_restore() early exits, no attributes
#> # A tibble: 1 x 6
#> 1 vec_split_list(z)    236µs    280µs     3343.    58.7KB     29.5

bench::mark(vec_split_list(z_with_stuff)) # vec_restore() has to do work
#> # A tibble: 1 x 6
#> 1 vec_split_list(z_with_stuff)   781µs   965µs      988.    39.1KB     19.2
```

Having any attributes on `z` and then calling `as_df_row()` actually did not work before either, because `as.list()` nukes those attributes.

```r
# before
z_row <- as_df_row(z_with_stuff, quiet = TRUE)
z_row[[1]]
#> [1] 1

# after
z_row <- as_df_row(z_with_stuff, quiet = TRUE)
z_row[[1]]
#> [1] 1
#> attr(,"stuff")
#> [1] "stuff"
```

<sup>Created on 2019-07-26 by the [reprex package](https://reprex.tidyverse.org) (v0.2.1)</sup>

DavisVaughan:
This also tweaks `as_df_row()` to use `vec_names()` rather than `r_names()`, and `vec_unique_names()` to use `vec_names()` as well. This was due to what would happen if you try and `as_df_row()` a POSIXlt object. You'd end up ripping off the names of the underlying list, rather than getting back `NULL`.

lionel-:
@hadley Could you take a look and see if you agree with the new semantics for turning lists into rows please?

DavisVaughan:
The clearest summary of the new semantics is here:
https://github.com/r-lib/vctrs/issues/480#issuecomment-514755539

DavisVaughan:
@lionel-, refactored `vec_split_list()` into multiple sub functions. I like how each sub function has minimal arguments, is easy to read on its own, and makes the main function clearer to follow, even if there is some code repetition between the sub functions to set up the required arguments.

hadley:
Could you please briefly summarise the motivation for this change, preferably by reference to invariants? (i.e. especially what new invariants are now satisfied)

DavisVaughan:
I believe this change enforces this invariant in `vec_rbind()`:

```
For `vec_rbind()`, each element of the vector becomes a column in a single row.
```

To me, an "element" of a list should be extracted with `vec_slice()` (as opposed to `[[`), so the current behavior is incorrect:

``` r
library(vctrs)
library(tibble)

as_tbl_row <- function(x, quiet = FALSE) {
  as_tibble(vctrs:::as_df_row(x, quiet = quiet))
} 

as_tbl_row(list(1, 2, 3), quiet = TRUE)
#> # A tibble: 1 x 3
#>    ...1  ...2  ...3
#>   <dbl> <dbl> <dbl>
#> 1     1     2     3
```

If following my interpretation of the invariant, the correct result is this, which is what the new version does:

``` r
as_tbl_row(list(1, 2, 3), quiet = TRUE)
#> # A tibble: 1 x 3
#>   ...1      ...2      ...3     
#>   <list>    <list>    <list>   
#> 1 <dbl [1]> <dbl [1]> <dbl [1]>
```

As more motivation for this change, when an item in the list has `vec_size(x[[i]]) > 1`, it currently fails because a corrupt data frame is created. Notice that in the first column there are 2 rows, and in the second column there is 1 row (which breaks the 1 row invariant of `as_df_row()`)

``` r
# this fails on the conversion to tibble
as_tbl_row(list(1:2, 3), quiet = TRUE)
#> Tibble columns must have consistent lengths, only values of length one are recycled:
#> * Length 1: Requested with `.rows` argument
#> * Length 2: Column `...1`

# why? because a corrupt df has been created
corrupt <- vctrs:::as_df_row(list(1:2, 3), quiet = TRUE)
corrupt
#> Warning in format.data.frame(if (omit) x[seq_len(n0), , drop = FALSE] else
#> x, : corrupt data frame: columns will be truncated or padded with NAs
#>   ...1 ...2
#> 1    1    3

# 1st col has 2 elements
# 2nd col has 1 element
str(corrupt)
#> 'data.frame':    1 obs. of  2 variables:
#>  $ ...1: int  1 2
#>  $ ...2: num 3

corrupt[[1]]
#> [1] 1 2

corrupt[[2]]
#> [1] 3
```

When following the invariant, this doesn't happen because `list(1:2, 3)` is first converted to `list(list(1:2), list(3))` by `vec_split_list()`. This list is then converted into the 1 row data frame. I think this is more correct behavior as each element of the list (extracted by `vec_slice()`) ends up in its own column, and only 1 row is created.

``` r
as_tbl_row(list(1:2, 3), quiet = TRUE)
#> # A tibble: 1 x 2
#>   ...1      ...2     
#>   <list>    <list>   
#> 1 <int [2]> <dbl [1]>
```

DavisVaughan:
`vec_split_list()` on its own has invariants:

```
# size stability
vec_size(vec_split_list(x)) == vec_size(x)

# each element is size 1
vec_size(vec_split_list(x)[[i]]) == 1

# each element is just the `i`th slice of `x`
vec_split_list(x)[[i]] == vec_slice(x, i)
```

DavisVaughan:
Oooh it just occurred to me that a better name for `vec_split_list()` might be `vec_split_along()`

lionel-:
> Oooh it just occurred to me that a better name for vec_split_list() might be vec_split_along()

I like it.

PR #513, opened 2019-07-30
Teach `vec_slice_impl()` about `compact_seq()` + implement decreasing `compact_seq()`s

### LINKED ISSUE #489: `vec_slice()` should know about `compact_seq()`s
I finally have a use case for `vec_slice()` needing to know about `compact_seq()`s with my `slide()` ideas. 

I probably know enough about this now to implement it when I have a little time.

### LINKED ISSUE #490: `compact_seq()`s should be able to be decreasing
After #489 

It would be nice if you could have a decreasing `compact_seq()` like `5:1`, and have `vec_slice()` / `vec_assign()` understand it.

The combination of this with #489 would have very nice performance implications when you have to repeatedly adjust a sequence to `vec_slice()` with. Rather than regenerate a full SEXP sequence at each iteration of a loop, you can just adjust the `compact_seq()` start/stop locations and pass that to `vec_slice()` instead.

DavisVaughan:
Closes #490 
Closes #489 

This PR does two things:

- It teaches `vec_slice_impl()` how to work with `compact_seq()` objects
- It allows `compact_seq()`s to be negative by redefining a `compact_seq()` object in terms of a start location, a total size, and a step size (`+/-1`). Both `vec_slice_impl()` and `vec_assign_impl()` know how to deal with this.

We don't have a use case for this in vctrs, but I do have a potential use case for this in a C impl of `slide()` so I figured I would add it. 

I have exposed `vec_slice_seq()` to the R side for testing purposes.

Notes: 

- `compact_seq()`s are now _inclusive_ on both endpoints: `[from, to]` rather than the previous behavior of `[from, to)`. The original behavior was a bit hard to reason about with decreasing sequences, and kept you from being able to compute a decreasing sequence with only the 1st value, i.e. you would try `[0, 0)` but the size of the sequence would be computed as 0 and the sequence would be empty. Now it would be computed as `[0, 0]`, which is a sequence of 1 value at position 1.

- You can't have a seq with `NA` in it, so you'll see that `NA_VALUE` is not passed through in the macros.

- The use of `memcpy()` in `ASSIGN_COMPACT` was replaced with an assignment loop to be compatible with decreasing compact sequences. A few bench marks didn't show any loss in performance.

Using `compact_seq()`s in `vec_slice_impl()` can have some nice performance improvements:

``` r
library(nycflights13)
library(vctrs)

x <- flights

nrow(x)
#> [1] 336776

vec_slice_seq <- vctrs:::vec_slice_seq

bench::mark(
  vec_slice_seq(x, 0L, 99999L),
  vec_slice(x, 1:100000),
  x[1:100000,],
  iterations = 1000
)
#> # A tibble: 3 x 6
#> 1 vec_slice_seq(x, 0L, 99999L)  1.66ms  2.34ms      407.    11.8MB    116. 
#> 2 vec_slice(x, 1:1e+05)         1.93ms  2.64ms      372.    11.8MB    101. 
#> 3 x[1:1e+05, ]                  4.38ms  5.93ms      170.    11.9MB     47.0

year <- x$year

bench::mark(
  vec_slice_seq(year, 0L, 99999L),
  vec_slice(year, 1:100000),
  year[1:100000],
  iterations = 5000
)
#> # A tibble: 3 x 6
#>   expression                          min  median `itr/sec` mem_alloc
#>   <bch:expr>                      <bch:t> <bch:t>     <dbl> <bch:byt>
#> 1 vec_slice_seq(year, 0L, 99999L)  36.8µs  62.4µs    16129.     391KB
#> 2 vec_slice(year, 1:1e+05)        143.9µs 219.7µs     4418.     781KB
#> 3 year[1:1e+05]                   186.3µs 263.1µs     3908.     781KB
```

### DIFF
--- a/R/slice.R
+++ b/R/slice.R
@@ -197,3 +197,8 @@ vec_init <- function(x, n = 1L) {
 vec_split_along <- function(x) {
   .Call(vctrs_split_along, x)
 }
+
+# Exposed for testing (`start` is 0-based)
+vec_slice_seq <- function(x, start, size, increasing = TRUE) {
+  .Call(vctrs_slice_seq, x, start, size, increasing)
+}
--- a/src/bind.c
+++ b/src/bind.c
@@ -73,7 +73,7 @@ static SEXP vec_rbind(SEXP xs, SEXP ptype, SEXP names_to, enum name_repair_arg n
   }
 
   SEXP out = PROTECT_N(vec_init(ptype, nrow), &nprot);
-  SEXP idx = PROTECT_N(compact_seq(0, 0), &nprot);
+  SEXP idx = PROTECT_N(compact_seq(0, 0, true), &nprot);
   int* idx_ptr = INTEGER(idx);
 
   SEXP names_to_col = R_NilValue;
@@ -104,7 +104,7 @@ static SEXP vec_rbind(SEXP xs, SEXP ptype, SEXP names_to, enum name_repair_arg n
     }
 
     SEXP tbl = PROTECT(vec_cast(VECTOR_ELT(xs, i), ptype, args_empty, args_empty));
-    init_compact_seq(idx_ptr, counter, counter + size);
+    init_compact_seq(idx_ptr, counter, size, true);
     df_assign(out, idx, tbl, false);
 
     // Assign current name to group vector, if supplied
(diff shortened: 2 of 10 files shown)


DavisVaughan:
@lionel- I think this worked very well! 

- We can now use `compact_seq(size = 0)` to generate a 0-length sequence. There are new tests for this
- The end result is very clean.
   - `vec_rbind()` now just does `init_compact_seq(idx_ptr, counter, size, true);`, no need for the `counter + size - 1` like before
   - The slicing functions are essentially the same as before, but I think they are easy enough to follow.

I did a squash + cherry-pick to resolve a few merge conflicts that had cropped up which might make this harder to review. I should have waited until you took a look at the changes. Sorry about that.

PR #518, opened 2019-07-31
Implement `new_list_of()` and `init_list_of()` in C

### LINKED ISSUE #517: Implement simplified `new_list_of()` constructor in C
To construct `list_of<>` objects from C, it would be useful to have a simplified constructor there. 

This would also have performance benefits for lists/data frames since right now `new_list_of()` called from R goes through `new_vctr()` and `vec_set_attributes()`, which calls `attributes(x) <- attrib` resulting in 2 copies of the list. This can be painful if the list is very large.

See this comment/issue for more information on how it should be implemented:
https://github.com/r-lib/vctrs/pull/515#issuecomment-516841230

Also replace this ad hoc section with the constructor when implemented:
https://github.com/r-lib/vctrs/blob/59783eebbd885b4f0474a7352b07f6b77f8c1509/src/dictionary.c#L435

I'm imagining: `new_list_of(SEXP x, SEXP ptype)` that checks that `x` is a list and that `ptype` has size `0`, calls `r_maybe_duplicate(x)`, then runs the two `setAttrib()` lines already used in the section above

DavisVaughan:
Closes #517 

I tried following the `new_data_frame()` + `init_data_frame()` model. This is nice because we can do the bare minimum in `vctrs_split_id()` by using `init_list_of()` where we _know_ the input is correct. That will be tested by `vec_split_id()` tests which compare against objects created with the R level `list_of()`.

Questions:

`new_list_of()` currently has no use, but I feel like it will be used eventually. So there are no tests for it. I'm not sure what to do here.

Should I put these in a new `type-list-of.c` file? Or keep them in `utils.c`?

### DIFF
--- a/src/utils.c
+++ b/src/utils.c
@@ -161,6 +161,28 @@ SEXP vctrs_set_attributes(SEXP x, SEXP attrib) {
   return x;
 }
 
+SEXP new_list_of(SEXP x, SEXP ptype) {
+  if (TYPEOF(x) != VECSXP) {
+    Rf_errorcall(R_NilValue, "Internal error: `x` must be a list.");
+  }
+
+  if (vec_size(ptype) != 0) {
+    Rf_errorcall(R_NilValue, "Internal error: `ptype` must be a prototype with size 0.");
+  }
+
+  x = PROTECT(r_maybe_duplicate(x));
+
+  init_list_of(x, ptype);
+
+  UNPROTECT(1);
+  return x;
+}
+
+void init_list_of(SEXP x, SEXP ptype) {
+  Rf_setAttrib(x, R_ClassSymbol, classes_list_of);
+  Rf_setAttrib(x, syms_ptype, ptype);
+}
+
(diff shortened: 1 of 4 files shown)


src/utils.c
```diff
@@ -161,6 +161,28 @@ SEXP vctrs_set_attributes(SEXP x, SEXP attrib) {
   return x;
 }
 
+SEXP new_list_of(SEXP x, SEXP ptype) {
```
lionel-: Maybe in `type-list-of.c`?

lionel-: We should probably move the data frame utils to `type-dataframe.c` eventually.

DavisVaughan: ✅ Maybe in `type-list-of.c`?

> move the data frame utils to type-dataframe.c eventually.

Hasn't that been done already? https://github.com/r-lib/vctrs/blob/8d12bfc0e29e056966e0549af619253253752a64/src/type-data-frame.c#L6

PR #526, opened 2019-08-05
Fix data frame & proxied object name handling in `vec_c()`

### LINKED ISSUE #520: Do we need `vec_set_names()`?
I see we have `vec_names<-` but there is no C equivalent. Looking at it + looking at what we do in `vec_names()`, I came up with this function. Would it be useful?

I needed it in `slide()` to set the names on the output container to be the same as the names of the input. It's more complicated than `purrr::map()` though, because the output can be any type so we have to be careful about setting names on, say, a POSIXlt object, by calling back to `names<-`.

I wanted to make it a `void` function because most of the time it just does `Rf_setAttrib(x, R_NamesSymbol, names);`, but I don't know how to make that compatible with `names<-` which won't modify in place like that and actually returns a SEXP. So its in this weird state where it modifies in place sometimes, and doesn't other times. (I guess if I did `r_maybe_duplicate()` before doing `Rf_setAttrib()` then that would have consistent behavior)

`syms_set_rownames = Rf_install("rownames<-")`

`syms_set_names = Rf_install("names<-")`

I realize I could use the `vctrs_dispatch` helpers if we implemented this here

```c
SEXP vec_set_names(SEXP x, SEXP names) {
  if (names == R_NilValue) {
    return x;
  }

  // Never on a data frame
  if (OBJECT(x) && Rf_inherits(x, "data.frame")) {
    return x;
  }

  // rownames(x) <- names
  if (vec_dim_n(x) > 1) {
    SEXP env = PROTECT(r_new_environment(R_GlobalEnv, 3));

    Rf_defineVar(syms_set_rownames, fns_set_rownames, env);
    Rf_defineVar(syms_x, x, env);
    Rf_defineVar(syms_names, names, env);

    SEXP call = PROTECT(Rf_lang3(syms_set_rownames, syms_x, syms_names));

    UNPROTECT(2);
    return Rf_eval(call, env);
  }

  // names(x) <- names
  if (OBJECT(x)) {
    SEXP env = PROTECT(r_new_environment(R_GlobalEnv, 3));

    Rf_defineVar(syms_set_names, fns_set_names, env);
    Rf_defineVar(syms_x, x, env);
    Rf_defineVar(syms_names, names, env);

    SEXP call = PROTECT(Rf_lang3(syms_set_names, syms_x, syms_names));

    UNPROTECT(2);
    return Rf_eval(call, env);
  }

  Rf_setAttrib(x, R_NamesSymbol, names);
  return x;
}
```

### LINKED ISSUE #524: Is vec_c() name repair with data frames correct?
The third and fourth examples here are inconsistent, I think. I think outer names should probably be ignored for data frames here? `vec_rbind()` allows you to create a `.names_to` column, but I don't think we want that. The alternative is using outer names as row names but I don't think we want that either.

``` r
library(vctrs)

# (looks good) inner names used
vec_c(data.frame(xx = 1), data.frame(yy = 2))
#>   xx yy
#> 1  1 NA
#> 2 NA  2

# (looks good) row names ignored - i think this is on purpose?
vec_c(data.frame(xx = 1, row.names = "r1"), data.frame(yy = 2, row.names = "r2"))
#>   xx yy
#> 1  1 NA
#> 2 NA  2

# outer names used for columns?
# I think these should either have been used for row names (ew)
# or ignored? Then xx and yy would be the actual column names
vec_c(x = data.frame(xx = 1), y = data.frame(yy = 2))
#>    x  y
#> 1  1 NA
#> 2 NA  2

# I assume this error has something to do with how names are treated above
# It's trying to construct a column name vector of `c("x", "y")` but the resulting
# data frame should only have 1 column. Again, they should either become row
# names or get ignored
vec_c(x = data.frame(xx = 1), y = data.frame(xx = 2))
#> Error in vec_c(x = data.frame(xx = 1), y = data.frame(xx = 2)): 'names' attribute [2] must be the same length as the vector [1]
```

<sup>Created on 2019-08-03 by the [reprex package](https://reprex.tidyverse.org) (v0.2.1)</sup>

DavisVaughan:
Closes #524 
Closes #520 

This PR:

- Implements `vec_set_names()`. It has the following semantics:
   - Never sets row names on data frames
   - Sets names on unclassed vectors
   - Sets row names on unclassed matrices/arrays
   - Falls back to `names<-` on classed vectors
   - Falls back to `rownames<-` on classed matrices/arrays
- Uses `vec_set_names()` in `vec_c()` for better name handling with proxied objects
- Tweaks `has_names` in `vec_c()` to completely avoid outer name handling when the `ptype` is a data frame. Outer names are completely ignored here.

Here are a few examples of what this fixes:

``` r
library(vctrs)
library(rlang)
```

```r
# before
vec_c(x = data.frame(xx = 1), y = data.frame(yy = 2))
#>    x  y
#> 1  1 NA
#> 2 NA  2

# after
vec_c(x = data.frame(xx = 1), y = data.frame(yy = 2))
#>   xx yy
#> 1  1 NA
#> 2 NA  2
```

```r
# before
vec_c(x = data.frame(xx = 1), y = data.frame(xx = 2))
#> Error in vec_c(x = data.frame(xx = 1), y = data.frame(xx = 2)): 'names' attribute [2] must be the same length as the vector [1]

# after
vec_c(x = data.frame(xx = 1), y = data.frame(xx = 2))
#>   xx
#> 1  1
#> 2  2
```

```r
datetime_lt <- as.POSIXlt(new_datetime(0))
datetimes_lt <- as.POSIXlt(new_datetime(c(0, 1)))
named_datetime_lt <- set_names(datetime_lt, "dt_name")
```

```r
# before
vec_c(x = datetime_lt)
#> [1] "1969-12-31 19:00:00 EST"

# after
vec_c(x = datetime_lt)
#>                         x 
#> "1969-12-31 19:00:00 EST"
```

```r
# before
vec_c(x = datetimes_lt, .name_spec = "{outer}_{inner}")
#> [1] "1969-12-31 19:00:00 EST" "1969-12-31 19:00:01 EST"

# after
vec_c(x = datetimes_lt, .name_spec = "{outer}_{inner}")
#>                       x_1                       x_2 
#> "1969-12-31 19:00:00 EST" "1969-12-31 19:00:01 EST"
```

```r
# before
vec_c(x = named_datetime_lt, .name_spec = "{outer}_{inner}")
#> [1] "1969-12-31 19:00:00 EST"

# after
vec_c(x = named_datetime_lt, .name_spec = "{outer}_{inner}")
#>                 x_dt_name 
#> "1969-12-31 19:00:00 EST"
```

### DIFF
--- a/R/names.R
+++ b/R/names.R
@@ -463,8 +463,14 @@ tick_if_needed <- function(x) {
 }
 
 # Used in names.c
-set_rownames <- function(x, names) {
+set_rownames_fallback <- function(x, names) {
   rownames(x) <- names
+  x
+}
+
+# Used in names.c
+set_names_fallback <- function(x, names) {
+  names(x) <- names
   x
 }
 
--- a/R/names.R
+++ b/R/names.R
@@ -474,6 +474,10 @@ set_names_fallback <- function(x, names) {
   x
 }
 
+vec_set_names <- function(x, names) {
+  .Call(vctrs_set_names, x, names)
+}
+
 #' Repair names with legacy method
 #'
 #' This standardises names with the legacy approach that was used in
--- a/src/c.c
+++ b/src/c.c
@@ -64,6 +64,7 @@ static SEXP vec_c(SEXP xs,
 
   SEXP xs_names = PROTECT(r_names(xs));
   bool has_names = xs_names != R_NilValue || list_has_inner_names(xs);
+  has_names = has_names && !is_data_frame(ptype);
   SEXP out_names = has_names ? Rf_allocVector(STRSXP, out_size) : R_NilValue;
   PROTECT(out_names);
 
--- a/src/c.c
+++ b/src/c.c
@@ -112,8 +112,9 @@ static SEXP vec_c(SEXP xs,
 
   if (has_names) {
     out_names = PROTECT(vec_as_names(out_names, name_repair, false));
-    out = PROTECT(vec_set_names(out, out_names));
-    UNPROTECT(2);
+    out = vec_set_names(out, out_names);
+    REPROTECT(out, out_pi);
+    UNPROTECT(1);
   }
 
   UNPROTECT(6);
--- a/src/c.c
+++ b/src/c.c
@@ -109,15 +109,8 @@ static SEXP vec_c(SEXP xs,
 
   if (has_names) {
     out_names = PROTECT(vec_as_names(out_names, name_repair, false));
-
-    if (is_shaped) {
-      out = set_rownames(out, out_names);
-      REPROTECT(out, out_pi);
-    } else {
-      Rf_setAttrib(out, R_NamesSymbol, out_names);
-    }
-
-    UNPROTECT(1);
+    out = PROTECT(vec_set_names(out, out_names));
+    UNPROTECT(2);
   }
 
   out = vec_restore(out, ptype, R_NilValue);
--- a/src/c.c
+++ b/src/c.c
@@ -107,13 +107,13 @@ static SEXP vec_c(SEXP xs,
     UNPROTECT(1);
   }
 
+  out = vec_restore(out, ptype, R_NilValue);
+
   if (has_names) {
     out_names = PROTECT(vec_as_names(out_names, name_repair, false));
     out = PROTECT(vec_set_names(out, out_names));
     UNPROTECT(2);
   }
-
-  out = vec_restore(out, ptype, R_NilValue);
 
   UNPROTECT(6);
   return out;
--- a/src/init.c
+++ b/src/init.c
@@ -54,6 +54,7 @@ extern SEXP vctrs_as_minimal_names(SEXP);
 extern SEXP vec_names(SEXP);
 extern SEXP vctrs_is_unique_names(SEXP);
 extern SEXP vctrs_as_unique_names(SEXP, SEXP);
+extern SEXP vec_set_names(SEXP, SEXP);
 extern SEXP vctrs_df_as_dataframe(SEXP, SEXP, SEXP, SEXP);
 extern SEXP vctrs_type2_df_df(SEXP, SEXP, SEXP, SEXP);
 extern SEXP vctrs_type_info(SEXP);
@@ -123,6 +124,7 @@ static const R_CallMethodDef CallEntries[] = {
   {"vctrs_names",                      (DL_FUNC) &vec_names, 1},
   {"vctrs_is_unique_names",            (DL_FUNC) &vctrs_is_unique_names, 1},
   {"vctrs_as_unique_names",            (DL_FUNC) &vctrs_as_unique_names, 2},
+  {"vctrs_set_names",                  (DL_FUNC) &vec_set_names, 2},
   {"vctrs_df_as_dataframe",            (DL_FUNC) &vctrs_df_as_dataframe, 4},
   {"vctrs_type2_df_df",                (DL_FUNC) &vctrs_type2_df_df, 4},
   {"vctrs_type_info",                  (DL_FUNC) &vctrs_type_info, 1},
--- a/src/names.c
+++ b/src/names.c
@@ -659,6 +659,8 @@ SEXP vec_set_rownames(SEXP x, SEXP names) {
   return x;
 }
 
+// FIXME: Do we need to get the vec_proxy() and only fall back if it doesn't
+// exist? See #526 and #531 for discussion and the related issue.
 // [[ include("utils.h"); register() ]]
 SEXP vec_set_names(SEXP x, SEXP names) {
   // Never on a data frame
--- a/src/names.c
+++ b/src/names.c
@@ -603,7 +603,6 @@ static SEXP set_names_fallback(SEXP x, SEXP names) {
                          syms_names, names);
 }
 
-// [[ include("utils.h") ]]
 SEXP vec_set_rownames(SEXP x, SEXP names) {
   if (!has_dim(x)) {
     Rf_errorcall(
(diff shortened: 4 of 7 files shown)


lionel-:
Regarding data frames, I think we should set the row names of data frames, consistently with `vec_rbind()`. If the names are not repaired and duplicates/empty names are detected, this should be an error.

Regarding classed objects, I think ideally we'd have an internal generic mechanism for setting names, via the proxy. If the data frame class accepted duplicate row names, we could use this for record types. Since that's not the case, we might need support for a `.name` column vector in the proxy. Though that would make elements with different names compare unequal, which we probably don't want (the name is not part of the scalar value). So maybe a different `rownames` attribute.

DavisVaughan:
Also worth mentioning that POSIXlt works by assigning the names onto the `$year` slot. This way, they get sliced whenever the year vector gets sliced. That is actually the only reason `vec_slice(<named POSIXlt>, index)` actually keeps the names as it should.

I'm not sure how this could generalize to all record types.

``` r
library(vctrs)
library(rlang)

x <- set_names(as.POSIXlt(new_datetime(c(0, 1))), c("a", "b"))
x
#>                         a                         b 
#> "1969-12-31 19:00:00 EST" "1969-12-31 19:00:01 EST"

names(x)
#> [1] "a" "b"

# names are on `year`
str(unclass(x))
#> List of 11
#>  $ sec   : num [1:2] 0 1
#>  $ min   : int [1:2] 0 0
#>  $ hour  : int [1:2] 19 19
#>  $ mday  : int [1:2] 31 31
#>  $ mon   : int [1:2] 11 11
#>  $ year  : Named int [1:2] 69 69
#>   ..- attr(*, "names")= chr [1:2] "a" "b"
#>  $ wday  : int [1:2] 3 3
#>  $ yday  : int [1:2] 364 364
#>  $ isdst : int [1:2] 0 0
#>  $ zone  : chr [1:2] "EST" "EST"
#>  $ gmtoff: int [1:2] -18000 -18000
#>  - attr(*, "tzone")= chr [1:3] "" "EST" "EDT"

# ...which is the only reason this works
# (names get sliced along with the vector `year`)
vec_slice(x, 1)
#>                         a 
#> "1969-12-31 19:00:00 EST"

# can simulate it by doing this:
vec_restore(vec_slice(vec_proxy(x), 1L), x)
#>                         a 
#> "1969-12-31 19:00:00 EST"


# but then you need methods that know about this
names.POSIXlt
#> function (x) 
#> names(x$year)
#> <bytecode: 0x7fe9028f38b0>
#> <environment: namespace:base>

`names<-.POSIXlt`
#> function (x, value) 
#> {
#>     names(x$year) <- value
#>     x
#> }
#> <bytecode: 0x7fe923771528>
#> <environment: namespace:base>
```

<sup>Created on 2019-08-06 by the [reprex package](https://reprex.tidyverse.org) (v0.2.1)</sup>

lionel-:
How about the suggestion above, of a `vctrs_rownames` attribute on data frame proxies? `names.vctrs_rcrd` would handle it.

DavisVaughan:
I assume we'd need `df_slice()` to be aware of this? It would then handle data frames and data frame proxies. I suppose we would use `slice_names()` rather than `slice_rownames()` inside `df_slice()` because we do not want it to check for uniqueness.

df_slice()
https://github.com/r-lib/vctrs/blob/master/src/slice.c#L181

slice_names()
https://github.com/r-lib/vctrs/blob/master/src/slice.c#L229

slice_rownames()
https://github.com/r-lib/vctrs/blob/master/src/slice.c#L252

hadley:
I think we should merge this (after the code has been reviewed), and then carefully consider how vctrs should handle rownames for data frames and tibbles.

lionel-:
If we're going to fall back to `names<-`, we probably need to use the same strategy as the `[` fallback in `vec_slice()`: Get the `vec_proxy()` first, and if it exists, we don't fall back. vctrs classes with a special names vectors would need to implement a record type, following the discussion in #529.

DavisVaughan:
I'm not quite sure that I see how to do this. Say for a POSIXlt object we get the `vec_proxy()`. It exists, so we don't fall back. With the resulting data frame, we somehow have to know to update the names on the `$year` column. I'm not sure how we could know that without using `names<-` on the non-proxied object.

``` r
library(vctrs)

x <- as.POSIXlt(new_datetime(c(0, 1)))
names(x) <- c("a", "b")

x
#>                         a                         b 
#> "1969-12-31 19:00:00 EST" "1969-12-31 19:00:01 EST"

# Say we don't fall back to `names<-`.
# Now what? Somehow we have to know that
# the names to update are on $year
vec_proxy(x)
#>   sec min hour mday mon year wday yday isdst zone gmtoff
#> 1   0   0   19   31  11   69    3  364     0  EST -18000
#> 2   1   0   19   31  11   69    3  364     0  EST -18000

vec_proxy(x)$year
#>  a  b 
#> 69 69
```

Even `vctrs_vctr` has a `names<-vctrs_vctr` implementation. I don't think we'd want to bypass it even though `vctrs_vctr` implements `vec_proxy()`.

lionel-:
Maybe we should take POSIXlt as a test case for the details attributes and the `.name` column, and adjust the proxy and restore methods accordingly.

> Even vctrs_vctr has a names<-vctrs_vctr implementation. I don't think we'd want to bypass it even though vctrs_vctr implements vec_proxy().

I think we want to bypass it just like we bypass `[.vctrs_vctr`. These methods are for compatibility of vctrs classes with base R, and should call into the vctrs API.

lionel-:
Maybe add a FIXME comment that the fallback needs to be sorted out later, and open an issue? This way we can merge this PR and do things in steps.

PR #534, opened 2019-08-07
vec_recycle() throws classed error

### LINKED ISSUE #471: Should `vec_recycle()` throw a classed error?
`vec_recycle()` throws a standard error when the sizes are incompatible.
https://github.com/r-lib/vctrs/blob/010f3a22d574b398a28d0b9e27447c04e77912a4/src/size.c#L182

There is a spot in `tibble` where it would be nice to catch this specific error and rethrow it with a tibble specific error message (currently `vec_recycle_common()` is being called there but it will be changed to `vec_recycle()`). 
https://github.com/tidyverse/tibble/commit/7728a8b1c8e9ce0d7c881bc6fc64ead7b34572e4#diff-540c0e3edd23d1e0767808051997d469R238

I'm not quite sure what to call that error message. We have `stop_incompatible_size()` but that is for comparing between two vectors.

krlmlr:
Closes #471.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,6 +1,6 @@
 Package: vctrs
 Title: Vector Helpers
-Version: 0.2.0.9000
+Version: 0.2.0.9001
 Authors@R: 
     c(person(given = "Hadley",
              family = "Wickham",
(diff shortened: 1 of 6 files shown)


lionel-:
Thanks!

Issue #535, opened 2019-08-08
vec_size() takes a "long time" for POSIXlt


DavisVaughan:
``` r
library(vctrs)

ct <- new_datetime(0)
lt <- as.POSIXlt(ct)

bench::mark(
  vec_size(ct),
  vec_size(lt)
)
#> # A tibble: 2 x 6
#>   expression        min   median `itr/sec` mem_alloc `gc/sec`
#>   <bch:expr>   <bch:tm> <bch:tm>     <dbl> <bch:byt>    <dbl>
#> 1 vec_size(ct)    727ns    944ns  1013283.     6.3KB    101. 
#> 2 vec_size(lt)   18.6µs   21.8µs    42695.    29.5KB     25.6
```

Discovered while repeatedly calling `vec_restore(x, to, R_NilValue)` where `x` was a list holding a POSIXlt and `to` was a data frame. Because no size was passed in, it goes through `df_raw_size(x)`, which in this case calls `vec_size()` on the first element of the list (the POSIXlt). Totally crippled performance 😬 . Luckily I am able to pre-compute the size ahead of time to fix it for my use case.

The main "issue" is that `vec_size()` needs the proxy:
https://github.com/r-lib/vctrs/blob/f736b2353d248820847a09b808cb7370ab8fc2f6/src/size.c#L47

and `vec_proxy.POSIXlt` calls `new_data_frame()`, which is quite slow:

``` r
library(vctrs)

ct <- new_datetime(0)
lt <- as.POSIXlt(ct)

# imitate vec_proxy.POSIXlt
vctrs:::vec_proxy.POSIXlt
#> function(x, ...) {
#>   new_data_frame(unclass(x))
#> }
#> <bytecode: 0x7f99f69716f8>
#> <environment: namespace:vctrs>

# no problem here:
bench::mark(unclass(lt))
#> # A tibble: 1 x 6
#>   expression       min   median `itr/sec` mem_alloc `gc/sec`
#>   <bch:expr>  <bch:tm> <bch:tm>     <dbl> <bch:byt>    <dbl>
#> 1 unclass(lt)    308ns    373ns  2350164.        0B     235.

lt_unclassed <- unclass(lt)

# ouch
bench::mark(new_data_frame(lt_unclassed))
#> # A tibble: 1 x 6
#>   expression                       min  median `itr/sec` mem_alloc `gc/sec`
#>   <bch:expr>                   <bch:t> <bch:t>     <dbl> <bch:byt>    <dbl>
[...]

PR #536, opened 2019-08-08
Remove usage of `stopifnot()` and `structure()` in `new_data_frame()`

DavisVaughan:
Helps with #535 (maybe closes?)

`new_data_frame()` is _relatively_ slow, and since it can potentially be called many times through the use of `vec_proxy.POSIXlt()`, it can sometimes hurt performance. I've removed the usage of `stopifnot()` and `structure()` to get better performance.

``` r
library(vctrs)

ct <- new_datetime(0)
lt <- as.POSIXlt(ct)
lt_unclassed <- unclass(lt)
```

`new_data_frame()` on a list is now faster:

```r
# before
bench::mark(new_data_frame(lt_unclassed), iterations = 100000)
#> # A tibble: 1 x 6
#> 1 new_data_frame(lt_unclassed)  11.6µs  16.9µs    58050.    24.8KB     63.9

# after
bench::mark(new_data_frame(lt_unclassed), iterations = 100000)
#> # A tibble: 1 x 6
#> 1 new_data_frame(lt_unclassed)  4.36µs  6.05µs   162583.    30.2KB     79.7
```

Resulting in better performance when using `vec_size()` on a POSIXlt, which has to get the `vec_proxy.POSIXlt()`, which calls `new_data_frame()`.

```r
# before
bench::mark(vec_size(ct), vec_size(lt), iterations = 100000)
#> # A tibble: 2 x 6
#> 1 vec_size(ct)    522ns    645ns  1442711.     6.3KB     14.4
#> 2 vec_size(lt)     14µs   18.7µs    54743.    4.68KB     23.5

# after
bench::mark(vec_size(ct), vec_size(lt), iterations = 100000)
#> # A tibble: 2 x 6
#> 1 vec_size(ct)    501ns    572ns  1370221.     6.3KB     13.7
#> 2 vec_size(lt)   5.79µs   7.08µs   134353.    4.68KB     18.8
```

### DIFF
--- a/R/type-data-frame.R
+++ b/R/type-data-frame.R
@@ -16,9 +16,15 @@
 #' @examples
 #' new_data_frame(list(x = 1:10, y = 10:1))
 new_data_frame <- function(x = list(), n = NULL, ..., class = character()) {
-  stopifnot(is.list(x))
-  n <- n %||% df_size(x)
-  stopifnot(is.integer(n), length(n) == 1L)
+  if (!is.list(x)) {
+    abort("`x` must be a list.")
+  }
+
+  if (is.null(n)) {
+    n <- df_size(x)
+  } else if (!is.integer(n) || length(n) != 1L) {
+    abort("`n` must be an integer of size 1.")
+  }
 
   # names() should always be a character vector, but we can't enforce that
   # because as.data.frame() returns a data frame with NULL names to indicate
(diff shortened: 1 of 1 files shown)


R/type-data-frame.R
```diff
@@ -27,12 +33,15 @@ new_data_frame <- function(x = list(), n = NULL, ..., class = character()) {
...
 
-  structure(
-    x,
+  new_attributes <- list(
     ...,
     class = c(class, "data.frame"),
     row.names = .set_row_names(n)
   )
+
+  attributes(x) <- c(attributes(x), new_attributes)
```
hadley: I think a constructor should overwrite existing attributes?

DavisVaughan: `structure()` doesn't strip existing attributes, and that's how we get the `names()` off of `x` for the new data frame, but I can alter it to just be like `c(list(names = names(x), new_attributes)` (which feels appropriate because right now I think the `tzone` attribute is coming along too, and it shouldn't)

PR #541, opened 2019-08-14
Register C callables

DavisVaughan:
This PR registers a handful of C callables that are used by slurrr. It also creates a `inst/include/vctrs.h` API file so I just have to do `<vctrs.h>` in slurrr.

- `vctrs_cast()` is exported rather than `vec_cast()` because the last two arguments of `vec_cast()` are `struct vctrs_arg` types, but are `SEXP`s for `vctrs_cast()`. I have been told this might be changed in the next release anyways, so I'm not too worried about it for now.

- `compact_seq()` and `init_compact_seq()` are incredibly useful for efficient slicing with `vec_slice_impl()`, but aren't really part of the rest of the API. I've separated them a bit and marked them as "extremely" experimental.

### DIFF
--- a/inst/include/vctrs.h
+++ b/inst/include/vctrs.h
@@ -5,115 +5,19 @@
 #include <R_ext/Rdynload.h>
 #include <stdbool.h>
 
-static R_INLINE R_len_t vec_size(SEXP x) {
-  static R_len_t (*fn)(SEXP) = NULL;
+R_len_t (*vec_size)(SEXP);
+SEXP (*vec_proxy)(SEXP);
+SEXP (*vec_restore)(SEXP, SEXP, SEXP);
+SEXP (*vec_init)(SEXP, R_len_t);
+SEXP (*vec_assign_impl)(SEXP, SEXP, SEXP, bool);
+SEXP (*vec_slice_impl)(SEXP, SEXP);
+SEXP (*vec_names)(SEXP);
+SEXP (*vec_set_names)(SEXP, SEXP);
 
-  if (fn == NULL) {
-    fn = (R_len_t (*)(SEXP)) R_GetCCallable("vctrs", "vec_size");
-  }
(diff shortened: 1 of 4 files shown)


inst/include/vctrs.h
```diff
@@ -0,0 +1,119 @@
+#ifndef VCTRS_H
+#define VCTRS_H
+
+#include <Rinternals.h>
+#include <R_ext/Rdynload.h>
+#include <stdbool.h>
+
+static R_INLINE R_len_t vec_size(SEXP x) {
```
lionel-: It bothers me that we're not returning a long length here. It was fine internally, but for exporting we might want to return the larger type right away, otherwise we might end up with redundant functions in the API, one for each type. It is easier to directly tackle large vector than after the fact. On the other hand since vctrs primitives don't support large vectors, it wouldn't be consistent and possibly dangerous...

inst/include/vctrs.h
```diff
@@ -0,0 +1,119 @@
...
+
+  if (fn == NULL) {
+    fn = (SEXP (*)(SEXP, SEXP)) R_GetCCallable("vctrs", "vec_set_names");
+  }
+
+  return fn(x, names);
+}
+
+
+static R_INLINE SEXP compact_seq(R_len_t start, R_len_t size, bool increasing) {
```
lionel-: We should probably remove the more experimental functions from this file, e.g. the compact seq stuff and `vctrs_cast`. Still export them as C callables but they'll have to be declared manually in the client package.

DavisVaughan:
- There are now 2 files, as requested. This seems cleaner, I think, even if it is a _little_ more work for the user of the API.
- Extremely experimental functions have been removed from the api header, but are still registered callables

See the companion slurrr PR where this is all working correctly:
https://github.com/DavisVaughan/slurrr/pull/6

__I'll wait until you decide on whether or not you want to change `R_len_t` -> `R_xlen_t` for `vec_size()`.__ Just let me know!

PR #542, opened 2019-08-15
[<- assignment requires data frame on the rhs

krlmlr:
Fixes tidyverse/tibble#636.


hadley:
Could you add a unit test too?

krlmlr:
We have tests for `vec_proxy_compare.data.frame()` . This is to satisfy future requirements of `[<-` for tibbles -- updating multiple columns requires a data frame on the rhs.

krlmlr:
Let's wait a bit here.

PR #555, opened 2019-08-30
Fix string comparison with non-utf8 encodings

### LINKED ISSUE #553: vec_unique() is sensitive to the marked encoding, where base::unique() is not
*See below for a more minimal reprex.*

<details>
<summary>Long backstory</summary>
The troublesome object comes from https://github.com/tidyverse/tidyr/issues/722. It presents as a problem with `tidyr::unnest()` but I've narrowed it down to a very weird phenomenon with `vctrs::vec_rbind()`. I've pulled out the relevant list-column here, as just a list of tibbles.

``` r
result_ok <- list(
  tibble::tibble(
    `Max Temp Flag` = character(0),
    `Min Temp (°C)` = character(0),
    `Min Temp Flag` = character(0),
    `Mean Temp (°C)` = character(0)
  ),
  tibble::tibble(
    `Max Temp Flag` = rep(NA_character_, 6),
    `Min Temp (°C)` = rep(NA_character_, 6),
    `Min Temp Flag` = rep(NA_character_, 6),
    `Mean Temp (°C)` = rep(NA_character_, 6)
  )
)

bad_df_file <- tempfile(fileext = ".rds")
curl::curl_download(
  "https://gist.github.com/paleolimbot/ec9b62b758ae57a5b4669fa771fc40a0/raw/e96b55f54d68b1cb3877bb358b28b99dc8836ceb/bad_df.rds",
  bad_df_file
)
result_bad <- readRDS(bad_df_file)[["result"]]
```

The only apparent difference is in the attributes of the tibble components,
i.e. the presence of `flag_info`.

``` r
str(result_ok)
#> List of 2
#>  $ :Classes 'tbl_df', 'tbl' and 'data.frame':    0 obs. of  4 variables:
#>   ..$ Max Temp Flag : chr(0) 
#>   ..$ Min Temp (°C) : chr(0) 
#>   ..$ Min Temp Flag : chr(0) 
#>   ..$ Mean Temp (°C): chr(0) 
#>  $ :Classes 'tbl_df', 'tbl' and 'data.frame':    6 obs. of  4 variables:
#>   ..$ Max Temp Flag : chr [1:6] NA NA NA NA ...
#>   ..$ Min Temp (°C) : chr [1:6] NA NA NA NA ...
#>   ..$ Min Temp Flag : chr [1:6] NA NA NA NA ...
#>   ..$ Mean Temp (°C): chr [1:6] NA NA NA NA ...
str(result_bad)
#> List of 2
#>  $ :Classes 'tbl_df', 'tbl' and 'data.frame':    0 obs. of  4 variables:
#>   ..$ Max Temp Flag : chr(0) 
#>   ..$ Min Temp (°C) : chr(0) 
#>   ..$ Min Temp Flag : chr(0) 
#>   ..$ Mean Temp (°C): chr(0) 
#>   ..- attr(*, "flag_info")=Classes 'tbl_df', 'tbl' and 'data.frame': 0 obs. of  2 variables:
#>   .. ..$ flag       : chr(0) 
#>   .. ..$ description: chr(0) 
#>  $ :Classes 'tbl_df', 'tbl' and 'data.frame':    6 obs. of  4 variables:
#>   ..$ Max Temp Flag : chr [1:6] NA NA NA NA ...
#>   ..$ Min Temp (°C) : chr [1:6] NA NA NA NA ...
#>   ..$ Min Temp Flag : chr [1:6] NA NA NA NA ...
#>   ..$ Mean Temp (°C): chr [1:6] NA NA NA NA ...
#>   ..- attr(*, "flag_info")=Classes 'tbl_df', 'tbl' and 'data.frame': 13 obs. of  2 variables:
#>   .. ..$ flag       : chr [1:13] "A" "C" "E" "F" ...
#>   .. ..$ description: chr [1:13] "Accumulated" "Precipitation occurred, amount uncertain" "Estimated" "Accumulated and estimated" ...
```

In particular, the sub-tibble names appear to be the same.

``` r
nms_ok <- lapply(result_ok, names)
nms_bad <- lapply(result_bad, names)
identical(nms_ok, nms_bad)
#> [1] TRUE
identical(nms_ok[[1]], nms_ok[[2]])
#> [1] TRUE
identical(nms_bad[[1]], nms_bad[[2]])
#> [1] TRUE
```

But we get a different result from `vec_rbind()`. The columns with a special
character in the name aren’t correctly “matched up” with `result_bad` and we
get two copies.

``` r
vctrs::vec_rbind(!!!result_ok)   # 4 variables --> correct
#> # A tibble: 6 x 4
#>   `Max Temp Flag` `Min Temp (°C)` `Min Temp Flag` `Mean Temp (°C)`
#>   <chr>           <chr>           <chr>           <chr>           
#> 1 <NA>            <NA>            <NA>            <NA>            
#> 2 <NA>            <NA>            <NA>            <NA>            
#> 3 <NA>            <NA>            <NA>            <NA>            
#> 4 <NA>            <NA>            <NA>            <NA>            
#> 5 <NA>            <NA>            <NA>            <NA>            
#> 6 <NA>            <NA>            <NA>            <NA>
vctrs::vec_rbind(!!!result_bad)  # 6 variables --> wrong
#> # A tibble: 6 x 6
#>   `Max Temp Flag` `Min Temp (°C)` `Min Temp Flag` `Mean Temp (°C)`
#>   <chr>           <chr>           <chr>           <chr>           
#> 1 <NA>            <NA>            <NA>            <NA>            
#> 2 <NA>            <NA>            <NA>            <NA>            
#> 3 <NA>            <NA>            <NA>            <NA>            
#> 4 <NA>            <NA>            <NA>            <NA>            
#> 5 <NA>            <NA>            <NA>            <NA>            
#> 6 <NA>            <NA>            <NA>            <NA>            
#> # … with 2 more variables: `Min Temp (°C)` <chr>, `Mean Temp (°C)` <chr>
```

Stripping the `flag_info` attribute doesn’t rescue this. Seems irrelevant.

``` r
result_stripped <- result_bad
attr(result_stripped[[1]], "flag_info") <- NULL
attr(result_stripped[[2]], "flag_info") <- NULL
vctrs::vec_rbind(!!!result_stripped)
#> # A tibble: 6 x 6
#>   `Max Temp Flag` `Min Temp (°C)` `Min Temp Flag` `Mean Temp (°C)`
#>   <chr>           <chr>           <chr>           <chr>           
#> 1 <NA>            <NA>            <NA>            <NA>            
#> 2 <NA>            <NA>            <NA>            <NA>            
#> 3 <NA>            <NA>            <NA>            <NA>            
#> 4 <NA>            <NA>            <NA>            <NA>            
#> 5 <NA>            <NA>            <NA>            <NA>            
#> 6 <NA>            <NA>            <NA>            <NA>            
#> # … with 2 more variables: `Min Temp (°C)` <chr>, `Mean Temp (°C)` <chr>
```

The problem goes away with less challenging names, even without removing the
`flag_info` attribute.

``` r
result_renamed <- result_bad
names(result_renamed[[1]]) <- letters[1:4]
names(result_renamed[[2]]) <- letters[1:4]
vctrs::vec_rbind(!!!result_renamed)
#> # A tibble: 6 x 4
#>   a     b     c     d    
#>   <chr> <chr> <chr> <chr>
#> 1 <NA>  <NA>  <NA>  <NA> 
#> 2 <NA>  <NA>  <NA>  <NA> 
#> 3 <NA>  <NA>  <NA>  <NA> 
#> 4 <NA>  <NA>  <NA>  <NA> 
#> 5 <NA>  <NA>  <NA>  <NA> 
#> 6 <NA>  <NA>  <NA>  <NA>
```

Directly assigning the exact same names fixes it.

``` r
result_renamed_direct <- result_bad
nms <- c("Max Temp Flag", "Min Temp (°C)", "Min Temp Flag", "Mean Temp (°C)")
names(result_renamed_direct[[1]]) <- nms
names(result_renamed_direct[[2]]) <- nms
vctrs::vec_rbind(!!!result_renamed_direct)
#> # A tibble: 6 x 4
#>   `Max Temp Flag` `Min Temp (°C)` `Min Temp Flag` `Mean Temp (°C)`
#>   <chr>           <chr>           <chr>           <chr>           
#> 1 <NA>            <NA>            <NA>            <NA>            
#> 2 <NA>            <NA>            <NA>            <NA>            
#> 3 <NA>            <NA>            <NA>            <NA>            
#> 4 <NA>            <NA>            <NA>            <NA>            
#> 5 <NA>            <NA>            <NA>            <NA>            
#> 6 <NA>            <NA>            <NA>            <NA>
```

Re-assigning the same names this way does not fix it.

``` r
result_renamed_reassign <- result_bad
names(result_renamed_reassign[[1]]) <- names(result_bad[[1]])
names(result_renamed_reassign[[2]]) <- names(result_bad[[2]])
vctrs::vec_rbind(!!!result_renamed_reassign)
#> # A tibble: 6 x 6
#>   `Max Temp Flag` `Min Temp (°C)` `Min Temp Flag` `Mean Temp (°C)`
#>   <chr>           <chr>           <chr>           <chr>           
#> 1 <NA>            <NA>            <NA>            <NA>            
#> 2 <NA>            <NA>            <NA>            <NA>            
#> 3 <NA>            <NA>            <NA>            <NA>            
#> 4 <NA>            <NA>            <NA>            <NA>            
#> 5 <NA>            <NA>            <NA>            <NA>            
#> 6 <NA>            <NA>            <NA>            <NA>            
#> # … with 2 more variables: `Min Temp (°C)` <chr>, `Mean Temp (°C)` <chr>
```

BUT … re-assigning the same names with one level of indirection DOES fix it.

``` r
result_renamed_indirect <- result_bad
nms <- names(result_bad[[1]])
names(result_renamed_indirect[[1]]) <- nms
names(result_renamed_indirect[[2]]) <- nms
vctrs::vec_rbind(!!!result_renamed_indirect)
#> # A tibble: 6 x 4
#>   `Max Temp Flag` `Min Temp (°C)` `Min Temp Flag` `Mean Temp (°C)`
#>   <chr>           <chr>           <chr>           <chr>           
#> 1 <NA>            <NA>            <NA>            <NA>            
#> 2 <NA>            <NA>            <NA>            <NA>            
#> 3 <NA>            <NA>            <NA>            <NA>            
#> 4 <NA>            <NA>            <NA>            <NA>            
#> 5 <NA>            <NA>            <NA>            <NA>            
#> 6 <NA>            <NA>            <NA>            <NA>
```

I can’t see any differences in these names with `rawToChar()`.

Putting some distinguishing data in makes it easier to see that the column
names aren’t being correctly “matched up”. Also indicates that the problem isn't due to the first tibble having zero rows.

``` r
result_augmented <- result_bad
result_augmented[[1]][1, ] <- rep("one", 4)
result_augmented[[2]][1, ] <- rep("two", 4)
vctrs::vec_rbind(!!!result_augmented)
#> # A tibble: 7 x 6
#>   `Max Temp Flag` `Min Temp (°C)` `Min Temp Flag` `Mean Temp (°C)`
#>   <chr>           <chr>           <chr>           <chr>           
#> 1 one             one             one             one             
#> 2 two             <NA>            two             <NA>            
#> 3 <NA>            <NA>            <NA>            <NA>            
#> 4 <NA>            <NA>            <NA>            <NA>            
#> 5 <NA>            <NA>            <NA>            <NA>            
#> 6 <NA>            <NA>            <NA>            <NA>            
#> 7 <NA>            <NA>            <NA>            <NA>            
#> # … with 2 more variables: `Min Temp (°C)` <chr>, `Mean Temp (°C)` <chr>
```

<sup>Created on 2019-08-30 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0.9000)</sup>

</details>

DavisVaughan:
Related to #553 

CC @jennybc @paleolimbot

I did some investigation into this, and have a solution, but would really love another pair of eyes on this.

3 things:

- `chr_equal_scalar()` now supports comparing strings with different encodings. This is a part of the dictionary hashing, and also fixes these problems with `vec_equal()`.
- For the hashing, I added `hash_char()` to translate to UTF-8 before hashing a `CHARSXP`.
- Apparently, `CHARSXP` objects have attributes related to the global string pool. This caused me quite the headache, as I was unaware. I special cased them in `hash_object()`, so we don't check their attributes. I can't think of a better solution.

Base R skipping over `CHARSXP` attributes:
https://github.com/wch/r-source/blob/5a156a0865362bb8381dcd69ac335f5174a4f60c/src/main/identical.c#L99

Rationale for the `!strcmp(x,y)` approach used in `chr_equal_scalar_impl()`
https://stackoverflow.com/questions/37712672/and-in-differ-based-on-character-encoding

Feel free to modify/reject/redo this PR. Just wanted to get my thoughts down after investigating.

``` r
library(purrr)
library(vctrs)

bad_df_file <- tempfile(fileext = ".rds")
curl::curl_download(
  "https://gist.github.com/paleolimbot/ec9b62b758ae57a5b4669fa771fc40a0/raw/e96b55f54d68b1cb3877bb358b28b99dc8836ceb/bad_df.rds",
  bad_df_file
)
result_bad <- readRDS(bad_df_file)[["result"]]

names <- result_bad %>% 
  map(names) %>% 
  unlist()

# before
vec_unique(names)
#> [1] "Max Temp Flag"  "Min Temp (°C)"  "Min Temp Flag"  "Mean Temp (°C)"
#> [5] "Min Temp (°C)"  "Mean Temp (°C)"

# after
vec_unique(names)
#> [1] "Max Temp Flag"  "Min Temp (°C)"  "Min Temp Flag"  "Mean Temp (°C)"
```

```r
names[c(2, 6)]
#> [1] "Min Temp (°C)" "Min Temp (°C)"

# before
vec_equal(names[2], names[6])
#> [1] FALSE

# after
vec_equal(names[2], names[6])
#> [1] TRUE
```

### DIFF
--- a/src/equal.c
+++ b/src/equal.c
@@ -138,31 +138,39 @@ static int cpl_equal_scalar(const Rcomplex* x, const Rcomplex* y, bool na_equal)
   }
 }
 
-// This follows the same rules as base R's Seql()
-// - Try pointer comparison, if equal, done.
-// - If encoding was the same, or one was UTF8 and the other was LATIN1, done
-//   as they are either different with the same encoding, or just incompatible.
-// - We are then left with the possibility of:
-//   - unknown / utf8
-//   - unknown / latin1
-//   - bytes / utf8 (error, can't translate bytes)
-//   - bytes / latin1 (error, can't translate bytes)
+// The minimal number of checks required to determine if we need to translate.
+// - CHAR_ENC_TYPE - Don't translate these cases:
+//   (utf8 + utf8), (latin1 + latin1), (unknown + unknown), (bytes + bytes),
+//   (unknown + bytes)
+// - CHAR_IS_BYTES - Don't translate these cases:
+//   (bytes + utf8), (bytes + latin1)
+// - Translate these:
+//   (utf8 + latin1), (unknown + utf8), (unknown + latin1)
+static bool scalar_requires_translation(const SEXP x, const SEXP y) {
+  if (CHAR_ENC_TYPE(x) == CHAR_ENC_TYPE(y)) {
+    return false;
+  }
+
+  if (CHAR_IS_BYTES(x) || CHAR_IS_BYTES(y)) {
+    return false;
+  }
+
+  return true;
+}
+
 static int chr_equal_scalar_impl(const SEXP x, const SEXP y) {
   if (x == y) {
     return 1;
   }
 
-  int x_ce = Rf_getCharCE(x);
-  int y_ce = Rf_getCharCE(y);
-
-  bool x_known_encoding = (x_ce == CE_UTF8) || (x_ce == CE_LATIN1);
-  bool y_known_encoding = (y_ce == CE_UTF8) || (y_ce == CE_LATIN1);
-
-  if (x_known_encoding == y_known_encoding) {
-    return 0;
+  if (scalar_requires_translation(x, y)) {
+    const void *vmax = vmaxget();
+    int out = !strcmp(Rf_translateCharUTF8(x), Rf_translateCharUTF8(y));
+    vmaxset(vmax);
+    return out;
   }
 
-  return !strcmp(Rf_translateCharUTF8(x), Rf_translateCharUTF8(y));
+  return 0;
 }
 
 static int chr_equal_scalar(const SEXP* x, const SEXP* y, bool na_equal) {
--- a/src/equal.c
+++ b/src/equal.c
@@ -139,7 +139,11 @@ static int cpl_equal_scalar(const Rcomplex* x, const Rcomplex* y, bool na_equal)
 }
 
 static int chr_equal_scalar_impl(const SEXP x, const SEXP y) {
-  return !strcmp(Rf_translateCharUTF8(x), Rf_translateCharUTF8(y));
+  if (Rf_getCharCE(x) == Rf_getCharCE(y)) {
+    return x == y;
+  } else {
+    return !strcmp(Rf_translateCharUTF8(x), Rf_translateCharUTF8(y));
+  }
 }
 
 static int chr_equal_scalar(const SEXP* x, const SEXP* y, bool na_equal) {
--- a/src/equal.c
+++ b/src/equal.c
@@ -137,14 +137,18 @@ static int cpl_equal_scalar(const Rcomplex* x, const Rcomplex* y, bool na_equal)
     return real_equal && imag_equal;
   }
 }
+
+static int chr_equal_scalar_impl(const SEXP x, const SEXP y) {
+  return !strcmp(Rf_translateCharUTF8(x), Rf_translateCharUTF8(y));
+}
+
 static int chr_equal_scalar(const SEXP* x, const SEXP* y, bool na_equal) {
   const SEXP xi = *x;
   const SEXP yj = *y;
   if (na_equal) {
-    // Ignoring encoding for now
-    return xi == yj;
+    return chr_equal_scalar_impl(xi, yj);
   } else {
-    return (xi == NA_STRING || yj == NA_STRING) ? NA_LOGICAL : xi == yj;
+    return (xi == NA_STRING || yj == NA_STRING) ? NA_LOGICAL : chr_equal_scalar_impl(xi, yj);
   }
 }
 
--- a/src/equal.c
+++ b/src/equal.c
@@ -138,17 +138,28 @@ static int cpl_equal_scalar(const Rcomplex* x, const Rcomplex* y, bool na_equal)
   }
 }
 
-// TODO - doesn't work when comparing a string with "bytes" encoding
-// to another string with a different encoding. `Rf_translateCharUTF8()`
-// fails to convert with a nice error.
+// This follows the same rules as base R's Seql()
+// - Try pointer comparison, if equal, done.
+// - If encoding was the same, or one was UTF8 and the other was LATIN1, done
+//   as they are either different with the same encoding, or just incompatible.
+// - We are then left with the possibility of:
+//   - unknown / utf8
+//   - unknown / latin1
+//   - bytes / utf8 (error, can't translate bytes)
+//   - bytes / latin1 (error, can't translate bytes)
 static int chr_equal_scalar_impl(const SEXP x, const SEXP y) {
-  // String pointers are the same. Always means equivalent.
+  // String pointers are the same. Always equivalent.
   if (x == y) {
     return 1;
   }
 
-  // String pointers were different, but encoding is the same
-  if (Rf_getCharCE(x) == Rf_getCharCE(y)) {
+  int x_ce = Rf_getCharCE(x);
+  int y_ce = Rf_getCharCE(y);
+
+  bool x_known_encoding = (x_ce == CE_UTF8) || (x_ce == CE_LATIN1);
+  bool y_known_encoding = (y_ce == CE_UTF8) || (y_ce == CE_LATIN1);
+
+  if (x_known_encoding == y_known_encoding) {
     return 0;
   }
 
--- a/src/equal.c
+++ b/src/equal.c
@@ -138,12 +138,21 @@ static int cpl_equal_scalar(const Rcomplex* x, const Rcomplex* y, bool na_equal)
   }
 }
 
+// TODO - doesn't work when comparing a string with "bytes" encoding
+// to another string with a different encoding. `Rf_translateCharUTF8()`
+// fails to convert with a nice error.
 static int chr_equal_scalar_impl(const SEXP x, const SEXP y) {
-  if (Rf_getCharCE(x) == Rf_getCharCE(y)) {
-    return x == y;
-  } else {
-    return !strcmp(Rf_translateCharUTF8(x), Rf_translateCharUTF8(y));
+  // String pointers are the same. Always means equivalent.
+  if (x == y) {
+    return 1;
   }
+
+  // String pointers were different, but encoding is the same
+  if (Rf_getCharCE(x) == Rf_getCharCE(y)) {
+    return 0;
+  }
+
+  return !strcmp(Rf_translateCharUTF8(x), Rf_translateCharUTF8(y));
 }
 
 static int chr_equal_scalar(const SEXP* x, const SEXP* y, bool na_equal) {
--- a/src/equal.c
+++ b/src/equal.c
@@ -148,7 +148,6 @@ static int cpl_equal_scalar(const Rcomplex* x, const Rcomplex* y, bool na_equal)
 //   - bytes / utf8 (error, can't translate bytes)
 //   - bytes / latin1 (error, can't translate bytes)
 static int chr_equal_scalar_impl(const SEXP x, const SEXP y) {
-  // String pointers are the same. Always equivalent.
   if (x == y) {
     return 1;
   }
--- a/src/equal.c
+++ b/src/equal.c
@@ -138,32 +138,19 @@ static int cpl_equal_scalar(const Rcomplex* x, const Rcomplex* y, bool na_equal)
   }
 }
 
-// The minimal number of checks required to determine if we need to translate.
-// - CHAR_ENC_TYPE - Don't translate these cases:
-//   (utf8 + utf8), (latin1 + latin1), (unknown + unknown), (bytes + bytes),
-//   (unknown + bytes)
-// - CHAR_IS_BYTES - Don't translate these cases:
-//   (bytes + utf8), (bytes + latin1)
-// - Translate these:
-//   (utf8 + latin1), (unknown + utf8), (unknown + latin1)
-static bool scalar_requires_translation(const SEXP x, const SEXP y) {
-  if (CHAR_ENC_TYPE(x) == CHAR_ENC_TYPE(y)) {
-    return false;
-  }
-
-  if (CHAR_IS_BYTES(x) || CHAR_IS_BYTES(y)) {
-    return false;
-  }
-
-  return true;
-}
+// UTF-8 translation is successful in these cases:
+// - (utf8 + latin1), (unknown + utf8), (unknown + latin1)
+// UTF-8 translation fails purposefully in these cases:
+// - (bytes + utf8), (bytes + latin1), (bytes + unknown)
+// UTF-8 translation is not attempted in these cases:
+// - (utf8 + utf8), (latin1 + latin1), (unknown + unknown), (bytes + bytes)
 
 static int chr_equal_scalar_impl(const SEXP x, const SEXP y) {
   if (x == y) {
     return 1;
   }
 
-  if (scalar_requires_translation(x, y)) {
+  if (CHAR_ENC_TYPE(x) != CHAR_ENC_TYPE(y)) {
     const void *vmax = vmaxget();
     int out = !strcmp(Rf_translateCharUTF8(x), Rf_translateCharUTF8(y));
     vmaxset(vmax);
--- a/src/hash.c
+++ b/src/hash.c
@@ -153,8 +153,8 @@ static uint32_t sexp_hash(SEXP x);
 uint32_t hash_object(SEXP x) {
   uint32_t hash = sexp_hash(x);
 
-  // CHARSXP object have attributes that are purely related to the cache
-  // https://github.com/wch/r-source/blob/5a156a0865362bb8381dcd69ac335f5174a4f60c/src/main/identical.c#L99
+  // Skip attribute check on CHARSXP, which might have attribute related
+  // to the global string pool (#555)
   if (TYPEOF(x) == CHARSXP) {
     return hash;
   }
--- a/src/hash.c
+++ b/src/hash.c
@@ -42,7 +42,57 @@ static uint32_t hash_double(double x) {
 }
 
 static uint32_t hash_char(SEXP x) {
-  return hash_int64((int64_t) *Rf_translateCharUTF8(x));
+  return hash_int64((int64_t) x);
+}
+
+static uint32_t hash_char_translate(SEXP x) {
+  const void *vmax = vmaxget();
+
+  const char* x_utf8 = CHAR_IS_UTF8(x) ? CHAR(x) : Rf_translateCharUTF8(x);
+
+  uint32_t hash = 0;
+
+  while(*x_utf8++) {
+    hash = hash_combine(hash, hash_int64((int64_t) *x_utf8));
+  }
+
+  vmaxset(vmax);
+  return hash;
+}
+
+// Determine if `x` must be translated to UTF-8
+// - CHAR_IS_BYTES - Don't translate these cases:
+//   (bytes + bytes), (bytes + unknown), (bytes + utf8), (bytes + latin1)
+// - CHAR_ENC_TYPE - Don't translate these cases:
+//   (utf8 + utf8), (latin1 + latin1), (unknown + unknown)
+// - Translate these:
+//   (utf8 + latin1), (unknown + utf8), (unknown + latin1)
+static bool chr_requires_translation(SEXP x, R_len_t size) {
+  const SEXP* xp = STRING_PTR_RO(x);
+  SEXP xp_val;
+
+  bool translate = false;
+  int first_encoding = CHAR_ENC_TYPE(*xp);
+
+  for (R_len_t i = 0; i < size; ++i, ++xp) {
+    xp_val = *xp;
+
+    if (CHAR_IS_BYTES(xp_val)) {
+      return false;
+    }
+
+    if (translate) {
+      continue;
+    }
+
+    if (CHAR_ENC_TYPE(xp_val) == first_encoding) {
+      continue;
+    }
+
+    translate = true;
+  }
+
+  return translate;
 }
 
 // Hashing scalars -----------------------------------------------------
@@ -52,36 +102,10 @@ static uint32_t int_hash_scalar(const int* x);
 static uint32_t dbl_hash_scalar(const double* x);
 static uint32_t cpl_hash_scalar(const Rcomplex* x);
 static uint32_t chr_hash_scalar(const SEXP* x);
+static uint32_t chr_hash_scalar_translate(const SEXP* x);
 static uint32_t raw_hash_scalar(const Rbyte* x);
-static uint32_t df_hash_scalar(SEXP x, R_len_t i);
 static uint32_t list_hash_scalar(SEXP x, R_len_t i);
-static uint32_t shaped_hash_scalar(SEXP x, R_len_t i);
 
-
-// [[ include("vctrs.h") ]]
-uint32_t hash_scalar(SEXP x, R_len_t i) {
-  if (has_dim(x)) {
-    return shaped_hash_scalar(x, i);
-  }
-
-  switch(TYPEOF(x)) {
-  case LGLSXP: return lgl_hash_scalar(LOGICAL(x) + i);
-  case INTSXP: return int_hash_scalar(INTEGER(x) + i);
-  case REALSXP: return dbl_hash_scalar(REAL(x) + i);
-  case CPLXSXP: return cpl_hash_scalar(COMPLEX(x) + i);
-  case STRSXP: return chr_hash_scalar(STRING_PTR(x) + i);
-  case RAWSXP: return raw_hash_scalar(RAW(x) + i);
-  case VECSXP: {
-    if (is_data_frame(x)) {
-      return df_hash_scalar(x, i);
-    } else {
-      return list_hash_scalar(x, i);
-    }
-  }
-  default:
-    Rf_errorcall(R_NilValue, "Unsupported type %s", Rf_type2char(TYPEOF(x)));
-  }
-}
 
 static uint32_t lgl_hash_scalar(const int* x) {
   return hash_int32(*x);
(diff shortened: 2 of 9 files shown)


DavisVaughan:
@lionel- why do you use `intptr_t` over `int64_t` here?
https://github.com/r-lib/vctrs/blob/f5fe6e0913ec17a5079cc6ceaca98e0a37e90e5e/src/hash.c#L189

Should I be using `intptr_t` here?
https://github.com/r-lib/vctrs/pull/555/files#diff-292b11162bf53ed7cca27b794e4b092fR45

jennybc:
Does vctrs already have a militant UTF-8 everywhere policy or anything like that?

DavisVaughan:
Not that I am aware of!

Possibly also related:

lionel-:
@DavisVaughan This is from https://github.com/r-lib/vctrs/commit/aadeac260d8f125aa0c3c61e6f38f665e442cfe2

I'm not sure what's best here but I guess coercing pointers to a 64 bit type should work almost everywhere. I don't think going through `intptr_t` matters though.

src/equal.c
```diff
@@ -137,14 +137,18 @@ static int cpl_equal_scalar(const Rcomplex* x, const Rcomplex* y, bool na_equal)
     return real_equal && imag_equal;
   }
 }
+
+static int chr_equal_scalar_impl(const SEXP x, const SEXP y) {
+  return !strcmp(Rf_translateCharUTF8(x), Rf_translateCharUTF8(y));
```
lionel-: Maybe do a fast pointer comparison path if `getCharCE()` is the same for both inputs?

hadley:
Would you mind also doing a quick benchmark in order to assess the performance cost?

DavisVaughan:
Bad news is that the hashing got substantially slower, so `vec_unique()` is slow. Really it would be nice to know ahead of time whether a "translation" step is required or not. That would substantially speed things up, and is essentially what base R does. Here, if `useUTF8 = FALSE`, they don't convert.

https://github.com/wch/r-source/blob/2480f4d76dbdce30db615ee8c2cd43d655cb1241/src/main/unique.c#L178

Here is where they determine if that translation is required or not:
https://github.com/wch/r-source/blob/2480f4d76dbdce30db615ee8c2cd43d655cb1241/src/main/unique.c#L493

I'm not entirely sure how to refactor the code to allow for this, as right now its pretty hard to pass extra information on in a way that would feed it down to `hash_char()`.

``` r
library(vctrs)

# 10000 random words of size 10, append with utf-8 character
utf8 <- stringi::stri_rand_strings(10000, 10)
utf8 <- paste0(utf8, "\u00B0")

unknown <- utf8
```

```r
# before
bench::mark(vec_unique(utf8), unique(utf8))
#> # A tibble: 2 x 6
#> 1 vec_unique(utf8)    327µs    517µs     2014.     366KB     10.0
#> 2 unique(utf8)        263µs    354µs     2788.     245KB     14.6

# after
bench::mark(vec_unique(utf8), unique(utf8))
#> # A tibble: 2 x 6
#> 1 vec_unique(utf8)   12.8ms   13.7ms      72.4     366KB      0  
#> 2 unique(utf8)      272.9µs  345.9µs    2939.      245KB     12.2
```

```r
# before
bench::mark(vec_unique(unknown), unique(unknown))
#> # A tibble: 2 x 6
#> 1 vec_unique(unknown)    410µs    504µs     2027.     348KB     12.4
#> 2 unique(unknown)        124µs    149µs     6529.     245KB     28.9

# after
bench::mark(vec_unique(unknown), unique(unknown))
#> # A tibble: 2 x 6
#> 1 vec_unique(unknown)   15.3ms   15.8ms      62.4     348KB     2.08
#> 2 unique(unknown)      136.1µs  183.4µs    5420.      245KB    24.7
```

```r
x <- c(utf8, unknown)

# before
# Error!

# after
bench::mark(vec_unique(x), unique(x))
#> # A tibble: 2 x 6
#> 1 vec_unique(x) 282.68ms  282.7ms      3.54     451KB     3.54
#> 2 unique(x)       5.56ms    6.2ms    158.       412KB     4.11
```

<sup>Created on 2019-08-30 by the [reprex package](https://reprex.tidyverse.org) (v0.2.1)</sup>

On the other hand `vec_equal()` is fairly close to `==` in speed, but they have access to a way to compare encodings that we don't have access to (`ENC_KNOWN()`), which does make it slower when the string pointers aren't equal (a common case). There are also some special cases with encoding that should be documented somewhere. This got a bit complicated, unfortunately, but essentially comparing a bytes encoded string + latin1/utf8 string is an error. We might be able to tweak the logic somehow to avoid this, I can't tell if it is a bug in base R or not.

<details>

``` r
library(vctrs)

# 10000 random words of size 10, append with utf-8 character
utf8 <- stringi::stri_rand_strings(10000, 10)
utf8 <- paste0(utf8, "\u00B0")

unknown <- utf8

bytes <- unknown
Encoding(bytes) <- "bytes"

# Encodings are the same, and pointer comparison is always TRUE because they
# are the same
bench::mark(
  vec_equal(utf8, utf8), 
  utf8 == utf8
)
#> # A tibble: 2 x 6
#> 1 vec_equal(utf8, utf8)   21.6µs   33.9µs    27780.    99.3KB     17.9
#> 2 utf8 == utf8            27.8µs   32.6µs    28920.    39.1KB     20.3

# Encodings are the same, and pointer comparison is always FALSE because they
# are all different, requiring an additional encoding check
no_match_utf8 <- rev(utf8)
bench::mark(
  vec_equal(utf8, no_match_utf8), 
  utf8 == no_match_utf8
)
#> # A tibble: 2 x 6
#>   <bch:expr>                     <bch:> <bch:>     <dbl> <bch:byt>    <dbl>
#> 1 vec_equal(utf8, no_match_utf8) 51.1µs 61.4µs    16019.    39.1KB     9.94
#> 2 utf8 == no_match_utf8          35.7µs 38.8µs    24002.    39.1KB    16.4

# Encodings are different here
# - Pointer comparison will be false
# - Requires the encoding check, which returns FALSE
# - Then we do the expensive translate check
bench::mark(
  vec_equal(unknown, utf8), 
  unknown == utf8
)
#> # A tibble: 2 x 6
#> 1 vec_equal(unknown, utf8)   2.52ms   2.68ms      366.    39.1KB     6.27
#> 2 unknown == utf8            2.44ms   2.57ms      388.    39.1KB     6.18

# bytes + unknown is allowed
bench::mark(
  vec_equal(bytes, unknown),
  bytes == unknown
)
#> # A tibble: 2 x 6
#> 1 vec_equal(bytes, unknown)   57.6µs   62.6µs    15213.    39.1KB     9.81
#> 2 bytes == unknown            38.6µs   44.1µs    21602.    39.1KB    16.8

# But base R throws an error for bytes+utf8/latin1
# - pointer comparison was FALSE
# - the encoding check returns FALSE
# - it tries to translate to utf8, but can't convert bytes
vec_equal(bytes, utf8) # or latin1
#> Error in vec_equal(bytes, utf8): translating strings with "bytes" encoding is not allowed
bytes == utf8
#> Error in eval(expr, envir, enclos): translating strings with "bytes" encoding is not allowed
```

</details>

DavisVaughan:
I've come up with a clean new way to do this, speeds look good. I'll finalize it all Monday probably. Don't spend any time thinking about this

DavisVaughan:
Big picture changes:

- Unused `hash_scalar()` was removed. This was the old rowwise version. The removal was required to cleanly implement the logic for switching between a character hasher that does / does not translate to UTF-8. 
   - This also meant `df_hash_scalar()` and `shaped_hash_scalar()` could be removed.
   - `vctrs_hash()` no longer has a `rowwise` argument because of this.
   - Tests were removed / updated if they used `rowwise = TRUE`.

- `chr_hash_scalar()` no longer calls `hash_object()`, but instead drops directly into either `hash_char()` or `hash_char_translate()`. This actually had some substantial performance benefits.

- Hashing a character vector with either `chr_hash()` or `chr_hash_fill()` now checks if a translation to UTF-8 is required, and calls the appropriate helper depending on the result.

- `vec_equal()` has also been updated to handle UTF-8 translations.
   - It first checks if the pointers are equal. If so, we are done. This is very fast.
   - If pointers aren't equal, and we don't need a translation, then we are done.
   - If pointers aren't equal, and we do need a translation, then translate and check with `strcmp`.
   - The order of operations here was inspired by base R, it was pretty hard to beat it with any other scheme. The idea is to avoid a translation as much as possible since its expensive.

- You'll see `vmaxget()` and `vmaxset()` being used. This is due to `Rf_translateCharUTF8()` which requires manual handling of the memory.

src/equal.c
```diff
@@ -137,14 +137,49 @@ static int cpl_equal_scalar(const Rcomplex* x, const Rcomplex* y, bool na_equal)
     return real_equal && imag_equal;
   }
 }
+
+// The minimal number of checks required to determine if we need to translate.
+// - CHAR_ENC_TYPE - Don't translate these cases:
+//   (utf8 + utf8), (latin1 + latin1), (unknown + unknown), (bytes + bytes),
+//   (unknown + bytes)
+// - CHAR_IS_BYTES - Don't translate these cases:
+//   (bytes + utf8), (bytes + latin1)
```
hadley: (bytes + unknown) ?

src/equal.c
```diff
@@ -137,14 +137,49 @@ static int cpl_equal_scalar(const Rcomplex* x, const Rcomplex* y, bool na_equal)
...
+}
+
+static int chr_equal_scalar_impl(const SEXP x, const SEXP y) {
+  if (x == y) {
+    return 1;
+  }
+
+  if (scalar_requires_translation(x, y)) {
+    const void *vmax = vmaxget();
+    int out = !strcmp(Rf_translateCharUTF8(x), Rf_translateCharUTF8(y));
```
hadley: I assume `Rf_translateCharUTF8()` is a no-op if the CHARSXP is already UTF-8?

src/hash.c
```diff
@@ -41,6 +41,59 @@ static uint32_t hash_double(double x) {
...
+  const SEXP* xp = STRING_PTR_RO(x);
+  SEXP xp_val;
+
+  bool translate = false;
+  int first_encoding = CHAR_ENC_TYPE(*xp);
+
+  for (R_len_t i = 0; i < size; ++i, ++xp) {
+    xp_val = *xp;
+
+    if (CHAR_IS_BYTES(xp_val)) {
```
hadley: So if any element of the vector has bytes encoding, we don't translate?

DavisVaughan:
`vec_equal()` benchmarks

Setup:

```r
library(vctrs)
vec_equal_old <- vctrs2:::vec_equal

# 500000 random words of size 10, append with utf-8 character
utf8 <- stringi::stri_rand_strings(500000, 10)
utf8 <- paste0(utf8, "\u00B0")

rev_utf8 <- rev(utf8)

unknown <- utf8

bytes <- utf8
Encoding(bytes) <- "bytes"

latin1 <- iconv(utf8, "UTF-8", "latin1")
```

Equal strings - very fast:

```r
bench::mark(
  old = vec_equal_old(utf8, utf8),
  new = vec_equal(utf8, utf8),
  base = utf8 == utf8,
  iterations = 500
)
#> # A tibble: 3 x 6
#> 1 old        778.98µs 936.56µs      950.    1.96MB     56.4
#> 2 new          1.27ms   1.55ms      624.    1.96MB     24.6
#> 3 base         1.61ms   2.07ms      495.    1.91MB     18.5
```


All different - but same encoding. Requires an expensive-ish check to see if the encoding is the same. This is slightly slower than base R because we also check `na_equal` at each iteration, but without more refactoring I don't see any benefit to optimizing this.

```r
bench::mark(
  old = vec_equal_old(utf8, rev_utf8),
  new = vec_equal(utf8, rev_utf8),
  base = utf8 == rev_utf8,
  iterations = 500
)
#> # A tibble: 3 x 6
#> 1 old        824.69µs 963.05µs     1021.    1.91MB    47.0 
#> 2 new          5.01ms   5.29ms      188.    1.91MB     8.22
#> 3 base         4.38ms   4.63ms      214.    1.91MB     8.91
```

All equal - but a translation is required to make it so.

```r
bench::mark(
  # wrong! old = vec_equal_old(utf8, unknown),
  new = vec_equal(utf8, unknown),
  base = utf8 == unknown,
  iterations = 50
)
#> # A tibble: 2 x 6
#> 1 new           136ms    143ms      6.99    1.91MB     2.46
#> 2 base          132ms    137ms      7.26    1.91MB     3.11
```

Equivalent latin1/utf8

```r
vec_equal(latin1[1], utf8[1])
#> [1] TRUE

latin1[1] == utf8[1]
#> [1] TRUE
```

Base R error with bytes, but I just never attempt a translation if the strings are bytes.

```r
vec_equal(bytes[1], utf8[1])
#> [1] FALSE

bytes[1] == utf8[1]
#> Error in eval(expr, envir, enclos): translating strings with "bytes" encoding is not allowed
```

src/hash.c
DavisVaughan: Correct, that's an error with base R (b/c its an error with `Rf_translateCharUTF8()`), see the last example with `vec_equal` below

hadley:
Benchmarks look good; I don't have any concerns from a performance perspective. If we decide performance is a problem in the future, I think the way forward would be to use @krlmlr's approach and create a utf8 vector class that enforces a utf8 encoding.

src/hash.c
hadley: I think the only disadvantage is that it means that you can't short-circuit and return `true` as soon as you see an encoding that's different to the first.

src/equal.c
DavisVaughan: It's a bit tricky, but that case is captured by `CHR_ENC_TYPE()` already. See right above I listed `(unknown + bytes)`. Maybe the name is a bit misleading but it returns either:
- `0` if unknown / bytes
- `4` if latin1
- `8` if utf8

src/equal.c
```diff
@@ -137,14 +137,49 @@ static int cpl_equal_scalar(const Rcomplex* x, const Rcomplex* y, bool na_equal)
...
 }
+
+// The minimal number of checks required to determine if we need to translate.
+// - CHAR_ENC_TYPE - Don't translate these cases:
+//   (utf8 + utf8), (latin1 + latin1), (unknown + unknown), (bytes + bytes),
+//   (unknown + bytes)
+// - CHAR_IS_BYTES - Don't translate these cases:
+//   (bytes + utf8), (bytes + latin1)
+// - Translate these:
+//   (utf8 + latin1), (unknown + utf8), (unknown + latin1)
```
hadley: Do we need to translate unknown + utf8 and unknown + latin1? Doesn't unknown imply ASCII?

src/equal.c
DavisVaughan: Correct

src/hash.c
DavisVaughan: That's true. Luckily if you do need to short circuit and return `true`, then that means you need to alter encodings, which takes _much_ longer than this loop, even if it has to cycle through all the values.

src/equal.c
DavisVaughan: I think unknown+utf8 was the original error right? https://github.com/r-lib/vctrs/issues/553#issuecomment-526650247

hadley: Ah, I think I always get confused about how this works — you can't change the encoding of an ASCII string (it's always unknown), but you can give a UTF-8 string unknown encoding.

DavisVaughan:
For completeness, `vec_unique()` benchmarks and special cases:

Setup:

```r
library(vctrs)
vec_unique_old <- vctrs2:::vec_unique

# 500000 random words of size 10, append with utf-8 character
utf8 <- stringi::stri_rand_strings(500000, 10)
utf8 <- paste0(utf8, "\u00B0")

unknown <- utf8

bytes <- utf8
Encoding(bytes) <- "bytes"

latin1 <- iconv(utf8, "UTF-8", "latin1")
```

All values are unique, but same encoding

```r
bench::mark(
  old = vec_unique_old(utf8),
  new = vec_unique(utf8),
  base = unique(utf8),
  iterations = 20
)
#> # A tibble: 3 x 6
#> 1 old          31.1ms   37.9ms      27.2   15.64MB    22.2 
#> 2 new          37.1ms     46ms      21.6   15.65MB    11.6 
#> 3 base         44.5ms   62.2ms      16.9    9.72MB     7.23
```

All values are the same, with same encoding

```r
utf8_repeat <- rep(utf8[1], 500000)

bench::mark(
  old = vec_unique_old(utf8_repeat),
  new = vec_unique(utf8_repeat),
  base = unique(utf8_repeat),
  iterations = 100
)
#> # A tibble: 3 x 6
#> 1 old           7.8ms   9.24ms     109.     5.91MB     19.2
#> 2 new          8.25ms   9.92ms     102.     5.91MB     13.9
#> 3 base         8.75ms  11.98ms      84.0    5.91MB     11.5
```

Notable case - All values are the same, all latin1 encoding. Big improvement over `unique()`, which tries to convert to UTF-8 here

```r
latin1_repeat <- iconv(utf8_repeat, "UTF-8", "latin1")

bench::mark(
  old = vec_unique_old(latin1_repeat),
  new = vec_unique(latin1_repeat),
  base = unique(latin1_repeat),
  iterations = 20
)
#> # A tibble: 3 x 6
#> 1 old          8.07ms   9.32ms    107.      5.91MB    18.9 
#> 2 new          9.31ms  10.32ms     96.8     5.91MB    24.2 
#> 3 base       141.92ms 149.58ms      6.67    5.91MB     3.59
```

Mixed case - UTF8 and unknown encodings together. First convert everything to UTF-8.

```r
utf8_unknown <- c(utf8, unknown)

length(utf8_unknown)
#> [1] 1000000
length(unique(utf8_unknown))
#> [1] 500000

# Mixed utf8 / unknown - translation required
bench::mark(
  # wrong! old = vec_unique_old(utf8_unknown),
  new = vec_unique(utf8_unknown),
  base = unique(utf8_unknown),
  iterations = 20
)
#> # A tibble: 2 x 6
#> 1 new           447ms    450ms      2.22    21.5MB     9.44
#> 2 base          445ms    457ms      2.19    15.6MB     4.08
```

Can determine uniqueness among unknown / latin1 / utf8 encodings that all represent the same string.

```r
vec_unique(c(unknown[1], latin1[1], utf8[1]))
#> [1] "MkP6VHYFlx°"

unique(c(unknown[1], latin1[1], utf8[1]))
#> [1] "MkP6VHYFlx°"
```

_Special case_ - If any bytes string is detected, we never translate so the unknown/latin1/utf8 strings look different.

```r
vec_unique(c(bytes[1], unknown[1], latin1[1], utf8[1]))
#> [1] "MkP6VHYFlx\\xc2\\xb0" "MkP6VHYFlx°"        "MkP6VHYFlx°"       
#> [4] "MkP6VHYFlx°"

unique(c(bytes[1], unknown[1], latin1[1], utf8[1]))
#> [1] "MkP6VHYFlx\\xc2\\xb0" "MkP6VHYFlx°"        "MkP6VHYFlx°"       
#> [4] "MkP6VHYFlx°"
```

DavisVaughan:
Okay, I've decided to punt on special handling of bytes strings for now. They are too rare and require too much special handling for the benefit we'd get. 

Now we always error if a bytes string is mixed with another encoding, which forces a translation through `Rf_translateCharUTF8()`, which is where the error is thrown from.

I've since learned that while this PR is required to fix `vec_equal()` and `vec_unique()`, the real culprit is `vec_match()` (and also `vec_in()` will need a fix). I think those are complex enough to be another PR though, so I'm ready to merge this one.

src/hash.c
```diff
@@ -41,6 +41,44 @@ static uint32_t hash_double(double x) {
...
+}
+
+// Determine if `x` must be translated to UTF-8
+// UTF-8 translation will be successful in these cases:
+// - (utf8 + latin1), (unknown + utf8), (unknown + latin1)
+// UTF-8 translation will fail purposefully in these cases:
+// - (bytes + utf8), (bytes + latin1), (bytes + unknown)
+// UTF-8 translation is not attempted in these cases:
+// - (utf8 + utf8), (latin1 + latin1), (unknown + unknown), (bytes + bytes)
+static bool chr_requires_translation(SEXP x, R_len_t size) {
```
lionel-: 👍

PR #557, opened 2019-09-01
Stricter checking of index for vec_as_index()

krlmlr:
For tibble.

Closes https://github.com/tidyverse/tibble/issues/649.

### DIFF
--- a/src/slice.c
+++ b/src/slice.c
@@ -602,6 +602,11 @@ SEXP vec_as_index(SEXP i, R_len_t n, SEXP names) {
 
 SEXP vec_as_index_opts(SEXP i, R_len_t n, SEXP names,
                        struct vec_as_index_options* opts) {
+
+  if (vec_dim_n(i) != 1) {
+    Rf_errorcall(R_NilValue, "`i` must have one dimension, not %d.", vec_dim_n(i));
+  }
+
   switch (TYPEOF(i)) {
   case NILSXP: return vctrs_shared_empty_int;
   case INTSXP: return int_as_index(i, n, opts);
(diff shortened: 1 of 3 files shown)


tests/testthat/test-slice.R
```diff
@@ -1,9 +1,12 @@
-context("test-slice")
```
lionel-: Why this change?

src/slice.c
```diff
@@ -559,6 +559,14 @@ static SEXP chr_as_index(SEXP i, SEXP names) {
 }
 
 SEXP vec_as_index(SEXP i, R_len_t n, SEXP names) {
+  if (OBJECT(i)) {
+    Rf_errorcall(R_NilValue, "`i` must be an integer, character, or logical vector, not an object.");
```
lionel-: Probably better to use `vec_cast()` to integer? Or `vec_coerce_cast()`, so that only coercible types are allowed (e.g. not `c("1", "2")`).

src/slice.c
```diff
@@ -559,6 +559,14 @@ static SEXP chr_as_index(SEXP i, SEXP names) {
...
+  }
+
+  if (vec_dim_n(i) != 1) {
```
lionel-: 👍

tests/testthat/test-slice.R
krlmlr: `context()` is no longer needed in recent testthat.

lionel-: Looks like context are now generated from the file names, cool.

lionel-:
Thanks!

PR #565, opened 2019-09-05
Add encoding support for `vec_match()` and `vec_in()`

### LINKED ISSUE #553: vec_unique() is sensitive to the marked encoding, where base::unique() is not

DavisVaughan:
Closes #553 

Big picture ideas, I'll add some benchmarks in a bit:

- New `translate.c` file holds a number of translation utility functions.

- `vec_match()` and `vec_in()` are a bit tough to handle because they create 2 hashed dictionaries (one for `needles` and one for `haystack`) separate of each other. This is a problem if we have `vec_match(needles = <utf8>, haystack = <unknown>)` since `haystack` won't be translated to utf8. My solution is to have a `translate_encoding2()` that is run before the dictionary hashing, which translates both inputs to UTF8 if either requires it.

- `translate_encoding2()` is simple when the inputs are character vectors. It uses `translation_required_chr2()` to determine if either input needs translation, and if either does it translates them.

But we obviously want to be more general, since `vec_match()` should work with data frames / lists too. That's where the rest of the code comes in.

- In the case of data frame inputs, `translate_encoding2()` takes advantage of the fact that we ran `vec_cast()` to get `needles` and `haystack` to a common type beforehand. It loops over the columns (which are guaranteed to be in the same order and have the same type) and calls `translate_encoding2()` recursively.

- The case of having list inputs is the toughest. `vec_cast()` guarantees that `needles` and `haystack` are both lists, but there is no guarantee that if the first element of `needles` is a character vector, then the first element of `haystack` is too. I think the best solution is to perform an initial loop through `needles` and `haystack`, determine if any element of either is a character vector with a "known" encoding (utf8, latin1, bytes), and if that is true then we translate _every_ character vector in both lists to UTF-8. 

While this case with lists sounds expensive, I think this is better than the base R strategy. When you give `match()` lists or data frames, [it literally just runs `as.character()` on the inputs](https://github.com/wch/r-source/blob/2480f4d76dbdce30db615ee8c2cd43d655cb1241/src/main/unique.c#L906), which has some really odd looking intermediate results and can be pretty expensive. It seems to work for them though. There is even a note in `?match` of "Matching for lists is potentially very slow and best avoided except in simple cases."

### DIFF
--- a/src/vctrs.h
+++ b/src/vctrs.h
@@ -282,16 +282,6 @@ SEXP vec_as_unique_names(SEXP names, bool quiet);
 
 // Character translation ----------------------------------------
 
-// `sxpinfo.gp` is exposed through `LEVELS()`, and CHARSXP info is
-// extractable through a bitwise and. This copies the unexposed
-// interface in Defn.h
-#define BYTES_MASK 2
-#define LATIN1_MASK 4
-#define UTF8_MASK 8
-
-#define CHAR_IS_UTF8(x) (LEVELS(x) & UTF8_MASK)
-#define CHAR_ENC_TYPE(x) (LEVELS(x) & (BYTES_MASK | UTF8_MASK | LATIN1_MASK))
-
 bool translation_required_chr(SEXP x, R_len_t size);
 SEXP translate_encoding2(SEXP x, R_len_t x_size, SEXP y, R_len_t y_size);
 
--- a/src/vctrs.h
+++ b/src/vctrs.h
@@ -282,8 +282,8 @@ SEXP vec_as_unique_names(SEXP names, bool quiet);
 
 // Character translation ----------------------------------------
 
-bool translation_required_chr(SEXP x, R_len_t size);
-SEXP translate_encoding2(SEXP x, R_len_t x_size, SEXP y, R_len_t y_size);
+bool chr_translation_required(SEXP x, R_len_t size);
+SEXP obj_translate_encoding2(SEXP x, R_len_t x_size, SEXP y, R_len_t y_size);
 
 // Growable vector ----------------------------------------------
 
--- a/src/vctrs.h
+++ b/src/vctrs.h
@@ -6,16 +6,6 @@
 #include <stdbool.h>
 #include <stdint.h>
 
-// `sxpinfo.gp` is exposed through `LEVELS()`, and CHARSXP info is
-// extractable through a bitwise and. This copies the unexposed
-// interface in Defn.h
-#define BYTES_MASK 2
-#define LATIN1_MASK 4
-#define UTF8_MASK 8
-
-#define CHAR_IS_UTF8(x) (LEVELS(x) & UTF8_MASK)
-#define CHAR_ENC_TYPE(x) (LEVELS(x) & (BYTES_MASK | UTF8_MASK | LATIN1_MASK))
-
 typedef R_xlen_t r_ssize_t;
 
 
@@ -290,6 +280,20 @@ SEXP vec_as_names(SEXP names, enum name_repair_arg type, bool quiet);
 bool is_unique_names(SEXP names);
 SEXP vec_as_unique_names(SEXP names, bool quiet);
 
+// Character translation ----------------------------------------
+
+// `sxpinfo.gp` is exposed through `LEVELS()`, and CHARSXP info is
+// extractable through a bitwise and. This copies the unexposed
+// interface in Defn.h
+#define BYTES_MASK 2
+#define LATIN1_MASK 4
+#define UTF8_MASK 8
+
+#define CHAR_IS_UTF8(x) (LEVELS(x) & UTF8_MASK)
+#define CHAR_ENC_TYPE(x) (LEVELS(x) & (BYTES_MASK | UTF8_MASK | LATIN1_MASK))
+
+bool chr_requires_translation(SEXP x, R_len_t size);
+SEXP translate_common_encoding(SEXP x, R_len_t x_size, SEXP y, R_len_t y_size);
 
 // Growable vector ----------------------------------------------
 
--- a/src/vctrs.h
+++ b/src/vctrs.h
@@ -292,8 +292,8 @@ SEXP vec_as_unique_names(SEXP names, bool quiet);
 #define CHAR_IS_UTF8(x) (LEVELS(x) & UTF8_MASK)
 #define CHAR_ENC_TYPE(x) (LEVELS(x) & (BYTES_MASK | UTF8_MASK | LATIN1_MASK))
 
-bool chr_requires_translation(SEXP x, R_len_t size);
-SEXP translate_common_encoding(SEXP x, R_len_t x_size, SEXP y, R_len_t y_size);
+bool translation_required_chr(SEXP x, R_len_t size);
+SEXP translate_encoding2(SEXP x, R_len_t x_size, SEXP y, R_len_t y_size);
 
 // Growable vector ----------------------------------------------
 
--- a/src/dictionary.c
+++ b/src/dictionary.c
@@ -229,7 +229,7 @@ SEXP vctrs_match(SEXP needles, SEXP haystack) {
   R_len_t n_haystack = vec_size(haystack);
   R_len_t n_needle = vec_size(needles);
 
-  SEXP translated = PROTECT_N(translate_common_encoding(needles, n_needle, haystack, n_haystack), &nprot);
+  SEXP translated = PROTECT_N(translate_encoding2(needles, n_needle, haystack, n_haystack), &nprot);
   needles = VECTOR_ELT(translated, 0);
   haystack = VECTOR_ELT(translated, 1);
 
@@ -282,7 +282,7 @@ SEXP vctrs_in(SEXP needles, SEXP haystack) {
   R_len_t n_haystack = vec_size(haystack);
   R_len_t n_needle = vec_size(needles);
 
-  SEXP translated = PROTECT_N(translate_common_encoding(needles, n_needle, haystack, n_haystack), &nprot);
+  SEXP translated = PROTECT_N(translate_encoding2(needles, n_needle, haystack, n_haystack), &nprot);
   needles = VECTOR_ELT(translated, 0);
   haystack = VECTOR_ELT(translated, 1);
 
--- a/src/dictionary.c
+++ b/src/dictionary.c
@@ -226,12 +226,18 @@ SEXP vctrs_match(SEXP needles, SEXP haystack) {
   needles = PROTECT_N(vec_proxy_equal(needles), &nprot);
   haystack = PROTECT_N(vec_proxy_equal(haystack), &nprot);
 
+  R_len_t n_haystack = vec_size(haystack);
+  R_len_t n_needle = vec_size(needles);
+
+  SEXP translated = PROTECT_N(translate_common_encoding(needles, n_needle, haystack, n_haystack), &nprot);
+  needles = VECTOR_ELT(translated, 0);
+  haystack = VECTOR_ELT(translated, 1);
+
   dictionary d;
   dict_init(&d, haystack);
   PROTECT_DICT(&d, &nprot);
 
   // Load dictionary with haystack
-  R_len_t n_haystack = vec_size(haystack);
   for (int i = 0; i < n_haystack; ++i) {
     uint32_t hash = dict_hash_scalar(&d, i);
 
@@ -244,7 +250,6 @@ SEXP vctrs_match(SEXP needles, SEXP haystack) {
   dict_init_partial(&d_needles, needles);
 
   // Locate needles
-  R_len_t n_needle = vec_size(needles);
   SEXP out = PROTECT_N(Rf_allocVector(INTSXP, n_needle), &nprot);
   int* p_out = INTEGER(out);
 
@@ -274,12 +279,18 @@ SEXP vctrs_in(SEXP needles, SEXP haystack) {
   needles = PROTECT_N(vec_proxy_equal(needles), &nprot);
   haystack = PROTECT_N(vec_proxy_equal(haystack), &nprot);
 
+  R_len_t n_haystack = vec_size(haystack);
+  R_len_t n_needle = vec_size(needles);
+
+  SEXP translated = PROTECT_N(translate_common_encoding(needles, n_needle, haystack, n_haystack), &nprot);
+  needles = VECTOR_ELT(translated, 0);
+  haystack = VECTOR_ELT(translated, 1);
+
   dictionary d;
   dict_init(&d, haystack);
   PROTECT_DICT(&d, &nprot);
 
   // Load dictionary with haystack
-  R_len_t n_haystack = vec_size(haystack);
   for (int i = 0; i < n_haystack; ++i) {
     uint32_t hash = dict_hash_scalar(&d, i);
 
@@ -293,7 +304,6 @@ SEXP vctrs_in(SEXP needles, SEXP haystack) {
   PROTECT_DICT(&d_needles, &nprot);
 
   // Locate needles
-  R_len_t n_needle = vec_size(needles);
   SEXP out = PROTECT_N(Rf_allocVector(LGLSXP, n_needle), &nprot);
   int* p_out = LOGICAL(out);
 
--- a/src/dictionary.c
+++ b/src/dictionary.c
@@ -229,7 +229,7 @@ SEXP vctrs_match(SEXP needles, SEXP haystack) {
   R_len_t n_haystack = vec_size(haystack);
   R_len_t n_needle = vec_size(needles);
 
-  SEXP translated = PROTECT_N(translate_encoding2(needles, n_needle, haystack, n_haystack), &nprot);
+  SEXP translated = PROTECT_N(obj_translate_encoding2(needles, n_needle, haystack, n_haystack), &nprot);
   needles = VECTOR_ELT(translated, 0);
   haystack = VECTOR_ELT(translated, 1);
 
@@ -282,7 +282,7 @@ SEXP vctrs_in(SEXP needles, SEXP haystack) {
   R_len_t n_haystack = vec_size(haystack);
   R_len_t n_needle = vec_size(needles);
 
-  SEXP translated = PROTECT_N(translate_encoding2(needles, n_needle, haystack, n_haystack), &nprot);
+  SEXP translated = PROTECT_N(obj_translate_encoding2(needles, n_needle, haystack, n_haystack), &nprot);
   needles = VECTOR_ELT(translated, 0);
   haystack = VECTOR_ELT(translated, 1);
 
--- a/src/equal.c
+++ b/src/equal.c
@@ -150,7 +150,7 @@ static int chr_equal_scalar_impl(const SEXP x, const SEXP y) {
     return 1;
   }
 
-  if (CHAR_ENC_TYPE(x) != CHAR_ENC_TYPE(y)) {
+  if (Rf_getCharCE(x) != Rf_getCharCE(y)) {
     const void *vmax = vmaxget();
     int out = !strcmp(Rf_translateCharUTF8(x), Rf_translateCharUTF8(y));
     vmaxset(vmax);
--- a/src/hash.c
+++ b/src/hash.c
@@ -184,7 +184,7 @@ static uint32_t dbl_hash(SEXP x) {
   HASH(double, REAL_RO, dbl_hash_scalar);
 }
 static uint32_t chr_hash(SEXP x) {
-  if (translation_required_chr(x, Rf_length(x))) {
+  if (chr_translation_required(x, Rf_length(x))) {
     HASH(SEXP, STRING_PTR_RO, chr_hash_scalar_translate);
   } else {
     HASH(SEXP, STRING_PTR_RO, chr_hash_scalar);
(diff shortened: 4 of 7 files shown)


DavisVaughan:
`vec_match()` benchmarks. There are a number of them, but I'm trying to cover the cases of character vectors, lists, and data frames.

Setup:

```r
vec_match_old <- vctrs2::vec_match
library(vctrs)
library(tibble)
set.seed(123)

# table - 200000 words, 10 letters each
# x - A subset of 20000 of them, with replacement
table_unknown <- stringi::stri_rand_strings(200000, 10)
x_unknown <- sample(table_unknown, 20000, replace = TRUE)
```

`match()` has always been faster than `vec_match()` in the ascii case:

```r
bench::mark(
  vec_match_old(x_unknown, table_unknown),
  vec_match(x_unknown, table_unknown),
  match(x_unknown, table_unknown),
  iterations = 300
)
#> # A tibble: 3 x 6
#>   expression                                  min  median `itr/sec`
#>   <bch:expr>                              <bch:t> <bch:t>     <dbl>
#> 1 vec_match_old(x_unknown, table_unknown)  13.2ms 14.88ms      65.8
#> 2 vec_match(x_unknown, table_unknown)     13.61ms 14.53ms      67.3
#> 3 match(x_unknown, table_unknown)          4.81ms  5.24ms     186. 
```

With any UTF8 characters, `match()` attempts a conversion, slowing it down.

```r
table_utf8 <- paste0(table_unknown, "\u00B0C")
x_utf8 <- sample(table_utf8, 20000, replace = TRUE)

bench::mark(
  vec_match_old(x_utf8, table_utf8),
  vec_match(x_utf8, table_utf8),
  match(x_utf8, table_utf8),
  iterations = 300
)
#> # A tibble: 3 x 6
#>   expression                           min median `itr/sec` mem_alloc
#>   <bch:expr>                        <bch:> <bch:>     <dbl> <bch:byt>
#> 1 vec_match_old(x_utf8, table_utf8) 12.6ms 14.3ms      68.1    1.92MB
#> 2 vec_match(x_utf8, table_utf8)     13.4ms 14.4ms      67.5    1.92MB
#> 3 match(x_utf8, table_utf8)         12.6ms 14.4ms      66.8    3.75MB
```

This means `match()` converts latin1 to utf8 when it don't need to:

```r
table_latin1 <- iconv(table_utf8, "UTF-8", "latin1")
x_latin1 <- iconv(x_utf8, "UTF-8", "latin1")

# match() converts to UTF8, but vec_match() doesn't
bench::mark(
  vec_match_old(x_latin1, table_latin1),
  vec_match(x_latin1, table_latin1),
  match(x_latin1, table_latin1)
)
#> # A tibble: 3 x 6
#>   expression                               min median `itr/sec` mem_alloc
#>   <bch:expr>                            <bch:> <bch:>     <dbl> <bch:byt>
#> 1 vec_match_old(x_latin1, table_latin1) 13.6ms 14.7ms      67.5    1.92MB
#> 2 vec_match(x_latin1, table_latin1)     14.6ms 15.9ms      62.1    1.92MB
#> 3 match(x_latin1, table_latin1)         94.7ms 96.1ms      10.4    3.75MB
```

Here we actually have to translate to UTF8 because the table was UTF8

```r
x_utf8_but_unknown <- x_utf8
Encoding(x_utf8_but_unknown) <- "unknown"

# Having to translate `needles` to UTF8
bench::mark(
  # wrong! vec_match_old(x_utf8_but_unknown, table_utf8),
  vec_match(x_utf8_but_unknown, table_utf8),
  match(x_utf8_but_unknown, table_utf8),
  iterations = 100
)
#> # A tibble: 2 x 6
#>   expression                                   min median `itr/sec`
#>   <bch:expr>                                <bch:> <bch:>     <dbl>
#> 1 vec_match(x_utf8_but_unknown, table_utf8) 26.2ms 27.6ms      35.8
#> 2 match(x_utf8_but_unknown, table_utf8)     26.8ms 30.4ms      32.2
```

Here we have a list that contains UTF8 characters in the `haystack`, which means we have to convert every character vector to UTF8. Not sure why this is slower, it's a strange case anyways, but I don't think it has to do with the new code. 

```r
lst_x_utf8_but_unknown <- as.list(x_utf8_but_unknown)
lst_table_utf8 <- as.list(table_utf8)

bench::mark(
  # wrong! vec_match_old(lst_x_utf8_but_unknown, lst_table_utf8),
  vec_match(lst_x_utf8_but_unknown, lst_table_utf8),
  match(lst_x_utf8_but_unknown, lst_table_utf8),
  iterations = 50
)
#> # A tibble: 2 x 6
#>   expression                                           min median `itr/sec`
#>   <bch:expr>                                        <bch:> <bch:>     <dbl>
#> 1 vec_match(lst_x_utf8_but_unknown, lst_table_utf8) 94.1ms 98.6ms      10.2
#> 2 match(lst_x_utf8_but_unknown, lst_table_utf8)     35.8ms 40.3ms      24.7
```

The speed of matching with tibbles containing character vectors wasn't really affected:

```r
table_df <- tibble(x = table_unknown, y = table_utf8)
x_df <- vec_slice(table_df, sample(vec_size(table_df), 20000, replace = TRUE))

bench::mark(
  vec_match_old(x_df, table_df),
  vec_match(x_df, table_df),
  # wrong! match(x_df, table_df),
  iterations = 50
)
#> # A tibble: 2 x 6
#>   <bch:expr>                    <bch:t> <bch:>     <dbl> <bch:byt>    <dbl>
#> 1 vec_match_old(x_df, table_df)    50ms   56ms      17.8    1.94MB    1.14 
#> 2 vec_match(x_df, table_df)        54ms 57.6ms      17.2    1.93MB    0.715
```

But since the translation check is recursive we can now correctly translate the `y` column to UTF8, and leave the `x` column alone:

```r
x_df_unknown <- x_df
x_df_unknown$y <- `Encoding<-`(x_df_unknown$y, "unknown")

bench::mark(
  # wrong! vec_match_old(x_df_unknown, table_df),
  vec_match(x_df_unknown, table_df),
  # wrong! match(x_df_unknown, table_df),
  iterations = 50
)
#> # A tibble: 1 x 6
#>   expression                           min median `itr/sec` mem_alloc
#>   <bch:expr>                        <bch:> <bch:>     <dbl> <bch:byt>
#> 1 vec_match(x_df_unknown, table_df) 67.1ms 72.2ms      13.5    3.59MB
```

src/vctrs.h
```diff
@@ -290,6 +280,20 @@ SEXP vec_as_names(SEXP names, enum name_repair_arg type, bool quiet);
 bool is_unique_names(SEXP names);
 SEXP vec_as_unique_names(SEXP names, bool quiet);
 
+// Character translation ----------------------------------------
+
+// `sxpinfo.gp` is exposed through `LEVELS()`, and CHARSXP info is
+// extractable through a bitwise and. This copies the unexposed
+// interface in Defn.h
+#define BYTES_MASK 2
```
lionel-: Do we have evidence that other R implementations use the same protocol for encoding? Good evidence would be that it's used in Rcpp, for example.

lionel-: Have you checked that it's worth it to use these instead of `getCharCE()`? We need to go through `LEVELS()` anyway, which is a function call because of the memory barrier, so it's not clear a priori we gain much. I'd prefer not relying on unexported API. (This should be checked with compiler optimisations enabled, maybe the branched mask inspection is optimised.)

DavisVaughan: I got the idea from data.table actually https://github.com/Rdatatable/data.table/blob/4d355d65353eabfb2a5b58549d96b66e2266a951/src/data.table.h#L19

I'll give it a test

DavisVaughan:
> use prefixes to indicate input type. Suffixes are for output type.

I hadn't picked up that this was the convention! I'll change it.

Would you call `translate_encoding()` -> `object_translate_encoding()`. I thought about that before and was curious.

lionel-:
It could be fine as is, or maybe `obj_` or `sexp_`?

src/vctrs.h
DavisVaughan: Hmm, I think you might be right. I swear I thought the performance difference was more prominent, but it doesn't seem to matter _too_ much. I'll remove in favor of `Rf_getCharCE()`. The good news is that it should do exactly the same thing as what I was doing based on [its implementation](https://github.com/wch/r-source/blob/0e291d10c6631984f149320b6b30c05ce7da79ae/src/main/sysutils.c#L779).

`vec_equal()` benchmark - I switched out `CHAR_ENC_TYPE()` for `Rf_getCharCE()` [here](). This should be a "worst case scenario" because basically all of the strings are different, so that encoding check is run for every element. This would also be the worst case scenario for `Rf_getCharCE()` in general since returning `CE_NATIVE` is the last thing it does, after determining it isnt utf8/latin1/bytes.

```r
library(vctrs)
set.seed(123)

x <- stringi::stri_rand_strings(1000000, 10)
x_rev <- rev(x)

# With CHAR_ENC_TYPE()
bench::mark(
  vec_equal(x, x_rev),
  iterations = 2000
)
#> # A tibble: 1 x 6
#> 1 vec_equal(x, x_rev)   10.4ms   10.8ms      91.8    3.87MB     9.92

# with Rf_getCharCE()
bench::mark(
  vec_equal(x, x_rev),
  iterations = 2000
)
#> # A tibble: 1 x 6
#> 1 vec_equal(x, x_rev)   10.7ms   11.2ms      88.7    3.87MB     9.58
```

`vec_match()` benchmark - I switched out `CHAR_ENC_TYPE()` for `Rf_getCharCE()` with the most important spot being the loop [here](https://github.com/r-lib/vctrs/pull/565/files#diff-853c27e1af01a8b6f3bd37411c2bd81bR15) which is run on both `needles` and `haystack`. I think the difference is small enough to just use the interface supplied by R directly.

```r
library(vctrs)
set.seed(123)

table_unknown <- stringi::stri_rand_strings(500000, 10)
x_unknown <- sample(table_unknown, 250000, replace = TRUE)

# with CHAR_ENC_TYPE()
bench::mark(
  vec_match(x_unknown, table_unknown),
  iterations = 2000
)
#> # A tibble: 1 x 6
#>   expression                             min median `itr/sec` mem_alloc
#>   <bch:expr>                          <bch:> <bch:>     <dbl> <bch:byt>
#> 1 vec_match(x_unknown, table_unknown) 38.4ms 56.8ms      18.2    7.82MB

# with Rf_getCharCE()
bench::mark(
  vec_match(x_unknown, table_unknown),
  iterations = 2000
)
#> # A tibble: 1 x 6
#>   expression                             min median `itr/sec` mem_alloc
#>   <bch:expr>                          <bch:> <bch:>     <dbl> <bch:byt>
#> 1 vec_match(x_unknown, table_unknown) 54.9ms 59.4ms      16.8    7.82MB
```

PR #581, opened 2019-09-23
Export `vec_recycle()` as callable from C

DavisVaughan:
One more callable that I will need in {slide}. See [this PR](https://github.com/DavisVaughan/slide/pull/25) if interested in why.

@lionel- does this suffer from the `R_len_t` vs `R_xlen_t` problem like `vec_size()`? If so, should the exported callable for `vec_init()` also be changed?

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,6 +1,6 @@
 Package: vctrs
 Title: Vector Helpers
-Version: 0.2.0.9002
+Version: 0.2.0.9003
 Authors@R: 
     c(person(given = "Hadley",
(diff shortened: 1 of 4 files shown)


lionel-:
Yes, maybe add a `short` suffix/prefix?

DavisVaughan:
@lionel- I prefixed them all like `short_vec_recycle()` now. I think that is a clearer indication that they are "special" as opposed to having `vec_recycle_short()`, which sort of makes it look like that is a normal vctrs function, with nothing special going on.

PR #585, opened 2019-09-24
Use `intptr_t` rather than `int64_t` to avoid 32-bit pointer->int cast warning

DavisVaughan:
@lionel- this does fix the Windows warning. I assume what is happening for now is that `intptr_t` is equivalent to `int32_t` on the 32-bit Windows build, which is then converted to `int64_t` when it gets to `hash_int64()`, but that is allowed without a warning.

I'll open another issue related to consistently using unsigned integer types

### DIFF
--- a/src/hash.c
+++ b/src/hash.c
@@ -42,7 +42,7 @@ static uint32_t hash_double(double x) {
 }
 
 static uint32_t hash_char(SEXP x) {
-  return hash_int64((int64_t) x);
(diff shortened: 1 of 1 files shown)


lionel-:
ah that makes sense!

I wonder if this might cause some hashing issues on 32 bit platforms (poor spreading across buckets) but I guess it's not worth worrying about as these platforms are on the way out.

PR #588, opened 2019-09-24
Ensure that only 1 or 0 is returned when checking if doubles are `NA`

### LINKED ISSUE #587: `vec_equal_na(NA_real_)` returns a logical that internally is -1
On a Windows Server 2019 build running 64-bit R, we get the following, where the `-1` should be a `1`:

```r
> .Internal(inspect((vec_equal_na(NA_real_))))
@0x00000241f470bbb0 10 LGLSXP g0c1 [] (len=1, tl=0) -1
```

This causes these tests to fail:

https://github.com/r-lib/vctrs/blob/da5db68f88cab710facc3603f1f6fddac84b2570/tests/testthat/test-equal.R#L176

https://github.com/r-lib/vctrs/blob/da5db68f88cab710facc3603f1f6fddac84b2570/tests/testthat/test-equal.R#L177

The problem is that `isnan()`, which is used [here](https://github.com/r-lib/vctrs/blob/da5db68f88cab710facc3603f1f6fddac84b2570/src/equal.c#L457) for checking if a double is an NA value, does _not_ consistently return 1 or 0, but instead just returns a non-zero integer or 0. The fix is to use base R's `ISNAN()`, which is a thin wrapper around `isnan()` defined as `isnan(x) != 0`.

I assume this means we will take a small performance hit because we have an extra comparison check, and a conversion from boolean to integer, but the conversion cost is probably quite low.

We also use `isnan()` [in a few other places](https://github.com/r-lib/vctrs/search?utf8=%E2%9C%93&q=isnan&type=), but never as the return value. We use it in an `if () {}` condition, which should be safe, so I will leave those as is.

(I will submit a PR)

DavisVaughan:
Closes #587 

Bench marks show that performance isn't affected much, which is great.

``` r
library(vctrs)
x <- rep(c(1, NA), times = 500000)

# before
bench::mark(vec_equal_na(x), iterations = 10000)
#> # A tibble: 1 x 6
#> 1 vec_equal_na(x)    376µs    742µs     1316.    3.82MB     132.

# after
bench::mark(vec_equal_na(x), iterations = 10000)
#> # A tibble: 1 x 6
#> 1 vec_equal_na(x)    472µs    749µs     1286.    3.82MB     129.
```

### DIFF
--- a/src/equal.c
+++ b/src/equal.c
@@ -454,7 +454,9 @@ static int int_equal_na_scalar(const int* x) {
 
 static int dbl_equal_na_scalar(const double* x) {
   // is.na(NaN) is TRUE
-  return isnan(*x);
+  // isnan() does not consistently return 1 and 0 on all platforms,
+  // but R's ISNAN() does
+  return ISNAN(*x);
 }
 
 static int chr_equal_na_scalar(const SEXP* x) {


DavisVaughan:
No new tests were added, as there were failing tests on the Window build already that are now fixed by this

PR #590, opened 2019-09-24
Only test unique/equal functions against base R equivalents

### LINKED ISSUE #589: Encoding test failures on Windows
There are a number of failures related to string encodings on Windows that don't occur on Mac/Linux. These aren't errors in the underlying code, but the tests don't take into account the fact that `Encoding(x) <- "unknown"` does different things on Windows than it does on Mac/Linux.

For example:

Mac:

``` r
utf8 <- "\u00B0C"
utf8
#> [1] "°C"

unknown <- utf8
unknown
#> [1] "°C"

utf8 == unknown
#> [1] TRUE

vctrs::vec_equal(utf8, unknown)
#> [1] TRUE
```

Windows:

``` r
utf8 <- "\u00B0C"
utf8
#> [1] "°C"

unknown <- utf8
unknown
#> [1] "Â°C"

utf8 == unknown
#> [1] FALSE

vctrs::vec_equal(utf8, unknown)
#> [1] FALSE
```

So, we are doing the same as base R, but a few tests expect the result to be `TRUE` because I didn't have access to a Windows computer to test against at the time. 

I think the right thing to do is just check the result against base R when the results should be the same (I do this already).

When we purposefully diverge from base R (iterating row wise on data frames), we can either skip on windows or have an `if(on_windows())` branch in the test

DavisVaughan:
Closes #589 

This PR fixes the failures experienced on Windows due to tests not anticipating that `Encoding(x) <- "unknown"` does different things than it does on Mac/Linux.

1) Where we are equivalent with base R, we now explicitly compare against the base equivalent rather than expecting a known result. This is robust against the OS differences.

2) Where we are different from base R, we use `on_windows()` to alter the expected output.

An alternative for 2) is to just skip those tests on windows.

### DIFF
--- a/tests/testthat/helper-os.R
+++ b/tests/testthat/helper-os.R
@@ -1,0 +1,3 @@
+on_windows <- function() {
+  tolower(Sys.info()[["sysname"]]) == "windows"
+}
--- a/tests/testthat/test-dictionary.R
+++ b/tests/testthat/test-dictionary.R
@@ -118,7 +118,6 @@ test_that("vec_unique() can detect uniqueness with the same string in various en
 
   x <- c(unknown, utf8, latin1)
 
-  expect_equal(vec_unique(x), x[1])
   expect_equal(vec_unique(x), unique(x))
 })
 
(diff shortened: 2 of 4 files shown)


DavisVaughan:
Closed in favor of #591 which takes a different approach and doesn't use any of the code here

PR #591, opened 2019-09-24
Use `iconv()` for portable encoding conversions

### LINKED ISSUE #589: Encoding test failures on Windows

DavisVaughan:
Closes #589 

This PR fixes the failures experienced on Windows due to tests not anticipating that `Encoding(x) <- "unknown"` does different things than it does on Mac/Linux.

We now always use `iconv()` to generate the differently encoded variables required for the tests (except bytes, which requires `Encoding<-` still as far as I can tell).

### DIFF
--- a/tests/testthat/test-dictionary.R
+++ b/tests/testthat/test-dictionary.R
@@ -109,42 +109,29 @@ test_that("unique functions take the equality proxy (#375)", {
 })
 
 test_that("vec_unique() can detect uniqueness with the same string in various encodings (#553)", {
-  string <- "\u00B0C"
-  utf8 <- iconv(string, from = Encoding(string), to = "UTF-8")
-  unknown <- iconv(string, from = Encoding(string), to = "", mark = FALSE)
-  latin1 <- iconv(string, from = Encoding(string), to = "latin1")
-
-  x <- c(unknown, utf8, latin1)
+  enc <- encodings()
+  x <- c(enc$unknown, enc$utf8, enc$latin1)
 
   expect_equal(vec_unique(x), x[1])
   expect_equal(vec_unique(x), unique(x))
 })
 
 test_that("vec_unique() returns differently encoded strings in the order they appear", {
-  string <- "\u00B0C"
-  utf8 <- iconv(string, from = Encoding(string), to = "UTF-8")
-  unknown <- iconv(string, from = Encoding(string), to = "", mark = FALSE)
-
-  x <- c(unknown, utf8)
-  y <- c(utf8, unknown)
+  enc <- encodings()
+  x <- c(enc$unknown, enc$utf8)
+  y <- c(enc$utf8, enc$unknown)
 
   expect_equal(Encoding(vec_unique(x)), "unknown")
   expect_equal(Encoding(vec_unique(y)), "UTF-8")
 })
 
 test_that("vec_unique() can determine uniqueness when the encoding is the same", {
-  string <- "\u00B0C"
-  utf8 <- iconv(string, from = Encoding(string), to = "UTF-8")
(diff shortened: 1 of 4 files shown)


hadley:
Maybe it’d be worthwhile to pull this out into a helper function? Maybe returning a named character vector with all three encodings?

DavisVaughan:
A named list where the actual elements were unnamed worked better than a named character vector, due to some weird inconsistencies with `match()` and named character vectors (see the character examples in #599)

tests/testthat/test-dictionary.R
```diff
@@ -233,12 +207,10 @@ test_that("can take the unique loc of 1d arrays (#461)", {
...
 
 test_that("can use matching functions with strings with different encodings", {
-  utf8 <- "\u00B0C"
-
-  unknown <- utf8
-  Encoding(unknown) <- "unknown"
-
-  latin1 <- iconv(utf8, "UTF-8", "latin1")
+  enc <- encodings()
+  utf8 <- enc$utf8
```
hadley: I think it'd be a little nicer to just use `enc$utf` in the tests

tests/testthat/test-dictionary.R
```diff
@@ -254,29 +226,21 @@ test_that("can use matching functions with strings with different encodings", {
...
-
-  bytes <- unknown
-  Encoding(bytes) <- "bytes"
+  enc <- encodings()
+  utf8 <- enc$utf8
+  unknown <- enc$unknown
+  latin1 <- enc$latin1
+  bytes <- enc$bytes
 
   expect_equal(vec_match(unknown, unknown), 1L)
```
hadley: Maybe it's worth turning this into a nested for loop to test every combination?

tests/testthat/test-dictionary.R
DavisVaughan: `enc$latin1` -> `enc$latin` too?

hadley: Yeah, ie don’t assign out the list elements.

DavisVaughan: Ah gotcha, I thought you meant actually change all occurrences of `utf8` to `utf`. I see what you mean now. Done.

tests/testthat/test-dictionary.R
DavisVaughan: Did this in a number of places, simplifying many of the tests

DavisVaughan:
- Simplified many tests with loops
- Extracted out `encoding_bytes()` as most tests don't need it mixed in with the other encodings, and it makes looping cleaner

PR #593, opened 2019-09-25
Add 'list' class to list_of vctrs

jeroen:
This adds the `list` class to `list_of` vctrs. This solves the issue with jsonlite (and possibly other packages implementing S4 list methods):

 - https://github.com/tidyverse/tidyr/issues/750
 - https://github.com/jeroen/jsonlite/issues/292

Perhaps jsonlite is ill-behaved, but I think adding the `list` class to `list_of` makes sense?

### DIFF
--- a/R/type-list-of.R
+++ b/R/type-list-of.R
@@ -62,7 +62,10 @@ new_list_of <- function(x = list(), ptype = logical(), ..., class = character())
   stopifnot(is.list(x))
   stopifnot(vec_size(ptype) == 0)
 
-  new_vctr(x, ..., ptype = ptype, class = c(class, "vctrs_list_of"))
+  # Append 'list' class after 'vctrs_vctr'
+  out <- new_vctr(x, ..., ptype = ptype, class = c(class, "vctrs_list_of"))
+  class(out) <- c(class(out), "list")
+  out
 }
 
 #' @export
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -37,6 +37,7 @@ Suggests:
     covr,
     crayon,
     generics,
+    jsonlite,
     knitr,
     pillar (>= 1.4.1),
     pkgdown,
--- a/NEWS.md
+++ b/NEWS.md
@@ -8,6 +8,8 @@
 * Equality and ordering methods are now implemented for raw and
   complex vectors (@romainfrancois).
 
+* `list_of` vctrs now include a 'list' class for compatibility with 
+   legacy S4 list methods, including jsonlite.
 
 # vctrs 0.2.0
 
--- a/src/utils.c
+++ b/src/utils.c
@@ -1029,13 +1029,14 @@ void vctrs_init_utils(SEXP ns) {
   SET_STRING_ELT(classes_tibble, 2, strings_data_frame);
 
 
-  classes_list_of = Rf_allocVector(STRSXP, 2);
+  classes_list_of = Rf_allocVector(STRSXP, 3);
   R_PreserveObject(classes_list_of);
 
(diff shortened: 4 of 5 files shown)


R/type-list-of.R
```diff
@@ -62,7 +62,10 @@ new_list_of <- function(x = list(), ptype = logical(), ..., class = character())
   stopifnot(is.list(x))
   stopifnot(vec_size(ptype) == 0)
 
-  new_vctr(x, ..., ptype = ptype, class = c(class, "vctrs_list_of"))
+  # Append 'list' class after 'vctrs_vctr'
+  out <- new_vctr(x, ..., ptype = ptype, class = c(class, "vctrs_list_of"))
+  class(out) <- c(class(out), "list")
```
hadley: I think you can just inline `list` into the class list — the point is not to preserve the existing class, but to lift the base type (which becomes the implicit class) into the subclass. 

I'm still not sure this is correct but I'm thinking about it.

hadley:
Ok, @lionel- has convinced me by saying that we could treat the presence of the list class as a signal that the object is a vector (not a scalar, like e.g. `lm()`). (@DavisVaughan — I think this is how we'll resolve the unnest problem in a user friendly way)

Can you please simplify the test to remove the dependency on a json? I think testing that `list_of` inherits from `list` should be adequate.

DavisVaughan:
I agree with this! This is what I was trying to say in the unnest issue as well

lionel-:
What is the unnest problem?

DavisVaughan:
https://github.com/tidyverse/tidyr/issues/761#issuecomment-534743783

lionel-:
I like this approach because it gives an easier way of making a class compatible with vctrs. Currently you can declare a list to be a vector by defining a proxy, but then vctrs assumes the class is fully implemented with vctrs primitives and no longer falls back to base methods, eg. to `[[` when slicing. Adding an explicit `"list"` class for compatibility with vctrs should be an easy change and undisruptive in most cases.

R/type-list-of.R
jeroen: Such that the resulting `list_of` class will be `c("vctrs_list_of", "list", "vctrs_vctr")` ?

lionel-: I think `"list"` should be last because we can't superclass foreign classes.

jeroen: Right, that's why I implemented it this way. So keep the PR as is?

hadley: I wonder if we need another argument to `new_vctr()`? Like `new_vctr(x, ..., ptype = ptype, class = c(class, "vctrs_list_of"), preserve_implict_class = TRUE)` (but a better name)

lionel-: oh yes that's occasionally useful.

```r
explicit_base_class = TRUE
base_class = TRUE
subclass_base_type = TRUE
```

Maybe the first one?

hadley: Or `explicit_type`? Otherwise, I like `explicit_base_class` best of the options so far.

hadley: Another option would be `include_type_in_class` (that's too long, but might suggest other ideas)

DavisVaughan: `explicit_typeof` might also work, depending on how we'd extract the base type

jeroen: Should I try to add something like this to the PR? Anyone with strong opinions want to take it from here?

lionel-: No worries @jeroen, I can finish the PR when I get back to work on vctrs.

hadley: How about `extends_type`?

PR #597, opened 2019-09-25
Issue 594

### LINKED ISSUE #594: `vec_size_common()` should enforce an integer return value
``` r
class(vctrs::vec_size_common(.size = 5))
#> [1] "numeric"
```

This is an easy fix, we just need to use the return value from `size_validate()` rather than ignoring it. I'll do a PR

DavisVaughan:
Closes #594 

We now return the validated size from `size_validate()`, which has been cast to an integer.

### DIFF
--- a/src/size-common.c
+++ b/src/size-common.c
@@ -8,9 +8,9 @@ SEXP vctrs_size_common(SEXP call, SEXP op, SEXP args, SEXP env) {
 
   SEXP size = PROTECT(Rf_eval(CAR(args), env)); args = CDR(args);
   if (size != R_NilValue) {
-    size_validate(size, ".size");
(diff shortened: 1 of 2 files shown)


hadley:
My only concern is that at some point we will need to add long vector support, making all sizes doubles.

lionel-:
I share Hadley's concern.

> making all sizes doubles.

Another approach is to make validated sizes either INTSXP or REALSXP depending on the magnitude of the index, since making them all double will impact performance with large index vectors.

DavisVaughan:
After discussing with Lionel, we decided to merge this for now and open a meta-issue to collect cases where we need to support long vectors.

PR #601, opened 2019-09-28
Add azure pipelines

DavisVaughan:
Thanks @jimhester!

Jim sent this in a few days ago. I've rebased on current master, which has all the fixes for Windows. We are now cleanly passing everywhere!


lionel-:
Awesome!

Issue #599, opened 2019-09-27
A clarification on `vec_match()` with attributes


DavisVaughan:
What is the correct behavior of `vec_match()` when attributes are involved? Consider a few base R examples:

- When the names are on the outermost vector, they are ignored.
- When the names are on the inner elements of a list, they are important since R calls `as.character()` on the inputs when they are lists.

``` r
x <- c(x_nm = 1)
y <- c(y_nm = 1)

match(x, y)
#> [1] 1

match(list(x), list(y))
#> [1] NA

as.character(list(x))
#> [1] "c(x_nm = 1)"

as.character(list(y))
#> [1] "c(y_nm = 1)"
```

Notice this has inconsistent behavior though. Consider what happens with characters. Notice how the `as.character()` call doesn't retain the names, so things look equal.

``` r
x <- c(x_nm = "1")
y <- c(y_nm = "1")

match(x, y)
#> [1] 1

match(list(x), list(y))
#> [1] 1

as.character(list(x))
#> [1] "1"

as.character(list(y))
#> [1] "1"
```

With other attributes, they are ignored (as far as i can tell) when converting to character.

``` r
x <- 1
x_attr <- structure(1, stuff = 1)

match(x, x_attr)
#> [1] 1

match(list(x), list(x_attr))
#> [1] 1

as.character(list(x))
#> [1] "1"

as.character(list(x_attr))
#> [1] "1"
```

What should `vec_match()` do? There are really two questions:

- Should names and attributes be ignored with this? My gut reaction is yes.

```r
vec_match(c(x_nm = 1), c(y_nm = 1))
#> [1] 1
```

- Should names and attributes be ignored in the nested version? My gut reaction is no.

```r
vec_match(list(c(x_nm = 1)), list(c(y_nm = 1)))
#> [1] NA
```

Similar questions arise with `vec_equal()` on these same examples.

DavisVaughan:
There might be nothing to do, as I think `vec_match()` is already doing the behavior I am proposing. But it might be worth adding a few more tests specifically about these edge cases to document the differences.

There is currently one bug with `vec_match()` and characters with different encodings, but I figured it out and its an easy fix. This is currently broken according to what I think `vec_match()` should be doing:

``` r
library(vctrs)

string <- "\u00B0C"

utf8 <- iconv(string, from = Encoding(string), to = "UTF-8")
latin1 <- iconv(string, from = Encoding(string), to = "latin1")

[...]

PR #602, opened 2019-09-29
Detangle hashing and encoding translation

### LINKED ISSUE #600: More dictionary + hashing + encoding issues
*Sigh* - More encoding issues. 

``` r
library(vctrs)

string <- "\u00B0C"

utf <- iconv(string, from = Encoding(string), to = "UTF-8")
unknown <- iconv(string, from = Encoding(string), to = "", mark = FALSE)
latin1 <- iconv(string, from = Encoding(string), to = "latin1")

x <- list(utf = utf, unknown = unknown, latin1 = latin1)

# bad
vec_unique(x)
#> $utf
#> [1] "°C"
#> 
#> $unknown
#> [1] "°C"
#> 
#> $latin1
#> [1] "°C"

# good
vec_unique(unlist(x))
#>  utf 
#> "°C"
```

This affects _all_ functions in `dictionary.c` except `vec_match()` and `vec_in()`.

The reason this currently fails for the list is because the code path looks like:

`vec_unique` -> `init_dict_impl` -> `hash_fill` -> sees a list -> for each list element call `hash_object`

Each list element is a character vector, but when `hash_object()` eventually calls `chr_hash()` on each one, which checks if a translation is required, they can't "see" each other, so it thinks no translation is required.

I _think_ the efficient way to solve this is to detangle the hashing code from the encoding translation code (right now it is tangled in the hashing of characters). Instead, the hashing should assume that any translation has already occurred. 

We should call `obj_translate_encoding()` from either inside `init_dict_impl()`, or in each of the functions that use the dictionary, like `vec_unique()`, `vec_count()`, etc. While the latter feels repetitive, I think it is the more efficient option. `vec_match()` and `vec_in()` create dictionaries with `init_dict()` and `init_dict_partial()`, but they have already pre-translated their inputs using `obj_translate_encoding2()`. So putting `obj_translate_encoding()` into `init_dict_impl()` would be repetitive for them. 

Removing the translation code from the hashing code would solve another similar repetitive problem that currently occurs with `vec_match()`/`vec_in()`. Currently, they pre-translate their input, but then build the dictionary, which calls `hash_fill()`. Inside there, each character vector is currently checked _again_ to see if it needs translation.

This is all a bit brain melting, but I think this solution would work nicely as a long term fix.

DavisVaughan:
Closes #600 
Fixes a bug with translation dropping attributes, mentioned in #599

I think this PR is a large improvement over my previous attempt at handling encodings. Hopefully it is the last of these for awhile 😬.

---

- `hash.c` is now completely unaware of encodings. Any required translation is done ahead of time before the dictionaries are created.

- `translate.c` has been restructured in a cleaner way:
   - `obj_maybe_translate_encoding()` _potentially_ translates the character elements of a vector, if required.

   - `obj_maybe_translate_encoding2()` _potentially_ translates the character elements of two vectors relative to each other, if required.

   - `obj_translate_encoding()` _always_ translates all character elements in the vector to UTF-8. It is the low level helper called by the two functions mentioned above.

   - `obj_maybe_translate_encoding()` and `obj_maybe_translate_encoding2()` are exported as part of the api in `vctrs.h`.

   - `chr_translation_required()` has been removed from the API as the hashing functions don't need it now.

- `translate.R` is a new file, with exported forms of the translation api functions, for testing (like `hash.R`).

- `test-translate.R` is a new file that now holds most of the translation specific tests. This allows for much more focused testing.

- `dictionary.c`: Every dictionary function now calls either `obj_maybe_translate_encoding()` or `obj_maybe_translate_encoding2()` when needed. This specifically closes #600 

- `test-dictionary.R`: Any tests that had to do with encoding but were not specific to that dictionary function have been moved to `test-translate.R`. Each dictionary function still has at least 1 test for encodings, just to ensure it works.

- There was a small bug in `chr_translate_encoding()` where attributes were being dropped. That has been fixed by duplicating the input rather than constructing a new `STRSXP` vector.

---

Things I like:

- C API is cleaner with `obj_maybe_translate_encoding()` and `obj_maybe_translate_encoding2()` being the main entry points for working with encodings.

- Testing is cleaner, with many encoding tests focused in `test-translate.R`.

- Hashing is cleaner now that the encoding pieces have been removed.


hadley:
Does this have any performance implications?

DavisVaughan:
For `vec_match()` and `vec_in()`, I went back and re-tested all of the benchmarks in this PR and nothing has changed, which was to be expected. https://github.com/r-lib/vctrs/pull/565

For other dictionary functions, they now all call `obj_maybe_translate_encoding()`. The worst case I could imagine is an all character data frame that is very wide. In that case, it has to loop through every element of each column and check if any encodings are different. Performance does not look to be affected much.

``` r
vec_unique_old <- vctrs2::vec_unique
library(vctrs)
library(tibble)

x <- as.character(1:10000)
lst <- as.list(x)
names(lst) <- x
wide_chr_tbl <- as_tibble(lst)
wide_chr_tbl <- vec_slice(wide_chr_tbl, rep(1, 100))

dim(wide_chr_tbl)
#> [1]   100 10000

wide_chr_tbl[1:5, 1:5]
#> # A tibble: 5 x 5
#>   `1`   `2`   `3`   `4`   `5`  
#>   <chr> <chr> <chr> <chr> <chr>
#> 1 1     2     3     4     5    
#> 2 1     2     3     4     5    
#> 3 1     2     3     4     5    
#> 4 1     2     3     4     5    
#> 5 1     2     3     4     5

bench::mark(
  vec_unique(wide_chr_tbl),
  vec_unique_old(wide_chr_tbl),
  iterations = 100
)
#> # A tibble: 2 x 6
#> 1 vec_unique(wide_chr_tbl)      21.8ms  22.7ms      43.9     254KB     1.36
#> 2 vec_unique_old(wide_chr_tbl)  20.8ms  21.4ms      46.1     177KB     1.43
```

<sup>Created on 2019-10-08 by the [reprex package](https://reprex.tidyverse.org) (v0.2.1)</sup>

DavisVaughan:
This is probably more targeted testing. `obj_maybe_translate_encoding()` is exported on the R side so we can test it directly.

This isn't a _new_ addition of 2ms, we can just test it directly now. Previously these encoding checks were mixed in with the hashing code, so it was still happening before.

``` r
library(vctrs)
library(tibble)

obj_maybe_translate_encoding <- vctrs:::obj_maybe_translate_encoding

x <- 1:10000
lst <- as.list(x)
names(lst) <- x
wide_int_tbl <- as_tibble(lst)
wide_int_tbl <- vec_slice(wide_int_tbl, rep(1, 100))


wide_chr_tbl <- wide_int_tbl
wide_chr_tbl[] <- lapply(wide_chr_tbl, as.character)

# 10000 integer cols, 100 rows
bench::mark(
  int = obj_maybe_translate_encoding(wide_int_tbl)
)
#> # A tibble: 1 x 6
#> 1 int           188µs    201µs     4929.    78.2KB     6.71

# 10000 character cols, 100 rows
bench::mark(
  chr = obj_maybe_translate_encoding(wide_chr_tbl)
)
#> # A tibble: 1 x 6
#> 1 chr          2.75ms   2.78ms      356.    78.2KB        0
```

<sup>Created on 2019-10-08 by the [reprex package](https://reprex.tidyverse.org) (v0.2.1)</sup>

PR #603, opened 2019-10-01
expression() is a vector

### LINKED ISSUE #559: expression() should be treated as a vector?
``` r
x <- expression()
y <- expression(1, 2 + 3)
length(x)
#> [1] 0
length(y)
#> [1] 2
y[[2]]
#> 2 + 3
vctrs::vec_is(x)
#> [1] FALSE
vctrs::vec_is(y)
#> [1] FALSE
```

<sup>Created on 2019-09-03 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

Use case: {bench} package (and perhaps others) store expressions in tibbles, which now require `vec_is()` for the columns. CC @jimhester.

krlmlr:
Closes tidyverse/tibble#657.

Closes #559.

Are there any downsides to supporting `expression()`? (Do you need a NEWS entry?)


R/type-expression.R
```diff
@@ -0,0 +1,5 @@
+#' @export
+vec_proxy.expression <- function(x, ...) {
+  browser()
```
lionel-:
```suggestion
```

lionel-:
Do srcrefs need to be updated after slicing?

krlmlr:
It's likely that srcrefs need an update when slicing. Do we need `vctrs_type_expression` ?

What's a full set of tests for a base datatype? Do you have an example data type to mimic, or a documentation?

lionel-:
I don't have an example besides the way we handle other base types. Adding a new base type requires complete coverage over all vctrs primitives.

I think the first question to answer is: Is this worth it?

krlmlr:
It's a corner case we need to prioritize against other issues. Too bad S3 dispatch doesn't take place with `"expression"` .

mikmart:
I think this would fix https://github.com/tidyverse/tidyr/issues/735, too.

I installed this branch to test, but it seems `vec_unique()` doesn't work yet:
``` r
vctrs::vec_unique(expression(x))
#> Warning: S3 method 'vec_proxy.expression' was declared in NAMESPACE but not found
#> Error in vec_unique_loc(x): Internal error: Unsupported type expression in `hash_fill()`.
```

<sup>Created on 2019-11-04 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

lionel-:
@mikmart This is a related but different problems, because in that case (hashing and equality check) the expression object should be treated as a recursive structure rather than a vector or or scalar. Could you file a new issue please?

mikmart:
Thanks! I opened #643, which hopefully captures the problem.

lionel-:
I now think adding support for expression vectors might be too much work because of the srcrefs handling.

krlmlr:
I work around in tibble, let's revisit later.

PR #604, opened 2019-10-01
Throw typed conditions for out-of-bounds indexing

lionel-:
Out-of-bounds indexing now throws errors of classes:

- `vctrs_error_index_oob_position` for OOB positions.
- `vctrs_error_index_oob_name` for OOB names.

Also tweaks [message wordings](https://github.com/r-lib/vctrs/compare/master...lionel-:add-oob-error-class?expand=1#diff-05f2962e6494e58a5cc1beaa5b135846).

Issue #578, opened 2019-09-23
vec_as_index() improvements


krlmlr:
Lessons learned from tibble.

1. Should `vec_slice()` be allowed to extend the vector?

    We should keep the following invariant: `` `vec_slice<-`(x, i, value)`` is the same as `` `vec_slice<-`(x, vec_as_index(i, vec_size(x), names2(x)), value)`` . 

2. Do we need `vec_as_scalar_index()` ? This would have the same semantics as `vec_as_index()` but only support positive scalar numerics and scalar names.

3. Do we really need both the `n` and the `names` argument to `vec_as_index()` and `vec_as_position()`? Can we allow a character vector as the `n` argument?

To be continued.

lionel-:
> Do we need vec_as_index_scalar() ?

Can you elaborate please.

hadley:
`[[` has different arguments to `[`:

* A logical vector doesn't make sense
* Numeric and character vectors must have length 1
* Numeric vectors must be positive

DavisVaughan:
This can probably be closed? We have `vec_as_location2()` now, and I think having both `n` and `names` isn't that bad.

I doubt we want `vec_slice()` to be able to extend a vector, but I'm not sure

lionel-:
Yes this API should be in a better state now.

PR #605, opened 2019-10-01
Add `vec_as_position()`

lionel-:
To standardise scalar index for extraction with `[[`.

Part of issue #578.
Includes PR #604.


R/slice.R
```diff
@@ -165,6 +174,81 @@ vec_as_index <- function(i, n, names = NULL) {
...
+  }
+
+  type <- typeof(i)
+  if (!type %in% c("integer", "character")) {
+    stop_position_incompatible_type(i)
+  }
+  if (length(i) != 1L) {
+    stop_position_incompatible_size(i)
+  }
+  if (type == "integer" && i < 1L) {
```
hadley: Do we need to check for `NA_integer_` here too? Or does `vec_as_index()` already handle that?

R/slice.R
```diff
@@ -165,6 +174,81 @@ vec_as_index <- function(i, n, names = NULL) {
...
+    stop_position_incompatible_sign(i)
+  }
+
+  vec_as_index(i, n, names = names)
+}
+
+stop_position_incompatible_type <- function(i) {
```
hadley: Do we really want one error class for each of this? It feels very fine grained

hadley: Thinking about the error messages:

```R
Error: Invalid type in argument `i`
* `i` is size 2
* `i` must be size 1

Error: Invalid type in argument `i`
* `i` is a logical vector
* `i` must be a character or integer vector

Error: Invalid value in argument `i`
* `i` is -1
* `i` must be greater than 0
```

I'm not sure the distinction between value and type is sufficient to require a different class.

lionel-: I agree it feels too fine-grained, but to merge these error classes we need:

* A concept that includes both type and size. I like the idea of working around this with the notion of invalid value.

* Perform the type checking again as part of the condition-message method. While a bit redundant, this also means the `stop_` constructor is useful in more cases. For instance in the case of vectorised errors, it can be called from a fast C routine that doesn't keep track of the bad values.

I will revisit.

What do you think about using `bad` instead of `invalid`? It makes for shorter error classes and messages.

lionel-: oh or were you suggesting making size errors a subset of type errors?

R/slice.R
lionel-: oops, I will make these OOB errors as we discussed on Slack.

lionel-:
I have reduced the error hierarchy to

- `vctrs_error_position_bad_type` for type and size errors
- `vctrs_error_position_bad_value` for sign and `NA` errors

lionel-:
One argument for merging the type and value classes is that "existing positive number || existing name" is the type of a position. A bad value is a value that doesn't conform to the type, so the current distinction between `vctrs_error_position_bad_type` and `vctrs_error_position_bad_value` is a bit artificial. I think you're right they should be merged.

R/slice.R
```diff
@@ -125,12 +125,19 @@ vec_assign_fallback <- function(x, i, value) {
...
 #'
-#' This provides a means of standardizing common indexing methods such as
-#' integer, character or logical indexing. The input is a vector of one
-#' of these three types, and the output is always an integer vector that can
-#' be used for subsetting.
+#' These helpers provide a means of standardizing common indexing
+#' methods such as integer, character or logical indexing.
+#'
+#' * `vec_as_index()` accepts integer, character, or logical vectors
+#'   of any sizes. The output is always an integer vector of the same
```
hadley:
```suggestion
#'   of any size. The output is always an integer vector of the same
```

hadley: The output isn't always the same size?

R/slice.R
```diff
@@ -125,12 +125,19 @@ vec_assign_fallback <- function(x, i, value) {
...
+#'   size that can be used for subsetting. This is suitable for
+#'   indexing with `[` or [vec_slice()].
```
hadley:
```suggestion
#'   subsetting with `[` or [vec_slice()].
```
?

R/slice.R
```diff
@@ -165,6 +174,93 @@ vec_as_index <- function(i, n, names = NULL) {
   vec_assert(n, integer(), 1L)
   .Call(vctrs_as_index, i, n, names)
 }
+#' @rdname vec_as_index
+#' @export
+vec_as_position <- function(i, n, names = NULL) {
+  if (is.object(i)) {
```
hadley: Is this too flexible? I'd want `vec_as_position(factor("a"))` to be an error. At least I think you should try coercing to character first.

R/slice.R
hadley: I was thinking something like:

```R
stop_bad_argument("i", vector_of_problems)
```

But I don't see how to move the logic into the `conditionMessage()` method

R/slice.R
lionel-: Either we restrict indices to be non-S3, or we coerce to int || chr. Factor indices come up because of stringsAsFactors, I got a request in tidyselect to make it work, and these are the same semantics that we discussed about over there.

I don't think the order matters because an S3 vector that is coercible to integer shouldn't be coercible to character. And for the rare case that it is, I think we should start with the most specific type integer before trying the least specific.

R/slice.R
lionel-: oh of course

lionel-: This is now:

```
#'   [...] The output is always an integer vector that is
#'   suitable for subsetting with `[` or [vec_slice()]. It might be a
#'   different size than the input because negative selections are
#'   transformed to positive ones and logical vectors are transformed
#'   to a vector of indices for the `TRUE` positions.
```

R/slice.R
lionel-: I wondered about that as well, maybe we could come up with a new generic that generates error bullets. Then the problems need to be wrapped up in objects, and probably error conditions are most suitable. The condition-message method for these conditions would invoke the bullet generic as well.

R/slice.R
hadley: The current behaviour gives exactly the wrong behaviour for factors. I don’t think any reasoning from tidyselect would apply here as this is a much lower level tool.

lionel-: What is the wrong behaviour for factors?

hadley: `list(a = 1, b = 2)[[factor(“b”)]]`

lionel-: This is with base R rules for coercion. I think it makes sense to treat factors as character when indexing, which is what is implemented in this PR based on vctrs coercion rules.

Tidyselect may be high level, but selecting positions is a low level task. The behaviour should be consistent with vctrs (and so it should become an error if we disallow S3 indices in vctrs).

hadley: Ok, I get it now.

R/slice.R
```diff
@@ -165,6 +174,93 @@ vec_as_index <- function(i, n, names = NULL) {
...
+    if (vec_is_coercible(i, int())) {
+      i <- vec_cast(i, int())
+    } else if (vec_is_coercible(i, chr())) {
+      i <- vec_cast(i, chr())
+    }
```
hadley: If an object isn't coercible to integer or character, I think we should error here, rather than using the underlying base type.

R/slice.R
```diff
@@ -165,6 +174,93 @@ vec_as_index <- function(i, n, names = NULL) {
   vec_assert(n, integer(), 1L)
   .Call(vctrs_as_index, i, n, names)
 }
+#' @rdname vec_as_index
+#' @export
+vec_as_position <- function(i, n, names = NULL) {
+  if (is.object(i)) {
+    if (vec_is_coercible(i, int())) {
```
hadley: I thought we were moving away from these abbreviations?

R/slice.R
```diff
@@ -165,6 +174,93 @@ vec_as_index <- function(i, n, names = NULL) {
...
+      i <- vec_cast(i, int())
+    } else if (vec_is_coercible(i, chr())) {
+      i <- vec_cast(i, chr())
+    }
+  } else if (typeof(i) == "double") {
```
hadley: `rlang::is_double()`?

R/slice.R
lionel-: oops, agreed!

R/slice.R
lionel-: I assumed from previous discussions that we were going to move them from rlang to vctrs?

lionel-:
* Drafted `vec_is_subtype()` to figure out direction of nodes in the subtyping graph.

* Positions can no longer be subtypes of logical.

* Improved error messages. They now all use the same leading sentence and have two i and x bullets:

```
> vec_as_position(na_int, 2L)
Error: Must extract with a single position or name.
ℹ Positions and names can't be missing.
✖ `i` can't be `NA`.
```

```
> vec_as_position(0, 2L)
Error: Must extract with a single position or name.
ℹ Positions must be positive integers.
✖ `i` can't be zero.
```

```
> vec_as_position(-1, 2L)
Error: Must extract with a single position or name.
ℹ Positions must be positive integers.
✖ `i` has the wrong sign: -1.
```

```
> vec_as_position(1:2, 2L)
Error: Must extract with a single position or name.
ℹ Positions and names must be size 1.
✖ `i` has the wrong size `2`.
```

```
> vec_as_position(TRUE, 10L)
Error: Must extract with a single position or name.
ℹ Positions and names must be integer or character.
✖ `i` has the wrong type `logical`.
```

R/slice.R
```diff
@@ -165,6 +176,96 @@ vec_as_index <- function(i, n, names = NULL) {
...
+  if (i < 1L) {
+    return(glue_lines(
+      lead,
+      glue_error_bullets(
+        i = "Positions must be positive integers.",
+        x =
+          if (i == 0L) {
+            "`{arg}` can't be zero."
+          } else {
+            "`{arg}` has the wrong sign: {i}."
```
DavisVaughan: Are you trying to use `:` whenever you say `has the wrong ___:`? If so, it is missing for `has the wrong type` and `has the wrong size`.

lionel-: This branch is different because it shows the whole value, which is why I used `:`

lionel-: Other ideas:

```
✖ `i` (with value -1) has the wrong sign.
✖ `i` has the wrong sign. Its value is -1.
```

PR #608, opened 2019-10-09
Subclass position type errors from index type errors

lionel-:
This makes bad position type errors a subclass of bad index type errors.

- Experiments with a sentinel-based approach. Instead of throwing when type is not right, we return an error object as sentinel. From this we derive predicate functions `vec_is_position()` and `vec_is_index()`, and coercion functions `vec_coerce_position()` and `vec_coerce_index()`.  Only the latter throw type errors.

- Index type errors have slightly different error messages:

```r
vec_as_position(foobar(), 2L)
#> Error: Must extract with a single position or name.
#> x `i` has the wrong type `vctrs_foobar`.
#> i Positions and names must be integer or character.

vec_as_index(foobar(), 2L)
#> Error: Must subset with positions or names.
#> x `i` has the wrong type `vctrs_foobar`.
#> i Positions and names must be integer, logical, or character.
```

Currently the types are:

- `vctrs_error_index_bad_type`
- `vctrs_error_position_bad_type`

I think this is cleaner to read but maybe we should make it a rule that error class hierarchies should be reflected in the class names, no matter the length? Then since positions are a subset of index, the class should probably become `vctrs_error_index_position_bad_type`.


R/slice.R
```diff
@@ -174,62 +174,129 @@ vec_assign_fallback <- function(x, i, value) {
 #' @export
 vec_as_index <- function(i, n, names = NULL) {
```
hadley: Not related to this PR, but coming back to it after a few days and I can't remember which way `as_index` and `as_position` are mapped to `[[` and `[`. Maybe it would better to use "extract" and "subset" like in the error messages?

R/utils.R
```diff
@@ -85,3 +85,26 @@ paste_line <- function (...) {
 has_dim <- function(x) {
   !is.null(attr(x, "dim"))
 }
+
+maybe <- function(value = NULL, error = NULL) {
```
hadley: This feels more like a Rust Result than a Haskell Maybe

hadley: Can we mark this clearly as experimental? I don't think we should start using it widely, even internally, until we're a bit more confident in the pattern.

R/slice.R
```diff
@@ -174,62 +174,129 @@ vec_assign_fallback <- function(x, i, value) {
...
 
 #' @export
-conditionMessage.vctrs_error_position_bad_type <- function(c) {
+conditionMessage.vctrs_error_index_bad_type <- function(c) {
   i <- c$i
   arg <- c$arg %||% "i"
+  issue <- "Must subset with positions or names."
 
-  lead <- "Must extract with a single position or name."
+  if (is.object(i) || !typeof(i) %in% c("integer", "character", "logical")) {
```
hadley: It feels wrong to me to have to repeat this logic in the error message generator. If you we have to  generate this message lazily, maybe it's better to create a closure in `vec_maybe_coerce_index()`?

tests/testthat/out/error-index-type.txt
```diff
@@ -0,0 +1,24 @@
+> vec_as_index(mtcars, 10L)
+Error: Must subset with positions or names.
+x `i` has the wrong type `data.frame<
```
hadley: It occurs to me that you only really need to show the details of the data frame if the expected type is another data frame. i.e. we only need to conditionally expand the details of the type.

```
vec_assert(data.frame(), integer())
#> Expected <integer> got <data.frame>

vec_assert(data.frame(x = integer()), data.frame(y = integer())
#> Expected <data.frame<x: integer>> got <data.frame<y: integer>>
```

We don't need to do anything about that now, but it's worth considering that how we display the actual type might depend on the expected type.

R/utils.R
lionel-: Good point, a Haskell maybe is just an option type with `NULL` as fallback value.

I'll mark it more clearly as experimental, I'm not sure I like this pattern either, compared to throwing and catching. It does feel boilerplaty. I guess try-catch is boilerplaty as well though.

tests/testthat/out/error-index-type.txt
lionel-: Great idea, I have opened an issue: https://github.com/r-lib/vctrs/issues/609

R/slice.R
lionel-: I like giving names to scalar types, like logical is a vector of boolean, integer is a vector of whole numbers, double is a vector of numbers, data frames are a vector of rows or of columns, lists are a vector of objects.

And then index vectors contain positions. Does this make sense?

lionel-: Or more precisely they contain `scalar_type` | `NA`.

R/utils.R
lionel-: I'll rename in a future PR to avoid conflicts with other work.

R/slice.R
lionel-: That's a great idea, I'll try that next, outside of this PR.

PR #519, opened 2019-07-31
Generalize `vec_split_along()` to implement `vec_split()` in C

DavisVaughan:
I've generalized `vec_split_along()` to be more useful. Rather than just mimicking a faster version of:

```
indices <- as.list(vec_seq_along(x))
map(indices, vec_slice, x = x)
```

It can now take an arbitrary list of integer `indices`. The signature is now `vec_split_along(x, indices = NULL)` with `NULL` defaulting to the original behavior. 

The reason this is useful is for using it in `vec_split()`. Currently there is a call to:

```
x_split <- map(out$id, vec_slice, x = x)
```

This can be completely replaced with the new version of `vec_split_along()`, and then a C version of `vec_split()` can be written.

Below are a large number of benchmarks. The tl;dr is that the new versions of `vec_split_along()` and `vec_split()` perform much better than their older alternatives in most cases, and never perform worse. Vectors have a much more dramatic performance boost than data frames. Generally, the more unique values you have to split by, the bigger the performance gap.

`vec_split_along()` benchmarks:

<details>

``` r
library(nycflights13)
library(purrr)
library(vctrs)

vec_split_along <- vctrs:::vec_split_along

x <- flights[1:3]

# `vec_split_along()` generally outperforms repeated `vec_slice()`ing
idx <- rep_len(list(1:50), 10000)
bench::mark(
  vec_split_along(x, idx),
  map(idx, vec_slice, x = x),
  iterations = 100
)
#> Warning: Some expressions had a GC in every iteration; so filtering is
#> disabled.
#> # A tibble: 2 x 6
#>   expression                      min   median `itr/sec` mem_alloc `gc/sec`
#>   <bch:expr>                 <bch:tm> <bch:tm>     <dbl> <bch:byt>    <dbl>
#> 1 vec_split_along(x, idx)      12.4ms   23.5ms     35.7     7.17MB     14.3
#> 2 map(idx, vec_slice, x = x)   66.2ms  104.6ms      9.20    7.22MB     24.1

# take a long factor and repeatedly subset the first 1000 values
z <- factor(rep(letters, times = 1000))
z_idx <- rep_len(list(1:1000), 1000)

bench::mark(
  vec_split_along(z, z_idx),
  map(z_idx, vec_slice, x = z)
)
#> # A tibble: 2 x 6
#>   expression                        min median `itr/sec` mem_alloc `gc/sec`
#>   <bch:expr>                   <bch:tm> <bch:>     <dbl> <bch:byt>    <dbl>
[...]

PR #612, opened 2019-10-11
Generalize `vec_split_along()`

DavisVaughan:
This PR replaces #519 as a better generalization of `vec_split_along()`.

`vec_split_along()` is currently an optimized:

```r
map(vec_seq_along(x), vec_slice, x = x)
```

With this PR, it becomes an optimized 

```r
map(indices, vec_slice, x = x)
``` 

where `indices` is a list of index vectors. The default of `indices = NULL` would be equivalent to how it works now.

This pattern has begun appearing in the dplyr 0.9.0 rewrite, for example, it is how group chunks are sliced out: https://github.com/tidyverse/dplyr/blob/16da75a2a043c44e6e2a6e9aac411f75dfb9baf7/R/tbl-df.r#L403

It also appears in the current `vec_split()`. https://github.com/r-lib/vctrs/blob/6c46673acaf9ad3db87dc50aee6cdcf13bdfacbf/R/dictionary.R#L264

---

- `vec_split()` has been rewritten in C now that we have the `indices` feature in `vec_split_along()`.

- In the rewrite of `vec_split_along()` I've added a `vctrs_split_info` struct that greatly reduces the amount of repeated code that was in each `split_along_*()` helper.

- The commit history isn't the best because at first I separated the split functions into their own `split.R` and `split.c` files. That was too much for this PR, so I walked it back in the end.

- To review the changes in `split.c`, I think it would be easiest to just review the changes to `vec_split_along()` as if they were completely new. It is essentially a full rewrite.

- One notable thing. `split_along_df()` used to use `df_slice()`, but this was quite slow because of the repeated proxy-ing of every column each time `vec_slice_impl()` was called. It now takes an inverted approach, where it goes over each column, proxies that column once, and completely splits it according to the `indices`. This greatly improved performance.

- ~A second PR will come that will restructure the files to combine the 3 splitting functions, `vec_split()`, `vec_split_id()` and `vec_split_along()` into their own `split.R` and `split.c` files. I will also export `vec_split_along()` there and combine them all into 1 Rd file. That is also where I will add a NEWS bullet.~

---

Here are two benchmarks regarding how the data frame splitting was inverted. There is a minor improvement with a few columns.

```r
df <- data.frame(x = rep(1, times = 20000))

bench::mark(
  vec_split_along(df),
  vec_split_along_old(df),
  iterations = 100
)
#> # A tibble: 2 x 6
#>   expression                   min   median `itr/sec` mem_alloc `gc/sec`
#>   <bch:expr>              <bch:tm> <bch:tm>     <dbl> <bch:byt>    <dbl>
#> 1 vec_split_along(df)       9.38ms   16.4ms      63.7     316KB     67.6
#> 2 vec_split_along_old(df)  12.12ms   20.9ms      50.9     160KB     48.9
```

But the difference is more dramatic with many columns.

```r
df_wide <- vec_cbind(!!! rep_len(list(df), 20))
dim(df_wide)
#> [1] 20000    20

bench::mark(
  vec_split_along(df_wide),
  vec_split_along_old(df_wide),
  iterations = 100
)
#> # A tibble: 2 x 6
#> 1 vec_split_along(df_wide)      33.7ms  46.3ms      20.0    7.17MB     52.1
#> 2 vec_split_along_old(df_wide)  66.8ms  81.5ms      12.3    4.12MB     84.1
```

Splitting a long vector by 1000 groups

```
x <- rep(1, times = 200000)

seq_along <- vec_seq_along(x)
indices <- vec_split(seq_along, cut(seq_along, breaks = 1000, labels = FALSE))$val

bench::mark(
  vec_split_along(x, indices),
  map(indices, vec_slice, x = x)
)
#> # A tibble: 2 x 6
#>   expression                          min   median `itr/sec` mem_alloc
#>   <bch:expr>                     <bch:tm> <bch:tm>     <dbl> <bch:byt>
#> 1 vec_split_along(x, indices)    514.24µs 626.84µs     1256.     1.6MB
#> 2 map(indices, vec_slice, x = x)   1.86ms   2.23ms      441.    1.63MB
```

Splitting a wide data frame into 1000 groups (`group_by()` ish)

```r
seq_along <- vec_seq_along(df_wide)
indices <- vec_split(seq_along, cut(seq_along, breaks = 1000, labels = FALSE))$val

bench::mark(
  vec_split_along(df_wide, indices),
  map(indices, vec_slice, x = df_wide),
  iterations = 20
)
#> # A tibble: 2 x 6
#>   expression                              min  median `itr/sec` mem_alloc
#>   <bch:expr>                           <bch:> <bch:t>     <dbl> <bch:byt>
#> 1 vec_split_along(df_wide, indices)    4.79ms  5.21ms     193.      4.5MB
#> 2 map(indices, vec_slice, x = df_wide) 9.89ms 12.09ms      83.9    4.23MB
```

Here is a benchmark of `vec_split()` being faster now.

```r
bench::mark(
  vec_split(mtcars, mtcars$mpg),
  vec_split_old(mtcars, mtcars$mpg)
)
#> # A tibble: 2 x 6
#>   expression                            min  median `itr/sec` mem_alloc
#>   <bch:expr>                        <bch:t> <bch:t>     <dbl> <bch:byt>
#> 1 vec_split(mtcars, mtcars$mpg)      58.6µs  74.6µs    13270.    34.5KB
#> 2 vec_split_old(mtcars, mtcars$mpg) 170.5µs 213.5µs     4705.    81.3KB
```


R/slice.R
```diff
@@ -304,8 +304,8 @@ vec_init <- function(x, n = 1L) {
 }
 
 # Used internally by `vec_rbind()`, but exported for testing
-vec_split_along <- function(x) {
-  .Call(vctrs_split_along, x)
+vec_split_along <- function(x, indices = NULL) {
```
hadley: This doesn't seem like the right name now — maybe `vec_split_by()`? or `vec_split_with()`? Might be best to leave `vec_split_along()` and add a new function

lionel-:
`id` has two disadvantages:

- It is 2 letters long whereas all other names are 3 letters
- More importantly, it makes me think of "identity" instead of "index"

However it has a precedent: `vec_duplicate_id()`. @hadley I'm not sure I understand the different suffixes of `vec_duplicate_id()` and `vec_unique_loc()`. Should we standardise?

I'd favour a 3 letters particle. I think I prefer `idx` over `loc`, but the latter would require fewer changes.

DavisVaughan:
Would you change `vec_split_id -> vec_split_idx` and have its return value column be `id -> idx`? It is a new function so we can still change that.

DavisVaughan:
Hmm, I've gotta say, now that it's implemented I'm not sure that it's worth it to return data frames for `vec_split_along()` and `vec_split_with()`. With `vec_split()` and `vec_split_id()`, both columns are very meaningful for the final result. Here, I think most users (or package developers) will really just want the result of splitting `x`, even if it seems less predictable. There isn't any new information in the `idx` column, and I don't see it being that useful in any downstream activities.

I think that `vec_split_along()` and `vec_split_with()` could both just return a `list_of<vec_ptype(x)>` instead.

__To be less confusing, would it make sense to call them `vec_slice_along()` and `vec_slice_with()`? Then you'd never be confused by the fact that `vec_split()` and `vec_split_id()` return data frames. This is also pretty evocative of what the functions do, they mimic some form of `map(indices, vec_slice, x = x)`.__

As a final piece of motivation for just returning a list, in the two places in vctrs where we use these new functions internally, I use the `_impl` versions that just return a list because I don't need the full data frame. And in the dplyr 0.9.0 branch, it would also just use the list, with no need for the `idx` column.

lionel-:
vec-slice would be confusing as well because I'd expect it to always return the type of the input.

I still think vec-split functions would be confusing if they don't all return the same sort of output.

DavisVaughan:
@hadley, 

@lionel- and I have settled on the following, but wanted to see if you agreed before moving forward:

- Rename `vec_split_with(x, indices)` -> `vec_chop(x, indices)`

- Remove `vec_split_along(x)` in favor of `vec_chop(x, indices = NULL)`, which would be the default. The combination of `vec_split_along()` and `vec_split_with()` into one function is what you argued against previously, but with this new name it seems to make more sense to have them combined.

- `vec_chop()` would return a `list_of<vec_ptype(x)>`. The different name means it won't get compared with `vec_split()`'s return value.

There is `tidyr::chop()` now, but I think there are some parallels between the two ideas so they should not clash.

hadley:
That proposal sounds reasonable to me. `indices` will always be an integer vector, right?

DavisVaughan:
A list of integer vectors, i.e. a list of `index` vectors that each get run through `vec_as_index()` to ensure validity

hadley:
Oh, yeah, makes sense.

DavisVaughan:
Alright, we now have `vec_chop()` returning a `list_of<vec_ptype(x)>` in place of both `vec_split_along()` and `vec_split_with()`. 

I've gone ahead and exported it in this PR with a NEWS bullet since the name has changed and it wouldn't make sense to group it with `vec_split()` / `vec_split_id()` now.

lionel-:
Looks good!

PR #616, opened 2019-10-15
Make `vec_as_index()` and `vec_as_position()` more flexible

lionel-:
Requires #615.

Tidyselect needed a bit more flexibility to use the vctrs tools for index vectors and positions:

1. Negative positions should be allowed in both as-index and as-position because they have a special meaning in `vars_pull()` and `vars_select()`. `vec_as_index()` now has `convert_negative = TRUE` and `vec_as_position()` has `allow_negative = FALSE`.

2. I needed a way to convert inputs to the set of allowed base types for as-position (integer and character). `vec_coerce_position()` is now exported to perform this task. It is in charge of converting S3 inputs and checking the base type, e.g. it throws with logical vectors.

Though it turned out not needed in tidyselect I also added `allow_missing` in `vec_as_position()`. It is equivalent to the parameter we have discussed adding in `is_bool()` and `is_string()`, and is connected to the issue of type-checking allowed scalar values in a given vector.

I'm not in love with this interface, so it's marked as experimental.


R/slice.R
```diff
@@ -151,6 +151,8 @@ vec_assign_fallback <- function(x, i, value) {
 #' @param arg The argument name to be displayed in error messages when
 #'   `vec_as_index()` and `vec_as_position()` are used to check the
 #'   type of a function input.
+#' @param convert_negative Should negative indices be converted to
```
DavisVaughan: Mark as Experimental as well?

lionel-:
Consolidated arguments into:

- `vec_as_index()`: `convert_values` taking `"negative"`

- `vec_as_position()`: `allow_values` taking `"negative"` or `"missing"`

- `vec_as_index()` and `vec_as_position()`: `allow_types` taking `"indicator"`, `"position"`, or `"name"`

These options are validated by `new_opts()`.

lionel-:
With `allow_types` taking all or any of `c("indicator", "position", "name")`, the error message is now generated rather than static:

- Only positions: `i These indices must be positions.`
- Only names: `These indices must be names.`
- Positions and names: `i These indices must be positions or names.`
- Any type: `i These indices must be indicators, positions or names.`

I think this is fine as long as the generation can be done at build time (no cross-package extensibility of error messages). If we ever want to translate these messages, we can come up with an API to customise the generation function per language.

PR #627, opened 2019-10-21
Implement `vec_group_id()` and `vec_group_rle()`

DavisVaughan:
After thinking about `vec_group()`, I think it's actually incredibly straightforward. @hadley if you think this is worth a standalone function then I can finish it out with better docs and tests. (it looks like this would cut `group_sum()` almost in half since most of the time is in `vec_group()`)

I guess it would be nice to return the number of groups, which is there in `d->used`, but I don't have a good data structure to return that info in. Maybe a two column data frame with 1 row? Col 1 is an integer column `n` and col 2 is a `list_of<int>` for `group`. Or something like that.

``` r
library(vctrs)

vec_group2 <- function(x) {
  unique <- vec_unique(x)
  id <- vec_match(x, unique)
  id
}

g1 <- rep(1:2000, length = 1e7)

bench::mark(
  vec_group(g1),
  vec_group2(g1),
  iterations = 50
)
#> # A tibble: 2 x 6
#>   expression          min   median `itr/sec` mem_alloc `gc/sec`
#>   <bch:expr>     <bch:tm> <bch:tm>     <dbl> <bch:byt>    <dbl>
#> 1 vec_group(g1)     147ms    150ms      6.60     140MB     50.0
#> 2 vec_group2(g1)    289ms    293ms      3.41     179MB     33.0
```

<sup>Created on 2019-10-21 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0.9000)</sup>

``` r
library(vctrs)

x <- mtcars[c("vs", "am")][1:5,]
x
#>                   vs am
#> Mazda RX4          0  1
#> Mazda RX4 Wag      0  1
#> Datsun 710         1  1
#> Hornet 4 Drive     1  0
#> Hornet Sportabout  0  0

vec_group(x)
#> [1] 1 1 2 3 4
```

<sup>Created on 2019-10-21 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0.9000)</sup>


hadley:
I think this would be useful to export, and I _think_ the right place for `n` is just as an attribute.

DavisVaughan:
I just realized how close this is to `vec_duplicate_id()`, but rather than giving the location of the first duplicate value it gives the identifier of the group it falls in, in the order that they appear.

And for `vec_group_id()` these truly are "identifiers" so I think the name works. With `vec_duplicate_id()` returning positional values, I do think it should eventually be renamed to `vec_duplicate_idx()` if possible. Same with `vec_unique_loc()`.

```r
> x <- c(1, 1, 2, 5, 6, 10, 1)
> vec_duplicate_id(x)
[1] 1 1 3 4 5 6 1
> vec_group_id(x)
[1] 1 1 2 3 4 5 1
attr(,"n")
[1] 5
```

hadley:
Sounds good!

lionel-:
The downside of an attribute is that we need to provide a getter function. Maybe the 1-row data frame is more consistent and easier to work with? It seems a bit weird to return a data frame instead of a list in that case, but I wonder if we would get used to it and like it eventually.

DavisVaughan:
It's a bit frustrating to get at the indices

``` r
groups <- tibble::tibble(
  n = 2L,
  id = list(c(1L, 2L, 1L))
)

groups$n
#> [1] 2

groups$id
#> [[1]]
#> [1] 1 2 1

groups$id[[1]]
#> [1] 1 2 1
```

<sup>Created on 2019-10-22 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0.9000)</sup>

hadley:
I don't see why we need to provide an accessor — I think we should just mark this API as experimental, and move on. It's not important enough to spend a bunch of time discussing right now.

For example, I am already wondering if it the output should be run length encoded; I think it'll be a little faster (since the output will be smaller, and in the common case where the groups are roughly contiguous, you'll only need to do an equality check with the previous value rather than a dictionary lookup).

DavisVaughan:
Alright, marked as experimental. Asking for review.

hadley:
For the "easy" case (where groups are contiguous) `vec_group_rle()` has a nice performance boost, although it's slower in the worst case:

``` r
library(vctrs)

vec_group2 <- function(x) {
  unique <- vec_unique(x)
  id <- vec_match(x, unique)
  id
}

g1 <- rep(1:2000, length = 1e7) # hard
g2 <- rep(1:2000, each = 1e7 / 2000) # easy

bench::mark(
  vec_group2(g1),
  vec_group_id(g1),
  vec_group_rle(g1),
  vec_group2(g2),
  vec_group_id(g2),
  vec_group_rle(g2),
  check = FALSE
)
#> disabled.
#> # A tibble: 6 x 6
#> 1 vec_group2(g1)       590ms    590ms      1.69     179MB     1.69
#> 2 vec_group_id(g1)     200ms    206ms      4.86     140MB     3.24
#> 3 vec_group_rle(g1)    384ms    390ms      2.57     178MB     2.57
#> 4 vec_group2(g2)       286ms    287ms      3.48     182MB     3.48
#> 5 vec_group_id(g2)     136ms    142ms      7.01     140MB     7.01
#> 6 vec_group_rle(g2)    120ms    127ms      7.43     178MB     7.43
```

<sup>Created on 2019-10-22 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

DavisVaughan:
There's a bug in there somewhere. This isn't right at the end.

``` r
vctrs::vec_group_rle(c(1, 1, 2, 3, 4, 3, 4))
#> $group
#> [1] 1 2 3 4 4 4
#> 
#> $length
#> [1] 2 1 1 1 1 1
#> 
#> attr(,"n")
#> [1] 4
```

<sup>Created on 2019-10-22 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0.9000)</sup>

DavisVaughan:
Hmm, I've fixed the bug. The `key` position is an index into `x`, not `g`. Another vector was required to correctly map into `g`. This hurts performance somewhat.

``` r
library(vctrs)

vec_group2 <- function(x) {
  unique <- vec_unique(x)
  id <- vec_match(x, unique)
  id
}

g1 <- rep(1:2000, length = 1e7) # hard
g2 <- rep(1:2000, each = 1e7 / 2000) # easy

bench::mark(
  vec_group2(g1),
  vec_group_id(g1),
  vec_group_rle(g1),
  vec_group2(g2),
  vec_group_id(g2),
  vec_group_rle(g2),
  check = FALSE,
  iterations = 30
)
#> disabled.
#> # A tibble: 6 x 6
#> 1 vec_group2(g1)       272ms    308ms      3.25     179MB     2.39
#> 2 vec_group_id(g1)   134.2ms    164ms      6.08     140MB     3.85
#> 3 vec_group_rle(g1)  261.5ms    283ms      3.49     242MB     3.61
#> 4 vec_group2(g2)     198.5ms    237ms      4.24     182MB     2.97
#> 5 vec_group_id(g2)    85.3ms    108ms      9.30     140MB     6.20
#> 6 vec_group_rle(g2)   99.9ms    117ms      8.51     242MB     8.51
```

<sup>Created on 2019-10-22 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0.9000)</sup>

R/dictionary.R
```diff
@@ -71,6 +71,76 @@ reset_rownames <- function(x) {
...
+#'   groups is also returned as an attribute, `n`.
+#'
+#' * `vec_group_rle()` locates groups in `x` and returns them run length
+#'   encoded in the order that they appear. The return value is a rcrd object
+#'   with fields for the `group` identifiers and the run `length` of the
+#'   corresponding group. The number of groups is also returned as an
+#'   attribute, `n`.
+#'
+#' @param x A vector
+#' @return
```
hadley: Cross reference to `vec_split_id()`? Or are you imagining that we'll rename that to `vec_group_something()`?

DavisVaughan: `vec_group_pos()`

R/dictionary.R
```diff
@@ -71,6 +71,89 @@ reset_rownames <- function(x) {
   x
 }
 
+# Groups -----------------------------------------------------------------
```
lionel-: Can you move this in group.R please?

lionel-: Same for the C code and unit tests, I think.

PR #628, opened 2019-10-22
Improve `vec_split_id()`

DavisVaughan:
Working on `vec_group_id()` gave me some insight into a better implementation of `vec_split_id()` that uses the `vec_group_id()` approach of constructing group identifiers. It uses less memory and is faster. It also feels more readable now.

``` r
library(vctrs)
vec_split_id_old <- vctrs2:::vec_split_id
```

2000 groups, 10 mil elements

```r
g1 <- rep(1:2000, length = 1e7)

bench::mark(
  vec_split_id(g1),
  vec_split_id_old(g1),
  iterations = 30
)
#> disabled.
#> # A tibble: 2 x 6
#> 1 vec_split_id(g1)        257ms    295ms      3.37     179MB     3.25
#> 2 vec_split_id_old(g1)    379ms    400ms      2.46     307MB     2.54
```

20000 groups, 10 mil elements

```r
g2 <- rep(1:20000, length = 1e7)

bench::mark(
  vec_split_id(g2),
  vec_split_id_old(g2),
  iterations = 20
)
#> disabled.
#> # A tibble: 2 x 6
#> 1 vec_split_id(g2)        341ms    390ms      2.56     180MB     2.81
#> 2 vec_split_id_old(g2)    617ms    713ms      1.41     308MB     2.82
```

1 million rows, 26000 groups

```r
set.seed(123)

# 1 million rows, ~26000 groups
df <- data.frame(
  x = sample(1000, 1000000, replace = TRUE),
  y = sample(letters, 1000000, replace = TRUE)
)

vec_unique_count(df)
#> [1] 26000

bench::mark(
  vec_split_id(df),
  vec_split_id_old(df),
  iterations = 40
)
#> # A tibble: 2 x 6
#> 1 vec_split_id(df)        142ms    161ms      6.02    20.7MB    0.488
#> 2 vec_split_id_old(df)    174ms    217ms      4.72    36.6MB    0.525
```

### DIFF
--- a/src/dictionary.c
+++ b/src/dictionary.c
@@ -452,10 +452,10 @@ SEXP vec_split_id(SEXP x) {
 
   int n_groups = d.used;
 
-  // Location of first occurence of each group in `x`
-  SEXP key_idx = PROTECT_N(Rf_allocVector(INTSXP, n_groups), &nprot);
-  int* p_key_idx = INTEGER(key_idx);
-  int key_id_current = 0;
+  // Position of first occurence of each group in `x`
+  SEXP key_pos = PROTECT_N(Rf_allocVector(INTSXP, n_groups), &nprot);
+  int* p_key_pos = INTEGER(key_pos);
+  int key_pos_current = 0;
 
   // Count of the number of elements in each group
   SEXP counts = PROTECT_N(Rf_allocVector(INTSXP, n_groups), &nprot);
@@ -465,9 +465,9 @@ SEXP vec_split_id(SEXP x) {
   for (int i = 0; i < n; ++i) {
     int group = p_groups[i];
 
-    if (group == key_id_current) {
-      p_key_idx[key_id_current] = i + 1;
-      key_id_current++;
+    if (group == key_pos_current) {
+      p_key_pos[key_pos_current] = i + 1;
+      key_pos_current++;
     }
 
     p_counts[group]++;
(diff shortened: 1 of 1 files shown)


src/dictionary.c
```diff
@@ -430,66 +430,71 @@ SEXP vctrs_split_id(SEXP x) {
...
+    int32_t hash = dict_hash_scalar(&d, i);
+    R_len_t key = d.key[hash];
 
-    if (d.key[hash] == DICT_EMPTY) {
-      p_tracker[hash] = d.used;
+    if (key == DICT_EMPTY) {
       dict_put(&d, hash, i);
-      p_count[hash] = 0;
+      p_groups[i] = g;
+      g++;
```
lionel-: Nitpick: prefix increment.

src/dictionary.c
```diff
@@ -431,66 +431,71 @@ SEXP vec_split_id(SEXP x) {
...
 
-    SET_VECTOR_ELT(out_id, p_tracker[hash], Rf_allocVector(INTSXP, p_count[hash]));
+    p_counts[group]++;
   }
 
-  // Fill index container and key locations
-  for (int i = 0; i < n; ++i) {
-    int j = p_out_pos[i];
-    int hash = p_counters[j];
+  SEXP out_id = PROTECT_N(Rf_allocVector(VECSXP, n_groups), &nprot);
```
lionel-: Should this be `out_pos`? We record the positions of each group elements, and `vec_split_id()` will be renamed to `vec_split_pos()` right?

DavisVaughan: Yea I was going to delay until we change the name, but I will go ahead and do it now so we don't forget. We will just have to remember to change the column name later. (Also, `vec_split_id` -> `vec_group_pos`)

PR #631, opened 2019-10-22
Odd rlang text output

DavisVaughan:
@lionel- did something change recently with rlang output for these error bullets? I don't know why the output has changed with rlang 0.4.1, but it doesn't look right here and it is causing other builds to fail on travis, which tries to install the most recent rlang where things have changed from their previous output

### DIFF
--- a/tests/testthat/out/error-position-na.txt
+++ b/tests/testthat/out/error-position-na.txt
@@ -1,12 +1,12 @@
 > vec_as_position(na_int, 2L)
 Error: Must extract with a single index.
-x `i` can't be `NA`.
-i This index can't be missing.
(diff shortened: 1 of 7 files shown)


tests/testthat/out/error-position-na.txt
```diff
@@ -1,19 +1,19 @@
 > vec_as_position(na_int, 2L)
 Error: Must extract with a single index.
-x `i` can't be `NA`.
-i This index can't be missing.
+x `i` has the wrong type `integer`.
```
DavisVaughan: Like, this looks odd right?

lionel-:
Yeah sorry, I'll fix first thing.

PR #632, opened 2019-10-23
Rename `vec_split_id()` to `vec_group_pos()` and move `vec_split()`

DavisVaughan:
* `vec_split_id()` -> `vec_group_pos()`

* The `$id` column has become `$pos`

* The `?vec_group` page now includes `vec_group_pos()` as well

* Tweaked NEWS to group the new `vec_group_*()` functions together

* `vec_split()` now has its own C / R / testing files since it now stands alone and doesn't make much sense in the dictionary files.


DavisVaughan:
Closed in favor of #635, which has better git history for `git blame -C`

PR #634, opened 2019-10-23
Don't call `vec_size()` on scalar types when translating

### LINKED ISSUE #633: Translation functions should be robust against scalar types
Currently this fails:

``` r
vctrs::vec_unique(list(formula1 = Sepal.Length ~ Sepal.Width))
#> `x` must be a vector, not a `formula` object
```

The tidyr vignette does something like this. It used to work.

With the translation helpers, we call `vec_size()` on each list element to get the size, which causes the error for formulas. We should be robust against scalar types and just not translate them.

This is where the problem comes from

``` r
vctrs:::obj_maybe_translate_encoding(list(formula1 = Sepal.Length ~ Sepal.Width))
#> `x` must be a vector, not a `formula` object
```

I'm working on a PR for this now.

DavisVaughan:
Closes #633

We now skip over any scalar types found inside of a list in the translation helpers so we don't accidentally call `vec_size()` on them.

``` r
library(vctrs)

lst <- list(y ~ x, y ~ x)

vctrs:::obj_maybe_translate_encoding(lst)
#> [[1]]
#> y ~ x
#> 
#> [[2]]
#> y ~ x

# so this works again
vec_unique(lst)
#> [[1]]
#> y ~ x
```

The change doesn't need to be made in the data frame functions because a scalar type will never directly be a column of a data frame, and it would pass on the data frame size anyways, rather than trying to call `vec_size()`.

There should be no performance implications.

### DIFF
--- a/src/translate.c
+++ b/src/translate.c
@@ -93,7 +93,6 @@ static bool obj_any_known_encoding(SEXP x, R_len_t size) {
 // For usage on list elements. They have unknown size, and might be scalars.
 static bool elt_any_known_encoding(SEXP x) {
   switch (vec_proxy_typeof(x)) {
-  case vctrs_type_scalar: return false;
   case vctrs_type_character: return chr_any_known_encoding(x, vec_size(x));
   case vctrs_type_list: return list_any_known_encoding(x, vec_size(x));
   case vctrs_type_dataframe: return df_any_known_encoding(x, vec_size(x));
@@ -163,7 +162,6 @@ static SEXP obj_translate_encoding(SEXP x, R_len_t size) {
 // For usage on list elements. They have unknown size, and might be scalars.
 static SEXP elt_translate_encoding(SEXP x) {
   switch (vec_proxy_typeof(x)) {
-  case vctrs_type_scalar: return x;
   case vctrs_type_character: return chr_translate_encoding(x, vec_size(x));
   case vctrs_type_list: return list_translate_encoding(x, vec_size(x));
   case vctrs_type_dataframe: return df_translate_encoding(x, vec_size(x));
--- a/src/translate.c
+++ b/src/translate.c
@@ -82,21 +82,39 @@ static bool list_any_known_encoding(SEXP x, R_len_t size);
 static bool df_any_known_encoding(SEXP x, R_len_t size);
 
 static bool obj_any_known_encoding(SEXP x, R_len_t size) {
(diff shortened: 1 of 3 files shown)


DavisVaughan:
> I'm not sure the vctrs typeof is right here, maybe it'd be better to use the lower level TYPEOF().

I did that so I can control the data frame path

DavisVaughan:
> Do we need to handle encoding in attributes as well?

I would imagine so. If we call `vec_unique()` with a list, the attributes of the list elements should be considered when determining uniqueness. Meaning that they need to be translated. I'm going to leave that for another PR though.

DavisVaughan:
I've swapped out `vec_proxy_typeof()` for `TYPEOF()`. It still makes sense to differentiate data frames from lists, so there is careful usage of `vec_size()` for data frames and `Rf_length()` elsewhere.

I mentioned in the first comment that we "skip over" scalar types found in a list, but I think that was incorrect. We now descend into scalar types found in a list and translate them as well if necessary. Using `Rf_length()` rather than `vec_size()` on lists allows this to work correctly for scalar lists.

Practically this means that something like this is now translated

``` r
x <- iconv("\u00B0C", from = "UTF-8", to = "latin1")
Encoding(x)
#> [1] "latin1"

# make a list holding a scalar type
scalar <- structure(list(x = x), class = "scalar_list")
lst <- list(scalar = scalar)

result <- vctrs:::obj_maybe_translate_encoding(lst)

Encoding(result$scalar$x)
#> [1] "UTF-8"
```

<sup>Created on 2019-11-12 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0.9000)</sup>

src/translate.c
```diff
@@ -82,21 +82,39 @@ static bool list_any_known_encoding(SEXP x, R_len_t size);
...
-  case vctrs_type_character: return chr_any_known_encoding(x, size);
-  case vctrs_type_list: return list_any_known_encoding(x, size);
-  case vctrs_type_dataframe: return df_any_known_encoding(x, size);
-  default: return false;
+  switch (TYPEOF(x)) {
+  case STRSXP: {
+    return chr_any_known_encoding(x, size);
+  }
+  case VECSXP: {
+    if (is_data_frame(x)) {
```
lionel-: Maybe add a comment that dfs and lists are treated differently for performance, because we can make size assumptions on df columns? IIUC.

PR #638, opened 2019-10-25
Treat positive and negative 0 as equivalent when hashing

### LINKED ISSUE #637: `vec_unique()` makes a distinction between +0 and -0 while `unique()` does not
Hi,

I noticed that vec_unique() seems to make a distinction between +0 and -0 while most R functions do not. This can affect the behavior of pivot_wider when +0 and -0 are present in the id_cols, causing it to generate two rows instead of one. 

This seems to be something about how the underlying C function vctrs_unique_loc() works, but that's as far as I got.

``` r
library(tidyverse)
library(vctrs)

posZero = 0
negZero = ceiling(-0.9)

unique(posZero, negZero)
#> [1] 0
vec_unique(c(posZero, negZero ))
#> [1] 0 0

# using `abs()` to convert -0 to a +0 fixes the problem
vec_unique(abs(c(posZero, negZero)))
#> [1] 0

# Similarly, `identical()` normally treats these as equivalent
identical(posZero, negZero, num.eq = T)
#> [1] TRUE

# However, setting num.eq = F does a bitwise comparison and differentiates between +0 and -0
identical(posZero, negZero, num.eq = F)
#> [1] FALSE

# This can cause `pivot_wider()` to generate extra rows
data.frame(id = c(posZero, negZero),
           names = c('a', 'b'), 
           vals = c(1,2)) %>%
  pivot_wider(id_cols = id, names_from =  names, values_from = vals )
#> # A tibble: 2 x 3
#>      id     a     b
#>   <dbl> <dbl> <dbl>
#> 1     0     1    NA
#> 2     0    NA     2

# using `abs()` to convert the -0 to a +0 again fixes the problem
data.frame(id = c(posZero, negZero),
           names = c('a', 'b'), 
           vals = c(1,2)) %>%
  mutate(id = abs(id)) %>%
  pivot_wider(id_cols = id, names_from =  names, values_from = vals )
#> # A tibble: 1 x 3
#>      id     a     b
#>   <dbl> <dbl> <dbl>
#> 1     0     1     2
```

<sup>Created on 2019-10-24 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

DavisVaughan:
Closes #637 

We now treat positive and negative 0 as equivalent, similar to how R does
https://github.com/wch/r-source/blob/800fa450bca0bef95d53ce705cbc70f6075fdec0/src/main/unique.c#L115

``` r
vctrs::vec_unique(c(0, -0))
#> [1] 0
```

The way that R does, by creating a tmp variable with `double tmp = (xi == 0.0) ? 0.0 : xi;` is slightly slower than what I do by skipping the re-assignment when the value is not 0.

FYI, integers do not have this problem. `0` and `-0` look the same at the C level.

This has minimal performance implications (it looks like it got faster but I think that's just proof that it doesn't have much of an impact, and you are just seeing noise).

``` r
library(vctrs)
x <- rep(1, times = 1e7)

vec_hash <- vctrs:::vec_hash
```

Before:

``` r
bench::mark(vec_hash(x), iterations = 100)
#> # A tibble: 1 x 6
#> 1 vec_hash(x)    117ms    131ms      7.28    38.1MB     7.58
```

After:

``` r
bench::mark(vec_hash(x), iterations = 100)
#> # A tibble: 1 x 6
#> 1 vec_hash(x)    119ms    124ms      8.01    38.1MB     8.34
```

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,6 +1,9 @@
 
 # vctrs 0.2.0.9000
 
+* Positive and negative 0 are now considered equivalent by all functions that
+  check for equality or uniqueness (#637).
+
 * New experimental `vec_group_rle()` for returning run length encoded groups.
 
 * New experimental `vec_group_id()` for constructing group identifiers from a
--- a/src/hash.c
+++ b/src/hash.c
@@ -32,6 +32,11 @@ static uint32_t hash_int64(int64_t x) {
 // Seems like something designed specificaly for doubles should work better
 // but I haven't been able to find anything
 static uint32_t hash_double(double x) {
+  // Treat positive/negative 0 as equivalent
+  if (x == 0.0) {
+    x = 0.0;
+  }
+
   union {
     double d;
     uint64_t i;
--- a/tests/testthat/test-dictionary.R
+++ b/tests/testthat/test-dictionary.R
@@ -112,7 +112,7 @@ test_that("unique functions take the equality proxy (#375)", {
   expect_identical(vec_match(tuple(2, 100), x), 2L)
 })
 
-test_that("unique functions treat positive and negative 0 as equivalent", {
+test_that("unique functions treat positive and negative 0 as equivalent (#637)", {
   expect_equal(vec_unique(c(0, -0)), 0)
   expect_equal(vec_unique_count(c(0, -0)), 1)
   expect_equal(vec_unique_loc(c(0, -0)), 1)
--- a/tests/testthat/test-dictionary.R
+++ b/tests/testthat/test-dictionary.R
@@ -112,6 +112,12 @@ test_that("unique functions take the equality proxy (#375)", {
   expect_identical(vec_match(tuple(2, 100), x), 2L)
 })
 
+test_that("unique functions treat positive and negative 0 as equivalent", {
+  expect_equal(vec_unique(c(0, -0)), 0)
+  expect_equal(vec_unique_count(c(0, -0)), 1)
+  expect_equal(vec_unique_loc(c(0, -0)), 1)
+})
+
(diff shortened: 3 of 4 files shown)


DavisVaughan:
🤔 I'm not quite sure what's happening with the azure linux builds. 

It says rlang 0.4.1.9000 is installed when remotes and rcmdcheck are installed

https://dev.azure.com/r-lib/vctrs/_build/results?buildId=601&view=logs&jobId=bb1c2637-64c6-57bd-9ea6-93823b2df951&taskId=4a1a54a7-28b3-5508-d0b8-dcd62167bb50&lineStart=254&lineEnd=255&colStart=1&colEnd=1

But then when the vctrs deps are actually installed it pulls in the 0.4.0 binary from rstudio package manager

https://dev.azure.com/r-lib/vctrs/_build/results?buildId=601&view=logs&jobId=bb1c2637-64c6-57bd-9ea6-93823b2df951&taskId=4a1a54a7-28b3-5508-d0b8-dcd62167bb50&lineStart=387&lineEnd=388&colStart=1&colEnd=1

So then later in the check it says rlang 0.4.0 is loaded, but rlang >= 0.4.1 is required

https://dev.azure.com/r-lib/vctrs/_build/results?buildId=601&view=logs&jobId=bb1c2637-64c6-57bd-9ea6-93823b2df951&taskId=9c62e7ff-391e-517d-ede1-af63bc6a25da&lineStart=62&lineEnd=63&colStart=1&colEnd=1

This must be some weird case where RPM hasn't refreshed to pull in the latest new CRAN rlang, but for some reason it isn't triggering a source install

PR #642, opened 2019-11-01
Allow integerish indexes in vec_as_index()

### LINKED ISSUE #577: Allow integerish indexes in vec_as_index()
Does this need to be an error?

``` r
vctrs::vec_as_index(1:3, 3)
#> `n` must be a vector with type <integer>.
#> Instead, it has type <double>.
```

<sup>Created on 2019-09-23 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

krlmlr:
With implicit tests.

Closes #577.

### DIFF
--- a/R/slice.R
+++ b/R/slice.R
@@ -193,6 +193,7 @@ vec_as_index <- function(i, n,
                          arg = "i") {
   if (!missing(...)) ellipsis::check_dots_empty()
 
+  n <- vec_coercible_cast(n, integer())
   vec_assert(n, integer(), 1L)
   i <- vec_coerce_index(i, arg = arg, allow_types = allow_types)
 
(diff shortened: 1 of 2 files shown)


lionel-:
I don't like the unnecessarily large diff in unit tests here. Can you just add a single test checking that double indices are allowed please, and maybe check that lossy casts (e.g. `n = 1.5`) are an error.

krlmlr:
I agree. 6212821 contains the removal of the `L`s as a separate commit, feel free to cherry-pick.

lionel-:
I will. But also feel free to rebase and force-push.

krlmlr:
I don't understand. Can we please merge this as is and discuss `L` afterwards?

lionel-:
No need to discuss `L` afterwards, I just meant to leave it out of the PR which is now the case, thanks!

PR #649, opened 2019-11-06
Improve `vec_compare()`

### LINKED ISSUE #648: Performance: `vec_compare()`
I think there could be a number of substantial improvements to `vec_compare()`

- Move the comparison loop in `vctrs_compare()` to be after the type check. So if we have a vector of ints it just checks that it is an integer once, and then loops through all of the values. This would use a similar approach as `vec_equal()`, but it will be made easier because we don't compare lists.

- Remove the `is_data_frame()`, length, and names checks from the `VECSXP` code path. This would probably require a switch to using `vec_proxy_typeof()` rather than `TYPEOF()` so we can know if we are in a data frame code path earlier on. These checks are pretty expensive, and we 100% don't need them because we called `vec_cast_common()` on the inputs already.

- Move the rest of the R side of `vec_compare()` to C. The only thing we need is the ability to call `vec_proxy_compare()` from the C level. Since we don't have a native implementation it could just call back to the R function.

- We also need to validate `na_equal`. i.e. this works when it shouldn't `vec_compare(1, 2, na_equal = list(1))`

DavisVaughan:
Closes #648 

Still working on this, but incredibly promising results with a rewrite of `vec_compare()`. 

It follows the ideas of `vec_equal()` for the most part. But importantly I'm also experimenting with the idea of doing the comparisons column wise on data frames. This is way more efficient than row wise and I think we might be able to port this idea over to `vec_equal()` too.

The way it generally works is that we construct a logical `skip` vector initialized to all false values and with size equal to the number of rows in the data frame. As we go through the data frame columns, we flip the `skip` values to `TRUE` when we "know" the comparison value (i.e. if it is less than or greater than) and we don't ever evaluate that row again when comparing downstream columns. If all `skip` values are ever `TRUE` (as measured by a decreasing `count` that hits `0` when we know everything), we exit immediately.

Before

``` r
library(vctrs)
library(nycflights13)

bench::mark(
  vec_compare(flights$year, flights$year)
)
#> # A tibble: 1 x 6
#>   expression                                 min median `itr/sec` mem_alloc
#>   <bch:expr>                              <bch:> <bch:>     <dbl> <bch:byt>
#> 1 vec_compare(flights$year, flights$year) 5.26ms 6.48ms      156.     135MB

bench::mark(
  vec_compare(flights, flights)
)
#> # A tibble: 1 x 6
#>   <bch:expr>                    <bch:t> <bch:>     <dbl> <bch:byt>    <dbl>
#> 1 vec_compare(flights, flights)   189ms  197ms      5.07    58.2MB     2.54

flights_sort <- dplyr::arrange(flights, dep_time)

bench::mark(
  vec_compare(flights, flights_sort)
)
#> # A tibble: 1 x 6
#> 1 vec_compare(flights, flights_sort) 39.1ms 42.3ms      23.9    11.6MB
```

After

``` r
library(vctrs)
library(nycflights13)

# vector speed up
bench::mark(
  vec_compare(flights$year, flights$year)
)
#> # A tibble: 1 x 6
#>   expression                                min median `itr/sec` mem_alloc
#>   <bch:expr>                              <bch> <bch:>     <dbl> <bch:byt>
#> 1 vec_compare(flights$year, flights$year) 351µs  509µs     1700.     133MB

# all equal "worst case"
bench::mark(
  vec_compare(flights, flights)
)
#> # A tibble: 1 x 6
#>   <bch:expr>                    <bch:t> <bch:>     <dbl> <bch:byt>    <dbl>
#> 1 vec_compare(flights, flights)  12.3ms   13ms      75.6    2.92MB     2.04

flights_sort <- dplyr::arrange(flights, dep_time)

# not all equal, we compare column wise and skip if we already
# know the result
bench::mark(
  vec_compare(flights, flights_sort)
)
#> # A tibble: 1 x 6
#> 1 vec_compare(flights, flights_sort) 2.52ms 3.13ms      306.    2.57MB
```

<sup>Created on 2019-11-06 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0.9000)</sup>

### DIFF
--- a/src/compare.c
+++ b/src/compare.c
@@ -259,13 +259,9 @@ static SEXP df_compare_impl(SEXP x, SEXP y, bool na_equal, SEXP info, R_len_t n)
 
 #define COMPARE_COL(CTYPE, CONST_DEREF, SCALAR_COMPARE)     \
 do {                                                        \
-  SEXP out = VECTOR_ELT(info, 0);                           \
-  SEXP skip = VECTOR_ELT(info, 1);                          \
-  SEXP count = VECTOR_ELT(info, 2);                         \
-                                                            \
-  int* p_out = INTEGER(out);                                \
-  int* p_skip = LOGICAL(skip);                              \
-  int* p_count = INTEGER(count);                            \
+  int* p_out = INTEGER(VECTOR_ELT(info, 0));                \
+  int* p_skip = LOGICAL(VECTOR_ELT(info, 1));               \
+  int* p_count = INTEGER(VECTOR_ELT(info, 2));              \
                                                             \
   const CTYPE* p_x = CONST_DEREF(x);                        \
   const CTYPE* p_y = CONST_DEREF(y);                        \
@@ -287,10 +283,6 @@ do {                                                        \
       }                                                     \
     }                                                       \
   }                                                         \
-                                                            \
-  SET_VECTOR_ELT(info, 0, out);                             \
-  SET_VECTOR_ELT(info, 1, skip);                            \
-  SET_VECTOR_ELT(info, 2, count);                           \
                                                             \
   return info;                                              \
 }                                                           \
--- a/src/compare.c
+++ b/src/compare.c
@@ -216,37 +216,69 @@ SEXP vctrs_compare(SEXP x, SEXP y, SEXP na_equal_) {
 
 // -----------------------------------------------------------------------------
 
-static SEXP vec_compare_col(SEXP x, SEXP y, bool na_equal, SEXP info, R_len_t n);
-static SEXP df_compare_impl(SEXP x, SEXP y, bool na_equal, SEXP info, R_len_t n);
+struct vctrs_df_compare_info {
+  SEXP out;
+  SEXP row_known;
+  R_len_t remaining;
+};
 
-static SEXP df_compare(SEXP x, SEXP y, bool na_equal, R_len_t n) {
-  // Skip nothing to begin with
-  SEXP skip = PROTECT(Rf_allocVector(LGLSXP, n));
-  int* p_skip = LOGICAL(skip);
-  memset(p_skip, 0, n * sizeof(int));
+#define PROTECT_DF_COMPARE_INFO(info, n) do {  \
+  PROTECT((info)->out);                        \
+  PROTECT((info)->row_known);                  \
+  *n += 2;                                     \
+} while (0)
 
-  SEXP count = PROTECT(Rf_allocVector(INTSXP, 1));
-  *INTEGER(count) = n;
+static struct vctrs_df_compare_info init_compare_info(R_len_t n) {
+  struct vctrs_df_compare_info info;
 
   // Initialize to "equality" value
-  SEXP out = PROTECT(Rf_allocVector(INTSXP, n));
-  int* p_out = INTEGER(out);
+  // and only change if we learn that it differs
+  info.out = PROTECT(Rf_allocVector(INTSXP, n));
+  int* p_out = INTEGER(info.out);
   memset(p_out, 0, n * sizeof(int));
 
-  SEXP info = PROTECT(Rf_allocVector(VECSXP, 3));
-  SET_VECTOR_ELT(info, 0, out);
-  SET_VECTOR_ELT(info, 1, skip);
-  SET_VECTOR_ELT(info, 2, count);
+  // To begin with, no rows have a known comparison value
+  info.row_known = PROTECT(Rf_allocVector(LGLSXP, n));
+  int* p_row_known = LOGICAL(info.row_known);
+  memset(p_row_known, 0, n * sizeof(int));
+
+  info.remaining = n;
+
+  UNPROTECT(2);
+  return info;
+}
+
+// -----------------------------------------------------------------------------
+
+static struct vctrs_df_compare_info vec_compare_col(SEXP x,
+                                                    SEXP y,
+                                                    bool na_equal,
+                                                    struct vctrs_df_compare_info info,
+                                                    R_len_t n);
+
+static struct vctrs_df_compare_info df_compare_impl(SEXP x,
+                                                    SEXP y,
+                                                    bool na_equal,
+                                                    struct vctrs_df_compare_info info,
+                                                    R_len_t n);
+
+static SEXP df_compare(SEXP x, SEXP y, bool na_equal, R_len_t n) {
+  int nprot = 0;
+
+  struct vctrs_df_compare_info info = init_compare_info(n);
+  PROTECT_DF_COMPARE_INFO(&info, &nprot);
 
   info = df_compare_impl(x, y, na_equal, info, n);
 
-  out = VECTOR_ELT(info, 0);
-
-  UNPROTECT(4);
-  return out;
+  UNPROTECT(nprot);
+  return info.out;
 }
 
-static SEXP df_compare_impl(SEXP x, SEXP y, bool na_equal, SEXP info, R_len_t n) {
+static struct vctrs_df_compare_info df_compare_impl(SEXP x,
+                                                    SEXP y,
+                                                    bool na_equal,
+                                                    struct vctrs_df_compare_info info,
+                                                    R_len_t n) {
   int n_col = Rf_length(x);
 
   if (n_col == 0) {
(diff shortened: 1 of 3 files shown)


lionel-:
> I'm also experimenting with the idea of doing the comparisons column wise on data frames. This is way more efficient than row wise and I think we might be able to port this idea over to vec_equal() 

yup we definitely need to finish vectorising data frame operations in vctrs, would be great to reach this point!

DavisVaughan:
@lionel- ready for review

- Added encoding support for `vec_compare()`, which I had forgotten about when I did it for `vec_equal()`
- We validate `na_equal` now
- Added various tests to flesh out existing coverage

There is a big drop in test coverage for this file because `compare_scalar()` is no longer used by anything. This means it has no tests, along with `df_compare_scalar()`, which is only ever used by `compare_scalar()`. We include `compare_scalar()` in the vctrs.h api file (like `equal_scalar()`), but we don't actually use it.

Not sure what to do about that. It seems like it might someday be useful internally, but I don't really feel like it would be worth it to export it to the R side just for testing now.

src/compare.c
```diff
@@ -1,136 +1,325 @@
...
+
+#undef COMPARE
+
+// -----------------------------------------------------------------------------
+
+static SEXP vec_compare_col(SEXP x, SEXP y, bool na_equal, SEXP info, R_len_t n);
+static SEXP df_compare_impl(SEXP x, SEXP y, bool na_equal, SEXP info, R_len_t n);
+
+static SEXP df_compare(SEXP x, SEXP y, bool na_equal, R_len_t n) {
+  // Skip nothing to begin with
```
lionel-: It would help to explain the algorithm here, e.g. what `skip` is for and what `count` is counting.

Maybe `count` should be `remaining_rows` and `skip` something like `row_done` to indicate it's about rows that are already compared?

src/compare.c
```diff
@@ -1,136 +1,325 @@
...
+}
+
+#undef COMPARE
+
+// -----------------------------------------------------------------------------
+
+static SEXP vec_compare_col(SEXP x, SEXP y, bool na_equal, SEXP info, R_len_t n);
+static SEXP df_compare_impl(SEXP x, SEXP y, bool na_equal, SEXP info, R_len_t n);
+
+static SEXP df_compare(SEXP x, SEXP y, bool na_equal, R_len_t n) {
```
lionel-: Maybe rename `n` to `nrow` everywhere, to help keep track of what it's about?

tests/testthat/test-compare.R
```diff
@@ -107,6 +147,59 @@ test_that("vec_proxy_compare() handles data frame with a POSIXlt column", {
...
+
+test_that("can compare different strings with different encodings", {
+  x <- "x"
+  y <- encodings()$latin1
+
+  expect_equal(vec_compare(x, y), -1L)
+})
+
+test_that("equality can always be determined when strings have identical encodings", {
+  encs <- c(encodings(), list(bytes = encoding_bytes()))
```
lionel-: I was puzzled for a bit here. I wonder if we should switch to this clearer way of combining lists:

```
encs <- list2(!!!encodings(), bytes = encoding_bytes())
```

tests/testthat/test-compare.R
```diff
@@ -107,6 +147,59 @@ test_that("vec_proxy_compare() handles data frame with a POSIXlt column", {
...
+
+test_that("can compare equal strings with different encodings", {
+  for (x_encoding in encodings()) {
+    for (y_encoding in encodings()) {
+      expect_equal(vec_compare(x_encoding, y_encoding), 0L)
+    }
+  }
+})
+
+test_that("can compare different strings with different encodings", {
```
lionel-:
```suggestion
test_that("can compare non-equal strings with different encodings", {
```

tests/testthat/test-compare.R
```diff
@@ -55,11 +55,51 @@ test_that("data frames are compared column by column", {
...
 test_that("C code doesn't crash with bad inputs", {
   df <- data.frame(x = c(1, 1, 1), y = c(-1, 0, 1))
 
   expect_error(.Call(vctrs_compare, df, df[1], TRUE), "not comparable")
-  expect_error(.Call(vctrs_compare, df, setNames(df, c("x", "z")), TRUE), "not comparable")
+
+  # Names are not checked, as `vec_cast_common()` should take care of the type.
+  # So if `vec_cast_common()` is not called, or is improperly specified, then
+  # this could result in false equality.
+  expect_equal(.Call(vctrs_compare, df, setNames(df, c("x", "z")), TRUE), c(0, 0, 0))
```
lionel-: Maybe also a test case for unequal rows?

```r
  df1 <- new_data_frame(list(x = 1:3, y = 1))
  df2 <- new_data_frame(list(y = 1:2, x = 1:2))
  .Call(vctrs_compare, df1, df2, TRUE)
```

src/compare.c
```diff
@@ -1,136 +1,325 @@
...
+  case STRSXP: return chr_compare_scalar(STRING_PTR(x) + i, STRING_PTR(y) + j, na_equal);
+  default: break;
+  }
 
-          if (cmp != 0)
-            return cmp;
-        }
-      }
+  switch (vec_proxy_typeof(x)) {
+  case vctrs_type_list: stop_not_comparable(x, y, "lists are not comparable");
```
lionel-: It seems weird that we can compare list-columns but not lists, like an invariant is broken. Perhaps we should be consistent.

DavisVaughan: I was _very_ surprised that we could do a `vec_compare()` on a data frame with list columns. This test was meant to be one saying "cannot compare data frames with list columns" until I realized it worked. 

https://github.com/r-lib/vctrs/pull/649/files#diff-190e0bcf1adfd9d01b92248770e25c0cR85

We don't actually do a true comparison. The list column gets replaced with `vec_seq_along()` in both `x` and `y`, so essentially it gets skipped.

This happens in `vec_proxy_compare.data.frame()` where we set `vec_proxy_compare(relax = TRUE)`.

lionel-: This makes sense when you consider `arrange()` on a data frame containing list columns. We can't have an error there, and a descent into list elements would be costly and would not produce meaningful ordering.

src/compare.c
```diff
@@ -1,136 +1,325 @@
...
-  case LGLSXP: {
-    int xi = LOGICAL(x)[i], yj = LOGICAL(y)[j];
-    if (na_equal) {
-      return icmp(xi, yj);
-    } else {
-      return (xi == NA_LOGICAL || yj == NA_LOGICAL) ? NA_INTEGER : icmp(xi, yj);
-    }
+// -----------------------------------------------------------------------------
+
+static int lgl_compare_scalar(const int* x, const int* y, bool na_equal) {
```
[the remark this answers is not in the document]
DavisVaughan: ✔️

src/compare.c
```diff
@@ -1,136 +1,325 @@
...
+
+  SEXP count = PROTECT(Rf_allocVector(INTSXP, 1));
+  *INTEGER(count) = n;
+
+  // Initialize to "equality" value
+  SEXP out = PROTECT(Rf_allocVector(INTSXP, n));
+  int* p_out = INTEGER(out);
+  memset(p_out, 0, n * sizeof(int));
+
+  SEXP info = PROTECT(Rf_allocVector(VECSXP, 3));
```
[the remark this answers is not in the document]
DavisVaughan: ✔️

src/compare.c
```diff
@@ -216,37 +219,82 @@ SEXP vctrs_compare(SEXP x, SEXP y, SEXP na_equal_) {
...
+};
+
+#define PROTECT_DF_COMPARE_INFO(info, n) do {  \
+  PROTECT((info)->out);                        \
+  PROTECT((info)->row_known);                  \
+  *n += 2;                                     \
+} while (0)
+
+static struct vctrs_df_compare_info init_compare_info(R_len_t n_row) {
+  struct vctrs_df_compare_info info;
```
lionel-: This seems a bit overkill for a one-off initialisation, but that's not a big deal.

src/compare.c
```diff
@@ -216,37 +219,82 @@ SEXP vctrs_compare(SEXP x, SEXP y, SEXP na_equal_) {
...
+ * @member row_known A logical vector of size `n_row`. Initially, all values
+ *   are initialized to `FALSE`. As we iterate along the columns, we flip the
+ *   corresponding row's `row_known` value to `TRUE` if we can determine it
+ *   from the current column comparison (i.e. the comparison returns less than
+ *   or greater than). Once a row's `row_known` value is `TRUE`, we never check
+ *   that row again as we continue through the columns.
+ * @member remaining The number of `row_known` values that are still `FALSE`.
+ *   If this hits `0` before we traverse the entire data frame, we can exit
+ *   immediately because all comparison values are already known.
+ */
```
lionel-: 👍

PR #652, opened 2019-11-08
Implement column wise `vec_equal()` for data frames

DavisVaughan:
Uses the same approach as `vec_compare()`

Before

``` r
library(vctrs)
library(nycflights13)

bench::mark(
  vec_equal(flights, flights)
)
#> # A tibble: 1 x 6
#>   <bch:expr>                  <bch:tm> <bch:t>     <dbl> <bch:byt>    <dbl>
#> 1 vec_equal(flights, flights)    113ms   118ms      8.56     138MB        0

flights_sort <- dplyr::arrange(flights, dep_time)

bench::mark(
  vec_equal(flights, flights_sort)
)
#> # A tibble: 1 x 6
#>   expression                          min median `itr/sec` mem_alloc
#>   <bch:expr>                       <bch:> <bch:>     <dbl> <bch:byt>
#> 1 vec_equal(flights, flights_sort) 7.91ms 8.85ms      112.    1.29MB
```

After:

``` r
library(vctrs)
library(nycflights13)

bench::mark(
  vec_equal(flights, flights)
)
#> # A tibble: 1 x 6
#>   <bch:expr>                  <bch:tm> <bch:t>     <dbl> <bch:byt>    <dbl>
#> 1 vec_equal(flights, flights)   9.54ms  10.5ms      91.1     134MB     2.07

flights_sort <- dplyr::arrange(flights, dep_time)

bench::mark(
  vec_equal(flights, flights_sort)
)
#> # A tibble: 1 x 6
#>   expression                          min median `itr/sec` mem_alloc
#>   <bch:expr>                       <bch:> <bch:>     <dbl> <bch:byt>
#> 1 vec_equal(flights, flights_sort) 2.34ms  2.6ms      353.    2.57MB
```

Also:

- Validates `na_equal`
- Fleshes out test coverage a bit
- Recognized that the structs we use in `vec_equal()` and `vec_compare()` are identical, so I pulled it out into `vctrs.h`. Let me know what you think about the name (`vctrs_df_rowwise_info`). It will also be used to vectorize `vec_equal_na()` and `vec_duplicate_all()` for data frames using the same approach so I thought it made sense to pull it out. It's nice because the `init_*()` functions can set the `info.out` SEXP to whatever type is required for that particular function. `vec_compare()` uses an integer, `vec_equal()` uses a logical.

### DIFF
--- a/src/equal.c
+++ b/src/equal.c
@@ -43,6 +43,9 @@ int equal_scalar(SEXP x, R_len_t i, SEXP y, R_len_t j, bool na_equal) {
   vctrs_stop_unsupported_type(vec_typeof(x), "equal_scalar()");
 }
 
+// -----------------------------------------------------------------------------
+
+static SEXP df_equal(SEXP x, SEXP y, bool na_equal, R_len_t n_row);
 
 #define EQUAL(CTYPE, CONST_DEREF, SCALAR_EQUAL)         \
   do {                                                  \
@@ -75,26 +78,6 @@ int equal_scalar(SEXP x, R_len_t i, SEXP y, R_len_t j, bool na_equal) {
   }                                                     \
   while (0)
 
-#define EQUAL_DF(SCALAR_EQUAL)                                                      \
-  do {                                                                              \
-    SEXP out = PROTECT(Rf_allocVector(LGLSXP, size));                               \
-    int32_t* p = LOGICAL(out);                                                      \
-                                                                                    \
-    int n_col = Rf_length(x);                                                       \
-                                                                                    \
-    if (n_col != Rf_length(y)) {                                                    \
-      Rf_errorcall(R_NilValue, "`x` and `y` must have the same number of columns"); \
-    }                                                                               \
-                                                                                    \
-    for (R_len_t i = 0; i < size; ++i) {                                            \
-      p[i] = SCALAR_EQUAL(x, i, y, i, na_equal, n_col);                             \
-    }                                                                               \
-                                                                                    \
-    UNPROTECT(3);                                                                   \
-    return out;                                                                     \
-  }                                                                                 \
-  while (0)
-
 // [[ register() ]]
 SEXP vctrs_equal(SEXP x, SEXP y, SEXP na_equal_) {
   x = PROTECT(vec_proxy_recursive(x, vctrs_proxy_equal));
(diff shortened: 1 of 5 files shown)


src/equal.c
```diff
@@ -418,6 +421,162 @@ bool equal_names(SEXP x, SEXP y) {
...
+static struct vctrs_df_rowwise_info init_rowwise_equal_info(R_len_t n_row) {
+  struct vctrs_df_rowwise_info info;
+
+  // Initialize to "equality" value
+  // and only change if we learn that it differs
+  info.out = PROTECT(Rf_allocVector(LGLSXP, n_row));
+  int* p_out = LOGICAL(info.out);
+
+  for (R_len_t i = 0; i < n_row; ++i, ++p_out) {
+    *p_out = 1;
```
lionel-: I know I wrote code like this in the past, but I think in such simple loops we should use for readability:

```suggestion
    p_out[i] = 1;
```

But here the more structured `memset()` seems better so that the reader does not need to inspect carefully the loop.

src/equal.c
```diff
@@ -418,6 +421,162 @@ bool equal_names(SEXP x, SEXP y) {
...
+  for (R_len_t i = 0; i < n_row; ++i, ++p_row_known) { \
+    if (*p_row_known) {                                \
+      continue;                                        \
+    }                                                  \
+                                                       \
+    int eq = SCALAR_EQUAL(x, i, y, i, na_equal);       \
+                                                       \
+    if (eq <= 0) {                                     \
+      p_out[i] = eq;                                   \
+      *p_row_known = true;                             \
```
lionel-: I think this should be `p_row_known[i]`, otherwise it suggests that it has a different increment rule than for `p_out` which is indexed with `i`.

src/equal.c
DavisVaughan: How would you memset() this?

lionel-: `memset(p_out, 1, sizeof(*p_out) * n_row)`? Or am I missing something?

DavisVaughan: I tried that first and then had to learn how memset worked 😬 https://stackoverflow.com/questions/20951369/how-to-use-memset-for-initializing-bufffers-with-values-other-than-0

lionel-: oh yeah that makes sense.

src/equal.c
DavisVaughan: Would it make sense for `row_known` to just be a `bool` array that we `R_alloc()`? I don't use the fact that it is a SEXP anywhere, and it might make this clearer:

```
struct vctrs_df_rowwise_info {
  SEXP out;
  SEXP row_known;
  R_len_t remaining;
};
```

```
struct vctrs_df_rowwise_info {
  SEXP out;
  bool* row_known;
  R_len_t remaining;
};
```

It would also nicely enforce the type that `row_known` is intended to be. Right now it can be any SEXP. This is a nice property of `out`, but not of `row_known`.

lionel-: I think that makes sense, we could try this pattern. It is easy to set and reset the vmax from our `PROTECT_` function right?

DavisVaughan:
@lionel- if the `RAWSXP` bit looks good then this should be good to merge

src/vctrs.h
```diff
@@ -276,19 +276,23 @@ bool duplicated_any(SEXP names);
 /**
  * @member out A vector of size `n_row` containing the output of the
  *   row wise data frame operation.
- * @member row_known A logical vector of size `n_row`. Initially, all values
- *   are initialized to `FALSE`. As we iterate along the columns, we flip the
- *   corresponding row's `row_known` value to `TRUE` if we can determine the
+ * @member row_known A boolean array of size `n_row`. It is intended to be
```
lionel-: I'm not sure we need to mention "managed by a RAWSXP", this seems like a distraction. Instead I'd just say something like "Allocated on the R heap". Also good to keep as is.

Issue #497, opened 2019-07-22
Unexpected failures when vector inherits from character


hadley:
``` r
library(vctrs)
library(glue)

x <- glue("abc")
vec_c(x, x)
#> Only bare vectors have shapes.
```

```
    █
 1. ├─vctrs::vec_c(x, x)
 2. ├─vctrs:::vec_type2_dispatch(x = x, y = y, x_arg = x_arg, y_arg = y_arg)
 3. ├─vctrs::vec_ptype2.character(x = x, y = y, x_arg = x_arg, y_arg = y_arg)
 4. └─vctrs:::vec_ptype2.character.character(...)
 5.   └─vctrs:::shape_match(character(), x, y)
 6.     └─vctrs:::shape_common(x, y)
 7.       ├─vctrs:::n_dim2(shape(x), shape(y))
 8.       └─vctrs:::shape(x)
```

<sup>Created on 2019-07-22 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

From https://github.com/tidyverse/tidyr/issues/676.

hadley:
@DavisVaughan relaxed in https://github.com/r-lib/rray/blob/90793f06821280b19fe02610dae0dcf73df1968c/R/shape.R#L3

yutannihilation:
Is this issue only about the error message? I think no error is the expected result.

lionel-:
```r
x <- structure(letters[1:3], class = "foo")
vctrs::vec_match(x, x)
#> [1] 1 2 3

x <- structure(letters[1:3], class = "character")
vctrs::vec_match(x, x)
#> <error/rlang_error>
#> Only bare vectors have shapes.
#> Backtrace:
#>  1. vctrs::vec_match(x, x)
#>  4. vctrs:::vec_ptype2.character.character(...)
#>  5. vctrs:::shape_match(character(), x, y)
#>  6. vctrs:::shape_common(x, y)
#>  8. vctrs:::shape(x)
```

PR #654, opened 2019-11-11
Add extends_type argument to new_vctr()

hadley:
Closes #593

Currently failing due to #497


NEWS.md
```diff
@@ -1,6 +1,13 @@
 
 # vctrs 0.2.0.9000
 
+* `new_vctr()` gains a `extends_type` argument which determines whether or not
+  the class of the underlying type will be include in the class. This generally
```
lionel-:
```suggestion
  the class of the underlying type will be included in the class. This generally
```

hadley:
That's fair. 

At some point we'll need to consider if this should use the implicit class. i.e. should `new_vctr(1)` get class numeric as well as double?

R/type-vctr.R
```diff
@@ -49,16 +49,21 @@
...
-new_vctr <- function(.data, ..., class = character()) {
+new_vctr <- function(.data, ..., class = character(), inherit_base_type = TRUE) {
   if (!is_vector(.data)) {
     abort("`.data` must be a vector type.")
   }
 
   nms <- validate_names(.data)
-  attrib <- list(names = nms, ..., class = c(class, "vctrs_vctr"))
+
+  class <- c(class, "vctrs_vctr", if (inherit_base_type) class(.data))
```
DavisVaughan: Would it make sense for it to be `typeof(.data)` so we don't drag all of the `.data` classes along with it? Ideally I guess the caller ensured it was a bare vector with no classes. Also, there is no `typeof()` result for `"data.frame"`. But the others would work.

hadley: Oh yeah, I forgot that people might be feeding this things other than base classes.

hadley: This also revealed a subtle bug because `class(1)` is numeric, not double.

DavisVaughan:
Potential alternative:

```
new_vctr(.data, ..., class = character(), super = typeof(.data)) {
  if (!is_vector(.data)) {
    abort("`.data` must be a vector type.")
  }
  
  nms <- validate_names(.data)
  
  class <- c(class, "vctrs_vctr", super)
  attrib <- list(names = nms, ..., class = class)
  
  vec_set_attributes(.data, attrib)
}
```

Would allow `super = NULL` to opt out, and `super = "data.frame"` if the user wanted that.

lionel-:
I think we shouldn't allow arbitrary super classes, it should generally be considered UB.

PR #655, opened 2019-11-11
Treat `list(NULL)` as a missing value in `vec_equal()`

### LINKED ISSUE #653: `vec_equal(list(NULL), list(NULL))` returns `TRUE`
This should return `NA` right? Unless `na_equal = TRUE`?

``` r
vctrs::vec_equal(list(NULL), list(NULL))
#> [1] TRUE
```

Problem line:

https://github.com/r-lib/vctrs/blob/673c9f1feb20ab517f99a7ba15f1e1155fdf83c2/src/equal.c#L256

DavisVaughan:
Closes #653 

``` r
library(vctrs)

vec_equal(list(NULL), list(NULL))
#> [1] NA

vec_equal(list(NULL), list(NULL), na_equal = TRUE)
#> [1] TRUE
```

<sup>Created on 2019-11-11 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0.9000)</sup>

### DIFF
--- a/tests/testthat/test-equal.R
+++ b/tests/testthat/test-equal.R
@@ -124,6 +124,7 @@ test_that("equality is known to fail when comparing bytes to other encodings", {
 
 test_that("can compare NULL",{
   expect_true(obj_equal(NULL, NULL))
+  expect_equal(obj_equal(NULL, NULL, na_equal = FALSE), NA)
 })
 
 test_that("can compare objects with reference semantics", {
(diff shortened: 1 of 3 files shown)


tests/testthat/test-equal.R
```diff
@@ -37,6 +38,10 @@ test_that("double special values", {
   expect_true(vec_equal(-Inf, -Inf))
 })
 
+test_that("`list(NULL)` is considered a missing value (#653)", {
```
lionel-: Maybe add a test for `obj_equal()` as well?

PR #656, opened 2019-11-11
Improve base type matching

### LINKED ISSUE #497: Unexpected failures when vector inherits from character

hadley:
* If neither x nor y are objects, preserve their shape
* Otherwise preserve the underlying base type

I think this is correct because if the class vector includes (e.g.) "character", that's an assertion that the object is a character vector. To preserve attributes and class, you'd need to provide a vec_ptype2 method.

Fixes #497

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,6 +1,10 @@
 
 # vctrs 0.2.0.9000
 
+* `vec_ptype()` has relaxed default behaviour for base types; now if two
+  vectors both inherit from (e.g.) "character", the common type is also
+  "character" (#497).
+
 * Positive and negative 0 are now considered equivalent by all functions that
   check for equality or uniqueness (#637).
 
--- a/R/shape.R
+++ b/R/shape.R
@@ -11,15 +11,6 @@ new_shape <- function(type, shape = NULL) {
   }
 }
 
-shape_match <- function(type, x, y) {
-  if (!is.object(x) && !is.object(y)) {
-    shape <- shape_common(x, y)
-    new_shape(type, shape)
-  } else {
-    type
-  }
-}
-
 shape_common <- function(x, y) {
   shape <- n_dim2(shape(x), shape(y))
   map2_int(shape$x, shape$y, axis2)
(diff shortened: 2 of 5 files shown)


DavisVaughan:
This might be a meaningful test. With `"character"` subclasses that have a `dim`, the common type is `"character"` with no `dim`.

``` r
x <- structure("x", dim = c(1, 1), class = c("foo", "character"))
vctrs::vec_ptype2(x, x)
#> character(0)
```

<sup>Created on 2019-11-11 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0.9000)</sup>

As you said, I think it would be up to the `foo` creator to ensure that a common shape is found if they want it.

NEWS.md
```diff
@@ -1,6 +1,10 @@
 
 # vctrs 0.2.0.9000
 
+* `vec_ptype()` has relaxed default behaviour for base types; now if two
```
lionel-: This feels more like a bug fix than a new rule? Since we normally have inheritance of common type methods.

Or perhaps it's just a workaround for the limitations our subtyping rule with container types (I'm starting to think of arrays/matrices as a separate type hierarchy).

 (No need to change the NEWS bullet, just to make sure I understand correctly)

hadley: Right. It's more of a bug fix, and I agree think that arrays are a separate hierarchy.

PR #658, opened 2019-11-11
Cast data frames to lists row wise to preserve size

### LINKED ISSUE #639: `vec_cast.list.default()` should work row wise on data frames
This currently changes the size of the result because `[[` is used in `vec_cast.list.default()`

``` r
vctrs::vec_cast(data.frame(x = 1:2), list())
#> [[1]]
#> [1] 1 2
```

I think this is a case where we should use `vec_get()`, which might be implemented to extract a single row from a data frame. (See #626)

In the mean time, we should special case data frames in `vec_cast.list.default()`.

This breaks the following, which is how I found it.

``` r
vctrs::vec_c(data.frame(x = 1:2), .ptype = list())
#> Error in vctrs::vec_c(data.frame(x = 1:2), .ptype = list()): Internal error in `vec_assign()`: `value` should have been recycled to fit `x`.
```

DavisVaughan:
Closes #639 

Casting data frames to lists should preserve the size, meaning it should happen row wise. This ensures that it works correctly with `vec_c()`, and generally is part of the contract of `vec_cast()`.

``` r
library(vctrs)

x <- data.frame(x = 1:2, row.names = c("a", "b"))
x
#>   x
#> a 1
#> b 2

vec_cast(x, list())
#> [[1]]
#>   x
#> 1 1
#> 
#> [[2]]
#>   x
#> 1 2

vec_c(x, .ptype = list())
#> [[1]]
#>   x
#> 1 1
#> 
#> [[2]]
#>   x
#> 1 2
```

I also think that eventually `vec_get()` will be the function that does this. I think that it should never include the names that we would return from `vec_names()`. Meaning that it won't return the row names for data frames. So I've imitated that here.


R/type-bare.R
```diff
@@ -390,7 +390,15 @@ vec_cast.list.default <- function(x, to, ...) {
     return(vec_init(to, length(x)))
   }
 
-  out <- lapply(seq_along(x), function(i) x[[i]])
+  # FIXME - Use real `vec_get()` implementation
+  if (is.data.frame(x)) {
```
hadley: I think it would be better to add `vec_cast.list.data.frame()`?

NEWS.md
```diff
@@ -1,6 +1,9 @@
 
 # vctrs 0.2.0.9000
 
+* `vec_cast()` now casts data frames to lists row wise to preserve the invariant
```
lionel-: Suggestion:

`vec_cast()` now casts data frames to lists rowwise, i.e. to a list of data frames of size 1. This preserves the invariant of ...

DavisVaughan:
For `vec_cast.list.data.frame()`, if we had `vec_get()` and whatever its equivalent to `vec_chop()` would be then the full implementation would just be `vec_get_chop(x)`. As it currently stands:

- I remove row names because I think that is what `vec_get()` will do with data frames
- I strip the list_of attributes back off of the result of `vec_chop()`

We now also have `vec_cast.data.frame.list()`. It doesn't work with `list_of<>` yet, but that should change with https://github.com/r-lib/vctrs/pull/654

``` r
library(vctrs)

df <- data.frame(x = 1)

vec_cast(list(df, df), df)
#>   x
#> 1 1
#> 2 1

vec_cast(list_of(df, df), df)
#> Error: Can't cast `x` <list_of<data.frame<x:double>>> to `to` <data.frame<x:double>>.
```

PR #659, opened 2019-11-12
Remove L prefix after numbers in tests

krlmlr:
This isn't strictly necessary, but makes the code easier to read.

Follow-up to #642.


lionel-:
As I mentioned in the previous PR I think this diff is unnecessary.

PR #661, opened 2019-11-13
Expression support for hashing / equality

### LINKED ISSUE #643: vec_unique() fails on lists of objects containing non-vectors
In the current dev version, `vec_unique()` doesn’t work with lists of objects that are not vectors, or that contain non-vector elements in a recursive structure:

``` r
m <- glm(mpg ~ wt, data = mtcars)
vctrs::vec_unique(list(m, m))
#> Error: `x` must be a vector, not a `glm/lm` object

o <- list(x = expression(x))
vctrs::vec_unique(list(o, o))
#> Error: `x` must be a vector, not an expression vector
```

`glm` objects are an example of both, and can frequently appear in list columns in data frames. As `vec_unique()` gets more use behind the scenes in packages such as tidyr, this threatens to break code that works with list columns: e.g. https://github.com/tidyverse/tidyr/issues/735 seems to be a symptom of this behaviour.

<sup>Created on 2019-11-04 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

DavisVaughan:
Closes #643 
Also fixes tidyverse/tidyr#735

With #634 and #644 out of the way, I think we are in a good enough place to be able to support lists of expressions in hashing and equality functions. This means that `vec_unique()` will work as well.

Note that this isn't true support of "expressions as vectors" (i.e. `vec_size(expression())` is still an error), but instead just fixes a few bugs where expressions should be treated as recursive types.


tests/testthat/test-equal.R
```diff
@@ -232,6 +239,16 @@ test_that("not equal if attributes not equal", {
   expect_false(obj_equal(x1, x2))
 })
 
+test_that("can compare expressions", {
+  expect_true(obj_equal(expression(x), expression(x)))
+  expect_false(obj_equal(expression(x), expression(y)))
+})
+
+test_that("`na_equal` works with expressions", {
+  expect_true(obj_equal(expression(NA), expression(NA)))
```
hadley: This is the only thing I have qualms about (and it probably isn't that important), but I'd think that the `NA` of an expression would be `NULL`, like lists.

DavisVaughan: Oh I think i have the test wrong, but it works just like lists

``` r
obj_equal <- vctrs:::obj_equal

obj_equal(expression(NA), expression(NA), na_equal = TRUE)
#> [1] TRUE
obj_equal(list(NA), list(NA), na_equal = TRUE)
#> [1] TRUE

obj_equal(expression(NA), expression(NA), na_equal = FALSE)
#> [1] NA
obj_equal(list(NA), list(NA), na_equal = FALSE)
#> [1] NA

obj_equal(list(NULL), list(NULL), na_equal = TRUE)
#> [1] TRUE
obj_equal(expression(NULL), expression(NULL), na_equal = TRUE)
#> [1] TRUE

obj_equal(list(NULL), list(NULL), na_equal = FALSE)
#> [1] NA
obj_equal(expression(NULL), expression(NULL), na_equal = FALSE)
#> [1] NA
```

<sup>Created on 2019-11-13 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0.9000)</sup>

hadley: Oh the `NA` result works because you're recursing down the tree.

lionel-: Now I am surprised to see `NA` propagation for scalar equality. I don't think `obj_equal()` should have an `na_equal` parameter at all because NA propagation is a property of vector operations.

PR #667, opened 2019-11-13
Implement optional reproducible (but slower) character hashing

### LINKED ISSUE #107: Optional hashing of string values
Making it possible to use vctrs hashing for (e.g.) factor levels display, and eliminating the dependency on digest.

### LINKED ISSUE #664: Expose vec_hash()
It would be useful in ggplot2 where we're currently using digest just to compute a unique id for the legends (based on a number of components).

DavisVaughan:
Closes #107 
Closes #664 

@hadley this is what it could look like. Mainly I've renamed `hash_char()` -> `hash_char_sexp()` and then added a new `hash_char()` that does the reproducible character hash.

I think `obj_hash(pool = FALSE)` could be used internally for the factor labels.

It's a bit of a pain to thread `pool` through all of the function calls.


DavisVaughan:
Should have made this a draft PR, it's not at all ready (needs tests, exporting of `vec_hash()`, etc)

hadley:
Yeah that's pretty much what I was thinking. Depending on how much time you want to put into this, you might include the replacement of digest in this PR.

src/hash.c
```diff
@@ -46,19 +46,41 @@ static uint32_t hash_double(double x) {
...
+
+static uint32_t hash_char_sexp(SEXP x) {
   return hash_int64((intptr_t) x);
 }
 
+// djb2 - http://www.cse.yorku.ca/~oz/hash.html
+static uint32_t hash_char(SEXP x) {
+  uint32_t hash = 5381;
+
+  const char *string = CHAR(x);
```
hadley: What happens with NAs here?

DavisVaughan: It seems to match `"NA"`

``` r
library(vctrs)

vec_hash(NA_character_, pool = FALSE)
#> [1] 4d ee 90 9e

vec_hash("NA", pool = FALSE)
#> [1] 4d ee 90 9e
```

<sup>Created on 2019-11-13 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0.9000)</sup>

https://github.com/wch/r-source/blob/8019f9819095f98d4069459aa787b0d222a248e2/src/main/names.c#L1189

DavisVaughan: It looks like `digest()` get's around this when they `serialize()` input, which they do by default. But their results are identical if you don't serialize.

``` r
library(digest)

# Look different by default
digest(NA_character_)
#> [1] "55611e71b358a30158c61810ad802435"
digest("NA")
#> [1] "279f2abfb74bae808fc0ff984b295c33"

# Without serializing they look the same
digest(NA_character_, serialize = FALSE)
#> [1] "d4cd0dabcf4caa22ad92fab40844c786"
digest("NA", serialize = FALSE)
#> [1] "d4cd0dabcf4caa22ad92fab40844c786"

# Different because of the serialization
serialize(NA_character_, connection = NULL, version = 2L)
#>  [1] 58 0a 00 00 00 02 00 03 06 00 00 02 03 00 00 00 00 10 00 00 00 01 00
#> [24] 00 00 09 ff ff ff ff
serialize("NA", connection = NULL, version = 2L)
#>  [1] 58 0a 00 00 00 02 00 03 06 00 00 02 03 00 00 00 00 10 00 00 00 01 00
#> [24] 04 00 09 00 00 00 02 4e 41
```

DavisVaughan: I wonder if this would work. The idea would be that normally when we convert an individual `char` into an `int` (on the line `int chr = *string++`), that `chr` integer value is a value between 0-127 on the ASCII table. So if we use `-1` for `NA` maybe that would be enough?

https://www.cs.cmu.edu/~pattis/15-1XX/common/handouts/ascii.html

```
static uint32_t hash_char(SEXP x) {
  uint32_t hash = 5381;

  if (x == NA_STRING) {
    return ((hash << 5) + hash) - 1;
  }

  const char *string = CHAR(x);
  int chr = *string++;

  while (chr) {
    hash = ((hash << 5) + hash) + chr; /* hash * 33 + chr */
    chr = *string++;
  }

  return hash;
}
```

lionel-: What about the empty string `""`? IIUC `obj_hash("")` and `obj_hash(c("", ""))` will both be `0` currently, when bool/stable is TRUE.

R/hash.R
```diff
@@ -1,12 +1,29 @@
...
-vec_hash <- function(x) {
-  .Call(vctrs_hash, x)
+#' Hash a vector
+#'
+#' `vec_hash()` hashes a vector and returns the hash as a raw vector. Each
+#' element of `x` is coded with 32-bit hashes. This results in a raw vector
+#' that is 4 times the size of `x`.
+#'
+#' @param x A vector
+#' @param pool A logical. Should R's global string pool be used to hash
```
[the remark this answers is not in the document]
lionel-: It might make sense to use serialisation as in digest when `stable = TRUE`, since the serialisation format is versioned and stable.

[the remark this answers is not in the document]
lionel-: Really I'm uneasy about declaring our format stable though. Digest uses serialisation + well defined hashing algorithms to achieve stability. It feels wrong to commit to stability because our hashing routines are pieced together from disparate sources, we might want to change these pieces in the future.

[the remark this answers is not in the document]
hadley: I really don't think we should claim this is stable, and I don't see the problem with mentioning R internals when they are stable and germane to the problem at hand.

src/hash.c
hadley: I'm not sure that `vec_hash(NA) == vec_hash("NA")` is a problem. And if `obj_hash("") == 0` is a problem then we should just run it through our hash scramble.

R/hash.R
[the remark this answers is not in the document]
lionel-: They may be stable but they are not set in stone. Other R implementations would prefer it if R gave up pointer equality for determining identity of objects, it's not entirely impossible that this changes in the future.

More importantly it seems like this option is trying to make hashing safe, or in any case more useful, for a particular kind of usage. Finding a word correctly describing that usage might help avoiding improper use.

src/hash.c
DavisVaughan: I'm not sure I understand the empty string problem

``` r
vctrs:::obj_hash("", pool = FALSE)
#> [1] 77 08 6f 3c

vctrs:::obj_hash(c("", ""), pool = FALSE)
#> [1] 0c 29 d6 6b
```

DavisVaughan:
Some arbitrary benchmarks just to show the difference as string length increases

``` r
library(vctrs)

x <- rep("a", times = 1e6)
y <- rep("aaaaaaaaaaaaaaaaaaaaaa", times = 1e6)

# very little difference because we hash pointers
bench::mark(
  vec_hash(x, pool = TRUE),
  vec_hash(y, pool = TRUE),
  check = FALSE
)
#> # A tibble: 2 x 6
#> 1 vec_hash(x, pool = TRUE)   2.17ms    2.6ms      355.    3.92MB     47.8
#> 2 vec_hash(y, pool = TRUE)      2ms   2.45ms      390.    3.81MB     38.7

# much slower to hash the long strings directly, as expected
bench::mark(
  vec_hash(x, pool = FALSE),
  vec_hash(y, pool = FALSE),
  check = FALSE
)
#> # A tibble: 2 x 6
#> 1 vec_hash(x, pool = FALSE)   2.61ms   3.17ms     312.     3.81MB    34.6 
#> 2 vec_hash(y, pool = FALSE)  13.42ms  13.88ms      70.5    3.81MB     6.61
```

<sup>Created on 2019-11-14 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0.9000)</sup>

src/hash.c
lionel-: It seems there is none.

DavisVaughan:
Unresolved questions:

- Argument name? `pool` or `stable` or something else.
- Mention R's global string pool in the documentation of `pool`?

lionel-:
Let's just keep `pool` since that's an internal tool. This way we don't have to use the same argument if other similar parameters are added, and then we won't have to worry about consistency of interface or semantics.

lionel-:
Each time I comment on the vctrs repo, I immediately get a new mail from Github actions with:

```
Run failed for master (fe176cf)
```

Anyone else has this?

PR #668, opened 2019-11-14
Update roxygen2 and re-document

DavisVaughan:


man/fields.Rd
```diff
@@ -19,10 +19,10 @@ field(x, i) <- value
 \item{x}{A \link{rcrd}, i.e. a list of equal length vectors with unique names.}
 }
 \description{
-A \link{rcrd} behaves like a vector, so \code{length()}, \code{names()}, and \code{$} can
+A \link{rcrd} behaves like a vector, so \code{length()}, \code{names()}, and \verb{$} can
```
lionel-: It seems all operators are now `\verb{$}` or `\verb{[[}` instead of `\code{}`. Can we change it back?

lionel-: It is possible that we should just leave them as `verb`, see https://github.com/r-lib/roxygen2/issues/971#issuecomment-554938791

PR #670, opened 2019-11-15
Return a bare list from `vec_chop()`

### LINKED ISSUE #660: Returning `list_of<>` vs `list` in "low level" functions
I have a bit of a tough time deciding when to return a `list_of<>` from a vctrs function vs just a bare list. For example, `vec_chop()` returns a `list_of<>`:

``` r
vctrs::vec_chop(letters[1:2])
#> <list_of<character>[2]>
#> [[1]]
#> [1] "a"
#> 
#> [[2]]
#> [1] "b"
```

But this is slightly problematic because it is often used as a "low level" function where I really just want a bare list back. Like here in `as_df_row_impl()`

https://github.com/r-lib/vctrs/blob/2c8dd41312a8eadd7daf9729851ab1a50c667193/src/bind.c#L175

And here where I planned to implement `vec_cast.list.data.frame`

```r
vec_cast.list.data.frame <- function(x, to, ...) {
  row.names(x) <- NULL
  
  # returns list_of()
  out <- vec_chop(x)
  attributes(out) <- NULL
  
  # alternative - much slower
  #out <- map(vec_seq_along(x), vec_slice, x = x)
  
  out
}
```

@lionel- and I think that `vec_chop()` should just return a bare list, but we would like to be consistent about this and make a systematic change in all vctrs functions. For example, would it still make sense for `vec_split()` to return a `list_of<>`?

``` r
vctrs::vec_split(letters[1:4], c(1, 1, 2, 2))$val
#> <list_of<character>[2]>
#> [[1]]
#> [1] "a" "b"
#> 
#> [[2]]
#> [1] "c" "d"
```

@hadley what do you think about this?

DavisVaughan:
Closes #660 

There are two other places we use `list_of` output (below). I have not touched these. I'm not sure if they should continue to return `list_of`s or not. They seem slightly less "low level" compared to `vec_chop()`, so it might be okay to continue returning a `list_of`.

``` r
library(vctrs)
library(tibble)

as_tibble(vec_split(c("a", "b", "a"), c(1, 1, 2)))
#> # A tibble: 2 x 2
#>     key         val
#>   <dbl> <list<chr>>
#> 1     1         [2]
#> 2     2         [1]

as_tibble(vec_group_pos(c("a", "b", "a")))
#> # A tibble: 2 x 2
#>   key           pos
#>   <chr> <list<int>>
#> 1 a             [2]
#> 2 b             [1]
```

Note that `tidyr::nest()` returns a `list_of` column because of a usage of `vec_split()`.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -6,6 +6,9 @@
 * `vec_proxy_equal()` is now applied recursively across the columns of
   data frames (#641).
 
+* `vec_split()` no longer returns the `val` column as a `list_of`. It is now
+  returned as a bare list.
+  
 * Complex numbers are coercible with the integer and
   double (#564).
 
--- a/R/group.R
+++ b/R/group.R
@@ -23,7 +23,7 @@
 #'   * `vec_group_pos()`: A two column data frame with size equal to
 #'     `vec_size(vec_unique(x))`.
 #'     * A `key` column of type `vec_ptype(x)`
-#'     * A `pos` column of type `list_of<integer>`
+#'     * A `pos` column of type list, with elements of type integer.
 #'   * `vec_group_rle()`: A `vctrs_group_rle` rcrd object with two integer
 #'     vector fields: `group` and `length`.
 #'
--- a/R/slice.R
+++ b/R/slice.R
@@ -670,8 +670,8 @@ vec_init <- function(x, n = 1L) {
 #'   would be valid as an index in [vec_slice()]. If `NULL`, `x` is split into
 #'   its individual elements, equivalent to using an `indices` of
 #'   `as.list(vec_seq_along(x))`.
-#' @return A vector of type `list_of<vec_ptype(x)>` and size `vec_size(indices)`
-#'   or, if `indices == NULL`, `vec_size(x)`.
+#' @return A list of size `vec_size(indices)` or, if `indices == NULL`,
+#'   `vec_size(x)`.
 #' @export
 #' @examples
 #' vec_chop(1:5)
--- a/R/split.R
+++ b/R/split.R
@@ -8,7 +8,8 @@
 #' @param by Vector whose unique values defines the groups.
 #' @return A data frame with two columns and size equal to
 #'   `vec_size(vec_unique(by))`. The `key` column has the same type as
-#'   `by`, and the `val` column has type `list_of<vec_ptype(x)>`.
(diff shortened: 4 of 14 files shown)


DavisVaughan:
I mentioned this change to @romainfrancois because I noticed that he was adding this PR to dplyr to ensure that `group_split()` and `group_nest()` return a `list_of` like `nest()` currently does (because it uses `vec_split()`). https://github.com/tidyverse/dplyr/pull/4623

He mentioned this example where it's especially useful to return a `list_of` in these higher level functions where we might try to `unnest()` when we have an empty tibble, and can use the `ptype` attribute to get meaningful information.

```r
library(tidyverse)

iris %>% 
  group_nest(Species) %>% 
  filter(Species == "setosa") %>% 
  unnest(cols = data)
#> # A tibble: 50 x 5
#>    Species Sepal.Length Sepal.Width Petal.Length Petal.Width
#>    <fct>          <dbl>       <dbl>        <dbl>       <dbl>
#>  1 setosa           5.1         3.5          1.4         0.2
#>  2 setosa           4.9         3            1.4         0.2
#>  3 setosa           4.7         3.2          1.3         0.2
#>  4 setosa           4.6         3.1          1.5         0.2
#>  5 setosa           5           3.6          1.4         0.2
#>  6 setosa           5.4         3.9          1.7         0.4
#>  7 setosa           4.6         3.4          1.4         0.3
#>  8 setosa           5           3.4          1.5         0.2
#>  9 setosa           4.4         2.9          1.4         0.2
#> 10 setosa           4.9         3.1          1.5         0.1
#> # … with 40 more rows

iris %>% 
  group_nest(Species) %>% 
  filter(Species == "zfzfz") %>% 
  unnest(cols = data)
#> # A tibble: 0 x 5
#> # … with 5 variables: Species <fct>, Sepal.Length <dbl>,
#> #   Sepal.Width <dbl>, Petal.Length <dbl>, Petal.Width <dbl>
```

So I guess the idea can be that vctrs functions like `vec_chop()` and `vec_split()` return bare lists, but higher level functions can add on the `list_of` class if they want it? I think I like that because it provides the flexibility for both approaches to work (vctrs can use the low level versions directly, dplyr can build on them to make higher level functions), and we don't sacrifice any speed in vctrs with dispatch on the list_of class.

DavisVaughan:
@lionel- just requesting one last review. All splitting functions now return a bare list.

PR #678, opened 2019-11-19
Use `R_len_t n` in `vec_restore()`

### LINKED ISSUE #650: Consider letting `vec_restore()` take an R_len_t `n` rather than a SEXP
A little benchmarking of slide revealed that when using `vec_slice_impl()` in a tight loop we really pay for the allocation of the restoration size here

https://github.com/r-lib/vctrs/blob/5a0e6965ae833c203c73a679f203b36917d8a656/src/slice.c#L269

It looks like it ends up being about 20% of the total time I spent in `vec_slice_impl()` in my loop.

Screen Shot 2019-11-07 at 5 24 05 PM

If `vec_restore()` allowed the restore size to be a `R_len_t` we wouldn't have to do this.

It does feel a bit awkward for this to be a SEXP value. In `vec_chop()` I also would have benefited from this just being a `R_len_t` value (I currently have to manage a SEXP and its pointer to continually update the restore size).

We would have to make a decision about what a missing size should be. Right now we use `R_NilValue` but I'd vote for just using `-1` instead.

DavisVaughan:
Closes #650 

The main benefits from `vec_restore()` taking a `R_len_t n` value are that `vec_slice_impl()` and `vec_chop()` are slightly simpler because they don't have to manage a `SEXP` value for `n`.

There are two changes that were made to support this:
- `VCTRS_UNKNOWN_SIZE` is the replacement for passing `R_NilValue` to `vec_restore()`, and is set to `-1`
- We export `short_vec_restore()` as a callable (for slide) rather than `vec_restore()`

In the third commit I also found a pre-existing bug, and I'm not sure how it wasn't breaking things already.

### DIFF
--- a/src/bind.c
+++ b/src/bind.c
@@ -305,7 +305,7 @@ static SEXP vec_cbind(SEXP xs, SEXP ptype, SEXP size, enum name_repair_arg name_
   names = PROTECT(vec_as_names(names, name_repair, false));
   Rf_setAttrib(out, R_NamesSymbol, names);
 
-  out = vec_restore(out, type, R_NilValue);
+  out = vec_restore(out, type, VCTRS_UNKNOWN_SIZE);
 
   UNPROTECT(8);
   return out;
--- a/src/bind.c
+++ b/src/bind.c
@@ -305,7 +305,7 @@ static SEXP vec_cbind(SEXP xs, SEXP ptype, SEXP size, enum name_repair_arg name_
   names = PROTECT(vec_as_names(names, name_repair, false));
   Rf_setAttrib(out, R_NamesSymbol, names);
 
-  out = vec_restore(out, type, VCTRS_UNKNOWN_SIZE);
+  out = vec_restore(out, type, NA_INTEGER);
 
   UNPROTECT(8);
   return out;
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,6 +1,6 @@
 Package: vctrs
 Title: Vector Helpers
-Version: 0.2.0.9007
+Version: 0.2.0.9008
 Authors@R: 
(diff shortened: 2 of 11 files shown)


src/bind.c
```diff
@@ -305,7 +305,7 @@ static SEXP vec_cbind(SEXP xs, SEXP ptype, SEXP size, enum name_repair_arg name_
   names = PROTECT(vec_as_names(names, name_repair, false));
   Rf_setAttrib(out, R_NamesSymbol, names);
 
-  out = vec_restore(out, type, R_NilValue);
+  out = vec_restore(out, type, VCTRS_UNKNOWN_SIZE);
```
hadley: Could we just use `NA_INTEGER` instead?

DavisVaughan:
Does the R level `vec_restore()` need to default to `n = NA` now? Or is `n = NULL` still ok there?

hadley:
I'm not actually sure that using `NA_INTEGER` makes sense so you should make sure @lionel- thinks its reasonable before going any further.

lionel-:
We recently discussed switching back from `n` to the full `i` index vector. Maybe we should keep the SEXP wrapper while the API is maturing?

DavisVaughan:
Fine by me

DavisVaughan:
Two commits worth pulling out and merging anyways:

https://github.com/r-lib/vctrs/pull/678/commits/a60b63684a48efe521b9726dbfccc0984870ac0e

https://github.com/r-lib/vctrs/pull/678/commits/ec26470d901b5e14fad1138fef31ac90091c1e3d

lionel-:
Sounds good.

DavisVaughan:
Closing this, I've got my own TODOs to go back and add fixes for those two commits

PR #680, opened 2019-11-20
Clarify unique repair of `...` + a sequence of digits

DavisVaughan:
I was slightly confused by the `vec_as_names()` docs because it states that for `unique` repair:

"Empty names, and `...` or `..` followed by a sequence of digits are banned."

Are `...1` and `...2` really banned as named? It does not seem like it.

```r
library(vctrs)

vec_as_names(c("..1", "..2"), repair = "unique")
#> New names:
#> * `..1` -> ...1
#> * `..2` -> ...2
#> [1] "...1" "...2"

vec_as_names(c("...1", "...2"), repair = "unique")
#> [1] "...1" "...2"
```

### DIFF
--- a/R/names.R
+++ b/R/names.R
@@ -15,7 +15,7 @@
 #'   `minimal`.
 #'
 #' * `unique` names are `minimal`, have no duplicates, and can be used
-#'   where a variable name is expected.  Empty names, and `...` or
+#'   where a variable name is expected.  Empty names, and
 #'   `..` followed by a sequence of digits are banned.
 #'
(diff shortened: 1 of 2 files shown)


DavisVaughan:
Oh wait, I see what it is saying. It means these are not allowed:

- Empty names
- `...`
- `..1` or `..2`

I'll tweak the wording because I think it is slightly ambiguous

jennybc:
Just for background, here is how we got to this point:

https://design.tidyverse.org/names-attribute.html#unique-names

`..j` is a reserved word in R.

PR #681, opened 2019-11-21
Dispatch correctly on `y`

DavisVaughan:
The vignette dispatches on `x` in the `vec_ptype2.vctrs_decimal` generic, when it should dispatch on the second argument.


hadley:
Ooops

PR #690, opened 2019-11-26
Add methods for grouped-df

lionel-:
- Coercion methods: data frames are promoted to grouped-df. The common type has the union of the groups.

- Proxy and restore methods: Group columns are instrumented in the proxy for grouped-df. They are wrapped in a private data frame subclass so that the restore method recognises which columns are groups.

- `vec_rbind()` and `vec_cbind()` are updated to work on the proxies of inputs, and restore at the end. This way they support the instrumented proxies described above.

```r
vec_c(
  dplyr::group_by(mtcars, cyl),
  mtcars,
  dplyr::group_by(mtcars, cyl, am)
)
#> # A tibble: 96 x 11
#> # Groups:   cyl, am [6]
#>     mpg   cyl  disp    hp  drat    wt  qsec    vs    am  gear  carb
#>   <dbl> <dbl> <dbl> <dbl> <dbl> <dbl> <dbl> <dbl> <dbl> <dbl> <dbl>
#> 1  21       6   160   110  3.9   2.62  16.5     0     1     4     4
#> 2  21       6   160   110  3.9   2.88  17.0     0     1     4     4
#> 3  22.8     4   108    93  3.85  2.32  18.6     1     1     4     1
#> 4  21.4     6   258   110  3.08  3.22  19.4     1     0     3     1
#> # … with 92 more rows


vec_cbind(
  dplyr::group_by(mtcars[1:3], cyl),
  dplyr::group_by(mtcars, am, vs)
)
#> # A tibble: 32 x 14
#> # Groups:   cyl...2, vs, am [7]
#>   mpg...1 cyl...2 disp...3 mpg...4 cyl...5 disp...6    hp  drat    wt  qsec    vs    am
#>     <dbl>   <dbl>    <dbl>   <dbl>   <dbl>    <dbl> <dbl> <dbl> <dbl> <dbl> <dbl> <dbl>
#> 1    21         6      160    21         6      160   110  3.9   2.62  16.5     0     1
#> 2    21         6      160    21         6      160   110  3.9   2.88  17.0     0     1
#> 3    22.8       4      108    22.8       4      108    93  3.85  2.32  18.6     1     1
#> 4    21.4       6      258    21.4       6      258   110  3.08  3.22  19.4     1     0
#> # … with 28 more rows, and 2 more variables: gear <dbl>, carb <dbl>
```


lionel-:
cc @romainfrancois

lionel-:
The implications of this approach are still unclear. For now it seems that:

- The restore method should only be called on proxies.

- When an input is inspected for its values rather than manipulated on raw data, the proxy shouldn't be taken. In this case one approach might be e.g. `as.data.frame(arg)` to validate the input, and `vec_cast(out, arg)` before returning.

lionel-:
Commit https://github.com/r-lib/vctrs/pull/690/commits/0a3526d239674e00583ace99a741b8abb8437409 adds a cast method for unspecified, which is needed for tidyverse/tidyr#812 because some tidyr verbs creates unspecified columns.

PR #697, opened 2019-12-06
Use `is_unspecified()` in `vec_cast.list.default()`

DavisVaughan:
Before:

``` r
vctrs::vec_cast(NA, list())
#> [[1]]
#> [1] NA
```

After:

``` r
vctrs::vec_cast(NA, list())
#> [[1]]
#> NULL
```

We had a test for the `list(NA)` result, but I'm not sure that's correct?

Open question, what should this return?

``` r
vctrs::vec_cast(c(NA, 1), list())
#> [[1]]
#> [1] NA
#> 
#> [[2]]
#> [1] 1
```

I feel like this should be `list(NULL, 1)`? It would still be reversible, i.e. `vec_cast()`ing back to a double would give the original result.

### DIFF
--- a/R/type-bare.R
+++ b/R/type-bare.R
@@ -413,7 +413,7 @@ vec_cast.list.list <- function(x, to, ...) {
 #' @export
 #' @method vec_cast.list default
 vec_cast.list.default <- function(x, to, ...) {
-  if (inherits(x, "vctrs_unspecified")) {
+  if (is_unspecified(x)) {
     return(vec_init(to, length(x)))
   }
 
--- a/R/type-bare.R
+++ b/R/type-bare.R
@@ -419,6 +419,8 @@ vec_cast.list.default <- function(x, to, ...) {
 
   out <- lapply(seq_along(x), function(i) x[[i]])
 
+  vec_slice(out, vec_equal_na(x)) <- list(NULL)
+
   if (!is.object(to)) {
     out <- shape_broadcast(out, to)
   }
@@ -433,6 +435,9 @@ vec_cast.list.data.frame <- function(x, to, ...) {
   # equivalent for `vec_get()`
   row.names(x) <- NULL
   out <- vec_chop(x)
+
+  vec_slice(out, vec_equal_na(x)) <- list(NULL)
+
   out
 }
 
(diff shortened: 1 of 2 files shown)


lionel-:
I think we used to consider `vec_cast(x, list())` would be equivalent to `as.list()`, and in this case we'd like to preserve the values of all elements. So the idea was that `NA` would be converted to `NULL` for the list-of case but not the list case.

However, nowadays we have `vec_chop()` for that purpose, which I think is a better primitive:

```r
vec_chop(c(NA, 2, 3))
#> <list_of<double>[3]>
#> [[1]]
#> [1] NA
#>
#> [[2]]
#> [1] 2
#>
#> [[3]]
#> [1] 3
```

Though it should probably return a bare list rather than a list-of.

lionel-:
It's a bit weird that in the other direction we have two possible missing values, but I guess that's ok. One kind of missing value comes from manipulating the atomic type, and the other kind comes from manipulating the list:

```r
vec_cast(list(NA, NULL, TRUE), logical())
#> [1]   NA   NA TRUE
```

DavisVaughan:
Regarding: https://github.com/r-lib/vctrs/pull/697#issuecomment-564935034

- I've made it so that `vec_cast(x, list())` replaces `NA` values in `x` with `NULL` in the resulting list.

- `vec_chop()` does now return a bare list, so it can be used as the way to preserve the values of all elements.

---

I've made an additional tweak to ensure that data frames can be cast to shaped lists.

Casting arrays to lists is not going to work right now. Part of that has to do with using `x[[i]]` and `seq_along(x)` on the array, which results in the wrong sizes. The other part has to do with `vec_equal_na()` not working for arrays (#722). It is too much work to fix all of that here.

---

Also noting that if we had the `vec_chop()` equivalent for `vec_get()`, then we could again get rid of `vec_cast.list.data.frame()` and only have:

```r
vec_cast.list.default <- function(x, to, ...) {
  if (is_unspecified(x)) {
    return(vec_init(to, length(x)))
  }
  
  out <- vec_get_chop(x) # `vec_chop()` equivalent for `vec_get()`
  
  vec_slice(out, vec_equal_na(x)) <- list(NULL)
  
  if (!is.object(to)) {
    out <- shape_broadcast(out, to)
  }
  
  out
}
```

PR #698, opened 2019-12-06
Export `vec_is_vector()` not `vctrs_is_vector()`

DavisVaughan:
I spoke with @romainfrancois and it seems he wants `vec_is_vector()` not `vctrs_is_vector()` (so it can just return a bool not a SEXP)

Used here where he does the reverse dance back to a bool
https://github.com/tidyverse/dplyr/pull/4646/files#diff-f5eb4fcb92142d5a2359648c78adb5ffR98

### DIFF
--- a/src/init.c
+++ b/src/init.c
@@ -96,6 +96,9 @@ extern SEXP vec_recycle(SEXP, R_len_t);
 extern SEXP compact_seq(R_len_t, R_len_t, bool);
 extern SEXP init_compact_seq(int*, R_len_t, R_len_t, bool);
 
+// Extremely experimental (for dplyr)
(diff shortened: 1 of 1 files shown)


lionel-:
@DavisVaughan Actually I'm not so sure about this, because bool in C99 and in C++ are not the same type. While there might not be any issue in practice, especially if the same compilers are used to compile vctrs and dplyr, I think it would be more prudent to return an `int` here.

cc @romainfrancois

lionel-:
oh or we can just say client packages have to use `extern "c"` to declare the C callables, and that's what we would do in our exported header file as well if we detect C++.

PR #699, opened 2019-12-06
Only run actions on PRs or the master branch

jimhester:
Instead of running checks on both the feature branch and the pull request this will only  them on pull requests against the master branch and commits to the master

### DIFF
--- a/.github/workflows/R-CMD-check.yaml
+++ b/.github/workflows/R-CMD-check.yaml
@@ -1,4 +1,10 @@
-on: [push, pull_request]
+on:
+  push:
+    branches:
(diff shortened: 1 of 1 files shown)


lionel-:
I sometimes send PRs to feature branches when I have dependent branches. Can we still get checks on those?

jimhester:
Ok, I believe it now will do what you want. If you typically use a common prefix for these feature branches we could also put that in there instead of the `**` wildcard to match everything. You have to use `**` because `*` will not match slashes.

lionel-:
Thanks!

PR #700, opened 2019-12-08
Make `vec_chop()` compatible with `compact_seq()`

DavisVaughan:
I would like to register the C callable for `vec_chop()` and be able to pass in a list of compact seqs. The first step to this is to make `vec_chop()` compatible with compact seqs. It was nearly there, this finishes it off and adds tests.

### DIFF
--- a/src/slice.c
+++ b/src/slice.c
@@ -805,6 +805,16 @@ static SEXP vec_chop_base(SEXP x, SEXP indices, struct vctrs_chop_info info) {
       Rf_errorcall(R_NilValue, "Can't slice a scalar");
     }
 
+    if (info.has_indices) {
+      for (int i = 0; i < info.out_size; ++i) {
+        SEXP index = VECTOR_ELT(indices, i);
+
+        if (is_compact(index)) {
+          SET_VECTOR_ELT(indices, i, compact_materialize(index));
+        }
+      }
+    }
+
     if (has_dim(x)) {
       return chop_fallback_shaped(x, indices, info);
     }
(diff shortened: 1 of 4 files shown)


src/slice.c
```diff
@@ -761,6 +761,27 @@ static SEXP vec_chop_base(SEXP x, SEXP indices, struct vctrs_chop_info info);
 
 static SEXP vec_as_indices(SEXP indices, R_len_t n, SEXP names);
 
+// [[ register() ]]
+SEXP vctrs_chop_seq(SEXP x, SEXP starts, SEXP sizes, SEXP increasings) {
+  int* p_starts = INTEGER(starts);
+  int* p_sizes = INTEGER(sizes);
+  int* p_increasings = LOGICAL(increasings);
```
lionel-: Is it a bit strange to have increasings as plural? It sounds odd to my (admittedly non-native) ears.

DavisVaughan: It is. If we had a way to make a compact seq on the R side I would just do `vec_chop(x, list(compact_seq()))` in the tests, but we don't so I figured this was the second best option. Since it is only used in testing I wasn't too worried about the name

DavisVaughan:
I was going to export `compact_seq()` and `compact_rep()` to the R side for direct use in testing, but then you'd actually need `vec_slice_impl()` and `vec_chop_impl()` exported as well, because they bypass the `vec_as_index()` checks that `vec_slice()` and `vec_chop()` do, which don't work with compact objects.

Because that is just as much work as what we currently have, I don't think it is worth any changes

PR #711, opened 2019-12-19
Draft tabular API

lionel-:


src/table.c
```diff
@@ -0,0 +1,104 @@
...
+SEXP tbl_slice(SEXP x, SEXP index) {
+  tbl_assert(x);
+  SEXP proxy = PROTECT(vec_proxy(x));
+
+  if (TYPEOF(proxy) != VECSXP) {
+    Rf_error("Internal error: Expected list in tabular proxy.");
+  }
+
+  SEXP names = PROTECT(r_names(x));
+  index = PROTECT(vec_as_index(index, vec_size(x), names));
```
DavisVaughan: `vec_size(x) -> vec_size(proxy)`? i.e. we don't want the number of rows, we want the number of columns?

DavisVaughan: This makes me wonder if we finally need a "getter" for "number of columns"

src/table.c
```diff
@@ -0,0 +1,104 @@
...
+
+  SEXP sliced_proxy = PROTECT(list_slice(x, index));
```
DavisVaughan: `list_slice(x, index) -> list_slice(proxy, index)`?

src/table.c
lionel-: oh yeah without this you can't reason about the final width after e.g. a cbind. I would call it `tbl_size()` though, unless we rename `vec_size()` to `vec_height()`.

src/table.c
lionel-: Good catch. Also need to repair names after slicing.

src/table.c
DavisVaughan: I thought about `tbl_size()` too, but I figured that a "table" still has a size that is the number of rows.

`vec_height()` feels wrong for a 1D vector like `1:5`.

One thought is to alias `vec_size()` as `tbl_height()` (maybe with a `tbl_assert()` check), and also have `tbl_width()`. Then you at least have a pair of functions that make sense together

src/table.c
```diff
@@ -0,0 +1,104 @@
...
+  Rf_copyMostAttrib(sliced_proxy, proxy);
+
+  names = PROTECT(slice_names(names, index));
+  Rf_setAttrib(sliced_proxy, R_NamesSymbol, names);
+
+  SEXP row_names = PROTECT(get_rownames(x));
+  Rf_setAttrib(sliced_proxy, R_RowNamesSymbol, row_names);
```
DavisVaughan: Are these set already by `Rf_copyMostAttrib()`?

lionel-: No, copy-most-attrib copies everything except names, row names, and dims.

src/table.c
lionel-: I think `vec_height()` makes sense for 1D vectors, because vctrs focuses on data frames.

> I figured that a "table" still has a size that is the number of rows.

When tables are conceived as vectors of rows, `vec_size()` is the relevant operator.

I don't really mind `tbl_width()`, though I prefer making the vec and tbl API symmetric.

lionel-:
Added `tbl_ptype2()` and `tbl_ptype_common()`. The latter is now used in `vec_cbind()` instead of `vec_ptype_common()`.

NEWS.md
```diff
@@ -1,6 +1,11 @@
 
 # vctrs 0.2.99.9000
 
+* New experimental API for tabular vectors. It currently consists in
```
hadley: Is "tabular vector" the right name for this? I'm slightly worried that `tbl` is going to be easily confused with tibbles.

R/table.R
```diff
@@ -0,0 +1,105 @@
+#' Generics for tabular vectors
+#'
+#' @description
+#'
+#' \Sexpr[results=rd, stage=render]{vctrs:::lifecycle("experimental")}
+#'
+#' These generics make it possible to work with tables, i.e. vectors
```
hadley: What calling them rectangular vectors? To me, that more clearly implies that each element is the same size.

R/table.R
```diff
@@ -0,0 +1,105 @@
...
+#' of dimension 2 like data frames and matrices. The main motivation
+#' of this API is to support colwise operations such as slicing
+#' columns. Operating across columns requires slightly different
+#' semantics and concepts than rowwise operations:
+#'
+#' - A tabular prototype is a zero-rows, zero-cols table. In order for
```
hadley: Hmmm, this implies that a `grouped_df` has to be able to record that it groups columns that it doesn't contain. That feels a bit weird.

R/type-tibble.R
```diff
@@ -20,3 +20,30 @@ vec_ptype2.tbl_df.data.frame <- function(x, y, ..., x_arg = "x", y_arg = "y") {
...
+
+tbl_ptype2.tbl_df.data.frame <- function(x, y, ..., x_arg = "x", y_arg = "y") {
+  if (inherits_only(y, "data.frame")) {
+    tibble::tibble()
+  } else {
+    stop_incompatible_type(x, y, x_arg = x_arg, y_arg = y_arg)
+  }
+}
+tbl_ptype2.data.frame.tbl_df <- function(x, y, ..., x_arg = "x", y_arg = "y") {
+  if (inherits_only(y, c("tbl_df", "tbl", "data.frame"))) {
```
hadley: Should this be `inherits_exactly()`?

R/type-tibble.R
```diff
@@ -20,3 +20,30 @@ vec_ptype2.tbl_df.data.frame <- function(x, y, ..., x_arg = "x", y_arg = "y") {
...
+  } else {
+    stop_incompatible_type(x, y, x_arg = x_arg, y_arg = y_arg)
+  }
+}
+tbl_ptype2.tbl_df.default <- function(x, y, ..., x_arg = "x", y_arg = "y") {
+  stop_incompatible_type(x, y, x_arg = x_arg, y_arg = y_arg)
+}
+
+tbl_ptype2.tbl_df.data.frame <- function(x, y, ..., x_arg = "x", y_arg = "y") {
+  if (inherits_only(y, "data.frame")) {
```
hadley: I find these a bit to hard to reason about because of `x` is checked in the generic, and the class of `y` is checked in the method.

R/table.R
```diff
@@ -0,0 +1,105 @@
...
+}
+
+#' @rdname tbl_is
+#' @export
+tbl_size <- function(x) {
+  .Call(vctrs_tbl_size, x)
+}
+
+#' @rdname tbl_is
+#' @inheritParams vec_slice
```
hadley: Why do we need a separate `tbl_slice()`?

R/table.R
```diff
@@ -0,0 +1,105 @@
...
+#' @rdname tbl_is
+#' @inheritParams vec_assert
+#' @export
+tbl_assert <- function(x, arg = as_label(substitute(x))) {
+  .Call(vctrs_tbl_assert, x, arg)
+}
+
+#' @rdname tbl_is
+#' @export
+tbl_size <- function(x) {
```
hadley: Can't we just use `vec_size()` here? Or is the size of a rectangle it's width?

R/table.R
lionel-: I'm not worried about confusion with tibbles, nor with data tables. A table is a generic term. 

So far I've enjoyed writing about "tabular" vectors to mean vectors with two directions. I don't think I would like "rectangular" as much as it's a long word, but if you prefer rectangular we can go for it. Regarding the prefix, `rect_` has the downside of being 4 letters, and `rec_` might evoke records, which is a different (though related) concept.

R/table.R
lionel-: I think that's only the case for static grouped-df, which makes sense. A prototype of dynamic grouped-dfs doesn't refer to any grouping variable.

R/table.R
lionel-: This is the width: `tbl_size(vec_cbind(x, y)) === tbl_size(x) + tbl_size(y)`.

R/table.R
lionel-: To slice across columns with vctrs semantics (for instance supporting proxies with virtual columns). This will be needed in tibble to implement `[`.

R/type-tibble.R
lionel-: After writing the grouped-df and tbl-df methods, I think this makes it easier to write these methods by following a simple rule. Otherwise this might get too boiler-platy, making it easier to make a predicate mistake in one of the branches.

R/type-tibble.R
lionel-: This is from an old version of rlang, 0.2.0 I believe. Do you think it's worth to rename?

R/table.R
lionel-: I think the main problem of "rectangular" is that it is a shape qualifier, thus it only makes sense as a qualifier of vector. In that regard I prefer "tabular" because it makes it natural to talk about tables as nouns, to collectively refer to matrices and data frames.

On the other hand it's not clear yet that we want to handle matrices as part of this API, currently they are an error. Maybe the API should be purely about data frames, and in this case maybe `df_` is better.

It would be too much work to rename in this PR as other branches are depending on it, so we still have time to think about this.

lionel-:
So far this approach works with dynamically grouped data frames that have instrumented columns in proxies, i.e. the grouping variables are wrapped in a special df-col by the proxy method, so that  their grouping status can be reinstated by the restore method.

However I'm hitting a wall with statically grouped data frames. In the following example the RHS can't be coerced to the statically grouped type because it doesn't contain the grouping variables:

```r
static_gdf <- dplyr::group_by(mtcars, cyl, .drop = FALSE)

vec_cbind(static_gdf, mtcars[10:11])
#> Error: Can't find grouping variables in data.
```

When we cast to the rectangle/tabular type of a statically grouped data frame, we apply the `group_data()` on the input. If the input does not contain the grouping variables, we can't compute group identifiers for each row. And these identifiers are not part of the `tbl_type()` of the statically grouped df since it is empty and contains no data. Hence the error.

The problem stems from having the grouping structure separate from the data. So if we want to allow removing grouping columns but still having them active, we'll have a similar error.

Not sure how to solve this. The easy way is to decide that:

1. You have to ungroup statically grouped data frames before concatenation (or other combining operations) with data frames of different types.

2. External grouping variables are not possible, or at least they prevent combination with vectors that have a different type.

Another way would be to include the row structure in tabular prototypes. In that case `new_data_frame(list(), n = 0L)` and `new_data_frame(list(), n = 3L)` would be two different types. It sort of makes sense as this would be symmetric to the vector types of data frames which include the column structure. This would make it a bit more tricky to write tabular/rectangle methods, but it might make the coercion system for colwise operations more flexible and useful.

R/table.R
hadley: The main problem is that `tbl` is the super class that tibbles use — I think the chance of confusion are high. And "table" already has existing meaning and class (`table()`)

(If I succeed in convincing you that rectangular or rectangle is a good name, I'd suggest `rct` as a prefix)

R/table.R
hadley: But it has to list the columns that it would apply to? Otherwise where is that information stored?

R/table.R
hadley: But isn't this already supported by `vec_slice()`?

R/table.R
hadley: I think using `size` here again might be a bad idea?

hadley:
Won't `vec_rbind()`ing two grouped data frames need tabular support?

R/table.R
lionel-: ok, I'll change to rectangle in an ulterior PR.

R/table.R
lionel-: For static gdf the grouping information is stored in the _type_. For dynamic gdf the grouping information is stored in the _data_ (instrumented df-cols in the proxy).

R/table.R
lionel-: `vec_slice()` slices across rows, `rct_slice()` slices across columns. If you coerce the input to list and use `vec_slice()` to slice across columns, you lose the proxy semantics of the input class.

R/table.R
lionel-: Davis suggested `rct_width()`. If we agree that keeping the API symmetric is better, this gives us the choice between `vec_size()` / `rct_size()` and `vec_height()` / `rct_width()`. In the latter case we'd have to rename `vec_size()` to `vec_height()`.

lionel-:
> Won't vec_rbind()ing two grouped data frames need tabular support?

No because this is a vector operation. The coercion to rows is done by an ad hoc function, not by the rectangle API. Then the relevant type for rowwise combination is the vector type, which includes the column structure.

By contrast, the relevant type for colwise combination is the rectangle type, which will includes the rowwise structure (once I switch to `rct_ptype(mtcars) === new_data_frame(list(), n = nrow(mtcars))` as outlined in the last paragraph of https://github.com/r-lib/vctrs/pull/711#issuecomment-568178172).

R/type-data-frame.R
```diff
@@ -195,7 +195,16 @@ tbl_cast.data.frame <- function(x, to, ..., x_arg = "x", to_arg = "to") {
...
 tbl_cast.data.frame.data.frame <- function(x, to, ...) {
-  as.data.frame(x)
+  # This is a departure from vector casting which is currently more
+  # liberal. Casting requires existence of a common type,
+  # e.g. compatible row names.
+  ptype <- tbl_ptype2(x, to)
+
+  x <- as.data.frame(x)
+  x <- vec_recycle(x, vec_size(to))
+  attr(x, "row.names") <- attr(x, "row.names")
```
DavisVaughan: Is this a typo? row names of x to row names of x?

lionel-: Fixed thanks

lionel-:
The row structure (nrow and row names) is now included in the frame prototypes. For data frames, row names must match (possibly out of order) to get a common type.

TODO:

* Custom conditions for failed cast and common type.
* tbl-cast should reorder input based on row names.

PR #726, opened 2020-01-03
Add `x_arg` parameter to `vec_recycle()`

### LINKED ISSUE #706: Add `x_arg` to vec_recycle()
So we can eventually generate informative errors like https://github.com/tidyverse/dplyr/pull/4416

lionel-:
Closes #706


R/conditions.R
```diff
@@ -398,20 +398,32 @@ stop_scalar_type <- function(x, arg = NULL) {
...
     .subclass = "vctrs_error_recycle_incompatible_size"
   )
 }
 
+#' @export
+cnd_header.vctrs_error_recycle_incompatible_size <- function(cnd, ...) {
+  x_arg <- cnd$x_arg
+  size <- cnd$size
+
+  glue::glue("`{x_arg}` can't be recycled to size {size}.")
```
hadley: Would be simpler to use `glue_data()` here (and below)

R/utils.R
```diff
@@ -121,3 +121,7 @@ new_opts <- function(x, opts, subclass = NULL, arg = NULL) {
     class = c(subclass, "vctrs_opts")
   )
 }
+
+glue_c <- function(..., env = caller_env()) {
```
hadley: What's the point of this function?

lionel-: That's glue + c, to create named vectors with glue interpolation.

hadley: My sense is that this should probably not yet be a function (just inline into `cnd_body.vctrs_error_recycle_incompatible_size`) or should be more specific: `glue_data_bullets()`. But I don't care that much.

R/conditions.R
lionel-: Good point

R/utils.R
lionel-: I like it, this is now `glue_error_bullets()`

PR #727, opened 2020-01-03
Pass on the restored data frame to `vec_restore_dispatch()`

DavisVaughan:
This was a typo from a while back.

When restoring to non-bare data frames, we first call `vctrs_df_restore()` then call the restore dispatch method, but we weren't passing through that restored df.

### DIFF
--- a/src/proxy-restore.c
+++ b/src/proxy-restore.c
@@ -131,7 +131,7 @@ SEXP vec_restore(SEXP x, SEXP to, SEXP n) {
     // first restore data frames as such before calling the restore
     // method, if any
     SEXP out = PROTECT(vctrs_df_restore(x, to, n));
-    out = vec_restore_dispatch(x, to, n);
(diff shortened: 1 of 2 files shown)


DavisVaughan:
Unrelated errors in the CI about a URL 404:

```r
Error in utils::download.file(url, path, method = download_method(), quiet = quiet,  : 
  cannot download all files
Calls: cat ... get_cran_deps -> download_json -> download -> <Anonymous>
In addition: Warning message:
In utils::download.file(url, path, method = download_method(), quiet = quiet,  :
  URL 'https://crandeps.r-pkg.org:443/deps/vctrs,R,ellipsis,digest,glue,rlang,bit64,covr,crayon,generics,knitr,pillar,pkgdown,rmarkdown,testthat,tibble,xml2,zeallot': status was '404 Not Found'
```

Issue #606, opened 2019-10-02
Make `vec_cast()` a directional coercion generic


lionel-:
There are two issues with the current casting rules in vctrs:

- They make vctrs complicated. Users and class implementors need to know about two diagrams for coercions and casts.

- The casting rules do not seem to be the right default most of the time. For instance input validation can't use `vec_cast()` because it is too flexible.

To fix this, we could make `vec_cast()` a directional coercion. 

- A __non-directional__ coercion converts an object to the common type, which is always the larger type (i.e. integer is coerced to double).

- A __directional__ coercion converts an object to the type supplied by the caller, but only if there exists a common type. A directional coercion is no longer guaranteed to convert to the larger type. If the cast is lossy, this is an error by default, unless the code is wrapped in `vctrs::allow_lossy_cast()`.

Essentially, we'd add the rule that `vec_cast()` methods can only be implemented for coercible types, i.e. types that have a common type specified by `vec_ptype2()` methods.

### Cast methods to remove

It is unclear yet how much of a breaking change it would be to remove these methods, we might need a deprecation strategy:

- character casts
- list casts
- double <-> date

Where should this conversion logic go? Either we move it to separate single-dispatch generics like `vec_as_double()`, or we add a new double-dispatch generic in vctrs. @davisvaughan suggested the name `vec_force()`.

- Even if we have a double-dispatch generic, it is nice to have the explicit is_ and as_ functions for a class, so we might end up with separate wrappers anyway.

- The separate generics have the advantage of requiring only a single dispatch, which makes things faster for users and simpler for class authors.

- At a syntax level, having a single double-dispatch generic seems to give a false sense of principled genericity, whereas most of the implemented behaviour is ad hoc.

- On the other hand, `vec_force()` could be useful as part of a deprecation plan.

DavisVaughan:
Just in terms of terminology, does this make: 

- `vec_cast_common()` non-directional
- `vec_cast_common(.ptype = <ptype>)` directional
- `vec_cast(x, ptype = <ptype>)` directional

DavisVaughan:
I just wanted to clarify something that I think is an important point here, and while I think @lionel- understands this well, it wasn't clear to me until now.

I'm going to call this directional coercion generic `vec_coerce()` for the moment, rather than `vec_cast()`.

As a directional coercion generic:

1. If there is a common type you are _always_ allowed to coerce up into the larger type. Ex. `integer() -> numeric()`

[...]

PR #728, opened 2020-01-03
Add missing `vec_cast()` methods for dates and datetimes

### LINKED ISSUE #715: Missing `vec_cast()` methods for date / datetime
- `vec_cast.character.Date()`
- `vec_cast.character.POSIXct()`
- `vec_cast.character.POSIXlt()`

Because their corresponding `double` methods exist:
- `vec_cast.Date.integer()`
- `vec_cast.POSIXct.integer()`
- `vec_cast.POSIXlt.integer()`

Should you be able to cast from Date/POSIXt -> double or integer as well to be symmetric here?

DavisVaughan:
Closes #715 

Should we also add the following methods for coercing to double/integer too?

- `vec_cast.integer.Date()`
- `vec_cast.double.Date()`
- `vec_cast.integer.POSIXct()`
- `vec_cast.double.POSIXct()`
- `vec_cast.integer.POSIXlt()`
- `vec_cast.double.POSIXlt()`

We don't have `vec_cast.integer.factor()` or `vec_cast.double.factor()` so I'd imagine no, but maybe we just don't have those because we really want a mental model of a factor as a restricted character vector?


hadley:
I think casting to and from integers and doubles might be too relaxed?

DavisVaughan:
I think that makes sense. That means that the double -> date/datetime methods that were already there should be removed, which I just did.

lionel-:
This might have to wait for vctrs 0.3.0 since this restricts existing coercions.

Also, adding coercion methods for character might conflict with #606. The current idea is to remove most coercions from and to characters.

lionel-:
Following last comments on Slack, it's probably less work to just keep `vec_cast()` as an arbitrary conversion generic, and introduce `vec_coerce()` for functions taking `.ptype` parameters, and for validating argument types in functions.

In that case, it's fine to include these character methods.

DavisVaughan:
Closed in favor of #741 

See https://github.com/r-lib/vctrs/issues/715#issuecomment-573666136 for reasoning

Issue #729, opened 2020-01-04
Reviewing the subscript conversion UI


lionel-:
The subscript/index conversion UI is challenging because it needs to be flexible enough to cover tidyselect, vctrs, and tibble usages. This is a proposal for making the naming more consistent and the functions more easily configurable. We'd rename:

```r
vec_coerce_index()    -> vec_as_subscript()
vec_as_index()        -> vec_as_location()

vec_coerce_location() -> vec_as_single_subscript()
vec_as_location()     -> vec_as_single_location()
```

The `as_location` functions convert to a numeric location in the vector. The `as_subscript` functions only convert to the base type of the three kinds of subscript vector:

* name → character
* indicator → logical
* location → integer (or double in the future when we add long vector support).

For example, factors are converted to character because all subtypes of characters are considered names.

`vec_as_subscript()` has options for ignoring or throwing an error for particular subscript types. `vec_as_location()` would have the same signature as `vec_as_subscript()`, with this definition:

```r
vec_as_location <- function(x,
                            n,
                            names = NULL,
                            indicator = c("convert", "error", "ignore"),
                            location = c("convert", "error", "ignore"),
                            name = c("convert", "error", "ignore")) {
  x <- vec_as_subscript(x, indicator = indicator, location = location, name = name)
  switch(typeof(x),
    logical = lgl_as_location(x, n, names),
    double = , # For long vectors
    integer = num_as_location(x, n, names),
    character = chr_as_location(x, n, names)
  )
}
```

It first calls `vec_as_subscript()` to get the base type, then calls specific conversion functions to convert to a numeric location. The specific conversion functions have additional options to control how values are converted:

```r
num_as_location <- function(x,
                            n,
                            names = NULL,
                            negative = c("invert", "error", "ignore"),
                            missing = c("error", "ignore")) {
  ...
}
num_as_single_location <- function(x,
                                   n,
                                   names = NULL,
                                   negative = c("error", "ignore"),
                                   missing = c("error", "ignore")) {
  ...
}
```

This is a departure from the current UI which tries to cram everything in `vec_as_location()` (currently called `vec_as_index()`). I think it's a good principle to split into multiple functions as soon as the parameters become trees (or flattened trees with prefixes).

`vec_as_index()` would go into a long deprecation cycle because it's used in CRAN tibble.

cc @DavisVaughan @hadley @krlmlr

krlmlr:
I can submit a patch release for tibble if it helps.

hadley:
[...]

PR #734, opened 2020-01-08
Review subscript UI

lionel-:
Part of #729.

* Split `allow_types` into separate `indicator`, `location`, and `name` parameters that can take one of `"coerce"` or `"error"`.

  (We might want to rename coerce to `"cast"` depending on what we decide for the semantics of `vec_cast()`).

* Narrowing coercions are now allowed, instead of just enriching promotions. I think this makes sense as long as the coercion is not lossy, and lossy coercions should be errors.

Tests are fixed in the next PR.


R/subscript.R
```diff
@@ -9,57 +9,65 @@
 #' not checked in any way (length, missingness, negative elements).
 #'
 #' @inheritParams vec_as_location
+#' @param indicator,location,name How to handle indicator (logical),
+#'   location (numeric), and name (character) subscripts. If
+#'   `"coerce"`, the subscript is coerced to its base type,
```
hadley: I found the coercion wording a bit hard to understand — I think you need a sentence describing the common case (i.e. it's a base vector) before talking about S3 vectors.

PR #735, opened 2020-01-08
Review location UI

### LINKED ISSUE #729: Reviewing the subscript conversion UI

lionel-:
Closes #729. Branched from #734.

The conversion of numeric locations (allowing negative values) is now parameterised in `num_as_location()` and `num_as_location2()`.


R/subscript-loc.R
```diff
@@ -62,70 +54,110 @@ vec_as_location <- function(i,
...
-#'   negative or missing.
+#' @param missing Whether to throw an `"error"` when `i` is a missing
+#'   value, or `"ignore"` it.
 #' @export
 vec_as_location2 <- function(i,
                              n,
                              names = NULL,
                              ...,
-                             allow_values = NULL,
+                             missing = c("error", "ignore"),
```
hadley: What does ignoring a value mean in this context? i.e. what gets returned instead?

lionel-: The value is returned, so a missing value in this case. I can see how `"ignore"` can be confusing. Would `"allow"` be better? The challenge is to find a term that describes the base case in other validation functions, for consistency.

hadley: That's fine; just documentation that it returns an `NA`

PR #748, opened 2020-01-14
Fix issue where vctrs functions imported in tibble can't be found

### LINKED ISSUE #747: Library loading issues with dev tibble and vctrs
I don't know exactly _why_ this happens, but I think I know a way to fix it. I'm more curious as to what's going on.

tibble used to use `@import vctrs`. If you have a version of tibble installed that does that, then you can `devtools::load_all()` vctrs with no problem.

tibble just recently switched to being more explicit, using `@importFrom vctrs ..fns..`. With the current dev tibble installed, I get this when loading vctrs with `load_all()`:

```r
> devtools::load_all(".")
Loading vctrs
Error: object ‘vec_slice<-’ is not exported by 'namespace:vctrs'
```

This isn't specific to `vec_slice<-`, it just happens to be `vec_slice<-` because of the order the imports are specified in tibble's `@importFrom` tags.

A traceback for the error sends me here, to where we call out to the tibble namespace in a test file (but not wrapped in a function)
https://github.com/r-lib/vctrs/blob/2e5416e759a65576e40daba9423c64fc95fb72f0/tests/testthat/helper-s3.R#L44

Wrapping that in a function like this seems to fix the issue

```r
tibble <- function(...) {
  tibble::tibble(...)
}
```

So I guess we can do that to fix it, I just don't know why the difference between `@import` and `@importFrom` was important.

<details>

```r
> traceback()
18: stop(sprintf(ngettext(length(miss), "object %s is not exported by 'namespace:%s'", 
        "objects %s are not exported by 'namespace:%s'"), paste(sQuote(miss), 
        collapse = ", "), getNamespaceName(expenv)), call. = FALSE, 
        domain = NA)
17: importIntoEnv(impenv, impnames, ns, impvars)
16: namespaceImportFrom(ns, loadNamespace(j <- i[[1L]], c(lib.loc, 
        .libPaths()), versionCheck = vI[[j]]), i[[2L]], from = package)
15: loadNamespace(name)
14: getNamespace(ns)
13: asNamespace(ns)
12: getExportedValue(pkg, name)
11: tibble::tibble at helper-s3.R#44
10: eval(exprs, env)
9: eval(exprs, env)
8: FUN(X[[i]], ...)
7: lapply(files, source_file, env = env, chdir = chdir, wrap = wrap)
6: source_dir(path, "^helper.*\\.[rR]$", env = env, wrap = FALSE)
5: testthat_source_test_helpers(find_test_dir(path), env = ns_env(package))
4: force(code)
3: withr_with_envvar(c(NOT_CRAN = "true", DEVTOOLS_LOAD = "true"), 
       testthat_source_test_helpers(find_test_dir(path), env = ns_env(package)))
2: pkgload::load_all(path = path, reset = reset, recompile = recompile, 
       export_all = export_all, helpers = helpers, quiet = quiet, 
       ...)
1: devtools::load_all(".")
```

</details>

DavisVaughan:
Closes #747 

@lionel- if you have a better solution I'm open to it 🤷‍♂ 

We don't really use the extra args to `tibble()` so I figured just passing dots is ok

### DIFF
--- a/tests/testthat/helper-s3.R
+++ b/tests/testthat/helper-s3.R
@@ -41,7 +41,9 @@ local_no_stringsAsFactors <- function(frame = caller_env()) {
   local_options(.frame = frame, stringsAsFactors = FALSE)
 }
 
-tibble <- tibble::tibble
+tibble <- function(...) {
+  tibble::tibble(...)
+}
 
 local_foobar_proxy <- function(frame = caller_env()) {
   local_methods(.frame = frame, vec_proxy.vctrs_foobar = identity)


lionel-:
Ah that makes sense! Creating an indirection like you did is the right approach.

PR #759, opened 2020-01-15
`vec_equal_all()`

### LINKED ISSUE #708: Rename vec_duplicate_all to vec_equal_all
And change the signature to `vec_equal_all(x, y = NULL, na_equal = TRUE)`

Move to `vec_equal` doc page

DavisVaughan:
Closes #708 

- Reworks `vec_duplicate_all(x)` to become `vec_equal_all(x, y = NULL, na_equal = FALSE, ptype = NULL)`.
  - If `x` and `y` are supplied, it tests if all elements _between_ the two vectors are equal. If only `x` is supplied, it tests if all elements _within_ `x` are equal.

- Also moved it from the duplicate functions help page to the equal functions one.
- Improved the speed of it with data frames dramatically using the column-wise approach we've been using elsewhere (`vec_duplicate_all()` never got that column-wise treatment).


DavisVaughan:
Hmm now that I've implemented it, I'm wondering if the _within_ case can just be treated as a special case of the _between_ case where `x = x, y = x`. I don't think there is much performance loss.

lionel-:
I have a few worries about this UI:

* `vec_equal()` is vectorised while `vec_equal_all()` isn't. The names are confusing in that regard.

* The semantic difference between 1 input and 2 inputs seems hard to understand.

* Programmatically, if `y` is replaced by `NULL` because of a problem, we'll get wrong results instead of an error.

DavisVaughan:
> vec_equal() is vectorised while vec_equal_all() isn't.

We do have `vec_duplicate_detect()` and `vec_duplicate_any()` that work similarly, so I wasn't too worried about that part.

---

Regarding the UI, I agree that it could be improved. Maybe the best idea is to break it into two functions, reverting back to using `vec_duplicate_all()` for the case where just `x` is supplied.

So we'd have:

- `vec_duplicate_all(x)` (within `x`)
- `vec_equal_all(x, y, na_equal = FALSE, ptype = NULL)` (between `x` and `y`)

It would be sort of odd that `vec_duplicate_all()` is implemented with equal tools, while `vec_duplicate_any()` is implemented with dictionary tools, but I guess thats okay.

`vec_duplicate_all()` would be implemented with a forced `na_equal = TRUE` to match semantics of `vec_duplicate_any()`.

```r
library(vctrs)

vec_duplicate_any(c(NA, NA))
#> [1] TRUE
vec_duplicate_any(c(NA, NaN))
#> [1] FALSE

# `na_equal` must be `TRUE` to match `vec_duplicate_any()`
vec_equal_all(c(NA, NA), na_equal = FALSE)
#> [1] NA

vec_equal_all(c(NA, NA), na_equal = TRUE)
#> [1] TRUE
vec_equal_all(c(NA, NaN), na_equal = TRUE)
#> [1] FALSE
```

Then you'd have two complete sets of functions where signatures are better aligned:

```r
vec_duplicate_all(x)
vec_duplicate_any(x)
vec_duplicate_id(x)
vec_duplicate_detect(x)

vec_equal(x, y, na_equal = FALSE, ptype = NULL)
vec_equal_all(x, y, na_equal = FALSE, ptype = NULL)
vec_equal_na(x)
```

lionel-:
I like the approach that `vec_equal_` functions are consistently binary operators while `vec_duplicate_` are consistently unary. (With the exception of `vec_equal_na()` - this seems fine.)

I wonder if we should also take the approach that summary operations are all prefixed in `vec_any_` and `vec_all_`. When you're programming, you generally know if you're expecting a `n -> n`, `n -> m`, or `n -> 1` operation. Having a common prefix for `n -> 1` operations might be more helpful than having a thematic prefix.

This would give `vec_any_duplicate()`, `vec_any_equal()`, `vec_all_duplicate()` and `vec_all_equal()`. These also read better I think, and the names are more suggestive of semantics.

DavisVaughan:
Superseded by #773. This was getting too complicated, and `vec_duplicate_all()` has been removed for 0.2.2. I will instead submit them as a sequence of smaller PRs for 0.3.0

PR #764, opened 2020-01-16
Rename `vec_duplicate_id()`, and new `vec_duplicate_loc()`

### LINKED ISSUE #763: Rethink `vec_duplicate_id()` name, and new `vec_duplicate_loc()`
We have:

```
vec_duplicate_any()
vec_duplicate_detect()
vec_duplicate_id()
```

```
vec_unique()
vec_unique_loc()
vec_unique_count()
```

`vec_duplicate_id()` returns "an integer vector giving the location of the first occurrence of the value". I no longer think this name is correct, for two reasons:

1. It returns locations, but `vec_duplicate_loc()` isn't the right name (see below).
2. It could be called `vec_unique_id()` and have the same meaning, suggesting that forcing it in with the unique/duplicated functions is wrong altogether.

I suggest soft-deprecating and using `vec_first_loc()` instead. "The location of the first occurrence of the current element in `x`".

---

I think that we could also use `vec_duplicate_loc()`, which to me would be the complement of `vec_unique_loc()`. Where `vec_unique_loc()` returns the locations of the first time a value is seen, I think `vec_duplicate_loc()` would return the locations of the second and subsequent times a value is seen (like `duplicated()`, but with locations).

```r
library(vctrs)

vec_duplicate_loc <- function(x) {
  vec_seq_along(x)[-vec_unique_loc(x)]
}

x <- c(5, 6, 5, 3, 6)

# Location of unique values, in order they appear
vec_unique_loc(x)
#> [1] 1 2 4

# - Location of non-unique values, i.e. the "duplicates".
# - This is like `duplicated()`, returning the loc of the second and subsequent
#   duplicate values
# - Notice how it is the complement of `vec_unique_loc()`
vec_duplicate_loc(x)
#> [1] 3 5


# I feel like `vec_duplicate_id()` and `vec_unique_id()` would be equivalent,
# suggesting that it has the wrong name altogether
vec_unique_id <- vec_duplicate_id

# i.e., both of these "make sense"

# "location of the first occurance of the value"
vec_duplicate_id(x)
#> [1] 1 2 1 4 2
vec_unique_id(x)
#> [1] 1 2 1 4 2

vec_first_loc <- vec_duplicate_id

vec_first_loc(x)
#> [1] 1 2 1 4 2
```

DavisVaughan:
Closes #763 

If approved, this is what the changes suggested in #763 would look like.


src/dictionary.c
```diff
@@ -141,6 +141,37 @@ SEXP vctrs_unique_loc(SEXP x) {
...
+  struct growable g = new_growable(INTSXP, 256);
+  PROTECT_GROWABLE(&g, &nprot);
+
+  for (int i = 0; i < n; ++i) {
+    uint32_t hash = dict_hash_scalar(&d, i);
+
+    if (d.key[hash] == DICT_EMPTY) {
+      dict_put(&d, hash, i);
+    } else {
+      growable_push_int(&g, i + 1);
```
DavisVaughan: This function is the same as `vctrs_unique_loc()` with the exception of this if statement

R/dictionary.R
```diff
@@ -96,7 +97,8 @@ reset_rownames <- function(x) {
 #'   * `vec_duplicate_any()`: a logical vector of length 1.
 #'   * `vec_duplicate_all()`: a logical vector of length 1.
 #'   * `vec_duplicate_detect()`: a logical vector the same length as `x`.
```
lionel-: With the new subscript nomenclature I wonder if this should be `vec_duplicate_ind()`, for "indicator". This has the nice relationship that the indicator variant is the `which()` of the location variant. Also the suffix has three letters.

R/dictionary.R
```diff
@@ -192,6 +193,27 @@ vec_unique_count <- function(x) {
   .Call(vctrs_n_distinct, x)
 }
 
+# First location ----------------------------------------------------------
+
+#' Locate first occurrences
+#'
+#' `vec_first_loc()` returns an integer vector giving the location of the
+#' first occurrence of the current element in `x`.
```
lionel-: I find "current" a bit confusing. Perhaps:

> returns an integer vector givin for each element of `x` the location of its first occurence.

src/dictionary.c
lionel-: Include as comment in the code?

R/dictionary.R
lionel-: It would help to compare this function to `vec_duplicate_loc()` and `vec_unique_loc()`.

R/dictionary.R
DavisVaughan: After discussion, I've soft deprecated `vec_duplicate_detect()` in favor of `vec_duplicate_flg(x, first = TRUE)`.

DavisVaughan:
Okay @lionel- we now have:

```
vec_duplicate_any(x)
vec_duplicate_all(x)
vec_duplicate_flg(x, first = FALSE)
vec_duplicate_loc(x, first = FALSE)
```

With a nice relationship of:

```
which(vec_duplicate_flg(x)) == vec_duplicate_loc(x)
which(vec_duplicate_flg(x, first = TRUE)) == vec_duplicate_loc(x, first = TRUE)
```

And soft deprecation of:

- `vec_duplicate_id()` -> `vec_first_loc()`
- `vec_duplicate_detect()` -> `vec_duplicate_flg(first = TRUE)`

`vec_duplicate_flg(first = TRUE)` is also a decent bit faster than `vec_duplicate_detect()` because I store and reuse the hashes rather than looking them up twice.

```r
library(nycflights13)
library(vctrs)

set.seed(123)

flights_with_repeats <- sample(nrow(flights), 1e6, TRUE)

# old

bench::mark(vec_duplicate_detect(flights_with_repeats), iterations = 100)
#> # A tibble: 1 x 6
#>   expression                                    min median `itr/sec` mem_alloc
#>   <bch:expr>                                 <bch:> <bch:>     <dbl> <bch:byt>
#> 1 vec_duplicate_detect(flights_with_repeats) 82.1ms 88.9ms      11.0    23.6MB

# new

bench::mark(vec_duplicate_flg(flights_with_repeats, first = TRUE), iterations = 100)
#> # A tibble: 1 x 6
#>   expression                                               min median `itr/sec`
#>   <bch:expr>                                            <bch:> <bch:>     <dbl>
#> 1 vec_duplicate_flg(flights_with_repeats, first = TRUE) 39.6ms 42.1ms      23.2
```

And if you don't want the first occurrence to be considered a duplicate, it is even faster:

```r
bench::mark(vec_duplicate_flg(flights_with_repeats, first = FALSE), iterations = 100)
#> # A tibble: 1 x 6
#>   expression                                                min median `itr/sec`
#>   <bch:expr>                                             <bch:> <bch:>     <dbl>
#> 1 vec_duplicate_flg(flights_with_repeats, first = FALSE) 25.6ms 30.6ms      32.4
```

hadley:
FWIW I really don't like the `flg` prefix

DavisVaughan:
Postponing until 0.3.0 until we can figure out suffix naming for `flg`

DavisVaughan:
It has come up elsewhere that we could also use the naming scheme:

- `vec_duplicate_detect` -> `vec_detect_duplicate()`

- `vec_duplicate_id()` -> `vec_locate_first()`

- `vec_duplicate_loc()` -> `vec_locate_duplicate()`

I think this also lends itself well to the proposed change of `vec_duplicate_any()` to `vec_any_duplicate()`, and the addition of `vec_all_duplicate()` and `vec_all_equal()`.

Also should think about maybe:

- `vec_unique_loc()` -> `vec_locate_unique()`

- `vec_unique_count()` -> `vec_count_unique()`

- `vec_group_id()` -> `vec_identity_group()`

- `vec_group_loc()` -> `vec_locate_group()`

lionel-:
I like this proposal.

hadley:
`vec_locate_first()` doesn't really suggest first duplicate to me.  Do you mind making a googlesheet with a few of the proposals on it?

DavisVaughan:

PR #765, opened 2020-01-16
Add customisation points for OOB errors

lionel-:
This is the first step to move generation of error messages from tibble to vctrs to ensure consistency and to avoid subclassing errors when there is no need.

The customisation mechanism consists in catching the vctrs error and adding specific condition fields:

```r
  tryCatch(
    expr,
    vctrs_error_subscript = function(cnd) {
      # Customise error generation
      cnd$arg <- "foo"
      cnd$subscript_elt <- "row"
      cnd$subscript_action <- "rename"

      # Rethrow error as is
      cnd_signal(cnd)
    }
  )
```

* `arg` can be a character vector, otherwise it is deparsed with `as_label()`.
* `subscript_elt` must be one of "element", "row", "column".
* `subscript_action` must be one of "select", "extract", "assign", "rename", "remove", or "negate".

"element" and "action" are restricted to specific values in order to make it possible to generate English sentences without constraints. This should also help translation.


lionel-:
tibble implementation in tidyverse/tibble#699

tests/testthat/test-subscript-loc.R
```diff
@@ -330,5 +369,16 @@ test_that("conversion to locations has informative error messages", {
...
     num_as_location(c(1, NA, 2, 3), 1)
     num_as_location(c(1, NA, 3), 1, oob = "extend")
+
+    "# can customise OOB errors"
+    vec_slice(set_names(letters), "foo")
+    "With tibble columns"
+    with_tibble_cols(vec_slice(set_names(letters), "foo"))
+    with_tibble_cols(vec_slice(set_names(letters), 30))
+    with_tibble_cols(vec_slice(set_names(letters), -30))
+    "With tibble columns"
```
krlmlr:
```suggestion
    "With tibble rows"
```

tests/testthat/error/test-subscript-loc.txt
```diff
@@ -301,12 +306,50 @@ missing values are supported in error formatters
...
+Error: Must rename existing columns in `tbl[i]`.
+x Can't rename location 30.
+i There are only 26 columns.
+
+> with_tibble_cols(vec_slice(set_names(letters), -30))
+Error: Must rename existing columns in `tbl[i]`.
+x Can't rename location 30.
+i There are only 26 columns.
+
+> # With tibble columns
```
krlmlr:
```suggestion
> # With tibble rows
```

lionel-:
The default action for negative subscripts is now "negate".

Before:

```r
vec_slice(letters, -100)
#> Error: Must index existing elements.
#> ✖ Can't subset location 100.
#> ℹ There are only 26 elements.
```

After:

```r
vec_slice(letters, -100)
#> Error: Must negate existing elements.
#> ✖ Can't negate location 100.
#> ℹ There are only 26 elements.
```

tests/testthat/error/test-conditions.txt
```diff
@@ -0,0 +1,20 @@
+
+can override arg in OOB conditions
+==================================
+
+> with_subscript_data(vec_slice(set_names(letters), "foo"), NULL)
```
hadley: These might be easier to understand if you just called `stop_subscript_oob_name()` directly?

tests/testthat/error/test-subscript-loc.txt
```diff
@@ -301,12 +306,50 @@ missing values are supported in error formatters
...
+can customise OOB errors
+========================
+
+> vec_slice(set_names(letters), "foo")
+Error: Must subset existing elements.
+x Can't subset element with unknown name `foo`.
+
+> # With tibble columns
+> with_tibble_cols(vec_slice(set_names(letters), "foo"))
+Error: Must rename existing columns in `tbl[i]`.
```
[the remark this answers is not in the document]
lionel-: Agreed this is confusing. I started with this and then noticed it should be something else but fixing it right away would have caused merge issues between branches.

This is just meant to test a call `arg`, I'll change it to `foo(bar)` when the PRs are merged.

tests/testthat/error/test-conditions.txt
lionel-: hmm I think my goal with these tests was to make sure that `vec_slice()` throws the right errors even when we change the internals of the errors. For instance `stop_subscript_oob_name()` doesn't exist anymore in the later PRs.

hadley: Yeah. I'm not sure how to improve it, but when reading through this file, I didn't really get any sense of what was being customised in the error message.

PR #766, opened 2020-01-17
Fix assign action

lionel-:
Branched from #765.

Add subscript action to the internal option struct, and set the action to "assign" when assigning.

Before:

```r
vec_slice(letters, 100) <- "foo"
#> Error: Must subset existing elements.
#> ✖ Can't subset location 100.
#> ℹ There are only 26 elements.

vec_assign(set_names(letters), "foo", "bar")
#> Error: Must subset existing elements.
#> ✖ Can't subset element with unknown name `foo`.
```

After:

```r
vec_slice(letters, 100) <- "foo"
#> Error: Must assign existing elements.
#> ✖ Can't assign location 100.
#> ℹ There are only 26 elements.

vec_assign(set_names(letters), "foo", "bar")
#> Error: Must assign existing elements.
#> ✖ Can't assign element with unknown name `foo`.
```

### DIFF
--- a/tests/testthat/error/test-slice-assign.txt
+++ b/tests/testthat/error/test-slice-assign.txt
@@ -38,6 +38,6 @@ x Can't negate location 100.
 i There are only 26 elements.
 
 > vec_assign(set_names(letters), "foo", "bar")
-Error: Must subset existing elements.
-x Can't subset element with unknown name `foo`.
+Error: Must assign existing elements.
+x Can't assign element with unknown name `foo`.
 
--- a/tests/testthat/error/test-slice-assign.txt
+++ b/tests/testthat/error/test-slice-assign.txt
@@ -20,3 +20,19 @@ Error: Logical subscripts must match the size of the indexed input.
 i The input has size 32.
 x The subscript has size 2.
 
+
+must assign existing elements
+=============================
+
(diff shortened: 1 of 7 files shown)


tests/testthat/error/test-slice-assign.txt
```diff
@@ -20,3 +20,24 @@ Error: Logical subscripts must match the size of the indexed input.
 i The input has size 32.
 x The subscript has size 2.
 
+
+must assign existing elements
+=============================
+
+> vec_assign(1:3, 5, 10)
+Error: Must assign existing elements.
```
hadley: Should be "assign to", I think (above and below)

tests/testthat/error/test-slice-assign.txt
```diff
@@ -20,3 +20,24 @@ Error: Logical subscripts must match the size of the indexed input.
...
+x Can't assign location 5.
+i There are only 3 elements.
+
+> vec_assign(1:3, "foo", 10)
+Error: Can't use character names to index an unnamed vector.
+
+> vec_slice(letters, -100) <- "foo"
+Error: Must negate existing elements.
```
hadley: I initially misunderstood this as implying that existing elements must be negated. Maybe "Can't negate elements that don't exist" ?

PR #767, opened 2020-01-18
Simplify error classes

lionel-:
This further simplifies the error hierarchy for subscripts.

These classes are removed:

- `vctrs_error_subscript2_bad_type`
- `vctrs_error_indicator_bad_size`
- `vctrs_error_location_negative_missing`
- `vctrs_error_location_negative_positive`
- `vctrs_error_subscript_oob_location_non_consecutive`

All the subscript errors are now one of these three classes:

- `vctrs_error_subscript_bad_type`
- `vctrs_error_subscript_bad_size`
- `vctrs_error_subscript_oob`

PR #768, opened 2020-01-20
Improve consistency of subscript errors customisation

lionel-:
Branched from #767 

* Use `subscript_arg` condition field instead of `arg`. This is more consistent and we might also want to provide arg info about subsetted input.

* More error messages now include `subscript_arg`.

* Removed default `arg = "i"` because it's most likely not relevant for the user. When the arg is not supplied, we refer to the subscript generically with "The subscript".

* Pass options struct more consistently across as-location routines. Add `subscript_arg` to option struct so it's properly forwarded.

* The errors for scalar locations are less redundant.

### DIFF
--- a/tests/testthat/error/test-subscript.txt
+++ b/tests/testthat/error/test-subscript.txt
@@ -3,32 +3,32 @@ vec_as_subscript() forbids subscript types
 ==========================================
 
 > vec_as_subscript(1L, indicator = "error", location = "error")
-Error: Must subset with a proper subscript vector.
+Error: Must subset elements with a proper subscript vector.
 x `i` has the wrong type `integer`.
 i The subscript must contain names.
 
 > vec_as_subscript("foo", indicator = "error", name = "error")
-Error: Must subset with a proper subscript vector.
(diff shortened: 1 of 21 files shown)


tests/testthat/error/test-subscript.txt
```diff
@@ -3,51 +3,60 @@ vec_as_subscript() forbids subscript types
 ==========================================
 
 > vec_as_subscript(1L, indicator = "error", location = "error")
-Error: Must subset with a proper subscript vector.
-x `i` has the wrong type `integer`.
-i The subscript must contain names.
+Error: Must subset elements with a proper subscript vector.
+x The subscript has the wrong type `integer`.
```
hadley: I think the lack of symmetry in these messages (integer vs names, character vs locations) is likely to cause confusion, and simply using the type name in the info bullet would be adequate. If you really wanted to continue using indicators/locations/names, I'd suggest putting them in parentheses.

lionel-: You're right, and we don't like "indicator" so much. Probably best to keep it simple. I'll make the changes in another PR.

PR #773, opened 2020-01-21
Add `vec_all_equal()`

DavisVaughan:
PR 1 of 3 for 0.3.0

Supersedes #759


DavisVaughan:

PR #774, opened 2020-01-21
Add `vec_all_duplicate()`

DavisVaughan:
PR 2 of 3 for 0.3.0

Adds `vec_all_duplicate()`, now correctly implemented using dictionary utilities.

The main problem with the one that was previously implemented with equal utilities was that it worked like `vec_all_same() ~== vec_all_equal(x, x[1])`. This is not what is expected, since we want to detect "all values are duplicated elsewhere", not "all values are the same in this vector".

Practically this means that this now correctly returns `TRUE`, where with `vec_all_same()` it would return `FALSE`:

```r
vctrs::vec_all_duplicate(c(1, 1, 2, 2))
#> [1] TRUE
```

A follow up PR will change `vec_duplicate_any()` to `vec_any_duplicate()`


DavisVaughan:

PR #778, opened 2020-01-22
Flip default of `inherit_base_type`

lionel-:
Fixes https://github.com/jessesadler/debkeepr

This default is closer to CRAN behaviour and makes sense from the perspective of https://github.com/r-lib/vctrs/pull/777#issuecomment-577110394

Also revealed a couple of problems:

- Casting `vctrs_vctr` to `POSIXlt` is now undefined as it should.
- Coercions between `vctrs_list_of` and `vctrs_unspecified` are fixed.

### DIFF
--- a/R/type-list-of.R
+++ b/R/type-list-of.R
@@ -171,10 +171,17 @@ as.character.vctrs_list_of <- function(x, ...) {
 # Type system -------------------------------------------------------------
 
 #' @rdname list_of
+#' @inheritParams vec_ptype2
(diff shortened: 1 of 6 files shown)


lionel-:
@jessesadler We're about to send a vctrs update to CRAN. Your package now passes R CMD check but can you make sure everything is ok regarding your RStudio talk please?

jessesadler:
@lionel- Thanks for letting me know. I have been following along and have set aside time to go through everything. I will let you know if I have any questions.

PR #784, opened 2020-01-28
Allow `""` as a valid name for `<vctrs_vctr>` objects

### LINKED ISSUE #780: Should size 1 vectors be allowed to have `""` as names?
We have a small inconsistency with this edge case.  `vec_slice(x, NA_integer_)` can generate a named vector of size 1 with `""` as names. But `new_vctr()` doesn't like this, as it errors if "all names are the empty string" (see `names_all_or_nothing()`)

Should `vec_slice()` be tweaked to not restore any names for this edge case, or should `names_all_or_nothing()` be more relaxed?

I think tweaking `vec_slice()` is correct. I don't think we want `vec_slice(c(x = 1), NA_integer_)` to return a vector with any names.

``` r
library(vctrs)

# Is this okay for names?
x <- c(foo = 1)

vec_slice(x, NA_integer_)
#>    
#> NA

names(vec_slice(x, NA_integer_))
#> [1] ""

# We have an inconsistency because `vec_slice()` can generate these size 1
# vectors where "all names are the empty string", but that is not allowed by
# new_vctr()'s `names_all_or_nothing()` check
y <- vec_slice(x, NA_integer_)
new_vctr(y)
#> Error: If any elements of `.data` are named, all must be named

# This mainly causes issues if you have a restore method that might eventually
# call `new_vctr()`

z <- new_vctr(c(foo = 1), class = "foobar", inherit_base_type = FALSE)

# Fine because default restore doesn't go through `new_vctr()`
vec_slice(z, NA_integer_)
#> <foobar[1]>
#>    
#> NA

vec_restore.foobar <- function(x, to, ..., n = NULL) {
  new_vctr(x, class = "foobar", inherit_base_type = FALSE)
}

vec_slice(z, NA_integer_)
#> Error: If any elements of `.data` are named, all must be named
```

<sup>Created on 2020-01-27 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0.9000)</sup>

### LINKED ISSUE #1268: Unexpected behavior of names with vctrs_vctr objects
I'm not sure if this belongs here or in the `dplyr` or `tibble` repos, but I am posting here because ultimately it is the opinionated behavior of names assignment to `vctrs_vctr` objects that causes the problem. 

Is the following expected behavior?

``` r
library(magrittr)
x <- tibble::tibble(a = c(1, NA), b = c(1, 2))
# error because `names<-.vctrs_vctr` requires all names be != "" and not NA
x %>% dplyr::group_nest(a) %>% tibble::deframe()
#> Error: If any elements are named, all elements must be named.
# what I expect
x %>% dplyr::group_nest(a) %>% dplyr::mutate(data = as.list(data)) %>% tibble::deframe()
#> $`1`
#> # A tibble: 1 x 1
#>       b
#>   <dbl>
#> 1     1
#> 
#> $<NA>
#> # A tibble: 1 x 1
#>       b
#>   <dbl>
#> 1     2
# what I expect
x %>% tidyr::nest(data = -a) %>% tibble::deframe()
#> $`1`
#> # A tibble: 1 x 1
#>       b
#>   <dbl>
#> 1     1
#> 
#> $<NA>
#> # A tibble: 1 x 1
#>       b
#>   <dbl>
#> 1     2
```

<sup>Created on 2020-09-23 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

DavisVaughan:
Closes #780 
Closes #1268
Closes https://github.com/tidyverse/haven/issues/526 (need to manually close that)
https://github.com/tidyverse/tidyr/issues/1212 requires this

Related to #1108, in that it fixes the issue originally brought up there, but I think there is still more work to do in terms of the order attributes are restored.

---

This PR adjusts the restriction that neither `NA_character_` nor `""` were allowed as names for `vctrs_vctr` objects. We now:
- Allow `""` names
- Repair `NA_character_` names by converting them to `""` silently

---

It turns out that `""` arises as a valid name in many scenarios, and disallowing it as a valid name has caused many issues.

- `""` is used as the name in `vec_slice(x, NA_integer_)` when `x` is a named vector
- Combining a named vector with an unnamed vector will force `""` as a name (see the hidden list-of comment below)

And `NA_character_` names seem to appear in practice (i.e. #1268 and https://github.com/tidyverse/haven/issues/526), so providing something that deals with them in a reasonable way is also useful.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,8 @@
 # vctrs (development version)
 
+* If `NA_character_` is specified as a name for `vctrs_vctr` objects, it is
+  now automatically repaired to `""` (#780).
+
 * `""` is now an allowed name for `vctrs_vctr` objects (#780).
 
 * `list_of()` is now much faster when many values are provided.
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,7 @@
 # vctrs (development version)
 
+* `""` is now an allowed name for `vctrs_vctr` objects (#780).
+
 * `list_of()` is now much faster when many values are provided.
 
 * `vec_as_location()` evaluates `arg` only in case of error, for performance
--- a/R/type-vctr.R
+++ b/R/type-vctr.R
@@ -100,17 +100,17 @@ new_vctr <- function(.data,
 validate_names <- function(.data) {
   nms <- names(.data)
 
-  if (!names_all_or_nothing(nms)) {
-    stop("If any elements of `.data` are named, all must be named", call. = FALSE)
+  if (any_na_names(nms)) {
+    abort("The names of `.data` must not be `NA`.")
   }
 
   nms
 }
-names_all_or_nothing <- function(names) {
+any_na_names <- function(names) {
   if (is.null(names)) {
-    TRUE
+    FALSE
   } else {
-    all(names != "" & !is.na(names))
+    any(is.na(names))
   }
 }
 
@@ -269,8 +269,8 @@ diff.vctrs_vctr <- function(x, lag = 1L, differences = 1L, ...) {
   if (length(value) != 0 && length(value) != length(x)) {
     abort("`names()` must be the same length as x.")
   }
-  if (!names_all_or_nothing(value)) {
-    abort("If any elements are named, all elements must be named.")
+  if (any_na_names(value)) {
+    abort("Names must not be `NA`.")
   }
   NextMethod()
 }
--- a/R/type-vctr.R
+++ b/R/type-vctr.R
@@ -72,8 +72,6 @@ new_vctr <- function(.data,
     abort("`.data` must be a vector type.")
   }
 
-  nms <- validate_names(.data)
-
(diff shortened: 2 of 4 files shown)


DavisVaughan:
We also get burned by this when combining named and unnamed elements together. I think this code is perfectly fine:

``` r
library(vctrs)

a <- list(x = 1L)
b <- list(2L)
both <- c(a, b)

new_list_of(both, ptype = integer())
#> Error: If any elements of `.data` are named, all must be named
```

This came up in tidyr https://github.com/tidyverse/tidyr/pull/1200#issuecomment-964901004

And this PR would fix that

lionel- (approved):
Awesome!

By the way, I've noticed that the `[<-` method uses `vec_cast()` and then the base method. I think it should use `vec_slice<-` instead? The casting allows the class to enforce pre-conditions (and thus repair names), but there is no restoration so post-conditions are not validated.

PR #788, opened 2020-01-30
extremely minor typo

coreyyanofsky:


lionel-:
Thanks! For future reference, the .Rd files are generated from the R file, so your change would have been overwritten at the next `document()`. I have updated the corresponding .R file.

Issue #785, opened 2020-01-28
Slightly confusing error message in vec_as_location2() with negative subscripts


DavisVaughan:
```r
vctrs::vec_as_location2(-1, 1)
#> Error: Must extract element with a single subscript.
#> x The subscript has value -1 but must be a positive location.
```

This was slightly confusing to me because I _am_ supplying a single subscript, it is just negative.

I think this would be clearer if we said `"a single valid subscript"` here:

https://github.com/r-lib/vctrs/blob/870db061e46e201dda21225f15db46608cba7d08/R/subscript.R#L268

This would also be consistent with the error message for `vec_as_location()` here:

https://github.com/r-lib/vctrs/blob/870db061e46e201dda21225f15db46608cba7d08/R/subscript.R#L267-L271

DavisVaughan:
For testing, it looks like we already have (at least) 1 test that verifies the output from this error message. It is here:

https://github.com/r-lib/vctrs/blob/ad92e664e095ae084e53993d07017d379d9b792c/tests/testthat/test-subscript-loc.R#L451

After making this change, you can run `devtools::test()` locally. The first time you run it after making the change it will error telling you that "results have changed". This will save the new results in the output text files that we use to verify that the error messages look correct. Run `test()` again and it should pass. Then commit both the changed R file and the updated txt files.

Read more about `testthat::verify_output()` (which is what powers this process) here:
https://www.tidyverse.org/blog/2019/11/testthat-2-3-0/

PR #789, opened 2020-01-31
Improve vec_as_location2 error message

808sAndBR:
Resolves: #785 

To be consistent with some of the other error messages I saw in the package I adjusted the suggested message slightly from "a single valid subscript" to "a single valid subscript vector".

### DIFF
--- a/R/subscript.R
+++ b/R/subscript.R
@@ -265,7 +265,7 @@ cnd_header.vctrs_error_subscript_type <- function(cnd) {
   action <- cnd_subscript_action(cnd)
   elt <- cnd_subscript_element(cnd)
   if (cnd_subscript_scalar(cnd)) {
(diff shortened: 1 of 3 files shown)


DavisVaughan:
Thanks! I think we still prefer, "a single valid subscript". I think the addition of "vector" makes it slightly less clear that it really is a single _scalar_ value

lionel-:
> To be consistent with some of the other error messages I saw in the package I adjusted the suggested message slightly from "a single valid subscript" to "a single valid subscript vector".

Hmm I think the original phrasing makes it clearer that a scalar is needed. From this perspective "vector" is giving the wrong signal. Can you remove it please?

808sAndBR:
Ohhhh I see now! Just pushed a fix to go back to the suggested phrasing.

DavisVaughan:
Thank you!

PR #798, opened 2020-02-07
Rearrange and rename files/functions related to ptype2

DavisVaughan:
- Rename `type2-dispatch.c` to `typeof2.c` (this holds `vec_typeof2()`)

- Rename `type2-dispatch-extension.c` to `typeof2-s3.c` (this holds `vec_typeof2_s3()`)

- Rename `type2-extension.c` to `type2-dispatch.c` and `vec_ptype2_extension()` to `vec_ptype2_dispatch()`

- Rename the callback to R of `vctrs_type2_dispatch()` to `vec_ptype2_dispatch_s3()` and move it to `type2-dispatch.c` with the non-callback equivalent

- Rename `vec_s3_typeof2()` to `vec_typeof2_s3()` for consistency with file names


DavisVaughan:
We are in a bit of a weird place with naming since we have some things that are type2 and new things that are ptype2.

@lionel- should `enum vctrs_type2_s3` be `enum vctrs_ptype2_s3`? It is the return value from `vec_typeof2_s3_impl()`. We also have `enum vctrs_type2` as the return value of `vec_typeof2()`, so I wasn't sure if we should be consistent with that or update

lionel-:
Maybe we can keep these enum as `_type2_`? This seems consisent with `enum vctrs_type`.

DavisVaughan:
That's what I was thinking too

PR #799, opened 2020-02-07
Factor performance - `vec_cast()`

DavisVaughan:
This PR adds internal `vec_cast()` dispatch for factors/ordered factors.

It does so in a way that is similar to `vec_ptype2()`, using the extension prototype idea @lionel- and I used there. I've added a `vec_cast_dispatch()` path in `vec_cast_switch()`, which allows us to hand off to specific C level casting routines for factors. The routines themselves live in `type-factor.c`.

I've maintained all the behavior from before, including the "template" idea of doing `vec_cast(factor("x"), factor())`.

A number of performance improvements were made, the largest being the easy case where `x` and `to` are both factors with the same levels, in which case there is nothing to do.

```r
library(vctrs)

vec_cast.factor.factor <- vctrs:::vec_cast.factor.factor
vec_cast.factor.character <- vctrs:::vec_cast.factor.character

a <- factor(rep("a", 1e5))
```

Casting to a factor with the same levels is now very fast

```r
bench::mark(
  vec_cast(a, a),
  vec_cast.factor.factor(a, a)
)
#> # A tibble: 2 x 6
#>   expression                        min   median `itr/sec` mem_alloc `gc/sec`
#>   <bch:expr>                   <bch:tm> <bch:tm>     <dbl> <bch:byt>    <dbl>
#> 1 vec_cast(a, a)                  897ns   1.21µs   790069.        0B     79.0
#> 2 vec_cast.factor.factor(a, a)   2.98ms   3.84ms      242.    4.61MB     24.2
```

Same with casting to a template

```r
template <- factor()

bench::mark(
  vec_cast(a, template),
  vec_cast.factor.factor(a, template)
)
#> # A tibble: 2 x 6
#>   expression                               min median `itr/sec` mem_alloc
#>   <bch:expr>                          <bch:tm> <bch:>     <dbl> <bch:byt>
#> 1 vec_cast(a, template)                  951ns 1.21µs   761576.        0B
#> 2 vec_cast.factor.factor(a, template)   1.16ms 1.54ms      637.    1.91MB
```

The factor to factor conversion tries to be somewhat smart when the levels are different, and the levels of `x` are a subset of the levels of `to`. If the levels of `x` are a contiguous subset of the levels of `to` then we can just construct a new factor immediately without "recoding" the underlying integer values.

```r
# no recoding required because levels of `a` are a contiguous subset of `ab`
# i.e. c("a") is a subset of c("a", "b") and is in the same order
ab <- factor(c(rep("a", 1e5/2), rep("b", 1e5/2)), levels = c("a", "b"))

bench::mark(
  vec_cast(a, ab),
  vec_cast.factor.factor(a, ab),
  factor(a, levels = levels(ab))
)
#> # A tibble: 3 x 6
#> 1 vec_cast(a, ab)                 31.03µs  57.17µs    14451.  390.67KB    122. 
#> 2 vec_cast.factor.factor(a, ab)    2.87ms   3.55ms      284.    4.58MB     29.4
#> 3 factor(a, levels = levels(ab))   1.36ms   1.48ms      672.    1.91MB     34.6
```

Here is a small comparison of recoding vs not recoding

```r
# recoding required because levels of `a` are NOT a contiguous subset of `ba`
# i.e. c("a") is a subset of c("b", "a") but is NOT in order
ba <- factor(c(rep("b", 1e5/2), rep("a", 1e5/2)), levels = c("b", "a"))

bench::mark(
  vec_cast(a, ab),
  vec_cast(a, ba),
  check = FALSE,
  iterations = 5000
)
#> # A tibble: 2 x 6
#> 1 vec_cast(a, ab)   30.2µs   55.5µs    15426.     391KB    118. 
#> 2 vec_cast(a, ba)   53.7µs   81.2µs    12208.     391KB     86.1
```

Here is an example of what is probably the worst case scenario. Every value is a different level. Faster than the original implementation still.

```r
# probably the slowest it would ever be
many_levels <- factor(paste0("a", 1:1e5), levels = paste0("a", 1:1e5))
many_levels_rev <- factor(rev(paste0("a", 1:1e5)), levels = rev(paste0("a", 1:1e5)))

bench::mark(
  vec_cast(many_levels, many_levels_rev),
  vec_cast.factor.factor(many_levels, many_levels_rev),
  factor(many_levels, levels = levels(many_levels_rev)),
  iterations = 50
)
#> # A tibble: 3 x 6
#>   expression                                               min median `itr/sec`
#>   <bch:expr>                                            <bch:> <bch:>     <dbl>
#> 1 vec_cast(many_levels, many_levels_rev)                10.6ms 12.2ms      80.7
#> 2 vec_cast.factor.factor(many_levels, many_levels_rev)  18.1ms 19.4ms      51.0
#> 3 factor(many_levels, levels = levels(many_levels_rev)) 12.4ms   13ms      76.3
```

I've also added character -> factor C routines. The one where `to` is a template takes the longest because we have to take the unique values of `x` to get the levels, run through them and remove the `NA` level if it exists, and then match that against `x`.

```r
chr <- paste0("a", 1:3e5)
fct <- factor(chr, levels = chr)

bench::mark(
  vec_cast(chr, fct),
  vec_cast.factor.character(chr, fct),
  iterations = 50
)
#> # A tibble: 2 x 6
#>   expression                             min median `itr/sec` mem_alloc `gc/sec`
#>   <bch:expr>                          <bch:> <bch:>     <dbl> <bch:byt>    <dbl>
#> 1 vec_cast(chr, fct)                  36.9ms 39.6ms      23.8    5.43MB     5.71
#> 2 vec_cast.factor.character(chr, fct) 74.2ms   85ms      11.3   34.89MB    14.7

bench::mark(
  vec_cast(chr, template),
  vec_cast.factor.character(chr, template),
  iterations = 50
)
#> # A tibble: 2 x 6
#>   expression                                  min median `itr/sec` mem_alloc
#>   <bch:expr>                               <bch:> <bch:>     <dbl> <bch:byt>
#> 1 vec_cast(chr, template)                  56.5ms 60.4ms      16.6      16MB
#> 2 vec_cast.factor.character(chr, template) 66.7ms 66.7ms      15.0    30.3MB
```

### DIFF
--- a/src/type-factor.c
+++ b/src/type-factor.c
@@ -295,6 +295,23 @@ static SEXP fct_as_factor_impl(SEXP x, SEXP x_levels, SEXP to_levels, bool* loss
     }
   }
 
+  // No recoding required if contiguous subset.
+  // Duplicate, strip non-factor attributes, and re-initialize with new levels.
+  // Using `r_maybe_duplicate()` avoids an immediate copy using ALTREP wrappers.
+  if (is_contiguous_subset) {
+    SEXP out = PROTECT(r_maybe_duplicate(x));
+    SET_ATTRIB(out, R_NilValue);
+
+    if (ordered) {
+      init_ordered(out, to_levels);
+    } else {
+      init_factor(out, to_levels);
+    }
+
+    UNPROTECT(1);
+    return out;
+  }
+
   const int* p_x = INTEGER_RO(x);
 
   SEXP out = PROTECT(Rf_allocVector(INTSXP, x_size));
@@ -304,13 +321,6 @@ static SEXP fct_as_factor_impl(SEXP x, SEXP x_levels, SEXP to_levels, bool* loss
     init_ordered(out, to_levels);
   } else {
     init_factor(out, to_levels);
-  }
-
-  // No recode required
-  if (is_contiguous_subset) {
-    memcpy(p_out, p_x, x_size * sizeof(int));
-    UNPROTECT(1);
-    return out;
   }
 
   SEXP recode = PROTECT(vec_match(x_levels, to_levels));
--- a/src/type-factor.c
+++ b/src/type-factor.c
@@ -155,17 +155,11 @@ static SEXP chr_as_factor_impl(SEXP x, SEXP levels, bool* lossy, bool ordered) {
 
   // Detect lossy no-matches, but allow `NA` values from `x`
   for (R_len_t i = 0; i < size; ++i) {
-    if (p_out[i] != NA_INTEGER) {
-      continue;
+    if (p_out[i] == NA_INTEGER && p_x[i] != NA_STRING) {
+      *lossy = true;
+      UNPROTECT(1);
+      return R_NilValue;
     }
-
-    if (p_x[i] == NA_STRING) {
-      continue;
-    }
-
-    *lossy = true;
-    UNPROTECT(1);
-    return R_NilValue;
   }
 
   if (ordered) {
--- a/src/type-factor.c
+++ b/src/type-factor.c
@@ -74,3 +74,289 @@ static SEXP levels_union(SEXP x, SEXP y) {
   UNPROTECT(2);
   return out;
 }
+
+// -----------------------------------------------------------------------------
+
+static void new_factor(SEXP x, SEXP levels);
+static void new_ordered(SEXP x, SEXP levels);
+
+
+// [[ include("vctrs.h") ]]
+SEXP fct_as_character(SEXP x, struct vctrs_arg* x_arg) {
+  SEXP levels = Rf_getAttrib(x, R_LevelsSymbol);
+
+  if (TYPEOF(levels) != STRSXP) {
+    stop_corrupt_factor_levels(x, x_arg);
+  }
+
+  return Rf_asCharacterFactor(x);
+}
+
+// [[ include("vctrs.h") ]]
+SEXP ord_as_character(SEXP x, struct vctrs_arg* x_arg) {
+  return fct_as_character(x, x_arg);
+}
+
+
+static SEXP chr_as_factor_from_self(SEXP x, bool ordered);
+static SEXP chr_as_factor_impl(SEXP x, SEXP levels, bool* lossy, bool ordered);
+
+// [[ include("vctrs.h") ]]
+SEXP chr_as_factor(SEXP x, SEXP to, bool* lossy, struct vctrs_arg* to_arg) {
+  SEXP levels = Rf_getAttrib(to, R_LevelsSymbol);
+
+  if (TYPEOF(levels) != STRSXP) {
+    stop_corrupt_factor_levels(to, to_arg);
+  }
+
+  // When `to` has no levels, it is treated as a template and the
+  // levels come from `x`
+  if (vec_size(levels) == 0) {
+    return chr_as_factor_from_self(x, false);
+  }
+
+  return chr_as_factor_impl(x, levels, lossy, false);
+}
+
+// [[ include("vctrs.h") ]]
+SEXP chr_as_ordered(SEXP x, SEXP to, bool* lossy, struct vctrs_arg* to_arg) {
+  SEXP levels = Rf_getAttrib(to, R_LevelsSymbol);
+
+  if (TYPEOF(levels) != STRSXP) {
+    stop_corrupt_ordered_levels(to, to_arg);
+  }
+
+  // When `to` has no levels, it is treated as a template and the
+  // levels come from `x`
+  if (vec_size(levels) == 0) {
+    return chr_as_factor_from_self(x, true);
+  }
+
+  return chr_as_factor_impl(x, levels, lossy, true);
+}
+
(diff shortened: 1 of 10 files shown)


src/type-factor.c
```diff
@@ -74,3 +74,287 @@ static SEXP levels_union(SEXP x, SEXP y) {
...
+  return fct_as_character(x, x_arg);
+}
+
+
+static SEXP chr_as_factor_from_self(SEXP x, bool ordered);
+static SEXP chr_as_factor_impl(SEXP x, SEXP levels, bool* lossy, bool ordered);
+
+// [[ include("vctrs.h") ]]
+SEXP chr_as_factor(SEXP x, SEXP to, bool* lossy, struct vctrs_arg* to_arg) {
+  SEXP levels = Rf_getAttrib(to, R_LevelsSymbol);
```
lionel-: Probably needs protection for rchk because `getAttrib()` potentially allocates.

src/type-factor.c
```diff
@@ -74,3 +74,287 @@ static SEXP levels_union(SEXP x, SEXP y) {
...
+static SEXP chr_as_factor_impl(SEXP x, SEXP levels, bool* lossy, bool ordered) {
+  SEXP out = PROTECT(vec_match(x, levels));
+  const int* p_out = INTEGER(out);
+
+  R_len_t size = vec_size(x);
+  const SEXP* p_x = STRING_PTR_RO(x);
+
+  // Detect lossy no-matches, but allow `NA` values from `x`
+  for (R_len_t i = 0; i < size; ++i) {
+    if (p_out[i] != NA_INTEGER) {
```
lionel-: I think it'd be a little clearer if the branch was `out == NA && x != NA` with the lossy `return` inside it, and no `continue` in the loop.

src/type-factor.c
```diff
@@ -74,3 +74,287 @@ static SEXP levels_union(SEXP x, SEXP y) {
...
+    new_factor(out, levels);
+  }
+
+  UNPROTECT(1);
+  return out;
+}
+
+static SEXP remove_na_levels(SEXP levels);
+
+// Factor levels are added in order of appearance
```
lionel-: The sentences are missing full stops, making them hard to read.

src/type-factor.c
```diff
@@ -74,3 +74,287 @@ static SEXP levels_union(SEXP x, SEXP y) {
...
+static SEXP remove_na_levels(SEXP levels) {
+  R_len_t size = vec_size(levels);
+  const SEXP* p_levels = STRING_PTR_RO(levels);
+
+  // There would only ever be 1 `NA` level
+  int na_loc;
+  bool any_na = false;
+
+  for (R_len_t i = 0; i < size; ++i) {
+    if (p_levels[i] != NA_STRING) {
```
lionel-: An early break might be easier to understand.

src/type-factor.c
```diff
@@ -74,3 +74,287 @@ static SEXP levels_union(SEXP x, SEXP y) {
...
+    }
+  }
+
+  const int* p_x = INTEGER_RO(x);
+
+  SEXP out = PROTECT(Rf_allocVector(INTSXP, x_size));
+  int* p_out = INTEGER(out);
+
+  if (ordered) {
+    new_ordered(out, to_levels);
```
lionel-: I would assume `new_` functions to return objects. Rename to `init_`?

src/type-factor.c
```diff
@@ -74,3 +74,287 @@ static SEXP levels_union(SEXP x, SEXP y) {
...
+  } else {
+    new_factor(out, to_levels);
+  }
+
+  // No recode required
+  if (is_contiguous_subset) {
+    memcpy(p_out, p_x, x_size * sizeof(int));
```
lionel-: `Rf_shallow_duplicate()` instead? Might help avoid a copy thanks to ALTREP wrappers.

DavisVaughan: That worked nicely!

DavisVaughan:
> Would it be much work to handle the lossy cases using the new lazy error infrastructure, instead of falling back to the R methods?

Together we decided to leave this for a separate PR

DavisVaughan:
Improvement on `vec_c()`-ing many factors with the same levels. This is related to one of the dplyr issues regarding slowness b/c of vctrs

```r
library(vctrs)

x <- factor(letters)
x <- rep_len(list(x), 10000)
head(x, 2)
#> [[1]]
#>  [1] a b c d e f g h i j k l m n o p q r s t u v w x y z
#> Levels: a b c d e f g h i j k l m n o p q r s t u v w x y z
#> 
#> [[2]]
#>  [1] a b c d e f g h i j k l m n o p q r s t u v w x y z
#> Levels: a b c d e f g h i j k l m n o p q r s t u v w x y z

# Before
bench::mark(vec_c(!!!x))
#> # A tibble: 1 x 6
#> 1 vec_c(!!!x)    647ms    647ms      1.55    43.6MB     20.1

# After
bench::mark(vec_c(!!!x))
#> # A tibble: 1 x 6
#> 1 vec_c(!!!x)   24.3ms   26.2ms      38.2    3.12MB     78.3
```

PR #802, opened 2020-02-10
Add base R fallback for `vec_c()`

### LINKED ISSUE #801: `vec_c()` fallback to `base::c()` for homogeneously classed inputs
To avoid issues like tidyverse/dplyr#4839.

lionel-:
Closes tidyverse/dplyr#4839
Closes #801

This fallback will help for the common case of concatenating S3 objects of the same class.

If the inputs all have the same class and the class is foreign, we fall back to `base::c()`. We consider classes to be foreign when they do not implement `vec_ptype2()`.

If this makes sense for c we should also look into fallbacks for rbind and cbind.


DavisVaughan:
> we should also look into fallbacks for rbind and cbind

That looks like it would help with https://github.com/r-lib/vctrs/issues/564#issuecomment-560117457

src/c.c
```diff
@@ -26,13 +26,33 @@ SEXP vctrs_c(SEXP call, SEXP op, SEXP args, SEXP env) {
...
+static SEXP vec_c_fallback(SEXP xs);
+
 // [[ include("vctrs.h") ]]
 SEXP vec_c(SEXP xs,
            SEXP ptype,
            SEXP name_spec,
            enum name_repair_arg name_repair) {
   R_len_t n = Rf_length(xs);
 
+  if (needs_vec_c_fallback(xs)) {
```
DavisVaughan: Should the `ptype` be considered here? What if I did `vec_c(<Period>, <Period>, .ptype = logical())`?

src/utils.c
```diff
@@ -320,6 +320,39 @@ SEXP s3_find_method(const char* generic, SEXP x) {
...
+// [[ include("utils.h") ]]
+bool list_is_s3_homogeneous(SEXP xs) {
+  R_len_t n = Rf_length(xs);
+  if (n == 0 || n == 1) {
+    return true;
+  }
+
+  SEXP first_class = PROTECT(r_class(VECTOR_ELT(xs, 0)));
+
+  for (R_len_t i = 2; i < n; ++i) {
```
DavisVaughan: `i = 1`?

src/c.c
lionel-: Good point. Probably best to give the same kind of error as with `name_spec`.

src/utils.c
lionel-: oops

PR #809, opened 2020-02-12
Fix name inconsistency in `vec_rbind()`

lionel-:
Vectors of size 1 with external names are now treated consistently.

```r
# Before
vec_rbind(a = 1, b = 2)
#> New names:
#> * `` -> ...1
#> New names:
#> * `` -> ...1
#>   ...1
#> 1    1
#> 2    2

# After
vec_rbind(a = 1, b = 2)
#>    a  b
#> 1  1 NA
#> 2 NA  2

# This is now the same as
vec_rbind(c(a = 1), c(b = 2))
#>    a  b
#> 1  1 NA
#> 2 NA  2
```

Also adds an output test to track the name-repair output. Currently suboptimal for `vec_rbind()`, but unclear how we could do better and probably good enough for such an edge case.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,7 +1,9 @@
 
 # vctrs (development version)
 
-* `vec_rbind()` now binds row names.
+* `vec_rbind()` now binds row names. When named inputs are supplied
+  and `names_to` is `NULL`, the names define row names. If `names_to`
+  is supplied, they are assigned in the column name as before.
 
 * The `c()` method for `vctrs_vctr` now throws an error when
   `recursive` or `use.names` is supplied (#791).
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,14 +1,6 @@
 
 # vctrs (development version)
 
-* `vec_rbind()` now consistently handles inputs of size 1 with
-  external names. These cases are now identical:
-
-  ```
-  vec_rbind(a = 1)
-  vec_rbind(c(a = 1))
-  ```
-
 * The `c()` method for `vctrs_vctr` now throws an error when
   `recursive` or `use.names` is supplied (#791).
 
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,6 +1,14 @@
 
 # vctrs (development version)
 
+* `vec_rbind()` now consistently handles inputs of size 1 with
+  external names. These cases are now identical:
+
+  ```
+  vec_rbind(a = 1)
+  vec_rbind(c(a = 1))
+  ```
+
 * The `c()` method for `vctrs_vctr` now throws an error when
   `recursive` or `use.names` is supplied (#791).
 
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,6 +1,8 @@
 
 # vctrs (development version)
 
+* `vec_rbind()` now binds row names.
+
 * The `c()` method for `vctrs_vctr` now throws an error when
   `recursive` or `use.names` is supplied (#791).
 
--- a/R/bind.R
+++ b/R/bind.R
@@ -31,9 +31,12 @@
 #' * `vec_ptype(vec_cbind(x, y)) == vec_cbind(vec_ptype(x), vec_ptype(x))`
 #' @param ... Data frames or vectors.
 #'
-#'   `vec_rbind()` ignores names unless `.names_to` is supplied.
-#'   `vec_cbind()` creates packed data frame columns with named
-#'   inputs.
+#'   When the inputs are named:
+#'   * `vec_rbind()` assigns names to row names unless `.names_to` is
+#'     supplied. In that case the names are assigned in the column
+#'     defined by `.names_to`.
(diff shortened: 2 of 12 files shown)


lionel-:
hmmm I think I was confused. I now realise external names should probably be treated as row names? Internal names are the column names, and external names would be the row names. This would require a non-trivial matching logic, perhaps it's best to make external names an error for now.  WDYT @hadley @DavisVaughan?

Relatedly, `vec_rbind()` strips row names atm:

```r
vec_rbind(mtcars[1:3, ], mtcars[4, ])
#>    mpg cyl disp  hp drat    wt  qsec vs am gear carb
#> 1 21.0   6  160 110 3.90 2.620 16.46  0  1    4    4
#> 2 21.0   6  160 110 3.90 2.875 17.02  0  1    4    4
#> 3 22.8   4  108  93 3.85 2.320 18.61  1  1    4    1
#> 4 21.4   6  258 110 3.08 3.215 19.44  1  0    3    1
```

lionel-:
> This would require a non-trivial matching logic,

In fact this just needs a round of "unique" name-repair on the row names after concatenation, this should be rather easy.

hadley:
Yeah, it makes sense for external names to become row names

DavisVaughan:
I thought external names were only used with `.names_to`? What would any of these do if they are treated as row names? The current behavior feels right to me, and flexible enough in all the right ways (the id column feels more useful than making row names).

``` r
library(vctrs)

vec_rbind(a = 1, b = 2)
#> New names:
#> * `` -> ...1
#> New names:
#> * `` -> ...1
#>   ...1
#> 1    1
#> 2    2

vec_rbind(a = 1, b = 2, .names_to = "id")
#> New names:
#> * `` -> ...1
#> New names:
#> * `` -> ...1
#>   ...1 id
#> 1    1  a
#> 2    2  b

vec_rbind(a = c(x = 1), b = c(x = 2), .names_to = "id")
#>   x id
#> 1 1  a
#> 2 2  b

vec_rbind(a = c(x = 1), b = c(y = 2), .names_to = "id")
#>    x  y id
#> 1  1 NA  a
#> 2 NA  2  b
```

<sup>Created on 2020-02-12 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

lionel-:
Good point, but external names to rownames also makes sense for symmetry with cbind. I don't think treating external names as rownames when `.names_to` is `NULL` is incompatible with the names-to behaviour, both could coexist. By default they go to rownames, and when id is set they go to a column.

lionel-:
@hadley @DavisVaughan 

`vec_rbind()` now binds row names:

```r
vec_rbind(
  mtcars[1:3, ],
  mtcars[4, ]
)
#>                 mpg cyl disp  hp drat    wt  qsec vs am gear carb
#> Mazda RX4      21.0   6  160 110 3.90 2.620 16.46  0  1    4    4
#> Mazda RX4 Wag  21.0   6  160 110 3.90 2.875 17.02  0  1    4    4
#> Datsun 710     22.8   4  108  93 3.85 2.320 18.61  1  1    4    1
#> Hornet 4 Drive 21.4   6  258 110 3.08 3.215 19.44  1  0    3    1
```

If external names are supplied, they define row names:

```r
vec_rbind(a = c(foo = 1), b = c(bar = 2))
#>   foo bar
#> a   1  NA
#> b  NA   2
```

Unless `.names_to` is supplied:

```r
vec_rbind(a = c(foo = 1), b = c(bar = 2), .names_to = "id")
#>   foo bar id
#> 1   1  NA  a
#> 2  NA   2  b
```

If there are multiple rows in a named input, the name is sequentialised as in tidyselect:

```r
vec_rbind(
  foo = unrownames(mtcars[1:3, ]),
  mtcars[4, ]
)
#>                 mpg cyl disp  hp drat    wt  qsec vs am gear carb
#> foo1           21.0   6  160 110 3.90 2.620 16.46  0  1    4    4
#> foo2           21.0   6  160 110 3.90 2.875 17.02  0  1    4    4
#> foo3           22.8   4  108  93 3.85 2.320 18.61  1  1    4    1
#> Hornet 4 Drive 21.4   6  258 110 3.08 3.215 19.44  1  0    3    1
```

If there are existing row names in a named input, the names are combined:

```r
vec_rbind(
  foo = mtcars[1:3, ],
  mtcars[4, ],
)
#>                      mpg cyl disp  hp drat    wt  qsec vs am gear carb
#> foo...Mazda RX4     21.0   6  160 110 3.90 2.620 16.46  0  1    4    4
#> foo...Mazda RX4 Wag 21.0   6  160 110 3.90 2.875 17.02  0  1    4    4
#> foo...Datsun 710    22.8   4  108  93 3.85 2.320 18.61  1  1    4    1
#> Hornet 4 Drive      21.4   6  258 110 3.08 3.215 19.44  1  0    3    1
```

DavisVaughan:
Oh yea, this makes sense

PR #810, opened 2020-02-12
Add internal FAQ about identity element of `vec_ptype2()`

lionel-:


man/faq/internal/ptype2-identity.Rmd
```diff
@@ -0,0 +1,81 @@
+
+```{r, child = "setup.Rmd", include = FALSE}
+```
+
+## Promotion monoid
+
+`vec_ptype2()` returns the _promotion_ type of two vectors, if any. This is typically the richer type of the two so that automatic coercions are never lossy. For example, the promotion type of integer and double is the latter because double covers a larger range of values than integer.
```
hadley: Can we just use richer type instead of promotion type?

man/faq/internal/ptype2-identity.Rmd
```diff
@@ -0,0 +1,81 @@
...
+  ""
+)
+```
+
+
+## Unspecified vectors
+
+In the vctrs coercion system, logical vectors of missing values are also automatically promoted to the type of any other vector, just like `NULL`. We call these vectors unspecified. The special coercion semantics of unspecified vectors serve two purposes:
+
+1. It makes it possible to assign vectors of `NA` inside any type of vectors, even when they are not coercible with logical:
```
DavisVaughan: Is this even for foreign classes? We had a test where `NA` assignment fails for `foobar()`, and I added another one that is highly specific to unspecified types here https://github.com/r-lib/vctrs/pull/812/files#diff-51c944df297abe15044974191b4f4ebbR271-R308

Practically, the problem with allowing any foreign type is that we have to cast `NA` to `<foreign-type>` at some point. I think ideally we'd just call `vec_init(to, vec_size(x))`, but this doesn't work for all types, i.e. integer64, which doesn't implement `NA_integer_` slicing in the CRAN version.

lionel-: This should work for foreign classes as well if they implement ptype2 correctly. `foobar` is not a real class so the methods might be buggy in tests, and need to be adjusted sometimes. As for integer64, it is buggy. `vec_init()` is a separate issue though, this is about ptype2 and unspecified, and perhaps proxy/restore.

lionel-: oh now I see what you mean about init, it's used in `vec_default_cast()` for the default coercion of unspecified to the `to` type.

PR #812, opened 2020-02-12
Add native handling of `unspecified()`

DavisVaughan:
This PR adds native handling of `unspecified()`, which is one of the things slowing down https://github.com/tidyverse/dplyr/issues/4561.

This PR is the result of much conversation with @lionel- about what unspecified should be doing, especially in relation to `vec_ptype2()` and `vec_ptype_common()`.

This PR adds 5 major things:

1. `vec_typeof()` can now return `vctrs_type_unspecified`. This is helpful for controlling ptype2 and cast methods. `vec_base_typeof()` will never return this, meaning that `vec_proxy_typeof()` will continue to return `vctrs_type_logical` for unspecified vectors. This is nice because this is what `vec_compare()` and `vec_equal()` end up using, so no additional special handling is required for this new internal type.

2. The cast dispatch works by catching if `x` is unspecified, and having special behavior if so. Most of the time it just calls `vec_init(to, vec_size(x))`. If `y` happens to be s3, then a dispatch routine specific to `x` being unspecified and `y` being s3 is run to allow us to handle any "known" s3 classes more efficiently.

3. The ptype2 dispatch works similarly to the cast dispatch, except it checks if either `x` or `y` is unspecified. There is special handling for `vec_ptype2(NA, list())` here, which _currently_ should return an error, although @lionel- and I are unsure of this.

4. `vec_typeof2()` and `vec_typeof2_s3()` know about `vctrs_type_unspecified`. Even though the enum values aren't really used in the ptype2 and cast dispatch, I think it makes sense to flesh these out.

5. After talking with @lionel-, I've changed `vec_ptype2(NA, NA)` so that it returns `unspecified()`. `vec_ptype_common()` will finalise this, so it will still return `logical()`. This fixes iterative `vec_ptype2()` calls like `vec_ptype2(vec_ptype2(NA, NA), "")`, which previously errored.

Below are some benchmarks. Improvements are pretty good, but I'd like to see more improvements in the `vec_ptype2(<unspecified>, <bare-factor>)` cases. This will end up calling `vec_type()`, which for S3 objects calls `vec_slice(<bare-factor>, NULL)`. An improved factor slicing method might help here, but I'm not sure yet. We could also special case our "known" classes in the `s3_type()` routine that `vec_type()` calls, avoiding the slice entirely.

``` r
library(vctrs)

na <- NA
uns <- unspecified()

x <- 1
y <- factor("x")
```

```r
# before
bench::mark(vec_ptype2(uns, uns))
#> # A tibble: 1 x 6
#> 1 vec_ptype2(uns, uns)   12.4µs     16µs    52382.    15.1KB     15.7

# after
bench::mark(vec_ptype2(uns, uns))
#> # A tibble: 1 x 6
#> 1 vec_ptype2(uns, uns)    945ns   1.23µs   564630.        0B        0
```

```r
# before
bench::mark(vec_ptype2(na, y))
#> # A tibble: 1 x 6
#> 1 vec_ptype2(na, y)   12.7µs   16.3µs    60953.    26.7KB     24.4

# after
bench::mark(vec_ptype2(na, y))
#> # A tibble: 1 x 6
#> 1 vec_ptype2(na, y)   7.79µs   10.3µs    91157.    10.7KB     18.2
```

```r
# before
bench::mark(vec_ptype2(uns, x))
#> # A tibble: 1 x 6
#> 1 vec_ptype2(uns, x)   3.78µs   5.07µs   194677.        0B     19.5

# after
bench::mark(vec_ptype2(uns, x))
#> # A tibble: 1 x 6
#> 1 vec_ptype2(uns, x)    758ns   1.01µs   866591.        0B        0
```

```r
# before
bench::mark(vec_ptype2(uns, y))
#> # A tibble: 1 x 6
#> 1 vec_ptype2(uns, y)   10.7µs   12.7µs    73955.        0B     22.2

# after
bench::mark(vec_ptype2(uns, y))
#> # A tibble: 1 x 6
#> 1 vec_ptype2(uns, y)   7.37µs   8.44µs   113308.        0B     34.0
```

```r
# before
bench::mark(vec_cast(na, y))
#> # A tibble: 1 x 6
#> 1 vec_cast(NA, y)   16.7µs   19.7µs    47827.    33.4KB     23.9

# after
bench::mark(vec_cast(na, y))
#> # A tibble: 1 x 6
#> 1 vec_cast(na, y)   5.89µs   6.88µs   133945.        0B     26.8
```

```r
# before
bench::mark(vec_cast(uns, x))
#> # A tibble: 1 x 6
#> 1 vec_cast(uns, x)   13.5µs   15.6µs    61904.    7.01KB     24.8

# after
bench::mark(vec_cast(uns, x))
#> # A tibble: 1 x 6
#> 1 vec_cast(uns, x)   1.19µs   1.66µs   590555.        0B        0
```

```r
# before
bench::mark(vec_cast(uns, y))
#> # A tibble: 1 x 6
#> 1 vec_cast(uns, y)   18.9µs   21.6µs    44425.        0B     17.8

# after
bench::mark(vec_cast(uns, y))
#> # A tibble: 1 x 6
#> 1 vec_cast(uns, y)   6.08µs   7.18µs   132693.        0B     26.5
```

### DIFF
--- a/R/type-bare.R
+++ b/R/type-bare.R
@@ -86,11 +86,11 @@ vec_ptype2.list <- function(x, y, ..., x_arg = "x", y_arg = "y") {
 #' @method vec_ptype2.logical logical
 #' @export
 vec_ptype2.logical.logical <- function(x, y, ..., x_arg = "x", y_arg = "y") {
-  # Special case `vec_ptype2(NA, NA)` to ensure that
-  # `unspecified()` is returned
   if (is.object(y)) {
     vec_default_ptype2(x, y, x_arg = x_arg, y_arg = y_arg)
   } else if (is_unspecified(x) && is_unspecified(y)) {
+    # Special case `vec_ptype2(NA, NA)` to ensure that
+    # `unspecified()` is returned
     unspecified()
   } else {
     shape_match(logical(), x, y)
--- a/R/type-bare.R
+++ b/R/type-bare.R
@@ -86,8 +86,12 @@ vec_ptype2.list <- function(x, y, ..., x_arg = "x", y_arg = "y") {
 #' @method vec_ptype2.logical logical
 #' @export
 vec_ptype2.logical.logical <- function(x, y, ..., x_arg = "x", y_arg = "y") {
+  # Special case `vec_ptype2(NA, NA)` to ensure that
+  # `unspecified()` is returned
   if (is.object(y)) {
     vec_default_ptype2(x, y, x_arg = x_arg, y_arg = y_arg)
+  } else if (is_unspecified(x) && is_unspecified(y)) {
+    unspecified()
   } else {
     shape_match(logical(), x, y)
   }
--- a/tests/testthat/test-slice-assign.R
+++ b/tests/testthat/test-slice-assign.R
@@ -93,9 +93,13 @@ test_that("atomics can't be assigned in lists", {
 # TODO - Revisit this behavior, as it seems inconsistent from our emerging
 # treatment of `NA` as `list(NULL)` and makes internal handling more
 # complicated
-test_that("unspecified() can be assigned in lists, but NA unspecified cannot", {
+test_that("NA vector of unspecified cannot be assigned into lists", {
   x <- list(1, 2)
   expect_error(vec_slice(x, 1) <- NA, class = "vctrs_error_incompatible_type")
+})
+
+test_that("monitoring test - unspecified() can be assigned in lists", {
+  x <- list(1, 2)
   expect_error(vec_slice(x, 1) <- unspecified(1), NA)
   expect_equal(x, list(NULL, 2))
 })
--- a/tests/testthat/test-slice-assign.R
+++ b/tests/testthat/test-slice-assign.R
@@ -294,6 +294,18 @@ test_that("slice-assign falls back to `[<-` when proxy is not implemented", {
   expect_identical(obj, c("dispatched", "dispatched", "baz"))
 })
 
+test_that("vec_assign() cannot assign unspecified values into foreign vector types", {
+  obj <- foobar(c("foo", "bar", "baz"))
+
+  # These are correctly incompatible cast errors, not type errors.
+  # There is a common type, it is `vec_ptype(obj)`, but you can't cast
+  # an unspecified vector to foreign types without implementing a method for it.
+  # (Just calling `vec_init(to, vec_size(x))` for all classes wouldn't work,
+  # since classes like integer64 might not support `NA_integer_` subsetting)
+  expect_error(vec_assign(obj, 1, NA), class = "vctrs_error_incompatible_cast")
+  expect_error(vec_assign(obj, 1, unspecified(1)), class = "vctrs_error_incompatible_cast")
+})
+
 test_that("slice-assign can assign unspecified values into foreign vector types", {
   obj <- foobar(c("foo", "bar", "baz"))
   expect_identical(vec_assign(obj, 1:2, NA), foobar(c(NA, NA, "baz")))
--- a/tests/testthat/test-slice-assign.R
+++ b/tests/testthat/test-slice-assign.R
@@ -306,12 +306,6 @@ test_that("vec_assign() cannot assign unspecified values into foreign vector typ
   expect_error(vec_assign(obj, 1, unspecified(1)), class = "vctrs_error_incompatible_cast")
 })
 
-test_that("slice-assign can assign unspecified values into foreign vector types", {
-  obj <- foobar(c("foo", "bar", "baz"))
-  expect_identical(vec_assign(obj, 1:2, NA), foobar(c(NA, NA, "baz")))
-  expect_identical(vec_assign(obj, 1:2, unspecified(1)), foobar(c(NA, NA, "baz")))
-})
-
 test_that("slice-assign restores value before falling back to `[<-` (#443)", {
   called <- FALSE
 
--- a/tests/testthat/test-slice-assign.R
+++ b/tests/testthat/test-slice-assign.R
@@ -83,11 +83,25 @@ test_that("can assign lists", {
 
 test_that("atomics can't be assigned in lists", {
   x <- list(NULL)
-  expect_error(vec_slice(x, 1) <- NA, class = "vctrs_error_incompatible_type")
-  expect_error(vec_assign(x, 1, NA), class = "vctrs_error_incompatible_type")
+  expect_error(vec_slice(x, 1) <- 1, class = "vctrs_error_incompatible_type")
+  expect_error(vec_assign(x, 1, 2), class = "vctrs_error_incompatible_type")
 
   expect_error(vec_slice(x, 1) <- "foo", class = "vctrs_error_incompatible_type")
   expect_error(vec_assign(x, 1, "foo"), class = "vctrs_error_incompatible_type")
+})
+
+test_that("unspecified can be assigned in lists", {
+  x <- list(1, 2)
+  expect_error(vec_slice(x, 1) <- NA, NA)
+  expect_equal(x, list(NULL, 2))
+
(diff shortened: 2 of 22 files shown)


R/type-bare.R
```diff
@@ -86,8 +86,12 @@ vec_ptype2.list <- function(x, y, ..., x_arg = "x", y_arg = "y") {
 #' @method vec_ptype2.logical logical
 #' @export
 vec_ptype2.logical.logical <- function(x, y, ..., x_arg = "x", y_arg = "y") {
+  # Special case `vec_ptype2(NA, NA)` to ensure that
```
hadley: I think this comment should go in the block below

tests/testthat/test-slice-assign.R
```diff
@@ -83,13 +83,23 @@ test_that("can assign lists", {
...
+  expect_error(vec_assign(x, 1, 2), class = "vctrs_error_incompatible_type")
 
   expect_error(vec_slice(x, 1) <- "foo", class = "vctrs_error_incompatible_type")
   expect_error(vec_assign(x, 1, "foo"), class = "vctrs_error_incompatible_type")
 })
 
+# TODO - Revisit this behavior, as it seems inconsistent from our emerging
+# treatment of `NA` as `list(NULL)` and makes internal handling more
+# complicated
+test_that("unspecified() can be assigned in lists, but NA unspecified cannot", {
```
lionel-: Why do we allow assigning unspecified? This seems odd to me, like unspecified has leaked where it shouldn't.

tests/testthat/test-slice-assign.R
```diff
@@ -280,10 +290,22 @@ test_that("slice-assign falls back to `[<-` when proxy is not implemented", {
...
   obj <- foobar(c("foo", "bar", "baz"))
-  vec_slice(obj, 1:2) <- NA
+  vec_slice(obj, 1:2) <- TRUE
   expect_identical(obj, c("dispatched", "dispatched", "baz"))
 })
 
+test_that("vec_assign() cannot assign unspecified values into foreign vector types", {
+  obj <- foobar(c("foo", "bar", "baz"))
+
+  # These are correctly incompatible cast errors, not type errors.
```
lionel-: This should work if we hard-code casts of unspecified vectors to vec-init right? (provided `[` or proxy/restore are correctly implemented). I think we should go in that direction.

tests/testthat/test-type-unspecified.R
```diff
@@ -77,3 +77,19 @@ test_that("unspecified() validates input", {
 test_that("tibble::type_sum() knows about unspecified", {
   expect_identical(tibble::type_sum(unspecified(3)), "???")
 })
+
+test_that("casting to a scalar type errors", {
+  expect_error(vec_cast(NA, quote(x)), "Not a vector")
+  expect_error(vec_cast(unspecified(1), quote(x)), "Not a vector")
+})
+
+test_that("can cast to unspecified from unspecified", {
```
lionel-: Should this be an error? When do we need to cast to unspecified? I see "unspecified" as a temporary type during the richer type reduction. Once we have the final type, it shouldn't be unspecified anymore, but logical.

lionel-: If it is needed somewhere because of internal type manipulations it is fine to keep it - but then I wonder if we should mark these tests as "monitoring" to distinguish them from definitional tests. I.e. the test doesn't define correct behaviour. I would also mark the slice(uns) tests as monitoring for instance.

tests/testthat/test-slice-assign.R
DavisVaughan: What should we do about integer64? `vec_init()` won't work because `[.integer64` doesn't work with `NA_integer_`.

I think I agree with you that I think this should be hardcoded to `vec_init(to, vec_size(x))`, but that would break the integer64 case. Normally I'd say that that would be the fault of the package author, and not us, but otherwise it would probably break legacy code in dplyr/tidyr that uses integer64

lionel-: I've been thinking we should just special-case the correct behaviour for integer64.

R/type-bare.R
```diff
@@ -86,8 +86,12 @@ vec_ptype2.list <- function(x, y, ..., x_arg = "x", y_arg = "y") {
...
+  # `unspecified()` is returned
   if (is.object(y)) {
     vec_default_ptype2(x, y, x_arg = x_arg, y_arg = y_arg)
+  } else if (is_unspecified(x) && is_unspecified(y)) {
```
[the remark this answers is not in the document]
DavisVaughan: Hmm, `shape_match()` returns `type` immediately rather than shape matching if either `x` or `y` are objects, which in this case they both are since unspecified vectors are classed.

I could unclass them?

```r
shape_match(unspecified(), unclass(x), unclass(y))
```

tests/testthat/test-slice-assign.R
DavisVaughan: Fixed now that #813 is supported

tests/testthat/test-slice-assign.R
DavisVaughan: I think our general stance is to always allow assigning unspecified.

`vec_ptype2.logical.list()` was special cased to not allow `vec_assign(list(1), 1, NA)`, but `vec_assign(list(1), 1, unspecified(1))` happened to slip through because it isn't caught by that method. This special case complicates the `vec_type2()` C code and is why I need `vec_ptype2_dispatch_unspecified_list()`.

I personally think that `vec_assign(list(1), 1, NA)` should be allowed, and is well defined. `NA` is cast to `list(NULL)`, and then assigned.

So my suggestion would be to:

- Remove `vec_ptype2.logical.list()` entirely, and let `vec_ptype2.logical.default()` be called

- Remove the C level `vec_ptype2_dispatch_unspecified_list()` special case

- Update this test to allow both `unspecified()` and `NA` to be assigned into lists

tests/testthat/test-type-unspecified.R
DavisVaughan: I think we definitely need `vec_cast(NA, NA)` to work, and at the `vec_typeof()` level that looks identical to `vec_cast(unspecified(1), unspecified(1))`. Both deal with `vctrs_type_unspecified` inputs.

So we can't just error outright.

Right now we have:

```r
vec_cast(unspecified(1), unspecified(1)) == unspecified(1)
vec_cast(NA, unspecified(1))             == unspecified(1)
vec_cast(NA, NA)                         == NA
vec_cast(unspecified(1), NA)             == NA
```

I could make it so that if both `x` and `to` are unspecified, we always return `NA`? Meaning we would call `vec_init(vctrs_shared_empty_lgl, vec_size(x))`. That way we end up with a finalised logical?

lionel-: It's fine if it works just because of an implementation detail, it's just strange to explicitly test for it rather than testing for the actual behaviour of interest. It gives the wrong signal to the reader of the test. It doesn't seem important to monitor this behaviour, but if you'd like to keep these tests, mark them as "monitoring tests"?

tests/testthat/test-slice-assign.R
lionel-: See https://github.com/r-lib/vctrs/pull/812#discussion_r379320081
I'd just not test `unspecified()` and test `NA` instead, otherwise mark the tests as monitoring instead of definitional.

R/type-bare.R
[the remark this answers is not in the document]
lionel-: > Should a logical matrix of all NA values be considered unspecified?

I think so. If it's too much work, we could deal with it another time.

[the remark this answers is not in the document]
DavisVaughan: It breaks a few tests (in particular, one in `vec_rbind()` that looks like it could use some thought), I'm going to leave it for a separate pr

PR #816, opened 2020-02-14
`vec_ptype()` and `vec_ptype_finalise()` improvements with data frames

### LINKED ISSUE #800: Can't cast <logical> to <vctrs_unspecified> in tibbles?
The following cast errors but only for logical NA and only for tibbles (see example below). Maybe this is related to #205? Sorry, not familiar enough yet with `vctrs`, could not figure out why this fails.

```r
vctrs::vec_rbind(
  dplyr::tibble(x = 1), 
  dplyr::tibble(x = 2, y = NA)
)
# Error: Can't cast `y` <logical> to `y` <vctrs_unspecified>.
```

If using e.g. an integer NA it works as expected. Same for logical NA in a base R data frame:
```r
vctrs::vec_rbind(
    dplyr::tibble(x = 1), 
    dplyr::tibble(x = 2, y = NA_integer_)
)
## A tibble: 2 x 2
#      x     y
#  <dbl> <int>
#1     1    NA
#2     2    NA

vctrs::vec_rbind(
    data.frame(x = 1), 
    data.frame(x = 2, y = NA)
)
#   x  y
#1 1 NA
#2 2 NA
```

DavisVaughan:
Closes #800 

This PR has two goals:

- Rework `vec_type()` to be recursive over all data frame classes, rather than calling `vec_slice(x, NULL)`. This ensures unspecified columns are handled correctly.

- Rework `vec_ptype_finalise()` to be recursive over all data frame classes. This ensures that unspecified columns are finalised to logicals.

Along the way I changed:

- Renamed `vctrs_df_restore()` -> `vec_bare_df_restore()`

- New `vec_df_restore()` that restores generically by dispatching to the R method if one exists

- These two things make `vec_restore()` a bit cleaner to follow

- Similarly, renamed `df_map()` to `bare_df_map()`

- New `df_map()` that restores with the generic `vec_df_restore()`. We use this to map over classed data frames

- `s3_type()` now uses `class_type()` to allow us to recursively map over data frames. Bare tibbles are optimized, and data frames with unknown classes go through the generic map. This fixes the first problem above.

- `vec_ptype_finalise()` has been simplified a bit, and now uses `class_type()` to be recursive over all data frame classes. This fixes the second problem. `vec_ptype_finalise()` also didn't have many tests, so I added a few more.

Mainly this now works correctly:

``` r
library(vctrs)
library(tibble, warn.conflicts = FALSE)

internal_ptype <- vec_ptype(tibble(x = 2, y = NA))
internal_ptype
#> # A tibble: 0 x 2
#> # … with 2 variables: x <dbl>, y <???>

vec_ptype_finalise(internal_ptype)
#> # A tibble: 0 x 2
#> # … with 2 variables: x <dbl>, y <lgl>
```

We get nice performance improvements in determining the type of tibbles. This isn't done directly that often, but is still useful:

```r
library(vctrs)
library(tibble, warn.conflicts = FALSE)
library(rlang)

# 1 row, 100 cols
x <- as_tibble(set_names(rep_len(list(1), 100), paste0("a", 1:100)))
dim(x)
#> [1]   1 100

# before
bench::mark(vec_ptype(x))
#> # A tibble: 1 x 6
#> 1 vec_ptype(x)   17.5µs   19.6µs    48136.    7.15KB     24.1

# after
bench::mark(vec_ptype(x))
#> # A tibble: 1 x 6
#> 1 vec_ptype(x)   2.99µs   3.63µs   253962.    7.15KB     25.4
```

### DIFF
--- a/src/type.c
+++ b/src/type.c
@@ -38,6 +38,23 @@ static SEXP vec_type_slice(SEXP x, SEXP empty) {
   }
 }
 static SEXP s3_type(SEXP x) {
+  switch(class_type(x)) {
+  case vctrs_class_bare_tibble:
+    return bare_df_map(x, &vec_type);
+
+  case vctrs_class_data_frame:
+    return df_map(x, &vec_type);
+
+  case vctrs_class_bare_data_frame:
+    Rf_errorcall(R_NilValue, "Internal error: Bare data frames should be handled by `vec_type()`");
+
+  case vctrs_class_none:
+    Rf_errorcall(R_NilValue, "Internal error: Non-S3 classes should be handled by `vec_type()`");
+
+  default:
+    break;
+  }
+
   if (vec_is_vector(x)) {
     return vec_slice(x, R_NilValue);
   } else {
--- a/src/type.c
+++ b/src/type.c
@@ -63,6 +63,7 @@ static SEXP s3_type(SEXP x) {
   }
 }
 
+static SEXP vec_type_finalise_unspecified(SEXP x);
 static SEXP vec_type_finalise_dispatch(SEXP x);
 
 // [[ include("vctrs.h"); register() ]]
@@ -72,11 +73,7 @@ SEXP vec_type_finalise(SEXP x) {
   }
 
   if (vec_is_unspecified(x)) {
-    R_len_t n = Rf_length(x);
-    SEXP out = PROTECT(Rf_allocVector(LGLSXP, n));
-    r_lgl_fill(out, NA_LOGICAL, n);
-    UNPROTECT(1);
-    return out;
+    return vec_type_finalise_unspecified(x);
   }
 
   if (vec_is_partial(x)) {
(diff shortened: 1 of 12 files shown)


src/type.c
```diff
@@ -53,33 +71,65 @@ static SEXP s3_type(SEXP x) {
...
+
+  case vctrs_class_none:
+    Rf_errorcall(R_NilValue, "Internal error: Non-S3 classes should have returned by now");
+
+  default:
+    return vec_type_finalise_dispatch(x);
+  }
+}
+
+static SEXP vec_type_finalise_unspecified(SEXP x) {
```
lionel-: Use the new naming scheme `ptype`? This will reduce the amount of commit noise when we do rename old functions.

lionel-: Feel free to also rename functions as you touch them, if they are not used in too many places and the renaming causes little commit noise.

PR #824, opened 2020-02-14
Implement `new_data_frame()` in C

DavisVaughan:
I have a feeling we will be using this heavily, so a C implementation will be helpful. It wasn't particularly slow before, but if we call it a lot this could be useful.

```r
library(vctrs)

lst <- list(x = 1:100, y = 1:100)
n <- 100L
```

```r
# before
bench::mark(new_data_frame(), iterations = 100000)
#> # A tibble: 1 x 6
#> 1 new_data_frame()   3.42µs   5.03µs   199830.    30.4KB     86.0

# after
bench::mark(new_data_frame(), iterations = 100000)
#> # A tibble: 1 x 6
#> 1 new_data_frame()   1.53µs   2.19µs   443495.    9.39KB     84.3
```

```r
# before
bench::mark(new_data_frame(lst), iterations = 100000)
#> # A tibble: 1 x 6
#> 1 new_data_frame(lst)   2.98µs   3.46µs   274274.        0B     21.9

# after
bench::mark(new_data_frame(lst), iterations = 100000)
#> # A tibble: 1 x 6
#> 1 new_data_frame(lst)   1.58µs   2.05µs   463825.        0B     23.2
```

```r
# before
bench::mark(new_data_frame(lst, n = n), iterations = 100000)
#> # A tibble: 1 x 6
#> 1 new_data_frame(lst, n = n)   2.77µs   3.18µs   298890.        0B     23.9

# after
bench::mark(new_data_frame(lst, n = n), iterations = 100000)
#> # A tibble: 1 x 6
#> 1 new_data_frame(lst, n = n)   1.53µs   1.78µs   532382.        0B     26.6
```

### DIFF
--- a/src/type-data-frame.c
+++ b/src/type-data-frame.c
@@ -39,6 +39,84 @@ SEXP new_data_frame(SEXP x, R_len_t n) {
   return x;
 }
 
+static R_len_t pull_data_frame_size(SEXP x, SEXP n);
+static void set_data_frame_class(SEXP x, SEXP cls);
+
+// [[ register() ]]
+SEXP vctrs_new_data_frame(SEXP x, SEXP n, SEXP attributes, SEXP cls) {
+  if (TYPEOF(x) != VECSXP) {
+    Rf_errorcall(R_NilValue, "`x` must be a list");
+  }
+
+  R_len_t size = pull_data_frame_size(x, n);
+
+  SEXP out = PROTECT(new_data_frame(x, size));
+
+  // Set extra attributes
+  while(attributes != R_NilValue) {
+    SEXP tag = TAG(attributes);
+
+    if (tag == R_NilValue) {
+      Rf_errorcall(R_NilValue, "Attributes supplied in `...` must be named");
+    }
+
+    if (tag != R_NamesSymbol &&
+        tag != R_RowNamesSymbol &&
+        tag != R_ClassSymbol) {
+      Rf_setAttrib(out, tag, CAR(attributes));
+    }
+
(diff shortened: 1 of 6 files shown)


src/type-data-frame.c
```diff
@@ -39,6 +39,84 @@ SEXP new_data_frame(SEXP x, R_len_t n) {
...
+static R_len_t pull_data_frame_size(SEXP x, SEXP n);
+static void set_data_frame_class(SEXP x, SEXP cls);
+
+// [[ register() ]]
+SEXP vctrs_new_data_frame(SEXP x, SEXP n, SEXP attributes, SEXP cls) {
+  if (TYPEOF(x) != VECSXP) {
+    Rf_errorcall(R_NilValue, "`x` must be a list");
+  }
+
+  R_len_t size = pull_data_frame_size(x, n);
```
lionel-: What does `pull_` mean? Why is this not consistent with `df_attr()` naming? To avoid these questions, maybe just inline the function since it's used only once?

DavisVaughan: Inlining makes it _slightly_ harder to read, I think. So I've just renamed it.

Inlined is probably:

```c
  R_len_t size;
  if (n == R_NilValue) {
    size = df_raw_size_from_list(x);
  } else {
    if (TYPEOF(n) != INTSXP || Rf_length(n) != 1) {
      Rf_errorcall(R_NilValue, "`n` must be an integer of size 1");
    }
    size = r_int_get(n, 0);
  }
```

PR #827, opened 2020-02-17
Support row names in `vec_cbind()`

lionel-:
Following row names support for `vec_rbind()`, `vec_cbind()` now has support as well:

```r
df <- mtcars[1:3, 1:3]
vec_cbind(df, foo = 1)
#>                mpg cyl disp foo
#> Mazda RX4     21.0   6  160   1
#> Mazda RX4 Wag 21.0   6  160   1
#> Datsun 710    22.8   4  108   1
```

Ideally we'd match row names in `vec_cbind()` the same way we match column names in `vec_rbind()`. However that require frame genericity, with the row names encoded in a frame prototype. For the moment this is an error:

```r
vec_cbind(df, df[3:1, ])
#> Error: Can't column-bind data frames with different row names.
```

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -33,8 +33,7 @@
   is supplied, they are assigned in the column name as before.
 
 * `vec_cbind()` now binds row names if they are congruent across
-  inputs. If the row names are not identical that's an error. If some
-  inputs do not have row names, they are propagated.
+  inputs. If the row names are not identical that's an error.
 
 * The `c()` method for `vctrs_vctr` now throws an error when
   `recursive` or `use.names` is supplied (#791).
--- a/NEWS.md
+++ b/NEWS.md
@@ -27,10 +27,14 @@
 * `vec_slice()` and `vec_chop()` now work correctly with `bit64::integer64()`
   objects when an `NA` subscript is supplied. By extension, this means that
   `vec_init()` now works with these objects as well (#813).
-  
+
 * `vec_rbind()` now binds row names. When named inputs are supplied
   and `names_to` is `NULL`, the names define row names. If `names_to`
   is supplied, they are assigned in the column name as before.
+
+* `vec_cbind()` now binds row names if they are congruent across
+  inputs. If the row names are not identical that's an error. If some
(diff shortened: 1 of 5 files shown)


NEWS.md
```diff
@@ -4,11 +4,15 @@
...
   `vec_init()` now works with these objects as well (#813).
-  
+
 * `vec_rbind()` now binds row names. When named inputs are supplied
   and `names_to` is `NULL`, the names define row names. If `names_to`
   is supplied, they are assigned in the column name as before.
 
+* `vec_cbind()` now binds row names if they are congruent across
+  inputs. If the row names are not identical that's an error. If some
+  inputs do not have row names, they are propagated.
```
hadley: What does propagated mean in this context?

src/bind.c
```diff
@@ -271,8 +271,9 @@ static SEXP vec_cbind(SEXP xs, SEXP ptype, SEXP size, struct name_repair_opts* n
   R_len_t n = Rf_length(xs);
 
   // Find the common container type of inputs
-  SEXP containers = PROTECT(map(xs, &cbind_container_type));
-  ptype = PROTECT(cbind_container_type(ptype));
+  SEXP rownames = R_NilValue;
+  SEXP containers = PROTECT(map_with_data(xs, &cbind_container_type, &rownames));
+  ptype = PROTECT(cbind_container_type(ptype, &rownames));
```
DavisVaughan: I guess making `map_with_data()` take `void* data` makes it generic, but seeing `cbind_container_type()` take `void* data` was pretty confusing at first. I really expected that to be a `SEXP` and was searching through `cbind_container_type()` to figure out why it was a `void*`.

lionel-: In general taking a `void*` signals the function is used as callback with a generic higher order function. I'm not sure making the `void*` function a wrapper around a `SEXP*` function would make things clearer overall because of the added noise of having multiple functions.

NEWS.md
lionel-: That they are assigned the existing rownames. I have removed this sentence as it doesn't seem that informative.

PR #829, opened 2020-02-17
Export `vec_recycle_common()` and `vec_size_common()` callables

DavisVaughan:
For https://github.com/tidyverse/tidyr/pull/888


DavisVaughan:
Likely not needed, as we have a better approach in https://github.com/tidyverse/tidyr/pull/888

Issue #625, opened 2019-10-19
Working with the `vec_assign_impl()` callable


DavisVaughan:
So in {slide} I use `vec_assign_impl(clone = false)` for building up my output container, which works great for the most part. But it doesn’t capture the case where the output container is a matrix / we don’t have a proxy method, because in those cases we fall back to `vec_assign_fallback()` earlier in `vec_assign()`, not in `vec_assign_impl()`
https://github.com/r-lib/vctrs/blob/8d12bfc0e29e056966e0549af619253253752a64/src/slice-assign.c#L46

This means that tests like these fail for me in `slide_vec()`, where the output container could be a matrix
https://github.com/DavisVaughan/slide/blob/a070c3025a88904348b302a7a77d2cfad0d52aa0/tests/testthat/test-slide-vec.R#L106

I see two possible solutions:

- We move this check into `vec_assign_impl()`, but that seems odd because if I set `clone = false` and we do have to fallback, because I’m not sure if it would work right (would it clone anyways?).
- I have to do this check myself in {slide}. That means I would need `vec_proxy_info()` and `vec_assign_fallback()` callables. And I guess I’d have to know how to work with the return value of a `struct vctrs_proxy_info` in slide.

DavisVaughan:
One consequence of the way this currently works is that we can't assign into a data frame with a matrix column.

``` r
library(vctrs)
library(tibble)

df <- tibble(x = 1:2, y = matrix(1:4, nrow = 2))
df
#> # A tibble: 2 x 2
#>       x y[,1]  [,2]
#>   <int> <int> <int>
#> 1     1     1     3
#> 2     2     2     4

vec_slice(df, 2) <- df[1,]
#> Error in `vec_slice<-`(`*tmp*`, 2, value = structure(list(x = 1L, y = structure(c(1L, : Internal error in `vec_assign()`: `value` should have been recycled to fit `x`.
```

<sup>Created on 2019-10-20 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

This is because `df_assign()` uses `vec_assign_impl()`, bypassing the fallback which should be called.

https://github.com/r-lib/vctrs/blob/142a257e4dbad97df4923ee72767b4df7a192d45/src/slice-assign.c#L239

lionel-:
Ideally vec-slice-impl would know how to slice arrays, and the fallback would only be used for OO objects.

Issue #838, opened 2020-02-19
Switch to `MAYBE_SHARED()` in `r_maybe_duplicate()`


lionel-:
`MAYBE_REFERENCED()` returns true if there is only 1 reference. `MAYBE_SHARED()` returns false in this case.

This won't matter much for a while even when R switches to reference count because the namedness/refcount of objects is bumped to NAMEDMAX when a promise is resolved. For this reason, even temporaries won't be considered safe for mutation:

```r
# `1:100` is wrapped in a promise, evaluated by `.Call()`, and bumped to NAMEDMAX
vec_assign(1:100, 1, 10)
```

However when R solves this issue with promises, we should be able to get perf improvements with temporaries. But for this we'll need to allow 1 reference, the one held by the closure wrapper.

Switching to it right away would also make it easier to assess performance in the no-copy cases, see https://github.com/r-lib/vctrs/pull/837#issuecomment-588124814.

lionel-:
```diff
diff --git a/src/utils.c b/src/utils.c
index f26d800a..adf703c2 100644
--- a/src/utils.c
+++ b/src/utils.c
@@ -1059,7 +1059,7 @@ bool r_is_function(SEXP x) {
 }
 
 SEXP r_maybe_duplicate(SEXP x) {
-  if (MAYBE_REFERENCED(x)) {
+  if (MAYBE_SHARED(x)) {
     return Rf_shallow_duplicate(x);
   } else {
     return x;
```

PR #842, opened 2020-02-19
Always call `r_maybe_duplicate()` in `vec_assign_impl()`

DavisVaughan:
After adding ALTREP support in `vec_slice_impl()` in #837, I discovered that we are going to need to change `vec_assign_impl()` if we want to merge that PR.

The issue is that, if `clone = false`, we modify `x` directly by accessing it with the corresponding `DATAPTR()` regardless of its NAMED status. This causes a problem with ALTREP reps.

Consider `vec_rbind()`. It calls `vec_init()` to initialize the output. Integer and double columns here are initialized as ALTREP reps.

When it goes to assign into these with `df_assign()` (and `vec_assign_impl()` for each column) it sets `clone = false`. When it hits an ALTREP `vctrs_compact_rep_int` column, it _should_ call the Duplicate method to generate the full column because we shouldn't be modifying this compact data (it is marked as not mutable). Instead it bypasses `Rf_shallow_duplicate()` and calls the corresponding `DEREF()`, which calls my `DATAPTR()` method. This expands the values, sets them in the ALTREP data2 slot, and then returns the `DATAPTR()` to that (exactly what the compact_intseq method does). So we end up overwriting this data, which I don't think is correct and ends up causing problems because we have a `vctrs_compact_rep_int` column that isn't guaranteed to always be a rep of the same value.

This note from R core explains the reasoning used in compact_intseq a little more. I think that what we are doing here would be classified as "mis-behaving C code" to them, since we have an object with NAMED-ness of 7, but we assign directly into it. https://github.com/wch/r-source/blob/8e1209bd032c8a3f88925221e4247ad019b8ebed/src/main/altclasses.c#L60-L67

My solution is to remove the `clone` argument to `vec_assign_impl()` and always call `r_maybe_duplicate()`. This should have the correct behavior, and shouldn't be any slower.

I think we should think about `vec_assign_impl()` as: _"it clones the proxy if it needs to, i.e. if it is referenced or marked as not mutable. otherwise it assigns directly into the proxy."_

I think this reasoning would actually allow us to refactor `vec_assign()` and `vec_assign_impl()` in a way that pushes the fallback handling into `vec_assign_impl()`, along with the coercible casting, value proxying, and value recycling. This would help with https://github.com/r-lib/vctrs/issues/625, fixing a few bugs that result from that, and make @romainfrancois's job in dplyr easier since the edge cases are all handled by `vec_assign_impl()` directly.

This would break {slider} because I use the `vec_assign_impl()` callable, but the benefits are more than worth it if we also make `vec_assign_impl()` able to handle the fallback cases

### DIFF
--- a/inst/include/vctrs.c
+++ b/inst/include/vctrs.c
@@ -2,7 +2,7 @@
 
 SEXP (*vec_proxy)(SEXP) = NULL;
 SEXP (*vec_restore)(SEXP, SEXP, SEXP) = NULL;
-SEXP (*vec_assign_impl)(SEXP, SEXP, SEXP, bool) = NULL;
+SEXP (*vec_assign_impl)(SEXP, SEXP, SEXP) = NULL;
 SEXP (*vec_slice_impl)(SEXP, SEXP) = NULL;
 SEXP (*vec_names)(SEXP) = NULL;
 SEXP (*vec_set_names)(SEXP, SEXP) = NULL;
@@ -11,7 +11,7 @@ SEXP (*vec_chop)(SEXP, SEXP) = NULL;
 void vctrs_init_api() {
   vec_proxy = (SEXP (*)(SEXP)) R_GetCCallable("vctrs", "vec_proxy");
   vec_restore = (SEXP (*)(SEXP, SEXP, SEXP)) R_GetCCallable("vctrs", "vec_restore");
-  vec_assign_impl = (SEXP (*)(SEXP, SEXP, SEXP, bool)) R_GetCCallable("vctrs", "vec_assign_impl");
+  vec_assign_impl = (SEXP (*)(SEXP, SEXP, SEXP)) R_GetCCallable("vctrs", "vec_assign_impl");
   vec_slice_impl = (SEXP (*)(SEXP, SEXP)) R_GetCCallable("vctrs", "vec_slice_impl");
   vec_names = (SEXP (*)(SEXP)) R_GetCCallable("vctrs", "vec_names");
   vec_set_names = (SEXP (*)(SEXP, SEXP)) R_GetCCallable("vctrs", "vec_set_names");
--- a/inst/include/vctrs.h
+++ b/inst/include/vctrs.h
@@ -7,7 +7,7 @@
 
 extern SEXP (*vec_proxy)(SEXP);
 extern SEXP (*vec_restore)(SEXP, SEXP, SEXP);
-extern SEXP (*vec_assign_impl)(SEXP, SEXP, SEXP, bool);
+extern SEXP (*vec_assign_impl)(SEXP, SEXP, SEXP);
 extern SEXP (*vec_slice_impl)(SEXP, SEXP);
 extern SEXP (*vec_names)(SEXP);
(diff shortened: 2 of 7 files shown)


lionel-:
Since there is no longer a guarantee of no duplication by `clone` to false, it means the result of vec-assign should always be reassigned and reprotected. Can you double check this stylistic change across the codebase please?

DavisVaughan:
> the result of vec-assign should always be reassigned and reprotected

Yes, part of this PR made the changes to ensure that that is now correct. I have double checked again, and we now protect the output from all calls to `vec_assign_impl()`, along with the lower level assign functions that we use elsewhere: `chr_assign()`, `list_assign()` and `df_assign()`.

---

> Is there any way to have mutable altreps or are they hardcoded as non mutable?

It is possible, I specifically mark them as immutable with `MARK_NOT_MUTABLE()`. In fact, R has a `#define` that can be flipped on to make the `compact_intseq`s mutable
https://github.com/wch/r-source/blob/5cd6111601fe754a2ec488d5fc7b6ffb4556989e/src/main/altclasses.c#L67

The last paragraph of this section of the ALTREP guide talks about that:
https://svn.r-project.org/R/branches/ALTREP/ALTREP.html#compact_integer_vectors

> An alternative approach would be to not mark these vectors as immutable but to treat them as standard vectors as soon as the data pointer is requested and the vector is expanded. If a writable data pointer is requested then the expanded data could be modified by an assignment. The Is_sorted and No_NA properties could no longer be assumed to be true, and the object would have to be serialized as a full vector.

It scares me a little to have them as mutable, because then if you `.Internal(inspect(x))`, where `x` is a `vctrs_compact_rep_int`, there is no guarantee that `x` has all the same values if it shows that it has been `expanded`. If it has been expanded, it can be modified by assignment but stays a `vctrs_compact_rep_int`. I like having these ALTREP reps as more ephemeral objects that make certain operations faster, but give you a standard R vector when you do anything else. Since R core chose to leave them as immutable, I assume they came to the same conclusion?

More arguments for leaving them as immutable:

- I haven't turned on the ALTREP methods for `Is_sorted` and `No_NA` yet, but I can probably make `No_NA` work if we keep it immutable. I don't think the sortedness enum R provides is set up correctly for determining the sortedness of a repeated value. 
- I do also have a serialization method that just serializes the compact data. 
- There is also an ALTREP sum method that we might could utilize. It would just be `value * size` assuming that we stay within the bounds of the data type we are using.

---

> Can you also think this through in light of #838 please?

I don't think anything would need to change. If we update to `MAYBE_SHARED()` then we'd (eventually) still get the nice behavior of not copying temporaries. ALTREP objects marked as immutable would be seen as "shared" so they'd still be duplicated correctly.

lionel-:
That makes sense.

DavisVaughan:
@romainfrancois I'm going to merge this in after the checks pass. If you are calling `vec_assign_impl()` anywhere in dplyr, you now need to `PROTECT()` and capture the result of the call, and you no longer need the `clone` argument.

Expect `vec_assign_impl()` to "clone if it needs to", i.e., if the object is referenced at all. Otherwise it modifies in place. But since we can't guarantee a modification in place, it is safest to always catch the result.

PR #847, opened 2020-02-21
Flatten df-cols in equality proxy

lionel-:
So that dictionary operations don't need to deal with recursion. This is to simplify code in an upcoming PR.

Data frames of 1 columns are unwrapped as before.

### DIFF
--- a/src/type-data-frame.c
+++ b/src/type-data-frame.c
@@ -241,3 +241,23 @@ SEXP df_poke_at(SEXP x, SEXP name, SEXP value) {
   UNPROTECT(1);
   return x;
 }
+
+// [[ include("type-data-frame.h") ]]
+R_len_t df_flat_width(SEXP x) {
+  R_len_t n = Rf_length(x);
+  R_len_t out = n;
+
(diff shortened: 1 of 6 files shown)


src/type-data-frame.c
```diff
@@ -241,3 +245,64 @@ SEXP df_poke_at(SEXP x, SEXP name, SEXP value) {
...
+  df_flatten_loop(x, out, out_names, 0);
+  init_data_frame(out, df_size(x));
+
+  UNPROTECT(2);
+  return out;
+}
+
+static R_len_t df_flatten_loop(SEXP x, SEXP out, SEXP out_names, R_len_t counter) {
+  R_len_t n = Rf_length(x);
+  SEXP x_names = PROTECT(r_names(x));
```
DavisVaughan: Do we need to worry about the very rare case of a data frame with no column names? You call `STRING_ELT()` on the names and they could be `NULL` in that case

lionel-: I think this would be a case of a corrupt data structure? In that case the `string-elt` error about wrong type-of seems sufficiently defensive.

PR #851, opened 2020-02-22
Switch to `data_frame()` when `data.frame()` is not required

DavisVaughan:
One of the R devel builds is failing because of the following, which it seems is a direct result of `stringsAsFactors` being flipped to `FALSE` by default in R 4.0.0.

To avoid any fallout, this PR preemptively swaps usage of `data.frame()` with `data_frame()` when it is not explicitly required (i.e. when we don't use row names, or when we aren't explicitly trying to check against the result of `data.frame()`)

```r
Result: ERROR
     Running 'testthat.R' [36s/41s]
    Running the tests in 'tests/testthat.R' failed.
    Complete output:
     > library(testthat)
     > library(vctrs)
     >
     > if (requireNamespace("xml2")) {
     + test_check("vctrs", reporter = MultiReporter$new(reporters = list(JunitReporter$new(file = "test-results.xml"), CheckReporter$new())))
     + } else {
     + test_check("vctrs")
     + }
     Loading required namespace: xml2
     -- 1. Failure: embedded data frames print nicely (@test-type-data-frame.R#26) -
     Results have changed from known value recorded in 'test-type-data-frame-embedded.txt'.
     2/16 mismatches
     x[6]: " b: character"
     y[6]: " b: factor<38ce1>"
    
     x[13]: " y: character"
     y[13]: " y: factor<38ce1>"
    
     == testthat results ===========================================================
     [ OK: 2618 | SKIPPED: 9 | WARNINGS: 0 | FAILED: 1 ]
     1. Failure: embedded data frames print nicely (@test-type-data-frame.R#26)
```

### DIFF
--- a/tests/testthat/helper-rational.R
+++ b/tests/testthat/helper-rational.R
@@ -25,7 +25,7 @@ vec_proxy_equal.vctrs_rational <- function(x) {
   n <- field(x, "n")
   d <- field(x, "d")
   gcd <- gcd(n, d)
-  data.frame(n = n / gcd, d = d / gcd)
+  data_frame(n = n / gcd, d = d / gcd)
 }
 gcd <- function(x, y) {
   r <- x %% y
--- a/tests/testthat/helper-types.R
+++ b/tests/testthat/helper-types.R
@@ -1,5 +1,5 @@
 # Don't call tibble::tibble() to avoid catch-22, because tibble now uses vctrs
-bare_tibble <- structure(data.frame(), class = c("tbl_df", "tbl", "data.frame"))
+bare_tibble <- structure(data_frame(), class = c("tbl_df", "tbl", "data.frame"))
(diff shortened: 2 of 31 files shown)


lionel-:
Something bothers me about this change but I'm not sure what. Probably the fact that we'd no longer be testing the data structure created by `data.frame()` in all these cases? Probably harmless but I think I'd rather avoid making this sort of switch unless it's on purpose.

We could perhaps add this to a helper file?

```r
If (old_r) {
  data.frame <- function(..., stringsAsFactors = NULL) {
    stringsAsFactors <- stringsAsFactors %||% FALSE
    base::data.frame(..., stringsAsFactors = stringsAsFactors)
  }
}
```

DavisVaughan:
Closed in favor of #856

Issue #846, opened 2020-02-21
Restrict `vec_unchop()` and `vec_chop()` to positive integer indices


DavisVaughan:
See https://github.com/r-lib/vctrs/pull/834#issuecomment-589435201

This would technically be a breaking change so we would mark it in the news

And then possibly rename `indices` to `locations`

PR #854, opened 2020-02-23
Implement `zero` handling in `num_as_location()`

### LINKED ISSUE #852: Control behavior of zeros in C level `vec_as_location()`
This would bubble up to R through `num_as_location()`

We need another option in `vec_as_location_opts` for controlling the behavior of zeros. I proposed this to @lionel- , who seemed to agree:

- remove: The current behavior, removes zeros. Problematic in some cases because this changes the size of the index.
- ignore: Leaves 0 values in place, unchanged.
- error: Errors with a custom classed error for this case

DavisVaughan:
Closes #852 

Adds a `zero` argument to `num_as_location()` with the ability to `"remove"`, `"ignore"`, or `"error"`. Will be used in the C level call to `vec_as_location()` to validate locations in #846

### DIFF
--- a/tests/testthat/error/test-subscript-loc.txt
+++ b/tests/testthat/error/test-subscript-loc.txt
@@ -41,12 +41,12 @@ num_as_location() optionally forbids zero indices
 > num_as_location(0L, 1L, zero = "error")
 Error: Must subset elements with a valid subscript vector.
 x The subscript can't contain `0` values.
-i It had a zero value at location 1.
+i It had a `0` value at location 1.
 
 > num_as_location(c(0, 0, 0, 0, 0, 0), 1, zero = "error")
 Error: Must subset elements with a valid subscript vector.
 x The subscript can't contain `0` values.
-i It has 6 zero values at locations 1, 2, 3, 4, 5, etc.
+i It has 6 `0` values at locations 1, 2, 3, 4, 5, etc.
 
 
 logical subscripts must match size of indexed vector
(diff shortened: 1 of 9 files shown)


tests/testthat/error/test-subscript-loc.txt
```diff
@@ -35,6 +35,20 @@ Error: Must subset elements with a valid subscript vector.
 x The subscript can't contain negative locations.
 
 
+num_as_location() optionally forbids zero indices
+=================================================
+
+> num_as_location(0L, 1L, zero = "error")
+Error: Must subset elements with a valid subscript vector.
+x The subscript can't contain `0` values.
+i It had a zero value at location 1.
```
lionel-: In the `x` bullet zero is numeric, in the `i` bullet it is textual. Use numeric everywhere?

tests/testthat/error/test-subscript-loc.txt
```diff
@@ -354,6 +368,11 @@ Error: Can't subset elements beyond the end with non-consecutive locations.
 i The input has size 2.
 x The subscript `foo` contains non-consecutive location 4.
 
+> num_as_location(0, 1, zero = "error", arg = "foo")
+Error: Must subset elements with a valid subscript vector.
+x The subscript `foo` can't contain `0` values.
+i It had a zero value at location 1.
```
lionel-: The other messages use the present "has", this one the past "had". Use "has" everywhere?

tests/testthat/test-subscript-loc.R
```diff
@@ -410,6 +427,10 @@ test_that("conversion to locations has informative error messages", {
     "# num_as_location() optionally forbids negative indices"
     num_as_location(dbl(1, -1), 2L, negative = "error")
 
+    "# num_as_location() optionally forbids zero indices"
+    num_as_location(0L, 1L, zero = "error")
```
lionel-: The header should correspond to a `test_that()` block containing `expect_error()` statements for the tested expressions, wrapped within `verify_errors()`.

This way there is a hard check on the expected error in addition to the output check, and the `verify_errors()` reminds us to update the corresponding block of expressions within `verify_output()`.

Issue #575, opened 2019-09-18
Ensure that we support x[[sym]]


hadley:
As base R does:

``` r
list(a = 1)[[quote(a)]]
#> [1] 1
c(a = 1)[[quote(a)]]
#> [1] 1
data.frame(a = 1)[[quote(a)]]
#> [1] 1
```

My intuition is that will tend to help with certain tidy evaluation tasks, and even if not, it's a relatively obscure feature so there shouldn't be downsides associated with implementing it.

krlmlr:
This seems to break the interaction package with dev tibble: https://github.com/tidyverse/tibble/blob/master/revdep/problems.md#interactions.

krlmlr:
Appears to work even in vctrs 0.2.3:

``` r
vctrs::vec_as_location(quote(c), 26, letters)
#> [1] 3
vctrs::vec_as_location2(quote(c), 26, letters)
#> [1] 3
```

<sup>Created on 2020-02-27 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

lionel-:
Not with `vec_slice()`, but this will be fixed on master soon once #868 is merged.

PR #857, opened 2020-02-24
Restrict `vec_unchop()` and `vec_chop()` to positive integer locations

### LINKED ISSUE #846: Restrict `vec_unchop()` and `vec_chop()` to positive integer indices

DavisVaughan:
Closes #846 

As discussed at https://github.com/r-lib/vctrs/pull/834#issuecomment-589435201, since `vec_unchop()`'s `indices` argument is all about placing elements of `x` in specified _locations_, it doesn't make sense to allow character/logical subscripts in that function. To align `vec_chop()` with `vec_unchop()`, I've restricted `vec_chop()` to only allow positive integer locations as well. `vec_unchop()` could technically allow character and logical, but it doesn't feel useful.

`NA_integer_` is still a valid location.

In the next PR I will rename `indices` to `locations`

### DIFF
--- a/src/slice-chop.c
+++ b/src/slice-chop.c
@@ -605,9 +605,28 @@ static SEXP vec_as_indices(SEXP indices, R_len_t n, SEXP names) {
 
   R_len_t size = vec_size(indices);
 
+  // Restrict index values to positive integer locations
+  struct vec_as_location_opts opts = {
+    .action = SUBSCRIPT_ACTION_DEFAULT,
+    .missing = SUBSCRIPT_MISSING_PROPAGATE,
+    .loc_negative = LOC_NEGATIVE_ERROR,
+    .loc_oob = LOC_OOB_ERROR,
+    .loc_zero = LOC_ZERO_ERROR,
+    .subscript_arg = R_NilValue
+  };
+
   for (int i = 0; i < size; ++i) {
     index = VECTOR_ELT(indices, i);
-    SET_VECTOR_ELT(indices, i, vec_as_location(index, n, names));
+
+    // Restrict the type to only numeric values.
+    // `vec_as_location_opts()` does not do this.
+    switch (TYPEOF(index)) {
+    case INTSXP: break;
+    case REALSXP: break;
+    default: Rf_errorcall(R_NilValue, "All elements of `indices` must be numeric vectors.");
+    }
+
+    SET_VECTOR_ELT(indices, i, vec_as_location_opts(index, n, names, &opts));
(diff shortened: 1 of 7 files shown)


tests/testthat/test-slice-chop.R
```diff
@@ -429,10 +417,24 @@ test_that("name repair is respected and happens after ordering according to `ind
   expect_named(vec_unchop(x, indices, name_repair = "unique"), c("a...1", "a...2"))
 })
 
-test_that("errors on OOB zero values", {
-  skip("until we can error on 0 values in vec_as_location()")
+test_that("errors on `0` locations", {
```
lionel-: Can you merge the two subscript type tests in one block and wrap in `verify_errors()` please?

src/slice-chop.c
```diff
@@ -612,9 +612,28 @@ static SEXP vec_as_indices(SEXP indices, R_len_t n, SEXP names) {
...
+    .loc_oob = LOC_OOB_ERROR,
+    .loc_zero = LOC_ZERO_ERROR,
+    .subscript_arg = R_NilValue
+  };
+
   for (int i = 0; i < size; ++i) {
     index = VECTOR_ELT(indices, i);
-    SET_VECTOR_ELT(indices, i, vec_as_location(index, n, names));
+
+    // Restrict the type to only numeric values.
```
lionel-: This is not completely right, for instance factors are implemented as numeric. We should check for OBJECT and call `vec_as_subscript()` for S3 inputs.

lionel-: I need to fix a similar problem for #575. I'll look into a C-level `vec_as_subscript()`.

DavisVaughan: I'm not sure I want S3 inputs to be allowed here. I really think I just want `num_as_location()`, but I'm now seeing that even that allows S3's

``` r
vctrs::num_as_location(factor("x"), 1)
#> [1] x
#> Levels: x
```

<sup>Created on 2020-02-25 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

lionel-: Not accepting S3 inputs would be inconsistent with other subscript handling. If we take locations we should take any type coercible to integer.

lionel-: I'll fix the num-as-location bug.

PR #858, opened 2020-02-25
Fix multiple-definition error with GCC 10.

QuLogic:
This variable is only used in one file, so it really should have remained `static`. To fix the unused-variable warning, it should be moved from the header to the source file.

### DIFF
--- a/src/altrep-rle.c
+++ b/src/altrep-rle.c
@@ -16,6 +16,8 @@ SEXP altrep_rle_Make(SEXP input) {
 
 #else
 
+static R_altrep_class_t altrep_rle_class;
+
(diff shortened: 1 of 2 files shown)


lionel-:
Thanks for looking into this! I have now added `-fno-common` to my `Makevars`.

> This variable is only used in one file, so it really should have remained static.

The intention of the author was to expose the variable in the header file, so I prefer to keep it this way. I have now marked the object `extern`.

QuLogic:
Oh, I didn't think this header was installed anywhere usable.

lionel-:
It is usable within vctrs (but not used yet).

Issue #723, opened 2020-01-02
Take care of known list classes


krlmlr:
From tibble's revdepchecks, I see the following S3 classes that wrap a list and are a vector:

Base classes should be fixed in vctrs:

- [x] `"package_version"`

These will need to be fixed in downstream packages, either by implementing vec-proxy or by adding an explicit `"list"` at the end of the vector class.

- [ ] bench: `"bench_expr"`
- [ ] concurve: `"data.frame"` (?????)
- [ ] feasts: `"lst_mdl"`
- [ ] heemod: `"lazy_dots"` (???)
- [x] jstor, textrecipes: `"quosures"`
- [ ] PML: `"trelliscope_panels"`
- [ ] psychmeta: `"ad_int_list"`
- [ ] units: `"mixed_units"`

Related: #666.

lionel-:
I think these should be fixed downstream.

krlmlr:
Can we keep this list here?

lionel-:
Yes, I've also added a "base" section to your post. Reprex for pkg-version:

``` r
vctrs::vec_is(packageVersion("tibble"))
#> [1] FALSE
data.frame(a = packageVersion("tibble"))
#>              a
#> 1 2.99.99.9011
```

<sup>Created on 2020-01-02 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

Required for {pkgsearch}.

lionel-:
Closing now, since tibble has been released.

PR #859, opened 2020-02-25
Add support for `numeric_version`

lionel-:
Part of #723

This is an interesting type because `c()` and `[` wrap vectors of size > 1 in a list-of-like type. To support those natively in vctrs, we'd need some way of customising the "wrapper" type of atomic classes.

I haven't implemented `vec_ptype2()` with character because it's unclear what is the richer type in this case [^1]. To reproduce base R semantics, it'd have to be numeric-version. Do we consider that numeric-version is a restriction on strings? Another class representing restricted strings is factor, for which character is considered to be richer:

```r
vec_ptype2(factor("foo"), "bar")
#> character(0)
```

Perhaps we need to distinguish between hard restrictions (like `list_of`) and soft restrictions (like `factor`). Choosing one or the other is a matter of balancing safety (programming) and friction (interactivity). Since numeric-version is mostly used for programming, safety seems to have greater weight.

But then, does ptype2 really always return the "richer" type? I think this touches the issue of richer behaviour (e.g. restrictions) versus richer data coverage. This is currently a weakness in our theory. Perhaps soft restrictions are not behaviour, but hard restrictions are.

[^1]: And also we'd need to implement proxy/restore and deal with the issue of wrapper type mentioned above.


tests/testthat/type-misc.R
```diff
@@ -0,0 +1,30 @@
+
+test_that("`numeric_version` is a vector (#723)", {
+  x <- utils::packageVersion("rlang")
```
hadley: Slightly safer to use the constructor, `numeric_version()`?

tests/testthat/type-misc.R
```diff
@@ -0,0 +1,30 @@
...
+  expect_identical(vec_unique(z), z)
+  expect_identical(vec_unique(c(y, z, x)), z[2:1])
+})
+
+test_that("`numeric_version` falls back to base methods", {
+  x <- utils::packageVersion("rlang")
+  y <- utils::packageVersion("vctrs")
+  z <- c(x, y)
+
+  # `z` is a `list-of`-like type but slicing 1 element returns the
```
hadley: I don't follow this.  Do you mean that `[` returns an atomic type or that `vec_slice()` does?

lionel-: They both do because `vec_slice()` falls back to `[`. To support this natively, without the fallback, we'd need to add a mechanism for wrapper types in vctrs.

krlmlr: I don't see how a scalar is different from a vector of length one.

``` r
x <- utils::packageVersion("rlang")
y <- utils::packageVersion("vctrs")
z <- c(x, y)

dput(z)
#> structure(list(c(0L, 4L, 4L), c(0L, 2L, 3L)), class = c("package_version", 
#> "numeric_version"))
dput(z[1:2])
#> structure(list(c(0L, 4L, 4L), c(0L, 2L, 3L)), class = c("package_version", 
#> "numeric_version"))
dput(z[1])
#> structure(list(c(0L, 4L, 4L)), class = c("package_version", "numeric_version"
#> ))
identical(x, x[1])
#> [1] TRUE
identical(x, x[[1]])
#> [1] TRUE
```

<sup>Created on 2020-02-25 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

krlmlr:
Thanks for looking into it. For tibble this is sufficient.

I feel that most list-like classes will be different conceptually from `list_of`, in the sense that `identical(x[1], x[[1]])` . Do we need a concept of the `of_type()` of a vector -- for `list_of` this will be the data type of the elements, but for `numeric_version` it will be `numeric_version`?

tests/testthat/type-misc.R
lionel-: oh you're right! I was thrown off by the output of `str()`:

```r
str(z)
#> List of 2
#>  $ :Classes 'package_version', 'numeric_version'  hidden list of 1
#>   ..$ : int [1:4] 0 4 4 9000
#>  $ :Classes 'package_version', 'numeric_version'  hidden list of 1
#>   ..$ : int [1:4] 0 2 99 9005
#>  - attr(*, "class")= chr [1:2] "package_version" "numeric_version"

str(z[1])
#> Classes 'package_version', 'numeric_version'  hidden list of 1
#>  $ : int [1:4] 0 4 4 9000
```

DavisVaughan: Agreed that the weirdness is limited to `str()`. Otherwise even a `numeric_version` of size 1 is a list_of-like object.

The weirdness comes from `irregCl`, which is `TRUE` for a `numeric_version` of size 1, but otherwise is `FALSE` for size >1.
https://github.com/wch/r-source/blob/ec6faae1911207f1cb9b2cf259238598dc1ffe17/src/library/utils/R/str.R#L335

So then the output is changed here based on that
https://github.com/wch/r-source/blob/ec6faae1911207f1cb9b2cf259238598dc1ffe17/src/library/utils/R/str.R#L348

lionel-: It's interesting that `str()` has built in support for lists where `[[` returns the same type.

PR #861, opened 2020-02-25
Move fallback handling into `vec_assign_impl()`

### LINKED ISSUE #625: Working with the `vec_assign_impl()` callable

### LINKED ISSUE #853: vec_slice<-() and factors
give unexpected results.

``` r
library(vctrs)

y <- factor(c("a", NA), levels = c("a", "b"))
`vec_slice<-`(y, 2, factor("b", levels = "b"))
#> [1] a    <NA>
#> Levels: a b
`vec_slice<-`(y, 1:2, factor("b", levels = "b"))
#> [1] <NA> <NA>
#> Levels: a b
```

<sup>Created on 2020-02-23 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

DavisVaughan:
Closes #853 
Closes #625 
Closes tidyverse/tidyr#788
Solves (partially) DavisVaughan/slider#2, will be completely solved with additional slider work on my end

This PR moves the fallback handling of `vec_assign()` into `vec_assign_impl()`. The separation is:

-  The caller of `vec_assign_impl()` handles
   - Coercible casting of `value`. If you already know the type is correct you don't have to cast.
   - Recycling `value`. Same reasoning as above.
   - Proxying and restoring the output container, so it is only done once. (Useful for things like `vec_c()`)

- `vec_assign_impl()` handles:
   - proxy-ing `value`. Required to be able to fallback using the non-proxied `value`
   - Handling fallback behavior

This PR also fixes a few bugs that came up with the awful corrupt objects that are `OBJECT()`s but don't have a class.

The one drawback here is that `vec_assign()` now calls `vec_recycle()` on the un-proxied `value`, then hands off to `vec_assign_impl()`, which proxies and might allocate. This is slightly less efficient than proxying first, then recycling, but I can live with this tradeoff to have a clear separation of responsibility.

The other change here is that we now `vec_restore()` after calling the fallback to `[<-`. I feel like this is the correct thing to do, and probably won't hurt, but it did require a tweak to a test that "expected" the object to not be restored.

### DIFF
--- a/inst/include/vctrs.c
+++ b/inst/include/vctrs.c
@@ -2,7 +2,7 @@
 
 SEXP (*vec_proxy)(SEXP) = NULL;
 SEXP (*vec_restore)(SEXP, SEXP, SEXP) = NULL;
-SEXP (*vec_assign_impl)(SEXP, SEXP, SEXP) = NULL;
+SEXP (*vec_proxy_assign)(SEXP, SEXP, SEXP) = NULL;
 SEXP (*vec_slice_impl)(SEXP, SEXP) = NULL;
 SEXP (*vec_names)(SEXP) = NULL;
 SEXP (*vec_set_names)(SEXP, SEXP) = NULL;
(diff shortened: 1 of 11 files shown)


DavisVaughan:
Renaming makes sense!

Issue #864, opened 2020-02-26
Move `vec_as_subscript()` to C


DavisVaughan:
<img width="765" alt="Screen Shot 2020-02-25 at 10 24 52 PM" src="https://user-images.githubusercontent.com/19150088/75308953-b14b0100-581d-11ea-92a0-998b9ca74f83.png">

It pops up a few times in a flame graph for `vars_select_eval()`, making it a somewhat significant portion of repeated tidyselect calls.

PR #866, opened 2020-02-26
Implement `vec_as_subscript()` in C

lionel-:
Branched from #865.
Part of #864.

### DIFF
--- a/src/cast.c
+++ b/src/cast.c
@@ -489,6 +489,7 @@ static void vec_coercible_cast_e_cb(void* data_) {
   data->out = vec_cast(data->x, data->to, data->x_arg, data->to_arg);
 }
 
+// [[ include("vctrs.h") ]]
 SEXP vec_coercible_cast_e(SEXP x,
                           SEXP to,
                           struct vctrs_arg* x_arg,
@@ -510,7 +511,7 @@ SEXP vec_coercible_cast_e(SEXP x,
 
   *err = r_try_catch(&vec_coercible_cast_e_cb,
                      &data,
-                     Rf_install("vctrs_error_cast_lossy"),
+                     syms_vctrs_error_cast_lossy,
                      NULL,
                      NULL);
   return data.out;
--- a/src/cast.c
+++ b/src/cast.c
@@ -423,8 +423,12 @@ static void vec_is_coercible_cb(void* data_) {
   vec_type2(data->x, data->y, data->x_arg, data->y_arg, data->dir);
 }
 
-// [[ include("vctrs.h") ]]
-bool vec_is_coercible(SEXP x, SEXP y, struct vctrs_arg* x_arg, struct vctrs_arg* y_arg, int* dir) {
+static void vec_is_coercible_e(SEXP x,
+                               SEXP y,
+                               struct vctrs_arg* x_arg,
+                               struct vctrs_arg* y_arg,
+                               int* dir,
+                               ERR* err) {
   struct vec_is_coercible_data data = {
     .x = x,
     .y = y,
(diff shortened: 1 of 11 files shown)


src/cast.c
```diff
@@ -462,6 +477,45 @@ SEXP vec_coercible_cast(SEXP x, SEXP to, struct vctrs_arg* x_arg, struct vctrs_a
...
+  struct vctrs_arg* x_arg;
+  struct vctrs_arg* to_arg;
+  SEXP out;
+};
+static void vec_coercible_cast_e_cb(void* data_) {
+  struct vec_coercible_cast_e_data* data = (struct vec_coercible_cast_e_data*) data_;
+  data->out = vec_cast(data->x, data->to, data->x_arg, data->to_arg);
+}
+
+SEXP vec_coercible_cast_e(SEXP x,
```
DavisVaughan: Mark with `// [[ include("vctrs.h") ]]`?

src/cast.c
```diff
@@ -462,6 +477,45 @@ SEXP vec_coercible_cast(SEXP x, SEXP to, struct vctrs_arg* x_arg, struct vctrs_a
...
+    .x = x,
+    .to = to,
+    .x_arg = x_arg,
+    .to_arg = to_arg,
+    .out = R_NilValue
+  };
+
+  *err = r_try_catch(&vec_coercible_cast_e_cb,
+                     &data,
+                     Rf_install("vctrs_error_cast_lossy"),
```
DavisVaughan: Worth it to generate this at init time? `syms_vctrs_error_cast_lossy`

src/subscript.c
```diff
@@ -0,0 +1,254 @@
...
+    *err = new_error_subscript_type(subscript,
+                                    opts,
+                                    dbl_cast_subscript_body,
+                                    err_obj);
+    UNPROTECT(2);
+    return R_NilValue;
+  }
+
+  // FIXME: Handle names in cast methods
+  SEXP nms = PROTECT(r_names(subscript));
```
DavisVaughan: Names are already handled with `orig_names` in `vec_as_subscript_opts()` right? Also, is it possible this could overwrite `subscript` directly if `vec_coercible_cast_e()`'s call to `vec_cast()` doesn't duplicate?

src/subscript.c
```diff
@@ -0,0 +1,254 @@
...
+  SEXP call = PROTECT(r_call(new_error_subscript_type_sym, syms, args));
+  SEXP out = Rf_eval(call, vctrs_ns_env);
+
+  UNPROTECT(3);
+  return out;
+}
+
+
+void vctrs_init_subscript(SEXP ns) {
+  new_error_subscript_type_sym = Rf_install("new_error_subscript_type");
```
DavisVaughan: would we normally call this `syms_new_error_subscript_type`?

src/utils.h
```diff
@@ -259,6 +265,65 @@ ERR r_try_catch(void (*fn)(void*),
...
+  SET_VECTOR_ELT(result, 1, err);
+
+  SET_ATTRIB(result, result_attrib);
+  SET_OBJECT(result, 1);
+
+  UNPROTECT(3);
+  return result;
+}
+
+// Inputs are protected so this can be used as return value
```
DavisVaughan: Does `x` explicitly need to be `PROTECT()`ed? I wasn't sure how this works

lionel-: Not sure I'm happy with the non-standard semantics, but it seemed to make sense since these functions are always tail calls. I thought we'd make an experiment of it and switch to normal protection later if we don't like it.

lionel-:
as-location will call as-subscript in an upcoming branch.

src/subscript.c
lionel-: oops I forgot to check this could be removed. I agree it would have been safer to maybe-duplicate, in case the method starts to return cached objects for opitimisation for instance.

src/utils.h
lionel-: I've changed my mind about this, these "result" operations are not common enough to be worth it. And should never be, since result objects are mostly temporary for prototyping C code in R.

PR #868, opened 2020-02-26
Call `vec_as_subscript()` from `vec_as_location()`

### LINKED ISSUE #575: Ensure that we support x[[sym]]

lionel-:
Merges R-level and C-level code paths.

Part of r-lib/rlang#836.
Closes #575.

### DIFF
--- a/src/subscript-loc.c
+++ b/src/subscript-loc.c
@@ -1,5 +1,6 @@
 #include "vctrs.h"
 #include "utils.h"
+#include "subscript.h"
 #include "subscript-loc.h"
 
 static SEXP int_invert_location(SEXP subscript, R_len_t n,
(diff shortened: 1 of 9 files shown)


src/subscript-loc.c
```diff
@@ -289,16 +290,35 @@ SEXP vec_as_location_opts(SEXP subscript, R_len_t n, SEXP names,
...
+  struct vec_as_subscript_opts subscript_opts = {
+    .logical = SUBSCRIPT_TYPE_ACTION_CAST,
+    .numeric = SUBSCRIPT_TYPE_ACTION_CAST,
+    .character = SUBSCRIPT_TYPE_ACTION_CAST,
+    .subscript_arg = opts->subscript_arg
+  };
+
+  ERR err = NULL;
+  subscript = vec_as_subscript_opts(subscript, &subscript_opts, &err);
+  PROTECT2(subscript, err);
```
DavisVaughan: It would be nice if there was a way for me to pass the `vec_as_subscript_opts()` through to `vec_as_location_opts()` if it is going to call it for me.

Otherwise I'm going to end up calling `vec_as_subscript_opts()` twice if I have to change any options, like this in `vec_chop()` where we limit to only numeric subscripts.

https://github.com/r-lib/vctrs/pull/857/files#diff-04334dad60bfe91381c3ac91164f5429R636-R642

`vec_as_chop_subscript()` has these opts set:
https://github.com/r-lib/vctrs/pull/857/files#diff-d6d1b31a2f6619cc28d5a053134fe50bR103-R105

lionel-: This would be inconsistent with the R interface, but I guess at the C level it makes sense?

lionel-: It now takes the subscript options.

DavisVaughan: Thank you!

PR #871, opened 2020-02-27
Avoid `stopifnot()` in `new_list_of()`

DavisVaughan:
Part of the slowness of `dplyr:::new_rowwise_df()`

### DIFF
--- a/R/type-list-of.R
+++ b/R/type-list-of.R
@@ -59,8 +59,13 @@ as_list_of.list <- function(x, ..., .ptype = NULL) {
 #' @keywords internal
 #' @export
 new_list_of <- function(x = list(), ptype = logical(), ..., class = character()) {
-  stopifnot(is.list(x))
-  stopifnot(vec_size(ptype) == 0)
+  if (!vec_is_list(x)) {
+    abort("`x` must be a list")
+  }
+
+  if (vec_size(ptype) != 0L) {
+    abort("`ptype` must have size 0")
+  }
 
   new_vctr(x, ..., ptype = ptype, class = c(class, "vctrs_list_of"))
 }
@@ -68,10 +73,14 @@ new_list_of <- function(x = list(), ptype = logical(), ..., class = character())
 #' @export
 #' @rdname list_of
 validate_list_of <- function(x) {
-  stopifnot(is.list(x))
+  if (!vec_is_list(x)) {
+    abort("`x` must be a list")
+  }
 
   ptype <- attr(x, "ptype")
-  stopifnot(vec_size(ptype) == 0)
+  if (vec_size(ptype) != 0L) {
+    abort("`ptype` must have size 0")
+  }
 
   walk(x, vec_cast, to = ptype)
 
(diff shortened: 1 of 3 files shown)


R/type-list-of.R
```diff
@@ -59,19 +59,28 @@ as_list_of.list <- function(x, ..., .ptype = NULL) {
 #' @keywords internal
 #' @export
 new_list_of <- function(x = list(), ptype = logical(), ..., class = character()) {
-  stopifnot(is.list(x))
-  stopifnot(vec_size(ptype) == 0)
+  if (!vec_is_list(x)) {
+    abort("`x` must be a list")
```
lionel-:
```suggestion
    abort("`x` must be a list.")
```

R/type-list-of.R
```diff
@@ -59,19 +59,28 @@ as_list_of.list <- function(x, ..., .ptype = NULL) {
...
+  }
+
+  if (vec_size(ptype) != 0L) {
+    abort("`ptype` must have size 0")
```
lionel-:
```suggestion
    abort("`ptype` must have size 0.")
```

R/type-list-of.R
```diff
@@ -59,19 +59,28 @@ as_list_of.list <- function(x, ..., .ptype = NULL) {
...
 
   new_vctr(x, ..., ptype = ptype, class = c(class, "vctrs_list_of"))
 }
 
 #' @export
 #' @rdname list_of
 validate_list_of <- function(x) {
-  stopifnot(is.list(x))
+  if (!vec_is_list(x)) {
+    abort("`x` must be a list")
```
lionel-:
```suggestion
    abort("`x` must be a list.")
```

R/type-list-of.R
```diff
@@ -59,19 +59,28 @@ as_list_of.list <- function(x, ..., .ptype = NULL) {
...
+  }
 
   ptype <- attr(x, "ptype")
-  stopifnot(vec_size(ptype) == 0)
+  if (vec_size(ptype) != 0L) {
+    abort("`ptype` must have size 0")
```
lionel-:
```suggestion
    abort("`ptype` must have size 0.")
```

src/type-list-of.c
```diff
@@ -3,7 +3,7 @@
 
 // [[ include("utils.h") ]]
 SEXP new_list_of(SEXP x, SEXP ptype) {
-  if (TYPEOF(x) != VECSXP) {
+  if (!vec_is_list(x)) {
```
lionel-: This explicit support for S3 lists makes me wonder if we should zap the attributes before initialising the list-of.

lionel-: We probably should.

DavisVaughan: The R level one does, so yea probably. I'll update. 

```
attributes(vctrs::new_list_of(structure(list(), x = 1)))
#> $ptype
#> logical(0)
#> 
#> $class
#> [1] "vctrs_list_of" "vctrs_vctr"
```

PR #874, opened 2020-02-28
Use `exp_` naming for extremely experimental callables

### LINKED ISSUE #843: Refactor exported callables on standardized `temp_` naming
This also changes them to not expose any `arg` arguments, as that API is not stable

```c
// from
vctrs_cast(SEXP x, SEXP to, SEXP x_arg, SEXP to_arg);

// to
tmp_vec_cast(SEXP x, SEXP to);
```

```c
// from
short_vec_size(SEXP x);

// to
temp_short_vec_size(SEXP x);
```

```c
// from
short_vec_recycle(SEXP x, R_len_t size, struct vctrs_arg*);

// to
temp_short_vec_recycle(SEXP x, R_len_t size);
```

```c
// from
short_vec_init(SEXP x, R_len_t size);

// to 
temp_short_vec_init(SEXP x, R_len_t size);
```

DavisVaughan:
Closes #843 

This PR attempts to add some structure to how we export extremely experimental callables. Extremely experimental ones now live in `experimental.c`, and have their own function prefixed with `exp_`. Anything that uses or returns a `R_len_t` lives here, and is also prefixed with `short`.

`vec_cast()` and `vec_recycle()` are exposed without their `struct* vctrs_arg` arguments.

@lionel- I ended up going with `exp_` rather than `tmp_` because I think it reads more naturally and works well with the `experimental.c` file name. I don't have a very strong preference though, and can change it if you don't like `exp_`

CC @romainfrancois dplyr will need some tweaking once this is merged


DavisVaughan:
@lionel- I have updated this to only include the 3 functions used by dplyr in the public header file. All of the rest are now callables. `vec_restore()` is now exposed without the `n` argument

PR #878, opened 2020-03-04
S3 class in a package

jennybc:
Written after adding a vctrs S3 class in googlesheets4 and realizing we could use some more official guidance on how to fit these classes into an R package, use with roxygen2, etc.

### DIFF
--- a/vignettes/s3-vector.Rmd
+++ b/vignettes/s3-vector.Rmd
@@ -931,7 +931,7 @@ vec_arith.vctrs_meter.MISSING <- function(op, x, y, ...) {
 +meter(1) 
 ```
 
-## Appendix: Implementing a vctrs S3 class in a package
+## Implementing a vctrs S3 class in a package
 
 Defining S3 methods interactively is fine for iteration and exploration, but if your class lives in a package, you need to do a few more things:
 
--- a/vignettes/s3-vector.Rmd
+++ b/vignettes/s3-vector.Rmd
@@ -92,7 +92,7 @@ new_percent <- function(x = double()) {
   new_vctr(x, class = "vctrs_percent")
 }
 
-x <- new_percent(c(seq(0, 1, length = 4), NA))
+x <- new_percent(c(seq(0, 1, length.out = 4), NA))
 x
 
 str(x)
--- a/vignettes/s3-vector.Rmd
+++ b/vignettes/s3-vector.Rmd
@@ -944,9 +944,11 @@ Let's assume that the `percent` class is implemented in the pizza package in the
 
 ### Getting started
 
-We begin by adding `Import: vctrs` to our `DESCRIPTION` (with `usethis::use_package("vctrs")`) and then make all vctrs functions available to us by adding `#' @import vctrs` to `R/pizza-doc.R` (created with `usethis::use_package_doc()`). Usually, it's not good practice to `@import` a specific package but vctrs is deliberately designed with this use case in mind.
+First, the pizza package needs to include vctrs in the `Imports` section of its `DESCRIPTION` (perhaps by calling`usethis::use_package("vctrs")`. Then we make all vctrs functions available within the pizza package by including the directive `#' @import vctrs` somewhere. You can use `usethis::use_package_doc()` to create package-level documentation in `R/pizza-doc.R`, in which case that is also the logical place to put `#' @import vctrs`. Usually, it's not good practice to `@import` the entire namespace of a package, but vctrs is deliberately designed with this use case in mind.
 
-Then we make a place to dump some internal documentation that's needed to avoid R CMD check complaints. We don't expect any human will ever read this documentation.
+Remember to call `devtools::document()` regularly, as you develop, to regenerate `NAMESPACE` and the `.Rd` files.
+
+Next we make a place to dump some internal documentation that's needed to avoid `R CMD check` complaints. We don't expect any human to ever read this documentation. This, and all subsequent code, presumably lives in `R/percent.R`:
 
 ```{r}
 #' Internal vctrs methods
@@ -958,7 +960,7 @@ NULL
 
 ### Low-level and user-friendly constructors
 
-Next we add our constructor 
+Next we add our constructor:
 
 ```{r}
 new_percent <- function(x = double()) {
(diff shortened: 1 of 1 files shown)


vignettes/s3-vector.Rmd
```diff
@@ -940,27 +931,223 @@ vec_arith.vctrs_meter.MISSING <- function(op, x, y, ...) {
...
-vec_cast.vctrs_percent <- function(x, to, ...) {
-} 
+vec_ptype_abbr.pizza_percent <- function(x, ...) {
+  "prcnt"
+}
+```
+
+### Double dispatch methods for coercion and casting
+
+Next we implement methods for `vec_ptype2()` and `vec_cast()`, along with all the roxygen2 boilerplate needed to make them work. `vec_ptype2()` and `vec_cast()` have exactly the same structure, so we'll explain everything via `vec_ptype2()` and just show the result for `vec_cast()`.
```
lionel-: > have exactly the same structure

Unfortunately this isn't true. ptype2 dispatches first on the first argument, whereas cast dispatches first on the second argument.

vignettes/s3-vector.Rmd
```diff
@@ -940,27 +931,223 @@ vec_arith.vctrs_meter.MISSING <- function(op, x, y, ...) {
...
+#' @export                                       
+vec_ptype2.vctrs_percent.double <- function(x, y, ...) double()
+
+#' @method vec_ptype2.double vctrs_percent
 #' @export
-vec_cast.binned.double <- function(x, y, ...) {
+vec_ptype2.double.vctrs_percent <- function(x, y, ...) double()
+```
+
+The exact same treatment applies to `vec_cast()`, which yields:
```
lionel-: I have fallen the trap many times, so I think describing the treatment as exactly the same might cause wrong copy/pasting.

vignettes/s3-vector.Rmd
```diff
@@ -940,27 +931,223 @@ vec_arith.vctrs_meter.MISSING <- function(op, x, y, ...) {
...
+#' @method vec_cast.pizza_percent double
+#' @export
+vec_cast.pizza_percent.double <- function(x, to, ...) percent(x)
+#' @method vec_cast.double pizza_percent
+#' @export
+vec_cast.double.pizza_percent <- function(x, to, ...) vec_data(x)
 ```
 
-Hopefully future versions of roxygen will make these exports less painful.
+For simple casting and coercion functions, this is the recommended formatting. If everything fits on one line, there's no need to use `{}` or add blank lines between method definitions. This compact formatting reinforces the common purpose of these methods and their documentation.
```
lionel-: I recommend the multi-line formatting. Otherwise the bodies don't align since the methods have names of varying width. This makes it harder to scan definitions and easier to get things wrong. Since it's hard to get good test coverage of behaviour for methods, eye-scanning and eye-checking are important.

The single-line formatting also makes it harder to change the code once a method becomes more complicated. And then you get inconsistent formatting in the file since once method has a multi-line `{ }` body and the others haven't.

vignettes/s3-vector.Rmd
```diff
@@ -940,27 +931,223 @@ vec_arith.vctrs_meter.MISSING <- function(op, x, y, ...) {
...
+
+* Make sure to test behaviour with zero-length inputs and missing values.
+
+* Use `testthat::verify_output()` to add some 
+  [golden tests](https://ro-che.info/articles/2017-12-04-golden-tests) for 
+  your format method. Customised printing is often a primary motivation for
+  creating your own S3 class in the first place, so it makes sense to actively
+  monitor this.
+
+* Check for method symmetry; use `expect_s3_class()` to
```
lionel-: `exact = TRUE`?

vignettes/s3-vector.Rmd
```diff
@@ -940,27 +931,223 @@ vec_arith.vctrs_meter.MISSING <- function(op, x, y, ...) {
...
+
+It's good practice to test your new class. Specific recommendations:
+
+* `R/percent.R` is the type of file where you really do want 100% test coverage.
+  You can use `devtools::test_file_coverage()` to check this.
+
+* Make sure to test behaviour with zero-length inputs and missing values.
+
+* Use `testthat::verify_output()` to add some 
+  [golden tests](https://ro-che.info/articles/2017-12-04-golden-tests) for
```
lionel-: I think https://www.tidyverse.org/blog/2019/11/testthat-2-3-0/ might be a better resource. The golden test blog post doesn't mention the most important reason for using `verify_output()`, which is to not depend on the printing/deparsing/messaging behaviour of other packages.

vignettes/s3-vector.Rmd
jennybc: This is a @Hadley opinion.

vignettes/s3-vector.Rmd
jennybc: OK I added this to the current PR refining this contribution.

vignettes/s3-vector.Rmd
jennybc: OK I added this to the current PR refining this contribution.

vignettes/s3-vector.Rmd
jennybc: Do you think I have the roxygen2 directives wrong? That is all I aimed to address here with the "same treatment" advice.

lionel-: Perhaps introduce with something like this?

> Whereas `vec_ptype2()` and `vec_cast()` methods have slightly different implementations because they dispatch in different order, they use the same pattern of roxygen2 tags.

vignettes/s3-vector.Rmd
jennybc: I weakly agree with @lionel-. Pragmatically, if one uses styler, for example, it's not so easy to have these one-off exceptions for when we prefer a different style.

vignettes/s3-vector.Rmd
jennybc: OK yes I have made the wording more precise now 5fdc470614aefb565ee24bef9a09398085cc889c

vignettes/s3-vector.Rmd
hadley: I don't care enough about this to fight y'all on it. Feel free to change.

jennybc: @lionel- I leave this in your hands. Obviously I'm OK with the advice, because I had a chance to push back. But I see your point.

Feel free to format in your preferred way. And let's remove any specific formatting advice.

lionel-: ok let's keep the formatting as is and just remove the recommendation, since we don't have a consensus on preferred formatting.

jennybc: I removed the recommendation, so I think we've resolved all the discussion points here. Agree that, at some point, splitting this vignette into topics will make sense. But it's still good have a start on this.

PR #879, opened 2020-03-04
Add `vec_proxy()` and `vec_restore()` methods for factors

### LINKED ISSUE #705: Handle factors in C
Once implemented, should make sure to rerun benchmarking code in https://github.com/tidyverse/dplyr/issues/4561

DavisVaughan:
Closes #705 

This PR implements `vec_proxy()` and `vec_restore()` methods for factors and ordered factors.

- `vec_proxy()` methods are trivial, and just exist to make `vec_slice()` not fallback to `[.factor`.

- `vec_restore()` methods do a few checks before handing off to `vec_restore_default()`

   - The main check is to ensure that the number of levels implied by the integer vector `x` aligns with the number of levels in the factor `to`. @lionel- had the idea to just check the maximum value of `x` against the length of the levels of `to`. `NA` values get ignored for this check.

   - I've intercepted the C level `vec_restore()` to hand off to `vec_restore_bare_factor()` directly to avoid R level dispatch.

I think if we wanted to we could also intercept `vec_proxy_method()` in some way for these known cases so we don't have to look to R to find the `vec_proxy.factor()` method.

Performance is a _little_ better for slicing, but I wasn't expecting much here.

``` r
library(vctrs)
set.seed(123)

x <- factor("x")

bench::mark(
  vec_slice(x, 1L),
  x[1L],
  iterations = 100000
)

# Master
#> # A tibble: 2 x 6
#> 1 vec_slice(x, 1L)   7.27µs   9.88µs    98447.    17.1KB     92.6
#> 2 x[1L]              3.02µs   3.51µs   258357.        0B     31.0

# This PR
#> # A tibble: 2 x 6
#> 1 vec_slice(x, 1L)   6.73µs   8.09µs   119642.    9.99KB    101. 
#> 2 x[1L]              3.15µs   4.05µs   228381.   10.66KB     27.4
```

```r
y <- factor(sample(letters, 1e5, replace = TRUE))
idx <- sample(1e4, replace = TRUE)

bench::mark(
  vec_slice(y, idx),
  y[idx],
  iterations = 10000
)

# Master
#> # A tibble: 2 x 6
#> 1 vec_slice(y, idx)     51µs   65.8µs    14893.    78.2KB     22.4
#> 2 y[idx]              32.4µs   36.2µs    25181.    39.1KB     20.2

# This PR
#> # A tibble: 2 x 6
#> 1 vec_slice(y, idx)   34.2µs   47.4µs    19200.    78.2KB     28.8
#> 2 y[idx]              32.9µs   40.6µs    24297.    39.1KB     19.5
```


src/type-factor.c
```diff
@@ -1,6 +1,88 @@
 #include "vctrs.h"
 #include "utils.h"
 
+static SEXP vec_restore_bare_factor_impl(SEXP x, SEXP to, SEXP to_levels) {
+  const int to_max = vec_size(to_levels);
```
lionel-:
```suggestion
  const R_len_t to_max = vec_size(to_levels);
```

Just for conformity of types if we swich vec-size to long sizes. This way we'll support the `factor(1:long_size)` case.

hadley:
>  So for factors we might not need to check invariants in the restore method because they are enforced through other means. Am I missing anything?

This matches my intuition too

lionel-:
> This matches my intuition too

And Davis' too, he tried to explain this to me yesterday but I missed it 🙈

DavisVaughan:
@lionel- this PR is ready to merge whenever you are ready. I'll leave it here for you to review again if you want. Feel free to take over if you are trying to get the release out and need to change something.

- I've removed the factor restore methods. I agree they are not needed.

- The fast path in C `vec_restore()` now just calls `vec_restore_default()`

- I like having `vec_restore.factor()` and `vec_restore.ordered()` implemented with `NextMethod()` just as reminders to future us that we have thought through the restore method for factors already. If you don't like that feel free to remove

lionel-:
> I like having vec_restore.factor() and vec_restore.ordered() implemented with NextMethod() just as reminders to future us that we have thought through the restore method for factors already. If you don't like that feel free to remove

I think this is correct. It would likely be better if all proxy and restore methods called their parents (in inverse order for restore methods).

PR #880, opened 2020-03-05
Give more specific advice for package developers

jennybc:
More thoughts about what goes where and why

### DIFF
--- a/vignettes/s3-vector.Rmd
+++ b/vignettes/s3-vector.Rmd
@@ -944,19 +944,43 @@ Let's assume that the `percent` class is implemented in the pizza package in the
 
 ### Getting started
 
-First, the pizza package needs to include vctrs in the `Imports` section of its `DESCRIPTION` (perhaps by calling`usethis::use_package("vctrs")`. Then we make all vctrs functions available within the pizza package by including the directive `#' @import vctrs` somewhere. You can use `usethis::use_package_doc()` to create package-level documentation in `R/pizza-doc.R`, in which case that is also the logical place to put `#' @import vctrs`. Usually, it's not good practice to `@import` the entire namespace of a package, but vctrs is deliberately designed with this use case in mind.
+First, the pizza package needs to include vctrs in the `Imports` section of its `DESCRIPTION` (perhaps by calling`usethis::use_package("vctrs")`. While vctrs is under very active development, it probably makes sense to state a minimum version.
 
-Remember to call `devtools::document()` regularly, as you develop, to regenerate `NAMESPACE` and the `.Rd` files.
+```
+Imports: 
+    a_package,
+    another_package,
+    ...
+    vctrs (>= x.y.z),
+    ...
+```
 
-Next we make a place to dump some internal documentation that's needed to avoid `R CMD check` complaints. We don't expect any human to ever read this documentation. This, and all subsequent code, presumably lives in `R/percent.R`:
+Then we make all vctrs functions available within the pizza package by including the directive `#' @import vctrs` somewhere. Usually, it's not good practice to `@import` the entire namespace of a package, but vctrs is deliberately designed with this use case in mind.
+
+Where should we put `#' @import vctrs`? There are two natural locations:
+
+* With package-level docs in `R/pizza-doc.R`. You can use
+  `usethis::use_package_doc()` to initiate this package-level documentation.
+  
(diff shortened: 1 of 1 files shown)


vignettes/s3-vector.Rmd
```diff
@@ -944,20 +944,44 @@ Let's assume that the `percent` class is implemented in the pizza package in the
...
 
-Remember to call `devtools::document()` regularly, as you develop, to regenerate `NAMESPACE` and the `.Rd` files.
+```
+Imports: 
+    a_package,
+    another_package,
+    ...
+    vctrs (>= x.y.z),
+    ...
+```
```
jennybc: I feel like it's important to feature obvious, copy-paste-able snippet for important things. Versus just embedding in prose.

vignettes/s3-vector.Rmd
```diff
@@ -944,20 +944,44 @@ Let's assume that the `percent` class is implemented in the pizza package in the
...
 ### Getting started
 
-First, the pizza package needs to include vctrs in the `Imports` section of its `DESCRIPTION` (perhaps by calling`usethis::use_package("vctrs")`. Then we make all vctrs functions available within the pizza package by including the directive `#' @import vctrs` somewhere. You can use `usethis::use_package_doc()` to create package-level documentation in `R/pizza-doc.R`, in which case that is also the logical place to put `#' @import vctrs`. Usually, it's not good practice to `@import` the entire namespace of a package, but vctrs is deliberately designed with this use case in mind.
+First, the pizza package needs to include vctrs in the `Imports` section of its `DESCRIPTION` (perhaps by calling`usethis::use_package("vctrs")`. While vctrs is under very active development, it probably makes sense to state a minimum version.
 
-Remember to call `devtools::document()` regularly, as you develop, to regenerate `NAMESPACE` and the `.Rd` files.
+```
+Imports: 
+    a_package,
+    another_package,
```
lionel-:
```suggestion
    onePackage,
    anotherPackage,
```

Looks more real this way :)

vignettes/s3-vector.Rmd
```diff
@@ -944,20 +944,44 @@ Let's assume that the `percent` class is implemented in the pizza package in the
...
+    ...
+    vctrs (>= x.y.z),
+    ...
+```
 
-Next we make a place to dump some internal documentation that's needed to avoid `R CMD check` complaints. We don't expect any human to ever read this documentation. This, and all subsequent code, presumably lives in `R/percent.R`:
+Then we make all vctrs functions available within the pizza package by including the directive `#' @import vctrs` somewhere. Usually, it's not good practice to `@import` the entire namespace of a package, but vctrs is deliberately designed with this use case in mind.
```
lionel-: I'm not sure about this advice. I think it makes sense for packages that heavily use vctrs to manipulate vectors (like tidyr), but not for packages that just implement a vctrs class (like pizza).

DavisVaughan: A somewhat awkward thing you have to do if you are implementing a vctrs class and you don’t import all of vctrs is to individually import `vec_ptype2.<foo>` methods like `vec_ptype2.factor` here https://github.com/tidymodels/probably/blob/389d3a576237d493d98c8d50f3d92e68893ae1b6/R/vctrs-compat.R#L256

lionel-: You'll also need to do that if your class is coercible with another foreign class.

vignettes/s3-vector.Rmd
hadley: But this isn't copy and pastable? I was inserting the usethis calls since you can actually run them on your package.

vignettes/s3-vector.Rmd
hadley: Resolved via local discussion

vignettes/s3-vector.Rmd
jennybc: But I also believe in keeping packages in alphabetical order!

vignettes/s3-vector.Rmd
jennybc: OK true not copy/pastable. But realistic re: how it should look and pulled into an obvious code chunk. I do still mention the usethis commands. I think when I read docs I often skip over the prose and just read the code (and I sometimes I suffer for that, but I can't be alone).

PR #884, opened 2020-03-06
Fix `vec_in()` to be consistent with `vec_match()`

DavisVaughan:


lionel-:
@hadley you implemented `na = "different"` instead of `na = "propagate"`. Now fixed in https://github.com/r-lib/vctrs/pull/884/commits/3d56f75fdc6172b5e92b8ed1a2ea2c341ccbac29.

src/dictionary.c
```diff
@@ -600,7 +601,7 @@ SEXP vctrs_in(SEXP needles, SEXP haystack, SEXP na_equal) {
   needles = VECTOR_ELT(translated, 0);
   haystack = VECTOR_ELT(translated, 1);
 
-  struct dictionary* d = new_dictionary_params(haystack, false, r_bool_as_int(na_equal));
+  struct dictionary* d = new_dictionary_params(haystack, false, na_equal);
```
krlmlr: Is this correct? I have:

```
static struct dictionary* new_dictionary_params(SEXP x, bool partial, bool na_equal) {
```

lionel-: yes see line 586.

hadley:
This needs _another_ version bump so I can depend on it in dplyr 😭

lionel-:
Done, we are now at 0.2.99.9010.

PR #886, opened 2020-03-06
Add backticks around types

krlmlr:
in error messages, for consistency.

### DIFF
--- a/R/subscript.R
+++ b/R/subscript.R
@@ -215,7 +215,7 @@ collapse_subscript_type <- function(cnd) {
     last <- ", or "
   }
 
-  glue::glue_collapse(types, sep = ", ", last = last)
+  glue::glue_collapse(tick(types), sep = ", ", last = last)
 }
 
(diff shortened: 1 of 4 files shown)


lionel-:
I don't think this is correct?

krlmlr:
``The subscript has the wrong type `environment`. It must be `logical`, `numeric`, or `character`.``

The new backticks are consistent with the backticks after "wrong type". Should we use `<>` instead?

lionel-:
In the first sentence this is about types, in the second this is about concepts. For instance a `<factor>` and a `<character>` are character types.

krlmlr:
Agree to omit the backticks around the concepts.

We already have this:

```
> vec_assert(lgl(), chr()):

Error: `lgl()` must be a vector with type <character>.
Instead, it has type <logical>.
```

This makes me think we should throw `The subscript has the wrong type <environment>.` Happy to do this in a new PR.

PR #889, opened 2020-03-06
Fix attribute handling in new_data_frame()

### LINKED ISSUE #883: `new_data_frame()` attribute issues
strips names if you provide extra attributes

``` r
library(vctrs)

new_data_frame(list(x = 1), n = 1L, stuff = "foo")
#>    
#> 1 1
```

won't let you add row names

``` r
library(vctrs)

row.names(new_data_frame(list(x = 1), n = 1L, row.names = "y"))
#> [1] "1"
```

I think the ordering of how attributes are added is a little off. The `SET_ATTRIB()` call will nuke all existing attributes, including names. So we could special case names by holding onto them and restoring them after calling that.

For manually added row names, currently they are added by the `SET_ATTRIB()` call, but then nuked by the `new_data_frame()` call.

krlmlr:
Needs tests, two existing tests fail.

Closes #883.


src/type-data-frame.c
```diff
@@ -59,21 +59,48 @@ SEXP vctrs_new_data_frame(SEXP args) {
...
+  R_len_t size = -1;
+
+  SEXP out = PROTECT(r_maybe_duplicate(x));
+
+  for (SEXP attr_in = attrib; attr_in != R_NilValue; attr_in = CDR(attr_in)) {
+    SEXP tag = TAG(attr_in);
+
+    if (tag == R_NamesSymbol) {
+      // names are ignored if the input is named
+      if (has_names) continue;
```
lionel-: Control flow statements must be on their own line.

src/type-data-frame.c
```diff
@@ -59,21 +59,48 @@ SEXP vctrs_new_data_frame(SEXP args) {
     Rf_errorcall(R_NilValue, "`x` must be a list");
   }
 
-  R_len_t size = df_size_from_list(x, n);
+  bool has_n = !Rf_isNull(n);
```
lionel-: We use `n == R_NilValue`

src/type-data-frame.c
```diff
@@ -59,21 +59,48 @@ SEXP vctrs_new_data_frame(SEXP args) {
...
+
+  SEXP out = PROTECT(r_maybe_duplicate(x));
+
+  for (SEXP attr_in = attrib; attr_in != R_NilValue; attr_in = CDR(attr_in)) {
+    SEXP tag = TAG(attr_in);
+
+    if (tag == R_NamesSymbol) {
+      // names are ignored if the input is named
+      if (has_names) continue;
+    } else if (tag == R_RowNamesSymbol) {
```
lionel-: _Style._ Shouldn't be an `else` branch here, since that suggests the previous branch hasn't `continue`'d.

src/type-data-frame.c
```diff
@@ -59,21 +59,48 @@ SEXP vctrs_new_data_frame(SEXP args) {
...
+      // row.names is ignored if n is provided
+      if (has_n) continue;
+      size = rownames_size(CAR(attr_in));
+    } else if (tag == R_ClassSymbol) {
+      // class in ... overrides class from argument, to support splicing
```
lionel-: This should be an error. If an attribute is a parameter it shouldn't be supplied through dots.

src/type-data-frame.c
```diff
@@ -59,21 +59,48 @@ SEXP vctrs_new_data_frame(SEXP args) {
...
+
+  for (SEXP attr_in = attrib; attr_in != R_NilValue; attr_in = CDR(attr_in)) {
+    SEXP tag = TAG(attr_in);
+
+    if (tag == R_NamesSymbol) {
+      // names are ignored if the input is named
+      if (has_names) continue;
+    } else if (tag == R_RowNamesSymbol) {
+      // row.names is ignored if n is provided
+      if (has_n) continue;
```
lionel-: It looks like this is not handling character rownames correctly. They should be assigned in the attributes list.

src/type-data-frame.c
krlmlr: Disagree. The previous branch doesn't always continue. The `else`-s make it clear that only one of the alternatives is valid.

src/type-data-frame.c
krlmlr: Except where we don't ;-)

https://github.com/r-lib/vctrs/blob/9ac7d408266e7cfcd24d3ab995497856724adcfb/src/utils.c#L204

Will send a separate PR.

src/type-data-frame.c
lionel-: You're right.

src/type-data-frame.c
lionel-: No need for a PR.

src/type-data-frame.c
krlmlr: The tests suggest it works.

lionel-: I don't understand how it works.

krlmlr: We copy over the `"row.names"` attribute only if `n` is not provided. If the caller passes `n`, it takes precedence over `row.names` in the dots.

krlmlr:
I wonder if we should unconditionally override names if the caller passes a `"names"` argument. This is how `new_tibble()` works, it's also simpler to communicate.

574c222, not yet part of this branch.

src/type-data-frame.c
lionel-: This is not correct. We should assign row names if they are character.

lionel-:
> I wonder if we should unconditionally override names if the caller passes a "names" argument. 

This sounds like a reasonable behaviour.

src/type-data-frame.c
krlmlr: What if `n` and `"row.names"` disagree? Why is this important?

It's easy to *not* pass `n`. It's more difficult, and slower, to avoid passing `"row.names"`.

lionel-: It is important to have a good and predictable UI.

If `n` and `row.names` disagree, please make it an error. This is a fast check.

src/type-data-frame.c
```diff
@@ -133,11 +167,23 @@ enum rownames_type rownames_type(SEXP x) {
...
+R_len_t rownames_size(SEXP rn) {
+  switch (rownames_type(rn)) {
+  case ROWNAMES_IDENTIFIERS:
+  case ROWNAMES_AUTOMATIC:
+    return Rf_length(rn);
+  case ROWNAMES_AUTOMATIC_COMPACT:
+    return compact_rownames_length(rn);
+  }
+
+  return -1;
```
lionel-:
```suggestion
  never_reached("rownames_size")
```

src/type-data-frame.c
```diff
@@ -59,29 +60,62 @@ SEXP vctrs_new_data_frame(SEXP args) {
...
+    if (tag == R_RowNamesSymbol) {
+      // "row.names" is checked for consistency with n,
+      // used only if character
+      if (n != R_NilValue) {
+        R_len_t n_size = df_size_from_n(n);
+        SEXP rn = CAR(attr_in);
+        R_len_t rn_size = rownames_size(rn);
+
+        if (n_size != rn_size) {
+          Rf_errorcall(R_NilValue, "`n` and `row.names` must be consistent");
```
lionel-:
```suggestion
          Rf_errorcall(R_NilValue, "`n` and `row.names` must be consistent.");
```

R/type-data-frame.R
```diff
@@ -11,6 +11,13 @@
 #' @param n Number of rows. If `NULL`, will be computed from the length of
 #'   the first element of `x`.
 #' @param ...,class Additional arguments for creating subclasses.
+#'   The `"names"` and `"row.names"` attributes override input in `x` and `n`,
+#'   respectively:
+#'
+#'   - `"names"` is used if provided, overriding existing names in `x`
+#'   - `"row.names"` is used if provided, if `n` is provided it must be
+#'     consistent
```
lionel-:
```suggestion
#'     consistent.
```

src/type-data-frame.c
```diff
@@ -59,29 +60,63 @@ SEXP vctrs_new_data_frame(SEXP args) {
...
+
+  SEXP out = PROTECT(r_maybe_duplicate(x));
+
   R_len_t size = df_size_from_list(x, n);
 
-  if (attrib != R_NilValue) {
-    x = r_maybe_duplicate(x);
-    SET_ATTRIB(x, attrib);
+  for (SEXP attr_in = attrib; attr_in != R_NilValue; attr_in = CDR(attr_in)) {
+    SEXP tag = TAG(attr_in);
```
lionel-: Should still check for `tag == R_ClassSymbol` and throw an error if supplied.

(Or at least `continue` to skip the `setAttrib()`, because right now the class in dots is ignored if `cls` is `NULL`, but not otherwise. But I think it's probably better to make it an error as mentioned earlier.)

krlmlr: How can a `"class"` sneak into the dots when `class` is part of the formals?

lionel-: oops sorry, I thought we had dynamic dots

krlmlr:
I have local branches for tibble and dplyr that build on this PR and test correctly. Will submit tomorrow.

lionel-:
@krlmlr Can this be merged now without breaking dplyr and tibble?

krlmlr:
I think so. dplyr currently has failing tests, about to submit a PR.

krlmlr:
Hang on.

krlmlr:
Not quite. This PR keeps all attributes of `x`, I think the original behavior was to strip attributes. Do we need to maintain original behavior?

lionel-:
oh yeah, attributes of `x` should be zapped before assigning the new ones.

krlmlr:
They weren't always before:

``` r
library(vctrs)

x <- structure(list(a = 1), foo = 2)

attributes(new_data_frame(x, n = 1L))
#> $names
#> [1] "a"
#> 
#> $foo
#> [1] 2
#> 
#> $class
#> [1] "data.frame"
#> 
#> $row.names
#> [1] 1
attributes(new_data_frame(x, n = 1L, bar = 3))
#> $bar
#> [1] 3
#> 
#> $class
#> [1] "data.frame"
#> 
#> $row.names
#> [1] 1
```

<sup>Created on 2020-03-06 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

Current behavior:

``` r
library(vctrs)

x <- structure(list(a = 1), foo = 2)

attributes(new_data_frame(x, n = 1L))
#> $names
#> [1] "a"
#> 
#> $foo
#> [1] 2
#> 
#> $row.names
#> [1] 1
#> 
#> $class
#> [1] "data.frame"
attributes(new_data_frame(x, n = 1L, bar = 3))
#> $names
#> [1] "a"
#> 
#> $foo
#> [1] 2
#> 
#> $bar
#> [1] 3
#> 
#> $row.names
#> [1] 1
#> 
#> $class
#> [1] "data.frame"
```

<sup>Created on 2020-03-06 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

Should the  "foo"` attribute really be stripped?

lionel-:
I think so.

krlmlr:
Tried that in 435872d1 (not part of this branch yet), this *also* breaks dev dplyr.

For a patch release we should replicate the original problematic behavior, and then revisit for a minor release.

DavisVaughan:
If the dev dplyr breakage has anything to do with the dplyr usage of `new_data_frame()` (like here https://github.com/tidyverse/dplyr/blob/53f5eebb98af3ee4557a48231c39d4ed25d1ac46/R/generics.R#L173) then we can fix that.

I agree that attributes of `x` should be zapped

lionel-:
I'm on it, along with a couple of other fixes. The behaviour was introduced in the last patch release by mistake. Thankfully we now have a much stronger set of unit tests thanks to this PR.

lionel-:
Thanks!

PR #893, opened 2020-03-07
Move protection code past Rf_error() calls

krlmlr:
Does `Rf_error()` clean up the protection stack?


lionel-:
> Does Rf_error() clean up the protection stack?

Yes. Is this patch still needed?

krlmlr:
Thanks. No.

PR #894, opened 2020-03-08
Fix corner case in new_data_frame(n = NULL)

krlmlr:
We should allow omitting `n` if `row.names` attribute is given.

Current behavior:

``` r
rlang::exec(
  vctrs::new_data_frame,
  list(),
  n = NULL, !!!attributes(data.frame(a = 1)[0])
)
#> Error: `n` and `row.names` must be consistent.
```

<sup>Created on 2020-03-08 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

With this patch:

``` r
rlang::exec(
  vctrs::new_data_frame,
  list(),
  n = NULL, !!!attributes(data.frame(a = 1)[0])
)
#> data frame with 0 columns and 1 row
```

<sup>Created on 2020-03-08 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

### DIFF
--- a/src/type-data-frame.c
+++ b/src/type-data-frame.c
@@ -86,10 +86,8 @@ SEXP vctrs_new_data_frame(SEXP args) {
 
     if (tag == R_RowNamesSymbol) {
       // "row.names" is checked for consistency with n (if provided)
-      if (size != rownames_size(CAR(node))) {
-        if (n != R_NilValue) {
-          Rf_errorcall(R_NilValue, "`n` and `row.names` must be consistent.");
-        }
+      if (size != rownames_size(CAR(node)) && n != R_NilValue) {
+        Rf_errorcall(R_NilValue, "`n` and `row.names` must be consistent.");
       }
 
       has_rownames = true;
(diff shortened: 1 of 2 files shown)


src/type-data-frame.c
```diff
@@ -87,7 +87,9 @@ SEXP vctrs_new_data_frame(SEXP args) {
     if (tag == R_RowNamesSymbol) {
       // "row.names" is checked for consistency with n (if provided)
       if (size != rownames_size(CAR(node))) {
-        Rf_errorcall(R_NilValue, "`n` and `row.names` must be consistent.");
+        if (n != R_NilValue) {
```
lionel-: Can you merge this if with the nesting if please.

tests/testthat/test-type-data-frame.R
```diff
@@ -281,9 +281,10 @@ test_that("attributes with special names are merged", {
     "rowname"
   )
 
-  expect_error(
+  # Can omit n if row.names attribute is given
```
lionel-: `n` is not omitted in this test. Should probably be its own `test_that()` block.

lionel-:
Thanks!

PR #897, opened 2020-03-09
Improve `vec_equal_na()` performance with data frames

DavisVaughan:
This PR updates `vec_equal_na()` to use the faster traversal method with data frames (looping over the columns one at a time, and early exiting if we know all the results).

It also adds support for `vec_equal_na()` with raw vectors. This always returns `FALSE`, but it doesn't error now.

This worst case scenario below only happens when all elements are `NA`. Even that is much faster than the current master implementation.

```r
library(vctrs)

# worst case
x <- data.frame(x = rep(NA, 1e5), y = rep(NA, 1e5), z = rep(NA, 1e5))

bench::mark(vec_equal_na(x))

# master
#> # A tibble: 1 x 6
#> 1 vec_equal_na(x)   2.57ms   2.75ms      349.     395KB     2.14

# this PR
#> # A tibble: 1 x 6
#> 1 vec_equal_na(x)    300µs    358µs     2596.     492KB     22.7
```

```r
# short circuit after the first column
y <- data.frame(x = rep(1, 1e5), y = rep(2, 1e5), z = rep(3, 1e5))

bench::mark(vec_equal_na(y))

# master
#> # A tibble: 1 x 6
#> 1 vec_equal_na(y)    894µs    953µs     1017.     391KB     6.24

# this PR
#> # A tibble: 1 x 6
#> 1 vec_equal_na(y)    117µs    150µs     6091.     488KB     54.7
```

### DIFF
--- a/src/compare.c
+++ b/src/compare.c
@@ -234,7 +234,7 @@ static struct vctrs_df_rowwise_info df_compare_impl(SEXP x,
                                                     struct vctrs_df_rowwise_info info,
                                                     R_len_t n_row);
 
-static struct vctrs_df_rowwise_info init_rowwise_compare_info(R_len_t n_row) {
+static struct vctrs_df_rowwise_info new_rowwise_compare_info(R_len_t n_row) {
   struct vctrs_df_rowwise_info info;
 
   // Initialize to "equality" value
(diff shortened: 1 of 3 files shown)


DavisVaughan:
@lionel- do you want more information than this?

I sort of feel like this should actually be called `vctrs_df_colwise_info` now, since we iterate over columns, but I'm not sure

Should this move to `colwise.h`?

https://github.com/r-lib/vctrs/blob/97c6fb4cd7ef0828ac299328d48c9515462140e6/src/vctrs.h#L473-L505

DavisVaughan:
> I imagine vec-complete-cases will work similarly but short-circuit as soon as it finds a missing value?

Yes, the code will be heavily duplicated, but I don't mind it that much

lionel-:
> do you want more information than this?

Ah no, that's good.

> I sort of feel like this should actually be called vctrs_colwise_info now, since we iterate over columns, but I'm not sure

Maybe something along the lines of `df_short_circuit_info`?

DavisVaughan:
I'm not sure `df_short_circuit_info` conveys exactly the right meaning because it carries the result object, `out`, as well, so its not just about short circuiting with `row_known` and `remaining`. I went with `df_col_traversal_info`.

DavisVaughan:
I have an idea to slightly simplify the info object, but I'll leave it and the renaming for a separate PR

PR #901, opened 2020-03-09
Use POSIXct for POSIXlt's comparison and equality proxy

DavisVaughan:
I am now convinced that we should just convert a POSIXlt object to POSIXct to get its comparison and equality proxy. It is too messy otherwise, with too many edge cases.

For the comparison proxy, we currently have a bug with DST because the comparison proxy doesn't include the `dst` field. But we can't just include that field because it won't be `NA` when the rest of the fields (like year, mon, etc) are `NA`, which would break `vec_equal_na()`.

I think the easiest way to do this is to just convert to POSIXct.

We will need to do this for the equality proxy as well, otherwise `vec_complete_cases()` won't work correctly. Currently the equality proxy for POSIXlt falls back to `vec_proxy()` which creates a data frame out of the entire record structure. This is a problem for `vec_complete_cases()` because the `gmtoffset` field can be `NA` if the offset is just "unknown" even if there is actual data for that element ([here](https://bugs.r-project.org/bugzilla/show_bug.cgi?id=16648) is an example), so it incorrectly will return `FALSE`. Again, converting to POSIXct removes this problem.

``` r
library(vctrs)

# 1am in EDT
x <- as.POSIXlt("2020-11-01 01:00:00", tz = "America/New_York")

# "falls back" to 1am again, but in EST
y <- as.POSIXlt(x + 3600)

x
#> [1] "2020-11-01 01:00:00 EDT"
y
#> [1] "2020-11-01 01:00:00 EST"

# look the same, but they are not!
vec_proxy_compare(x)
#>   year mon mday hour min sec
#> 1  120  10    1    1   0   0
vec_proxy_compare(y)
#>   year mon mday hour min sec
#> 1  120  10    1    1   0   0

# wrong!
vec_order(c(x, y))
#> [1] 1 2
vec_order(c(y, x))
#> [1] 1 2

order(c(x, y))
#> [1] 1 2
order(c(y, x))
#> [1] 2 1

x_ct <- as.POSIXct(x)
y_ct <- as.POSIXct(y)

# correct!
vec_order(c(x_ct, y_ct))
#> [1] 1 2
vec_order(c(y_ct, x_ct))
#> [1] 2 1
```

<sup>Created on 2020-03-09 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>


R/type-date-time.R
```diff
@@ -78,8 +78,12 @@ vec_proxy.POSIXlt <- function(x, ...) {
   new_data_frame(unclass(x))
 }
 #' @export
-vec_proxy_compare.POSIXlt <- function(x, ..., relax = FALSE) {
-  new_data_frame(vec_data(x)[c("year", "mon", "mday", "hour", "min", "sec")], n = length(x))
+vec_proxy_equal.POSIXlt <- function(x, ...) {
+  vec_data(as.POSIXct(x, tzone(x)))
```
lionel-: Is the `vec_data()` needed? I'd rather minimise its use in vctrs.

hadley: It should match the proxy for `POSIXct`, which returns a date time, so no 😄

DavisVaughan: The comparison proxy for `POSIXct` returns the underlying data, so maybe 

```
#' @export
vec_proxy_equal.POSIXlt <- function(x, ...) {
  # `vec_proxy_equal(<POSIXct>)` returns a POSIXct
  as.POSIXct(x, tzone(x))
}
#' @export
vec_proxy_compare.POSIXlt <- function(x, ...) {
  # `vec_proxy_compare(<POSIXct>)` returns the underlying data
  vec_data(as.POSIXct(x, tzone(x)))
}
```

DavisVaughan: Oh, this feels like less of a shortcut and should ensure consistency

```r
#' @export
vec_proxy_equal.POSIXlt <- function(x, ...) {
  x <- as.POSIXct(x, tzone(x))
  vec_proxy_equal(x)
}
#' @export
vec_proxy_compare.POSIXlt <- function(x, ...) {
  x <- as.POSIXct(x, tzone(x))
  vec_proxy_compare(x)
}
```

PR #903, opened 2020-03-10
Light refactoring of column iteration helpers

DavisVaughan:
This PR is a follow up of #897, and refactors the `struct vctrs_rowwise_info` into `struct df_short_circuit_info`.

- `SEXP out` has been removed from `df_short_circuit_info`. It didn't really belong, and this new name makes more sense without it.

- `R_len_t size` has been added to `df_short_circuit_info`. This makes it so we don't have to pass size through in all of the underlying functions that already take `info`.

- Now that we removed `out`, we can have a common `new_df_short_circuit_info()` constructor.

- General improvements in code quality, like using `size` rather than `n_row`.

### DIFF
--- a/src/compare.c
+++ b/src/compare.c
@@ -223,16 +223,16 @@ SEXP vctrs_compare(SEXP x, SEXP y, SEXP na_equal_) {
 // -----------------------------------------------------------------------------
 
 static void vec_compare_col(int* p_out,
+                            struct df_short_circuit_info* p_info,
                             SEXP x,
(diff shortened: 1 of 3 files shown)


DavisVaughan:
> pass the struct by pointer

I totally tried this earlier (after looking at how you do that with the dictionary now) and it seemed to hurt performance a bit. But I just added it in again and it seems to have essentially no effect on performance, so maybe I messed something up the first time. I even read that same SO post!

Issue #170, opened 2019-01-28
usethis-like function to create boilerplate code


krlmlr:
perhaps with RStudio add-in?

- [ ] construction
- [ ] validation
- [ ] casting
- [ ] coercion
- [ ] `as_()` and `is_()`
- [ ] `setOldClass()`
- [ ] tests (with a separate function?)

jennybc:
Tests?

krlmlr:
Would that be another function -- `use_vctr(name)` and `use_vctr_test(name = active_file_name())` ?

jennybc:
I'm not sure. I just want to put register the fact that there may be some standard test-able behaviours that could be scaffolded with an eventual template for vctrs S3 boilerplate.

hadley:
I think we can close this now, since the amount of boilerplate will dramatically decrease once #985 is implemented

PR #907, opened 2020-03-11
vctrs -> pizza

krlmlr:
for consistency, to simplify copy-pasting. Extends #880.


hadley:
I'm pretty sure we don't want to do this, since the bit at the bottom is likely to end up in another vignette.

krlmlr:
My use case is the adaptation of an existing class to vctrs. Currently copy-pasting and search-replacing from this text seems to be the fastest option; requires consistency, hence this PR. Would you consider reviewing as a stop-gap, until we have https://github.com/r-lib/vctrs/issues/170?

lionel-:
I was wondering if we should always use the `"my_"` namespace in examples.

hadley:
Maybe we could call the example package "my"?

lionel-:
Would you like to revisit this PR using the `my_` namespace @krlmlr?

krlmlr:
Will do.

krlmlr:
There's now :pizza: everywhere, rebased on master.

Now it's trivial to change this to "my" or to anything else. I think we should choose a longer word that will be unique in search+replace operations. There are [quite a few words ending with "my"](https://www.wordgamehelper.com/words-ending-with/my), including "dummy", not to mention [words that contain "my"](https://www.wordgamehelper.com/words-containing/my).

How about mypkg?

lionel-:
I don’t know, I like `my`. This is the conventional namespace in Emacs lisp for personal functions and variables and I have never had a search+replace issue with it. I’m not sure what kind of problems we could expect?

krlmlr:
It's probably fine, we only have one stray instance of "my" in the vignettes (and a few "MY" that need to be reviewed too). On the other hand, `mypkg` is 100% bulletproof now and in the future. Your call.

krlmlr:
Should I do `pizza` -> `mypkg` or `pizza` -> `<something else>` ?

lionel-:
The vctrs prefix is probably fine.

PR #908, opened 2020-03-11
Test S4 support

### LINKED ISSUE #550: `vec_restore()` should reset the S4 object bit
``` r
library(vctrs)

vec_restore(1, lubridate::seconds())
#> [1] 1
#> attr(,"year")
#> [1] 0
#> attr(,"month")
#> [1] 0
#> attr(,"day")
#> [1] 0
#> attr(,"hour")
#> [1] 0
#> attr(,"minute")
#> [1] 0
#> attr(,"class")
#> [1] "Period"
#> attr(,"class")attr(,"package")
#> [1] "lubridate"

isS4(vec_restore(1, lubridate::seconds()))
#> [1] FALSE
```

<sup>Created on 2019-08-29 by the [reprex package](https://reprex.tidyverse.org) (v0.2.1)</sup>

After doing something like this in `vec_restore_default()`:

```r
  if (IS_S4_OBJECT(to)) {
    SET_S4_OBJECT(x);
  }
```

we get correct results:

``` r
library(vctrs)

vec_restore(1, lubridate::seconds())
#> [1] "1S"

isS4(vec_restore(1, lubridate::seconds()))
#> [1] TRUE
```

<sup>Created on 2019-08-29 by the [reprex package](https://reprex.tidyverse.org) (v0.2.1)</sup>

### LINKED ISSUE #551: `vec_data()` might should unset the S4 object bit
``` r
library(vctrs)

vec_data(lubridate::seconds())
#> [1] 1

isS4(vec_data(lubridate::seconds()))
#> [1] TRUE

str(vec_data(lubridate::seconds()))
#> Formal class 'numeric' [package ""] with 0 slots
#>  list()

str(asS3(vec_data(lubridate::seconds())))
#>  num 1
```

<sup>Created on 2019-08-29 by the [reprex package](https://reprex.tidyverse.org) (v0.2.1)</sup>

This causes problems if `vctrs_vctr` objects are to be S4. In `str.vctrs_vctr` with a list-like vctrs object, we go through `obj_str_recursive`, which calls `utils::str(vec_data(x))`. When `x` is a `vctrs_vctr` object, `vec_data()` strips it down to the underlying `.Data`, but the S4-ness is not removed and `str.default` does something completely different than if it was S3 (as seen above)

krlmlr:
Fixes #550, fixes #551, fixes tidyverse/tibble#326.

### DIFF
--- a/R/proxy.R
+++ b/R/proxy.R
@@ -76,8 +76,12 @@ vec_data <- function(x) {
     x <- vec_set_attributes(x, list(names = names(x)))
   }
 
-  x
+  unset_s4(x)
 }
+unset_s4 <- function(x) {
+  .Call(vctrs_unset_s4, x)
+}
+
 #' @rdname vec_data
 #' @inheritParams ellipsis::dots_empty
 #' @export
--- a/R/proxy.R
+++ b/R/proxy.R
@@ -76,6 +76,7 @@ vec_data <- function(x) {
     x <- vec_set_attributes(x, list(names = names(x)))
   }
 
+  # Reset S4 bit in vector-like S4 objects
   unset_s4(x)
 }
 unset_s4 <- function(x) {
--- a/src/proxy-restore.c
+++ b/src/proxy-restore.c
@@ -11,7 +11,7 @@ static SEXP fns_vec_restore_dispatch = NULL;
 SEXP vec_restore_default(SEXP x, SEXP to) {
   SEXP attrib = ATTRIB(to);
 
-  if (attrib == R_NilValue) {
+  if (attrib == R_NilValue && !IS_S4_OBJECT(to)) {
     return x;
   }
 
@@ -85,6 +85,10 @@ SEXP vec_restore_default(SEXP x, SEXP to) {
 
   if (class != R_NilValue) {
     Rf_setAttrib(x, R_ClassSymbol, class);
+  }
+
+  if (IS_S4_OBJECT(to)) {
+    SET_S4_OBJECT(x);
   }
 
   UNPROTECT(n_protect);
--- a/src/proxy-restore.c
+++ b/src/proxy-restore.c
@@ -88,7 +88,7 @@ SEXP vec_restore_default(SEXP x, SEXP to) {
   }
 
   if (IS_S4_OBJECT(to)) {
-    SET_S4_OBJECT(x);
+    r_mark_s4(x);
   }
 
   UNPROTECT(n_protect);
--- a/src/proxy-restore.c
+++ b/src/proxy-restore.c
@@ -11,7 +11,9 @@ static SEXP fns_vec_restore_dispatch = NULL;
 SEXP vec_restore_default(SEXP x, SEXP to) {
   SEXP attrib = ATTRIB(to);
 
-  if (attrib == R_NilValue && !IS_S4_OBJECT(to)) {
+  const bool is_s4 = IS_S4_OBJECT(to);
+
+  if (attrib == R_NilValue && !is_s4) {
     return x;
   }
 
@@ -87,7 +89,7 @@ SEXP vec_restore_default(SEXP x, SEXP to) {
     Rf_setAttrib(x, R_ClassSymbol, class);
   }
 
-  if (IS_S4_OBJECT(to)) {
+  if (is_s4) {
     r_mark_s4(x);
   }
 
--- a/src/proxy.c
+++ b/src/proxy.c
@@ -58,8 +58,9 @@ static SEXP vec_proxy_unwrap(SEXP x) {
 
 // [[ register() ]]
 SEXP vctrs_unset_s4(SEXP x) {
-  UNSET_S4_OBJECT(x);
-  return(x);
+  x = r_maybe_duplicate(x);
+  r_unmark_s4(x);
+  return x;
 }
 
 SEXP vec_proxy_equal_dispatch(SEXP x) {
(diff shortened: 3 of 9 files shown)


krlmlr:
This checks that `vec_ptype()` has size 0. `vec_proxy()` has length > 0 in this case.

tests/testthat/test-s4.R
```diff
@@ -0,0 +1,113 @@
+.rando <- setClass("rando", contains = "numeric", slots = list(.Data = "numeric"))
+rando <- function(n = 0) {
+  withr::with_seed(20200311, .rando(.Data = runif(n)))
```
lionel-: Can this be simpler? E.g. no seed, no rng.

lionel-: Also please move this to a helper file.

tests/testthat/test-s4.R
```diff
@@ -0,0 +1,113 @@
...
+  x <- as_rando(1:10)
+  expect_equal(vec_ptype(vec_ptype2(x, x)), rando())
+
+  expect_equal(vec_ptype2(x, NA), rando())
+  expect_equal(vec_ptype2(NA, x), rando())
+
+  expect_equal(vec_ptype2(unspecified(), x), rando())
+  expect_equal(vec_ptype2(x, unspecified()), rando())
+
+  expect_error(vec_ptype2(x, 1))
```
lionel-: All these error expectations need `class = `.

tests/testthat/test-s4.R
```diff
@@ -0,0 +1,113 @@
...
+  expect_identical(vec_init(x, 2), vec_slice(x, idx))
+
+  dim(x) <- c(4, 2)
+  expect_identical(vec_init(x, 2), vec_slice(x, idx))
+
+  dim(x) <- c(2, 2, 2)
+  expect_identical(vec_init(x, 2), vec_slice(x, idx))
+})
+
+test_that("can chop rando objects with `NA_integer_` indices", {
```
lionel-: Are all these tests really necessary? This feels too comprehensive to me.

Now that they're there, no need to remove them, but in the future I think we can just assume that if the basic vctrs functionality works the rest works as well.

R/proxy.R
```diff
@@ -76,8 +76,12 @@ vec_data <- function(x) {
     x <- vec_set_attributes(x, list(names = names(x)))
   }
 
-  x
+  unset_s4(x)
```
lionel-: This doesn't seem necessary to me? The proxy of a S4 object should be a base vector.

src/type-info.c
```diff
@@ -82,13 +82,17 @@ static enum vctrs_type vec_base_typeof(SEXP x, bool proxied) {
   case CPLXSXP: return vctrs_type_complex;
   case STRSXP: return vctrs_type_character;
   case RAWSXP: return vctrs_type_raw;
+  case S4SXP:
+    // S4 objects are only vectors if they are proxied
+    if (proxied) return vctrs_type_list;
```
lionel-: 👍

src/proxy-restore.c
```diff
@@ -11,7 +11,7 @@ static SEXP fns_vec_restore_dispatch = NULL;
 SEXP vec_restore_default(SEXP x, SEXP to) {
   SEXP attrib = ATTRIB(to);
 
-  if (attrib == R_NilValue) {
+  if (attrib == R_NilValue && !IS_S4_OBJECT(to)) {
```
lionel-: If we support S4 objects here, it should be to throw an internal error recommending the class author to implement a restore method.

lionel-:
oops I wrote the review under the assumption that S4 classes are necessarily S4SXP. But there are actually vector-like S4 objects, as mentioned in the linked issues:

```r
x <- lubridate::seconds()

typeof(x)
#> [1] "double"

isS4(x)
#> [1] TRUE
```

R/proxy.R
lionel-: Can you add a comment that this is to support vector-like S4 objects, i.e. S4 objects that have a base vector type.

src/type-info.c
lionel-: Actually can you move this to a separate PR? The rando class is not a S4SXP.

src/proxy.c
```diff
@@ -56,6 +56,12 @@ static SEXP vec_proxy_unwrap(SEXP x) {
   return x;
 }
 
+// [[ register() ]]
+SEXP vctrs_unset_s4(SEXP x) {
```
lionel-: Can you make this `r_unmark_s4()` in utils.c please. The mark/unmark terminology conveys that the object is modified in place (like poke). Set and unset functions should make a copy if necessary.

Then add a `r_unset_s4()` variant that makes a copy.

R/proxy.R
lionel-: Please add a test to make sure the input object is not unmarked without a copy.

src/type-info.c
krlmlr: Done. We'll need to revert 15cbb55a44 in that separate PR.

tests/testthat/test-s4.R
krlmlr: Removed. Code is a burden and a liability, not an asset.

krlmlr:
Thanks for your review! Kept implementation in `utils.h` out of sheer laziness.

src/proxy-restore.c
DavisVaughan: Very minor, but you could do `bool is_s4 = IS_S4_OBJECT(to)` since you call `IS_S4_OBJECT(to)` later on as well

lionel-:
Thanks!

PR #911, opened 2020-03-12
More efficient `vec_group_loc()` vector access

DavisVaughan:
This PR further optimizes `vec_group_loc()`, which should directly result in performance improvements for `group_by()`.

In Instruments I noticed that the biggest killer in `vec_group_loc()` is the double deference in a tight loop of `INTEGER(VECTOR_ELT())`. That can be replaced with an array of integer pointers that index directly into those integer vectors. That really helped.

Side note: It is insane how fast this is if you order the data ahead of time. As pointed out by @brodieG

```r
library(vctrs)

set.seed(42)

g <- sample(1e6, 1e7, replace = TRUE)
tbl <- tibble(g = g)

o <- order(g)
ordered_tbl <- tbl[o,]
```

```r
bench::mark(
  vec_group_loc(tbl),
  vec_group_loc(ordered_tbl),
  check = FALSE,
  iterations = 20
)

# Master
#> # A tibble: 2 x 6
#> 1 vec_group_loc(tbl)            2.86s    2.99s     0.333     163MB    0.416
#> 2 vec_group_loc(ordered_tbl) 277.95ms 304.13ms     3.18      163MB    3.33

# This PR
#> # A tibble: 2 x 6
#> 1 vec_group_loc(tbl)            1.63s    1.71s     0.578     171MB    0.722
#> 2 vec_group_loc(ordered_tbl) 258.03ms 276.63ms     3.46      171MB    3.63
```

### DIFF
--- a/src/group.c
+++ b/src/group.c
@@ -154,8 +154,8 @@ SEXP vec_group_loc(SEXP x) {
 
   // Identify groups, this is essentially `vec_group_id()`
   for (int i = 0; i < n; ++i) {
-    int32_t hash = dict_hash_scalar(d, i);
-    R_len_t key = d->key[hash];
+    const int32_t hash = dict_hash_scalar(d, i);
+    const R_len_t key = d->key[hash];
 
     if (key == DICT_EMPTY) {
       dict_put(d, hash, i);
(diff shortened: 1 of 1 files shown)


brodieG:
Awesome!  Obviously the ratios are not quite as good once you include the ordering time, but at least for this 1e7/1e6 case it seems to be worth it.

If you do go ahead and end up using the order thing, please considering giving a nod to the data table folk that contributed the radix sort to R.  This trick is out of the question without that.

Also, as is probably obvious to you already, ordering might allow simplifying the group computation logic as well (maybe, I haven't looked at how you do it).

Thanks for keeping me in the loop, appreciate it.

PR #898, opened 2020-03-09
Improve `vec_as_names()` error messages

DavisVaughan:
Closes #882 

Adds improved error messages for:
- `stop_names_cannot_be_empty()`
- `stop_names_cannot_be_dot_dot()`
- `stop_names_must_be_unique()`

PR #915, opened 2020-03-12
Make locations available in stop_names()

krlmlr:
Follow-up to #898. The idea is to make the locations available in the `cnd` object when they are easy to compute.

### DIFF
--- a/R/conditions.R
+++ b/R/conditions.R
@@ -442,10 +442,11 @@ stop_names <- function(class = NULL, ...) {
   )
 }
 
-stop_names_cannot_be_empty <- function(names) {
+stop_names_cannot_be_empty <- function(names, locations) {
   stop_names(
     class = "vctrs_error_names_cannot_be_empty",
-    names = names
+    names = names,
+    locations = locations
   )
 }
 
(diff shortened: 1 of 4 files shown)


lionel-:
As you can see in #898, this is not the direction we're going for. You can store data (including locations) in the condition object at throw-time, but this shouldn't be a common parameter.

lionel-:
(If you need to store data in the condition for some feature, please add dots to the signature to that PR. No need for a separate PR.)

DavisVaughan:
@lionel-, sorry I should have let you know we talked about this. @krlmlr re-throws these messages in tibble and uses the `locations` data. For the empty names and dot dot names errors, we already have to compute that location data to determine if we need to signal the error, so we don't get any laziness benefits here. For that reason I thought it might be fine to just pass that information on through, since it was already computed.


@krlmlr if we do end up allowing this, please reuse the location data you store in the cnd object in these two places
https://github.com/krlmlr/vctrs/blob/75be1bd73dd98d7f3be1266c7d647e4c8eefb4c0/R/conditions.R#L460
https://github.com/krlmlr/vctrs/blob/75be1bd73dd98d7f3be1266c7d647e4c8eefb4c0/R/conditions.R#L489

krlmlr:
Yeah, I could have requested a review or motivate the PR more. I've worked around in tibble. There's a bit of redundancy between the tibble and vctrs codebases now (2 lines of code).

lionel-:
It seems like we're better off with slight redundancy? You can always store internal data in your error object (e.g. `cnd$tibble$locs <- locs`) and use it in your error display method.

I think it is preferable to keep the public API as small as possible. Having a public `locations` parameters when most errors are lazy and generate the messages in their own way seems to make the interface a bit sloppy.

PR #918, opened 2020-03-13
Add `x_arg` and `value_arg` to `vec_assign()`

### LINKED ISSUE #917: Should `vec_assign()` have `x_arg` and `value_arg`?
These would be passed down to `vec_coercible_cast()`, `vec_recycle()`, and `vec_assert()`
https://github.com/r-lib/vctrs/blob/5b08f1e6f4538cb75d13838e75669937acba2706/src/slice-assign.c#L40-L41
https://github.com/r-lib/vctrs/blob/5b08f1e6f4538cb75d13838e75669937acba2706/src/slice-assign.c#L30

DavisVaughan:
Closes #917 

CC @krlmlr 

Oh @lionel- I just had a thought. Should we add `...` to the signature before `x_arg` and `value_arg` and check that they are empty? (I'm going to guess yes)

### DIFF
--- a/R/slice.R
+++ b/R/slice.R
@@ -12,6 +12,10 @@
 #' @param value Replacement values. `value` is cast to the type of
 #'   `x`, but only if they have a common type. See below for examples
 #'   of this rule.
+#' @param x_arg,value_arg Argument names for `x` and `value`. These are used
+#'   in error messages to inform the user about the locations of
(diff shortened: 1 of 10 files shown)


lionel-:
Merging now so I can resolve diff conflicts with my branch for #689 right away.

PR #920, opened 2020-03-16
Return row names from `vec_names()`

lionel-:
Step towards #689, and general support of row names.

`vec_names()` returns row names, but only if they are of type character. Numeric row names are not treated as identifiers.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,6 +1,10 @@
 
 # vctrs (development version)
 
+* The internal function `vec_names()` now returns row names if the
+  input is a data frame. This is part of a general effort at
+  making row names the vector names of data frames in vctrs.
+
 * `vec_assign()` gains `x_arg` and `value_arg` parameters (#918).
 
 * `vec_group_loc()`, which powers `dplyr::group_by()`, now has more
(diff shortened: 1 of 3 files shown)


DavisVaughan:
@lionel- I believe we have discussed this, but I want to be sure.

`vec_set_names()` should continue to _never_ set data frame row names, correct?

lionel-:
hmm I think it should set names, with an additional unique-names check or repair.

NEWS.md
```diff
@@ -1,6 +1,10 @@
 
 # vctrs (development version)
 
+* The internal function `vec_names()` now returns row names, if
```
hadley: If the input is a data frame?

lionel-:
These are internal tools, so I don't expect them to be causing problems in themselves.

In the last version of vctrs we implemented row names support in vec-rbind, with user-visible consequences. The default way to handle uniqueness is to just repair the names verbosely, which avoids causing failures, but alerts users to potential issues of row identification.

hadley:
My primary concern is that we're now conflating two base R ideas (row names + vector names) that have slightly different semantics. I don't think it should block this PR, but I have a vague worry that it's going to create more work down the road.

lionel-:
You're right that we should be careful and reconsider this direction if it proves too much work.

PR #923, opened 2020-03-17
Set row names in `vec_set_names()`

lionel-:
Follow-up to #920. Step towards #689.

### DIFF
--- a/src/names.c
+++ b/src/names.c
@@ -709,18 +709,25 @@ SEXP vec_set_rownames(SEXP x, SEXP names) {
 }
 
 SEXP vec_set_df_rownames(SEXP x, SEXP names) {
-  x = PROTECT(r_maybe_duplicate(x));
-
   if (names == R_NilValue) {
+    if (rownames_type(df_rownames(x)) != ROWNAMES_IDENTIFIERS) {
+      return(x);
+    }
+
+    x = PROTECT(r_maybe_duplicate(x));
     init_compact_rownames(x, vec_size(x));
-  } else {
-    // Repair row names verbosely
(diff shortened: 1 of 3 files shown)


src/names.c
```diff
@@ -708,17 +708,32 @@ SEXP vec_set_rownames(SEXP x, SEXP names) {
   return x;
 }
 
+SEXP vec_set_df_rownames(SEXP x, SEXP names) {
+  x = PROTECT(r_maybe_duplicate(x));
+
+  if (names == R_NilValue) {
+    init_compact_rownames(x, vec_size(x));
+  } else {
+    // Repair row names verbosely
```
DavisVaughan: You say verbose here, but silent in the news bullet? Did I read something wrong? The default unique repair opts seem to have `quiet = false`

lionel-: It should say "verbosely" in NEWS (but without error).

src/names.c
```diff
@@ -708,17 +708,32 @@ SEXP vec_set_rownames(SEXP x, SEXP names) {
   return x;
 }
 
+SEXP vec_set_df_rownames(SEXP x, SEXP names) {
+  x = PROTECT(r_maybe_duplicate(x));
+
+  if (names == R_NilValue) {
+    init_compact_rownames(x, vec_size(x));
```
DavisVaughan: It might be worth special casing the `names == R_NilValue` case. If we are setting row names to `NULL` and we already have compact row names, we don't need to duplicate the input. We just return it as is. Both the array rownames setter and the atomic name setter make this check, so I think adding it for data frames would be nice to be consistent

PR #924, opened 2020-03-17
Preserve names when row-binding data frames

### LINKED ISSUE #689: vec_rbind() and names

lionel-:
Branched from #923.
Closes #689.

* The internal `vec_assign()` function gains an `assign_names` parameter. When `true`, names are assigned row-recursively.

* `vec_c()` and `vec_rbind()` set `assign_names` to `true` so that they preserve row names and inner names.

### DIFF
--- a/src/slice-assign.c
+++ b/src/slice-assign.c
@@ -141,22 +141,22 @@ SEXP vec_proxy_assign(SEXP proxy, SEXP index, SEXP value) {
 }
 SEXP vec_proxy_assign_opts(SEXP proxy, SEXP index, SEXP value,
                            const struct vec_assign_opts* opts) {
-  struct vctrs_proxy_info info = vec_proxy_info(value);
+  struct vctrs_proxy_info value_info = vec_proxy_info(value);
 
   // If a fallback is required, the `proxy` is identical to the output container
   // because no proxy method was called
   SEXP out = R_NilValue;
 
-  if (vec_requires_fallback(value, info) || has_dim(proxy)) {
+  if (vec_requires_fallback(value, value_info) || has_dim(proxy)) {
     index = PROTECT(compact_materialize(index));
     out = PROTECT(vec_assign_fallback(proxy, index, value));
   } else {
     PROTECT(index);
-    out = PROTECT(vec_assign_switch(proxy, index, info.proxy, opts));
+    out = PROTECT(vec_assign_switch(proxy, index, value_info.proxy, opts));
   }
 
   if (opts->assign_names) {
-    out = vec_proxy_assign_names(out, index, value);
+    out = vec_proxy_assign_names(out, index, value_info.proxy);
   }
 
   UNPROTECT(2);
--- a/src/slice-assign.c
+++ b/src/slice-assign.c
@@ -1,4 +1,5 @@
 #include "vctrs.h"
+#include "slice-assign.h"
 #include "subscript-loc.h"
 #include "utils.h"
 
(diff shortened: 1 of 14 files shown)


src/slice-assign.c
```diff
@@ -31,19 +32,29 @@ SEXP vctrs_assign(SEXP x, SEXP index, SEXP value, SEXP x_arg_, SEXP value_arg_)
...
+  };
+
+  return vec_assign_opts(x, index, value, &opts);
 }
 
 // [[ include("vctrs.h") ]]
-SEXP vec_assign(SEXP x, SEXP index, SEXP value,
-                struct vctrs_arg* x_arg,
-                struct vctrs_arg* value_arg) {
+SEXP vec_assign(SEXP x, SEXP index, SEXP value) {
```
DavisVaughan: I don't think we use this internally yet, but if we did I'd imagine that we'd want to pass `x_arg` and `value_arg` through?

src/slice-assign.c
```diff
@@ -52,39 +63,71 @@ SEXP vec_assign(SEXP x, SEXP index, SEXP value,
...
+  if (proxy_nms == R_NilValue) {
+    proxy_nms = Rf_allocVector(STRSXP, vec_size(proxy));
+    UNPROTECT(1);
+    PROTECT(proxy_nms);
+  }
+
+  proxy_nms = PROTECT(chr_assign(proxy_nms, index, value_nms));
+
+  proxy = PROTECT(r_maybe_duplicate(proxy));
+  vec_set_names(proxy, proxy_nms);
```
DavisVaughan: `vec_set_names()` works like `vec_proxy_assign()`. It might have to call back to `names<-` and duplicate the input, so it always returns it's input. So you won't need to duplicate the proxy, but I think you do need to do `proxy = PROTECT(vec_set_names())`

src/slice-assign.h
```diff
@@ -0,0 +1,29 @@
...
+}
+
+SEXP vec_assign_opts(SEXP x, SEXP index, SEXP value,
+                     const struct vec_assign_opts* opts);
+
+SEXP vec_proxy_assign_opts(SEXP proxy, SEXP index, SEXP value,
+                           const struct vec_assign_opts* opts);
+
+SEXP df_assign(SEXP x, SEXP index, SEXP value,
+               const struct vec_assign_opts* opts);
```
DavisVaughan: Does it make sense to go ahead and move `chr_assign()`, `list_assign()` and `vec_assign()` here too?

tests/testthat/test-bind.R
```diff
@@ -517,3 +517,27 @@ test_that("rbind repairs names of data frames (#704)", {
     class = "vctrs_error_names_must_be_unique"
   )
 })
+
+test_that("rbind supports names and inner names (#689)", {
```
DavisVaughan: This _should_ work with S3 objects like factors / POSIXct that `vec_set_names()` has to fall back to `names<-` on, but could you please add a test for that too? It would make me feel a bit better about it

src/slice-assign.c
lionel-: The idea was to keep the main function as easy to call as possible and to have another one with an option struct for additional parameters.

tests/testthat/test-bind.R
lionel-: This would have caught the unassigned return you mentioned above, so that test is definitely needed!

PR #925, opened 2020-03-17
Attempt to lookup S4 `c()` methods in `vec_c()`

### LINKED ISSUE #919: vec_c() for S4 classes
Introduced in #803, breaks as of d8efb4f, works in 361667d = d8efb4f^.

``` r
.rando <- setClass("rando", contains = "numeric", slots = list(.Data = "numeric"))
rando <- function(n = 0) {
  .rando(as.numeric(seq_len(n)))
}

as_rando <- function(x) {
  rando(length(x))
}

setMethod("[", "rando", function(x, i, j, ..., drop = TRUE) {
  new_n <- length(vec_as_location(i, length(x@.Data), names(x@.Data)))
  rando(new_n)
})


library(vctrs)
vec_c(rando(), rando())
#> Error: Can't find vctrs or base methods for concatenation.
#> vctrs methods must be implemented for class `rando`.
#> See <https://vctrs.r-lib.org/articles/s3-vector.html>.


library(lubridate)
#> 
#> Attaching package: 'lubridate'
#> The following object is masked from 'package:vctrs':
#> 
#>     new_duration
#> The following object is masked from 'package:base':
#> 
#>     date
x <- interval(today(), today())
vec_c(x, x)
#> Error: Can't find vctrs or base methods for concatenation.
#> vctrs methods must be implemented for class `Interval`.
#> See <https://vctrs.r-lib.org/articles/s3-vector.html>.
```

<sup>Created on 2020-03-15 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

DavisVaughan:
Closes #919 
Closes tidyverse/dplyr#4986
Related to https://github.com/tidyverse/lubridate/issues/721

This PR adds a way to check for S4 `c()` methods to fall back to in `vec_c()`. The implementation is derived from looking at `methods:::.showMethodsTable()`. S4 generic functions have an environment attached to them, and that environment contains another environment, `.MTable`, which is the methods table for that specific generic. That can be used to check if a method exists or not.

``` r
library(vctrs)
library(lubridate, warn.conflicts = FALSE)

vec_c(days(1:2), 3)
#> Error: No common type for `..1` <Period> and `..2` <double>.

vec_c(days(1:2), days(3))
#> [1] "1d 0H 0M 0S" "2d 0H 0M 0S" "3d 0H 0M 0S"
```

It seems to also work even if you don't explicitly load lubridate, which is nice

```r
library(vctrs)

vec_c(lubridate::days(1:2), lubridate::days(3))
#> [1] "1d 0H 0M 0S" "2d 0H 0M 0S" "3d 0H 0M 0S"
```

### DIFF
--- a/src/utils.c
+++ b/src/utils.c
@@ -10,6 +10,7 @@ SEXP (*rlang_env_dots_values)(SEXP) = NULL;
 SEXP (*rlang_env_dots_list)(SEXP) = NULL;
 SEXP vctrs_method_table = NULL;
 SEXP base_method_table = NULL;
+SEXP s4_c_method_table = NULL;
 
 SEXP strings_tbl = NULL;
 SEXP strings_tbl_df = NULL;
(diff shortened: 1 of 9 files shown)


src/utils.c
```diff
@@ -1354,6 +1395,9 @@ void vctrs_init_utils(SEXP ns) {
 
   base_method_table = r_env_get(R_BaseNamespace, Rf_install(".__S3MethodsTable__."));
 
+  s4_c_method_table = r_parse_eval("environment(methods::getGeneric('c'))$.MTable", R_GlobalEnv);
+  R_PreserveObject(s4_c_method_table);
```
DavisVaughan: Is the right thing to do here to preserve this object?

lionel-: It shouldn't hurt. If the table is replaced for some reason and gets GC'd, we'd be working with outdated methods, but we wouldn't crash. Since I don't know anything about S4 I'm not sure if that's a possibility, but I'd be surprised if it were.

PR #928, opened 2020-03-18
Fix indirections in child errors

lionel-:
Closes tidyverse/tibble#735.

@krlmlr This had to do with closure-inlined `cnd_body()` methods and child errors. The fix was rather tricky, but I think these rules might offer a general solution:

```
* Close over the parent when a child error is created outside of a
  given error class. For instance lossy cast errors need to consistently
  refer to their actual error object.

* Conversely, don't close over the parent when a child error is
  created within a given class. This allows subscript errors to pick up
  updated `subscript_arg` fields.
```

### DIFF
--- a/tests/testthat/error/test-subscript-loc.txt
+++ b/tests/testthat/error/test-subscript-loc.txt
@@ -174,6 +174,11 @@ i It must be numeric or character.
 Error: Must extract element with a single valid subscript.
 x Lossy cast from `foo` <double> to <integer>.
 
+> with_tibble_rows(vec_as_location2(TRUE))
+Error: Must remove row with a single valid subscript.
+x The subscript `foo(bar)` has the wrong type `logical`.
+i It must be numeric or character.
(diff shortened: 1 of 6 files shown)


tests/testthat/error/test-subscript-loc.txt
```diff
@@ -174,6 +174,11 @@ i It must be numeric or character.
 Error: Must extract element with a single valid subscript.
 x Lossy cast from `foo` <double> to <integer>.
 
+> with_tibble_rows(vec_as_location2(TRUE))
+Error: Must remove row with a single valid subscript.
+x The subscript `foo(bar)` has the wrong type `logical`.
```
krlmlr: `foo(bar)`?

lionel-: This is what `subscript_arg` is set to in the `with_tibble_rows()` tests.

Issue #922, opened 2020-03-17
Improve ptype2 and cast error messages


lionel-:
* [x] Instead of "Lossy cast from x to y", mention "Converting from x to y loses precision"?

* [ ] Error FAQ about lossy casts.

lionel-:
```r
#> Error: Can't cast `a` <fs_path> to `a` <fs_path>.
```

Should say "convert" instead of "cast". Would be nice to mention the type of input ("argument", "column", ...), in the same way we do for subscripts.

hadley:
Also in the case where the two classes are the same, maybe we should print some hint suggesting that a method is missing and needs to be provided by the package authort?

lionel-:
Good idea. Most of the time it should work once we finish implementing the fallback. But if `is_same_type()` returns `FALSE` even though the classes are the same, this hint would be useful.

lionel-:
```
Can't convert from `times` <character> to <integer>.
=>
Can't convert `times` from <character> to <integer>.
```

And then maybe:

```
Can't convert from `times` <character> to `foo` <integer>.
=>
Can't convert `times` from <character> to the type of `foo` <integer>.
```

lionel-:
We could improve the ptype2 errors along the same lines. Instead of:

> No common type for x and y.

It could be

> Can't combine x and y.

hadley:
Actually maybe the `from` isn't needed?

```
Can't convert `times` <character> to <integer>.
Can't convert `times` <character> to the type of `foo` <integer>.
```

hadley:
And do we need a hint?

```R
Can't combine `x` <character> to `y` <integer>.
Do you need to convert one of the arguments?
```

batpigandme:
I like the "loses precision" and using combine and convert as needed. 
```r
#> Error: Can't cast `a` <fs_path> to `a` <fs_path>.
```
This will definitely read like nonsense to an end user, even if it's convert instead of cast. Is there a way that a user could allow for things to be more permissive while waiting for the author to implement a method? How rare do we think this will be?

hadley:
I think the `can't cast <x> to <x>` problem is now (mostly) fixed so it should be rare for the user to see it.

PR #931, opened 2020-03-18
Improve support of foreign classes

### LINKED ISSUE #776: Can't cast <type> to <type>

lionel-:
In combining operations `vec_c()`, `vec_rbind()`, `vec_unchop()`, for homogeneously classed inputs. Closes #776.

* Like before, `c()` is used as fallback. However we check in advance for a `c()` implementation. If not found, we use the second fallback.

* Like `vec_ptype2()`, `vec_cast()` now uses `is_same_type()`. If the types are the same (their prototypes all have the same attributes), we no-op. Combining operations restore the attributes after concatenation.

I don't remember why I didn't implement this fallback immediately, I think it had to do with issues caused by method inheritance and `vctrs_vctr` inheriting from the base type, which we have since disabled.

Currently `vec_rbind()` doesn't fall back to `c()`, I'll implement this separately. Also still need to provide hints about implementing vctrs methods for incompatible homogeneously classed inputs (see #922).

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -6,6 +6,22 @@
   that it only supports lists. Atomic vectors and data frames result in an
   error.
 
+* Improved support for foreign classes in the combining operations
+  `vec_c()`, `vec_rbind()`, and `vec_unchop()`. A foreign class is a
+  class that doesn't implement `vec_ptype2()`. When all the objects to
+  combine have the same foreign class, one of these fallbacks is invoked:
+
+  -  If the class implements a `base::c()` method, the method is used
+    for the combination. (FIXME: `vec_rbind()` currently doesn't use
+    this fallback.)
+
+  - Otherwise if the objects have identical attributes and the same
+    base type, we consider them to be compatible. The vectors are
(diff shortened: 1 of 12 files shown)


lionel-:
Now that I think about it, `vec_rbind()` still is a bit tricky because we want to fall back to `c()` _before_ the ptype2 reduction...

If we do it after (the easy way), we're only compatible with simple foreign classes, not complex ones.

PR #934, opened 2020-03-19
Pull information from the proxy in `vec_chop()`

DavisVaughan:
This PR does two things:

- Slightly cleans up syntax by localizing the usage of the `SEXP elt` variable

- Tweaks `vec_chop()` to always pull information about the names / number of columns from the proxy. This previously was pulling from `x`, but that isn't right (my bad). If `x` is not a data frame, but has a proxy that is a data frame, this can cause issues. It hasn't bitten before because normally in these cases `x` is a rcrd type implemented internally as a list, and the length of the list matches the number of columns of the proxy data frame. However, if an S4 object is used `Rf_length()` doesn't match anymore, and we ended up crashing R or giving wrong results.

### DIFF
--- a/tests/testthat/test-slice-chop.R
+++ b/tests/testthat/test-slice-chop.R
@@ -169,30 +169,25 @@ test_that("vec_chop() falls back to `[` for shaped objects with no proxy when in
 })
 
 test_that("vec_chop() with data frame proxies always uses the proxy's length info", {
-  .Rcrd <- methods::setClass(
-    "vctrs_Rcrd",
-    contains = "integer",
-    slots = c(y = "integer")
-  )
-
   local_methods(
-    vec_proxy.vctrs_Rcrd = function(x) {
-      new_data_frame(list(x = x@.Data, y = x@y))
+    vec_proxy.vctrs_proxy = function(x) {
+      x <- proxy_deref(x)
+      new_data_frame(list(x = x$x, y = x$y))
     },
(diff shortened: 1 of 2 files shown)


tests/testthat/test-slice-chop.R
```diff
@@ -168,6 +168,33 @@ test_that("vec_chop() falls back to `[` for shaped objects with no proxy when in
   expect_equal(result, x)
 })
 
+test_that("vec_chop() with data frame proxies always uses the proxy's length info", {
```
lionel-: Just a note that you can also use the `vctrs_proxy` class from helper-s3.R for this kind of tests, see `local_proxy()`. This is a vector class implemented on top of an environment.

lionel-: I just realised that there's a tension between objects like `vctrs_proxy` and S4-style records. The former work better with names pulled from the proxy, whereas the latter works out of the box with names pulled from the input. I think that explains why we had different instincts about where to pull names in our call yesterday.

DavisVaughan: > The former work better with names pulled from the proxy

Oh, I see what you mean. For `vctrs_proxy`, we have no idea how to extract names from the input. We need the developer to give us a proxy with a standardized place for names to live, so we can slice them appropriately. Interesting.

That does suggest that names should always exist on the proxy, and should be pulled from the proxy. We should probably look at updating `vec_slice_impl()` to do this consistently.

PR #938, opened 2020-03-20
Copy "extending" vignette from tibble, with tweaks

krlmlr:
More work to do:

- Default rendering is already good as soon as a `format()` method is defined. Maybe start with only a print method?
- Use more intense coloring
- Rethink `is_vector_s3()` (https://github.com/r-lib/pillar/issues/181)
- Mention `verify_output()`

### DIFF
--- a/vignettes/pillar.Rmd
+++ b/vignettes/pillar.Rmd
@@ -1,0 +1,458 @@
+---
+title: "Printing vectors nicely in tibbles"
+author: "Kirill Müller, Hadley Wickham"
+output: rmarkdown::html_vignette
+vignette: >
+  %\VignetteIndexEntry{Printing vectors nicely in tibbles}
+  %\VignetteEngine{knitr::rmarkdown}
+  %\VignetteEncoding{UTF-8}
+---
+
+```{r setup, include = FALSE}
+options(crayon.enabled = TRUE)
+options(pillar.bold = TRUE)
+
+knitr::opts_chunk$set(collapse = TRUE, comment = pillar::style_subtle("#>"))
+
+colourise_chunk <- function(type) {
+  function(x, options) {
+    # lines <- strsplit(x, "\\n")[[1]]
+    lines <- x
+    if (type != "output") {
+      lines <- crayon::red(lines)
+    }
+    paste0(
+      '<div class="sourceCode"><pre class="sourceCode"><code class="sourceCode">',
+      paste0(
+        sgr_to_html(htmltools::htmlEscape(lines)),
+        collapse = "\n"
+      ),
+      "</code></pre></div>"
+    )
+  }
+}
+
+knitr::knit_hooks$set(
+  output = colourise_chunk("output"),
+  message = colourise_chunk("message"),
+  warning = colourise_chunk("warning"),
+  error = colourise_chunk("error")
+)
+
+# Fallback if fansi is missing
+sgr_to_html <- identity
+sgr_to_html <- fansi::sgr_to_html
+```
+
+To ensure that a vector prints nicely in a tibble, you need to understand how printing works.
+The presentation of a column in a tibble is powered by four S3 generics:
+
+* `type_sum()` determines what goes into the column header.
+* `pillar_shaft()` determines what goes into the body of the column.
+* `is_vector_s3()` and `obj_sum()` are used when rendering list columns.
+
+If you have created a vector class that can be used as a column, you can override these generics to make sure your data prints well in a tibble.
+To start, you must import the `pillar` package that powers the printing of tibbles.
+Either add `pillar` to the `Imports:` section of your `DESCRIPTION`, or call:
+
+```{r, eval = FALSE}
+usethis::use_package("pillar")
+```
+
+This short vignette assumes a package that implements a `"latlon"` vector and uses `roxygen2` to create documentation and the `NAMESPACE` file.
+To make the required methods available, we attach pillar.
+
+```{r attach}
+library(pillar)
+```
+
+
+
+## Prerequisites
+
+We define a class `"latlon"` that encodes geographic coordinates in a record.
+For simplicity, the values are printed as degrees and minutes only.
+
+```{r}
+library(vctrs)
+
+#' @export
+latlon <- function(lat, lon) {
+  new_rcrd(list(lat = lat, lon = lon), class = "latlon")
+}
+
+#' @export
+format.latlon <- function(x, ..., formatter = deg_min) {
+  x_valid <- which(!is.na(x))
+
+  lat <- field(x, "lat")[x_valid]
+  lon <- field(x, "lon")[x_valid]
+
+  ret <- rep("<NA>", vec_size(x))
+  ret[x_valid] <- paste(
+    formatter(lat, c("N", "S")),
+    formatter(lon, c("E", "W"))
+  )
+  format(ret, justify = "right")
+}
+
+deg_min <- function(x, pm) {
+  sign <- sign(x)
+  x <- abs(x)
+  deg <- trunc(x)
+  x <- x - deg
+  min <- round(x * 60)
+
+  ret <- sprintf("%d°%.2d'%s", deg, min, pm[ifelse(sign >= 0, 1, 2)])
+  format(ret, justify = "right")
+}
+
+latlon(32.7102978, -117.1704058)
+```
+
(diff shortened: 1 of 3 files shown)


krlmlr:
Can we please merge a variant of this so that I can change the tibble vignette to point to the new location (https://github.com/tidyverse/tibble/issues/560)?

vignettes/pillar.Rmd
```diff
@@ -0,0 +1,458 @@
...
+
+To ensure that a vector prints nicely in a tibble, you need to understand how printing works.
+The presentation of a column in a tibble is powered by four S3 generics:
+
+* `type_sum()` determines what goes into the column header.
+* `pillar_shaft()` determines what goes into the body of the column.
+* `is_vector_s3()` and `obj_sum()` are used when rendering list columns.
+
+If you have created a vector class that can be used as a column, you can override these generics to make sure your data prints well in a tibble.
+To start, you must import the `pillar` package that powers the printing of tibbles.
```
hadley: House style is to not put package names in code font

vignettes/pillar.Rmd
```diff
@@ -0,0 +1,458 @@
...
+Either add `pillar` to the `Imports:` section of your `DESCRIPTION`, or call:
+
+```{r, eval = FALSE}
+usethis::use_package("pillar")
+```
+
+This short vignette assumes a package that implements a `"latlon"` vector and uses `roxygen2` to create documentation and the `NAMESPACE` file.
+To make the required methods available, we attach pillar.
```
hadley:
```suggestion
To make the required methods available in this vignette, we attach pillar. 
You don't need to do this in your package.
```

hadley: Or maybe put this above the `use_package()` code so then you can say something like: "In a package, you instead do ..."

vignettes/pillar.Rmd
```diff
@@ -0,0 +1,458 @@
...
+```{r attach}
+library(pillar)
+```
+
+
+
+## Prerequisites
+
+We define a class `"latlon"` that encodes geographic coordinates in a record.
+For simplicity, the values are printed as degrees and minutes only.
```
hadley: Link back to other vignette here?

vignettes/pillar.Rmd
```diff
@@ -0,0 +1,458 @@
...
+
+Columns on this class can be used in a tibble right away, but the output will be less than ideal:
+
+```{r}
+library(tibble)
+data <- tibble(
+  venue = "rstudio::conf",
+  year  = 2017:2019,
+  loc   = latlon(
+    c(28.3411783, 32.7102978, NA),
```
hadley: Could you reduce the number of decimal places so this can fit on one line?

vignettes/pillar.Rmd
```diff
@@ -0,0 +1,458 @@
...
+```{r}
+#' @importFrom pillar type_sum
+#' @export
+type_sum.latlon <- function(x) {
+  "geo"
+}
+```
+
+Because the value shown there doesn't depend on the data, we just return a constant.
+(For date-times, the column info will eventually contain information about the timezone, see [#53](https://github.com/r-lib/pillar/pull/53).)
```
hadley: I don't think this parenthetical adds anything here, so can be removed.

vignettes/pillar.Rmd
```diff
@@ -0,0 +1,458 @@
...
+
+## Rendering the value
+
+To use our format method for rendering, we implement the `pillar_shaft()` method for our class.
+(A [*pillar*](https://en.wikipedia.org/wiki/Column#Nomenclature) is mainly a *shaft* (decorated with an *ornament*), with a *capital* above and a *base* below.
+Multiple pillars form a *colonnade*, which can be stacked in multiple *tiers*.
+This is the motivation behind the names in our API.)
+
+```{r include=FALSE}
+import::from(pillar, pillar_shaft)
```
hadley: Can you avoid using import?

vignettes/pillar.Rmd
```diff
@@ -0,0 +1,458 @@
...
+}
+
+#' @export
+format.latlon <- function(x, ..., formatter = deg_min) {
+  x_valid <- which(!is.na(x))
+
+  lat <- field(x, "lat")[x_valid]
+  lon <- field(x, "lon")[x_valid]
+
+  ret <- rep("<NA>", vec_size(x))
```
hadley: Shouldn't the format method return a `NA_character_` here?

vignettes/pillar.Rmd
```diff
@@ -0,0 +1,458 @@
...
+  ret[x_valid] <- paste(
+    formatter(lat, c("N", "S")),
+    formatter(lon, c("E", "W"))
+  )
+  format(ret, justify = "right")
+}
+
+deg_min <- function(x, pm) {
```
hadley: Why have a `pm` argument?

vignettes/pillar.Rmd
```diff
@@ -0,0 +1,458 @@
...
+data
+```
+
+Currently, ANSI escapes are not rendered in vignettes, so the display here isn't much different from earlier examples.
+This may change in the future.
+
+## Testing
+
+If you want to test the output of your code, you can compare it with a known state recorded in a text file.
+For this, pillar offers the `expect_known_display()` expectation which requires and works best with the testthat package.
```
hadley: I think we'd recommend `verify_output()` now?

vignettes/pillar.Rmd
```diff
@@ -0,0 +1,458 @@
...
+```
+
+You may want to create a series of output files for different scenarios:
+
+- Colored vs. plain (to simplify viewing differences)
+- With or without special Unicode characters (if your output uses them)
+- Different widths
+
+For this it is helpful to create your own expectation function.
+Use the tidy evaluation framework to make sure that construction and printing happens at the right time:
```
hadley: I think this is a bit too complicated for this vignette and is better omitted.

vignettes/pillar.Rmd
```diff
@@ -0,0 +1,458 @@
...
+  %\VignetteEncoding{UTF-8}
+---
+
+```{r setup, include = FALSE}
+options(crayon.enabled = TRUE)
+options(pillar.bold = TRUE)
+
+knitr::opts_chunk$set(collapse = TRUE, comment = pillar::style_subtle("#>"))
+
+colourise_chunk <- function(type) {
```
hadley: I'd prefer to leave this off to match the other vignettes (and you appear to have already included a caveat about colouring not working)

vignettes/pillar.Rmd
krlmlr: Not sure, this is meant to replace the other vignette.

vignettes/pillar.Rmd
krlmlr: I don't know:

``` r
format(1[NA])
#> [1] "NA"
format("1"[NA])
#> [1] "NA"
format(Sys.time()[NA])
#> [1] NA
format(Sys.Date()[NA])
#> [1] NA
format(hms::hms()[NA])
#> [1] "NA"
```

<sup>Created on 2020-03-20 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

vignettes/pillar.Rmd
krlmlr: Latitute is N/S, longitude is E/W. What's a better name?

vignettes/pillar.Rmd
hadley: Sorry, I meant the other vctrs vignette that tells you how to create a rcrd

vignettes/pillar.Rmd
hadley: Oh, now I get it. I think it would be slightly easier to understand if you did

```R
deg_min <- function(x, direction = c("lat", "lon")) {
  pm <- if (direction == "lat") c("N", "S") else c("E", "W")
  ...
```

krlmlr:
Rearranged, coloring comes first. The tests use `verify_output()`, `is_s3_vector()` is gone, `vec_ptype_abbr()` is used instead of `type_sum()` .

hadley:
I made quite a few tweaks — can you please take another pass through?

krlmlr:
Thanks, I like the new narrative. Did you remove left alignment on purpose?

hadley:
Yes, I thought it was ugly 😄

krlmlr:
Is this good for merging?

PR #939, opened 2020-03-20
Improve subscript errors of unexpected dimensionality

lionel-:
Closes #936.

```
> vec_as_location(array(TRUE, dim = c(1, 1, 1)), 3L)
Error: Must subset elements with a valid subscript vector.
x The subscript must be a simple vector, not an array.

> with_tibble_rows(vec_as_location(matrix(TRUE, nrow = 1), 3L))
Error: Must remove rows with a valid subscript vector.
x The subscript `foo(bar)` must be a simple vector, not a matrix.
```

PR #940, opened 2020-03-20
Check for dimensionality in `vec_as_subscript()`

lionel-:
Branched from #939.

Some internal reorganisation:

* Include `subscript_action` in subscript struct.
* Include subscript struct in location struct.
* Remove `subscript_arg` from location struct and use the one from the subscript struct instead.

This allows moving the dimensionality check in `vec_as_subscript()`, which now consistently checks for matrices and arrays in the same way as `vec_as_location()`.

### DIFF
--- a/src/subscript-loc.c
+++ b/src/subscript-loc.c
@@ -118,7 +118,9 @@ static SEXP int_invert_location(SEXP subscript, R_len_t n,
     j = -j;
     if (j > n) {
       struct vec_as_location_opts updated_opts = *opts;
-      updated_opts.action = SUBSCRIPT_ACTION_NEGATE;
+      struct vec_as_subscript_opts updated_subscript_opts = *updated_opts.subscript_opts;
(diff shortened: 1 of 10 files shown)


src/subscript-loc.c
```diff
@@ -283,30 +282,14 @@ SEXP vec_as_location(SEXP subscript, R_len_t n, SEXP names) {
...
-  } else {
-    struct vec_as_subscript_opts default_subscript_opts = {
-      .logical = SUBSCRIPT_TYPE_ACTION_CAST,
-      .numeric = SUBSCRIPT_TYPE_ACTION_CAST,
-      .character = SUBSCRIPT_TYPE_ACTION_CAST,
-      .subscript_arg = location_opts->subscript_arg
-    };
-    subscript = vec_as_subscript_opts(subscript, &default_subscript_opts, &err);
-  }
+  subscript = vec_as_subscript_opts(subscript, opts->subscript_opts, &err);
```
DavisVaughan: This is a nice simplification 👍

Issue #822, opened 2020-02-14
Infinite recursion with ptype2 and partial types


DavisVaughan:
This is the cran version:

```r
> library(vctrs)
> vec_ptype2(vec_ptype2(NULL, 1), partial_factor())
Error: C stack usage  7970384 is too close to the limit
```

Slightly different error on dev master, but same idea

```r
> library(vctrs)
> vec_ptype2(vec_ptype2(NULL, 1), partial_factor())
Error: evaluation nested too deeply: infinite recursion / options(expressions=)?
```

I think this comes from how `is_same_type()` is implemented. We eventually call it from `vec_ptype2()`, and then it calls into `vec_ptype_common()`, so we get recursion

lionel-:
Calling vctrs from vctrs methods is currently dangerous. We'll need clear guidelines about what is safe to be called, but currently this is not well defined and changing over time.

PR #942, opened 2020-03-20
Double dispatch manually in `vec_ptype2()`

lionel-:
We look up the left ptype2 method in the vctrs method table, then we look up the right method in the method table of the environment of the left method. This way the process for implementing methods mostly stays the same and we continue to benefit from registration via roxygen2 and R CMD check support. However implementation is now simplified:

* No longer need to worry about order of dispatch, a simple `UseMethod("vec_ptype2.class")` is sufficient.

* No longer need to implement default methods.

In a future PR, it will be easy to disable inheritance in ptype2 dispatch and avoid the ambiguities it causes.

I struggled for a while with an infloop involving partial frames and unimplemented ptype2 methods. I understand the infloop (ptype2-default calls ptype2 again for partial objects) but not how it worked before. Not sure how to solve it, so I just disabled the implicated test. Since partial types are not very useful right now, we can revisit these later.

The `AsIs` class was a bit problematic because it relies on a default ptype2 method. We'll probably want to remove support for those, but this will need to wait until vctrs supports native wrapper types.

I'll update the docs and vignette after the inheritance PR. Launching revdeps now so we monitor the backward compatibility of these important changes.


src/ptype2-dispatch.c
```diff
@@ -64,6 +63,135 @@ SEXP vec_ptype2_dispatch_s3(SEXP x,
...
+                                            &x_method_sym));
+
+  if (x_method == R_NilValue) {
+    SEXP out = vec_ptype2_default(x, y, x_arg_obj, y_arg_obj);
+    UNPROTECT(3);
+    return out;
+  }
+
+  const char* x_method_str = CHAR(PRINTNAME(x_method_sym));
+  SEXP x_table = s3_get_table(CLOENV(x_method));
```
DavisVaughan: So, this ensures that the double dispatch method can be found in a package that is extending vctrs? Because the S3 method table would have to come from that package, and not vctrs?

Update: Oh yea that is what this is doing. Nice!

src/ptype2-dispatch.c
```diff
@@ -64,6 +63,135 @@ SEXP vec_ptype2_dispatch_s3(SEXP x,
...
+  UNPROTECT(1);
+  *method_sym_out = R_NilValue;
+  return R_NilValue;
+}
+
+// [[ include("vctrs.h") ]]
+SEXP vec_ptype2_dispatch_s3(SEXP x,
+                            SEXP y,
+                            enum vctrs_type x_type,
+                            enum vctrs_type y_type,
```
DavisVaughan: Do you end up using these?

tests/testthat/test-partial-frame.R
```diff
@@ -54,25 +54,30 @@ test_that("incompatible data frames are an error", {
 })
 
 test_that("dispatch is symmetric with tibbles", {
+  # Now causes infloop with the new double-dispatch mechanism because
```
DavisVaughan: Likely related? https://github.com/r-lib/vctrs/issues/822

DavisVaughan:
For `AsIs`, would it make any sense to make it a known class and handle it internally like we do with `vctrs_unspecified`?

R/type-vctr.R
```diff
@@ -661,7 +660,6 @@ format.hidden <- function(x, ...) rep("xxx", length(x))
 local_hidden <- function(frame = caller_env()) {
   local_bindings(.env = global_env(), .frame = frame,
     vec_ptype2.hidden         = function(x, y, ...) UseMethod("vec_ptype2.hidden", y),
-    vec_ptype2.hidden.default = function(x, y, ...) stop_incompatible_type(x, y, ...),
```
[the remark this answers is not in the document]
lionel-: hmm the test in test-type-bare.R still seems useful, I'll update the comment.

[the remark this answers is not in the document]
lionel-: oh no, I see what you mean.

R/type-vctr.R
```diff
@@ -108,9 +108,8 @@ vec_restore.vctrs_vctr <- function(x, to, ..., i = NULL) {
 #' @method vec_ptype2 vctrs_vctr
 #' @export
 vec_ptype2.vctrs_vctr <- function(x, y, ..., x_arg = "x", y_arg = "y") {
-  # This method is redundant with `vec_ptype2.default()` but it
-  # instructs `vec_c()` that it isn't a foreign type. This avoids
-  # infinite recursion through `c.vctrs_vctr()`.
+  # This method instructs `vec_c()` that it isn't a foreign type. This
+  # avoids infinite recursion through `c.vctrs_vctr()`.
   vec_default_ptype2(x, y, ..., x_arg = x_arg, y_arg = y_arg)
```
lionel-: With the new system, the code in left methods (generic-methods) is never called. The consequence is that it's now possible to implement `vec_ptype2()` methods with `vctrs_vctr`. Also it's misleading to make it seem that the code inside the left-method is called.

I wonder if it'd make sense to parameterise dispatch via literal hints:

```r
vec_ptype2.vctrs_vctr <- function(x, y, ...) {
  "vctrs:::no_ptype2"
  UseMethod("vec_ptype2.vctrs_vctr")
}

vec_ptype2.AsIs <- function(x, y, ...) {
  "vctrs:::wrapper_type"
  UseMethod("vec_ptype2.AsIs")
}
```

src/ptype2-dispatch.c
```diff
@@ -38,20 +38,19 @@ SEXP vec_ptype2_dispatch(SEXP x, SEXP y,
...
 static SEXP fns_vec_ptype2_dispatch_s3 = NULL;
 static SEXP syms_vec_ptype2_dispatch_s3 = NULL;
 
-// [[ include("vctrs.h") ]]
-SEXP vec_ptype2_dispatch_s3(SEXP x,
-                            SEXP y,
-                            struct vctrs_arg* x_arg,
-                            struct vctrs_arg* y_arg) {
-                            SEXP x_arg_chr = PROTECT(vctrs_arg(x_arg));
+static SEXP vec_ptype2_dispatch_old(SEXP x,
```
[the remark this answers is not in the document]
lionel-: oops, this was a leftover from development.

src/ptype2-dispatch.c
```diff
@@ -64,6 +63,135 @@ SEXP vec_ptype2_dispatch_s3(SEXP x,
...
+  // Support corrupt objects where `x` is an OBJECT(), but the class is NULL
+  if (class == R_NilValue) {
+    class = base_dispatch_class(x);
+  }
+  PROTECT(class);
+
+  SEXP* class_ptr = STRING_PTR(class);
+  int n_class = Rf_length(class);
+
+  // FIXME: Disable inheritance
```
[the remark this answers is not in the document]
lionel-: Yes we'll only dispatch on the first class.

src/ptype2-dispatch.c
lionel-: I originally planned to hard-code behaviour for bare vectors, as an optimisation (we'd dispatch directly on the right method in the vctrs table). In the end I think this isn't worth it, I'll remove.

tests/testthat/test-partial-frame.R
lionel-: Probably, this is about calling `vec_ptype2()` again from `vec_ptype2_default()`, for partial objects.

lionel-:
@DavisVaughan I think adding classes to the type enum should be a last resort. I'd rather add a branch in ptype2 if needed. But if we can frame this special casing as a preliminary general mechanism for wrapper types, that'd be better. I'll look into it on Monday, but if it looks like it's too difficult I'll special-case.

NEWS.md
```diff
@@ -1,6 +1,12 @@
 
 # vctrs (development version)
 
+* Double-dispatch methods for `vec_ptype2()` are now easier to
```
hadley: Can you update the vignette too please?

src/ptype2-dispatch.c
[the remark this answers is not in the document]
hadley: When we do that, we should probably add some check for the symmetric method? Maybe a debug mode that would be activated in tests and would run (e.g.) `vec_cast.factor.character` and `vec_cast.character.factor` and error if not equal?

src/ptype2-dispatch.c
```diff
@@ -46,24 +46,107 @@ SEXP vec_ptype2_dispatch(SEXP x, SEXP y,
...
+                              const char* generic,
+                              SEXP table,
+                              SEXP* method_sym_out) {
+  SEXP class = R_NilValue;
+  if (OBJECT(x)) {
+    class = Rf_getAttrib(x, R_ClassSymbol);
+  }
+
+  // This also handles gremlins objects where `x` is an OBJECT(), but
+  // the class is NULL
```
hadley:
```suggestion
  // Handle gremlin objects where `x` is an OBJECT(), but class is NULL
```

src/ptype2-dispatch.c
```diff
@@ -46,24 +46,107 @@ SEXP vec_ptype2_dispatch(SEXP x, SEXP y,
...
+  if (class == R_NilValue) {
+    class = s3_dispatch_class(x);
```
hadley: This is the internal class?

NEWS.md
```diff
@@ -1,6 +1,12 @@
...
+  implement. vctrs takes care of implementing the default and
+  unspecified methods. Methods no longer have to worry about order of
+  dispatch, for instance it is no longer needed to supply a second
+  argument as in `UseMethod("vec_ptype2.mytype", y)`.
```
hadley: Mention explicitly that `NextMethod()` doesn't work, and it never worked correctly?

src/ptype2-dispatch.c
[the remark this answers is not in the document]
lionel-: Good idea. The debug mode probably shouldn't be activated automatically in tests though, in case the error comes from another package?

src/ptype2-dispatch.c
lionel-: I'll be more explicit that the other case handled here is bare objects.

src/ptype2-dispatch.c
lionel-: Yes. Maybe that'd be clearer as `s3_bare_class()`?

src/ptype2-dispatch.c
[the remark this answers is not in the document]
hadley: Sorry, yeah, I meant in _our_ tests.

src/ptype2-dispatch.c
hadley: Yeah

lionel-:
@hadley I updated the vignette.

PR #943, opened 2020-03-23
Don't restore row names if target is not a data frame

lionel-:
Fixes proxy/restore roundtrip for POSIXlt objects. Before the patch, they'd gain a `row.names` attribute.

### DIFF
--- a/tests/testthat/test-type-date-time.R
+++ b/tests/testthat/test-type-date-time.R
@@ -47,6 +47,13 @@ test_that("vec_proxy() returns a double for Dates with int representation", {
   expect_true(is.double(vec_proxy(x)))
 })
 
+test_that("POSIXlt roundtrips through proxy and restore", {
+  x <- as.POSIXlt("2020-01-03")
+  out <- vec_restore(vec_proxy(x), x)
+  expect_identical(out, x)
+})
+
+
 # coerce ------------------------------------------------------------------
 
 test_that("datetime coercions are symmetric and unchanging", {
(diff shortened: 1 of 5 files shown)


tests/testthat/test-type-date-time.R
```diff
@@ -47,6 +47,13 @@ test_that("vec_proxy() returns a double for Dates with int representation", {
   expect_true(is.double(vec_proxy(x)))
 })
 
+test_that("POSIXlt roundtrips through proxy and restore", {
+  x <- as.POSIXlt("2020-01-03")
+  out <- vec_restore(vec_proxy(x), x)
+  expect_identical(out, x)
```
lionel-: This is why I prefer `expect_identical()` in low level packages by the way. `expect_equal()` would not have checked for row names.

DavisVaughan:
Can we completely move the handling of data frame row names to `bare_df_restore_impl()`? We already query the row names there

https://github.com/r-lib/vctrs/blob/187e34375f158106afad4613234015157a569782/src/proxy-restore.c#L119

DavisVaughan:
If we do that, I _think_ you'd have to call `vec_names(x)` from `bare_df_restore_impl()`, and do it before `vec_restore_default()` is called

lionel-:
This is something to keep in mind, but it's tricky since the inheritance of restore methods isn't clear yet.

lionel-:
This broke `grouped_df` restoration and dplyr, looking into it.

lionel-:
Fixed in https://github.com/r-lib/vctrs/commit/71aedcf42d8fdedc2e3f344af18813d9cf71acff.

Issue #710, opened 2019-12-19
Avoid inheritance in double dispatch


lionel-:
By guarding our methods and generics with `inherits_only()`:

```r
vec_ptype2.data.table <- function(x, y, ..., x_arg = "x", y_arg = "y") {
  if (inherits_only(x, c("data.table", "data.frame"))) {
    UseMethod("vec_ptype2.data.table", y)
  } else {
    vec_default_ptype2(x, y, x_arg = x_arg, y_arg = y_arg)
  }
}
vec_ptype2.data.table.data.frame <- function(x, y, ..., x_arg = "x", y_arg = "y") {
  if (inherits_only(y, "data.frame")) {
    x
  } else {
    vec_default_ptype2(x, y, x_arg = x_arg, y_arg = y_arg)
  }
}

vec_ptype2.data.frame.data.table <- function(x, y, ..., x_arg = "x", y_arg = "y") {
  # The `x` check should be taken care of `vec_ptype2.data.frame()`
  if (inherits_only(x, "data.frame") && inherits_only(y, c("data.table", "data.frame"))) {
    y
  } else {
    vec_default_ptype2(x, y, x_arg = x_arg, y_arg = y_arg)
  }
}
```

This means that we force subclasses to write common type methods with their superclasses, but in return we gain type stability. If we allow inheritance we get:

```r
dt <- data.table::as.data.table(mtcars)
gdf <- dplyr::group_by(mtcars, cyl)

# This would return a `dt` because `gdf` inherits from data.frame
vec_ptype2(dt, gdf)
# And this would return a `gdf` because `dt` inherits from data.frame
vec_ptype2(gdf, dt)
```

We should try and offer a better common type mechanism in the long term, but this approach would at least make things more predictable in the short term.

DavisVaughan:
Small comment. I assume it would return `vec_ptype(x)` not `x` when the `inherits_only()` case is true.

lionel-:
Unclear whether we should return the base super type `data.frame()` instead of throwing a incompatible type error. This would make it possible to combine a dt and a gdf, but we'd silently lose the grouping structure. Probably better to be stricter for now.

Regarding `vec_cast()`, we should avoid inheritance for `to`, but it should be fine to allow it for `x`.

PR #950, opened 2020-03-25
Fix ptype2 inheritance

lionel-:
Branched from #942.
Part of #710.

`vec_ptype2()` now determines methods from the first component of the class vector. This disables inheritance, which methods no longer need to guard from.


DavisVaughan:
Just a comment as I'm reading through. Most R level `vec_ptype2()` methods are now only ever used for shaped input. So we will need to implement that at the C level too so we can eventually make the R level methods "dummy" methods

R/type-list-of.R
```diff
@@ -185,11 +185,7 @@ as.character.vctrs_list_of <- function(x, ...) {
 #' @method vec_ptype2 vctrs_list_of
 #' @export
 vec_ptype2.vctrs_list_of <- function(x, y, ..., x_arg = "x", y_arg = "y") {
-  if (inherits_only(x, c("vctrs_list_of", "vctrs_vctr"))) {
-    UseMethod("vec_ptype2.vctrs_list_of", y)
-  } else {
-    vec_default_ptype2(x, y, x_arg = x_arg, y_arg = y_arg)
-  }
+  UseMethod("vec_ptype2.vctrs_list_of", y)
```
DavisVaughan: I guess you don't need `y` here now right?

DavisVaughan: I guess you don't need `y` here now right?

R/type2.R
```diff
@@ -23,6 +23,27 @@
 #' return the same value as `vec_ptype2.y.x()`; this is not enforced
 #' for reasons of efficiency, but should be tested.
 #'
+#' `vec_ptype2()` are not inherited, classes must explicitly implement
+#' the methods. There are two reasons for this:
```
DavisVaughan:
```suggestion
#' `vec_ptype2()` methods are not inherited, classes must explicitly implement
#' them. There are two reasons for this:
```

tests/testthat/test-type2.R
```diff
@@ -158,11 +158,11 @@ test_that("Subclasses of data.frame dispatch to `vec_ptype2()` methods", {
   expect_identical(vec_ptype2(mtcars, quux), "dispatched!")
 })
 
-test_that("Subclasses of `tbl_df` have `tbl_df` common type (#481)", {
+test_that("Subclasses of `tbl_df` do not have `tbl_df` common type (#481)", {
```
DavisVaughan: Maybe update or remove the issue number so we aren't confused if we look back on this?

```suggestion
test_that("Subclasses of `tbl_df` do not have `tbl_df` common type (#710)", {
```

R/type-list-of.R
lionel-: Right

lionel-:
> Most R level vec_ptype2() methods are now only ever used for shaped input. So we will need to implement that at the C level too so we can eventually make the R level methods "dummy" methods

That'd be great. I was wondering if we could include shape-matching as part of our default handling, even for foreign S3 classes.

R/type-list-of.R
lionel-:
```suggestion
  UseMethod("vec_ptype2.vctrs_list_of")
```

NEWS.md
```diff
@@ -1,6 +1,12 @@
 
 # vctrs (development version)
 
+* Double-dispatch methods for `vec_ptype2()` are no longer inherited
```
hadley: I think this needs an example just to make this concrete

R/type-bare.R
```diff
@@ -7,77 +7,49 @@
 #' @method vec_ptype2 logical
 #' @export
 vec_ptype2.logical <- function(x, y, ..., x_arg = "", y_arg = "") {
-  if (is.object(x)) {
-    vec_default_ptype2(x, y, x_arg = x_arg, y_arg = y_arg)
-  } else {
-    UseMethod("vec_ptype2.logical")
-  }
+  UseMethod("vec_ptype2.logical")
```
hadley: Can you add a note in the vignette saying that these functions are never actually called?

src/ptype2-dispatch.c
```diff
@@ -116,26 +116,17 @@ SEXP vec_ptype2_dispatch_s3(SEXP x,
...
-                                    x_table,
-                                    &y_method_sym);
-
-  // FIXME: The `AsIs` class relies on a default ptype2 method. Should
-  // be changed with a general mechanism for wrapper types.
-  if (y_method == R_NilValue && Rf_inherits(x, "AsIs")) {
-    y_method_sym = s3_paste_method_sym(x_method_str, "default");
-    y_method = s3_sym_get_method(y_method_sym, x_table);
-  }
+  SEXP y_method = PROTECT(get_ptype2_method(y,
```
DavisVaughan: Nice!

PR #951, opened 2020-03-25
Add dplyr methods

lionel-:
Branched from #950.

Because `grouped_df` no longer inherit from the `tbl_df` methods, dplyr methods need to be implemented. With this PR all dplyr tests pass, except for those that count the number of regroups.

`vec_cbind()` with data frames that have special columns was problematic. One goal of frame API PRs was to find semantics that would allow propagating these special columns in colwise combinations. This is too involved for dplyr 1.0, so we use a different approach here:

- All inputs are converted to frame prototypes, i.e. zero-column inputs. `vec_restore()` is called as part of this conversion. I'm not entirely happy about using restore for this because this isn't what it was designed for. I'm not sure how we could do better, except by calling back the R level `[` method. Maybe that's a safer option.

- The restore method for `grouped_df` returns a simple tibble when the grouping columns have been lost, so the frame prototype of grouped data frames are tibbles.

- The common type is determined from the frame prototypes.

As a result, `vec_cbind(gdf, df)` returns a tibble (the simplified type of grouped data frames).


tests/testthat/test-type-dplyr.R
```diff
@@ -0,0 +1,144 @@
...
+
+test_that("can cbind grouped data frames", {
+  gdf <- dplyr::group_by(mtcars[-10], cyl)
+  df <- unrownames(mtcars)[10]
+
+  expect_identical(
+    unrownames(vec_cbind(gdf, df)),
+    tibble::as_tibble(mtcars)[c(1:9, 11, 10)]
+  )
+})
```
DavisVaughan: Also, it is somewhat interesting that `vec_ptype2(rww, gdf)` is an error, but `vec_cbind(rww, gdf)` works and returns a rowwise_df because the `df_container_type()` of `rww` is a rowwise_df, and for `gdf` it is a simple tibble.

This somehow feels wrong?

lionel-: Good point. I agree it doesn't feel right, but I'm not sure how to do better. The alternative is to decide the frame ptype of rowwise data frames is a tibble purely for consistency with grouped data frames. This feels arbitrary as well.

lionel-:
Updated tidyr revdeps. We have a bunch of issues involving subclasses of data frames. The package authors will need to implement methods before we can release the breaking changes.

amt (cbind)

```r
#> Error: No common type for `..1` <track_xyt<>> and `..2` <data.frame<>>.
```


tbl_time

```r
#> +    >> and `..2$nested.col` <tbl_time<
#> +      date     : date
#> +      observed : double
#> +      season   : double
#> +      trend    : double
#> +      remainder: double
```


comperes (cbind)

```r
#> Error: No common type for `..1` <longcr<>> and `..2` <data.frame<>>.
```


cutpointr (cbind)

```r
#> Error: No common type for `..1` <roc_cutpointr<>> and `..2` <data.frame<>>.
```


fable

```r
#> Error: No common type for `..1` <tbl_ts<>> and `..2` <tbl_df<>>.
```


tidygapminder (cbind)

```r
#> Error: No common type for `..1` <data.table<>> and `..2` <tbl_df<>>.
```

lionel-:
> All inputs are converted to frame prototypes, i.e. zero-column inputs. vec_restore() is called as part of this conversion. I'm not entirely happy about using restore for this because this isn't what it was designed for. I'm not sure how we could do better, except by calling back the R level [ method. Maybe that's a safer option.

One thing that bothers me is that we call `vec_restore()` without having taken the proxy first. If we take the proxy first, what happens when the proxy isn't a data frame? This is a case that isn't well thought out yet.

Maybe we should be prudent and use an ad hoc stopgap generic:

```r
experimental_frame_ptype <- function(x, ...) {
  UseMethod("experimental_frame_ptype")
}

experimental_frame_ptype.default <- function(x, ...) {
  x[0]
}
```

- This would handle classes like `grouped_df` by default.
- Classes that are not well behaved like sf data frames could override the behaviour to return a simplified class instead.

tests/testthat/test-type-dplyr.R
lionel-: My feeling is that we can't do better without a proper frame API. Unfortunately there's too much work left and uncertainty around it for the dplyr 1.0 release.

lionel-:
Now uses ad hoc `vec_cbind_frame_ptype()` generic to find the frame prototype.

DavisVaughan:
I guess you can also remove the hacks you used in the tests that kept `x[0]` with grouped-dfs from working right?

lionel-:
Following the dplyr dev fix? I don't think so because they need to work with CRAN dplyr too.

DavisVaughan:
I think it worked in CRAN dplyr?

DavisVaughan:
Confirming CRAN dplyr behavior

``` r
gdf <- dplyr::group_by(mtcars, cyl, am, vs)

dplyr::group_vars(gdf)
#> [1] "cyl" "am"  "vs"

dplyr::group_vars(gdf[0, ])
#> [1] "cyl" "am"  "vs"
```

<sup>Created on 2020-03-26 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

lionel-:
Tests are now updated.

hadley:
How urgent is this? Do we need it for dplyr 1.0.0? If not, I'd like to think about it a bit more.

lionel-:
This is needed for dplyr 1.0 if we want to go through with all the double dispatch changes and disable inheritance.

The only reason dplyr 1.0 and tidyr work right now is that grouped-df inherits from tbl-df, and the wrong ptype2 methods are used. Currently the ptype2 of a grouped-df and a df is a tbl-df.

I'm mostly worried about the `vec_cbind()` behaviour which isn't totally right, but the current behaviour isn't right either, and we probably can't do better in the dplyr release timeframe.

R/bind.R
```diff
@@ -167,3 +167,26 @@ as_df_row <- function(x, quiet = FALSE) {
...
+#'
+#' @param x A data frame.
+#' @inheritParams ellipsis::dots_empty
+#'
+#' @keywords internal
+#' @export
+vec_cbind_frame_ptype <- function(x, ...) {
+  UseMethod("vec_cbind_frame_ptype")
+}
+#' @export
```
hadley: Should we add a proactively add a sf method here? It's not correct in the long run, but seems reasonable since we know there are likely to be issue if we don't.

NEWS.md
```diff
@@ -1,6 +1,22 @@
 
 # vctrs (development version)
 
+* `vec_cbind()` now calls `vec_restore()` on inputs emptied of their
+  columns. This happens before taking the common type. This has
```
hadley:
```suggestion
  columns before computing the common type. This has
```

NEWS.md
```diff
@@ -1,6 +1,22 @@
...
+  devolve into simpler classes when the columns are subsetted
+  out. These classes are now always simplified by `vec_cbind()`.
+
+  For instance, column-binding a grouped data frame with a data frame
+  now produces a tibble (the simplified class of a grouped data
+  frame).
+
+* dplyr methods are now implemented for `vec_restore()`,
+  `vec_ptype2()`, and `vec_cast()`. The user-visible consequence (and
+  breaking change) is that combining a grouped data frame and a data
```
hadley: By combining you mean row-binding?

NEWS.md
```diff
@@ -1,6 +1,22 @@
 
 # vctrs (development version)
 
+* `vec_cbind()` now calls `vec_restore()` on inputs emptied of their
+  columns. This happens before taking the common type. This has
+  consequences for data frame classes with special columns that
+  devolve into simpler classes when the columns are subsetted
+  out. These classes are now always simplified by `vec_cbind()`.
+
+  For instance, column-binding a grouped data frame with a data frame
```
hadley: Thinking purely in terms of dplyr, this doesn't feel right to me, because it doesn't seem like adding columns to the RHS should change the grouping defined by the LHS. 

Crazy thought: could we say the the class of the output is defined purely by the LHS? (This might end up being more consistent with dplyr joins)

NEWS.md
lionel-: I meant any operation relying on a `vec_ptype2()` reduction, including row-binding.

R/bind.R
lionel-: Good idea.

hadley:
Can we revert this? I no longer think it's the right approach and it changes dplyr semantics in ways that are going to be tricky to fix.

hadley:
Oh actually it's not too tricky to fix, but it means I can no longer rely on `vec_slice()` as a low-level function. Hmm.

hadley:
It's not too tricky to fix in most places, but it is hard to fix in joins. I'm more convinced that this code shouldn't live in vctrs because managing groups really has to be done by dplyr.

lionel-:
If we revert this there will be a compatibility issue between tidyr and the next release of vctrs. I agree that they should be implemented in dplyr, they are conditionally registered to that effect.

Can you expand on the problems it causes in dplyr?

lionel-:
ah with #981 the breaking changes would be somewhat limited. Instead of returning a tibble when grouped-dfs are involved we'd return a data frame with a warning.

PR #954, opened 2020-03-25
Add support for numeric and character subscripts in `is.na<-.vctrs_vctr()`

### LINKED ISSUE #947: In is.na<-, value can be any index
`is.na<-.vctrs_vct` asserts value argument is a logical.

https://github.com/r-lib/vctrs/blob/69cff44aa55791a5c7275b8a4575be8ef8550c61/R/type-vctr.R#L640

However, in base::R merge.data.frame it may be an index of values to set to NA. See (for example) this section of the merge.data.frame base R function:

```{r}
        if (all.x) {
            zap <- (lxy + 1L):(lxy + nxx)
            for (i in seq_along(y)) {
                if (is.matrix(y[[1]])) 
                  y[[1]][zap, ] <- NA
                else is.na(y[[i]]) <- zap
            }
        }
```

DavisVaughan:
Closes #947 

`vec_assign()` automatically calls `vec_as_location()` internally, so the implementation is pretty trivial now.

Note that we had explicitly forbidden this, but I think with `vec_as_location()` being well-defined now, this makes sense

```r
`is.na<-.vctrs_vctr` <- function(x, value) {
  # No support for other arguments than logical for now,
  # even if base R is more lenient here.
  vec_assert(value, logical())

  vec_slice(x, value) <- vec_init(x)
  x
}
```

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,6 +1,9 @@
 
 # vctrs (development version)
 
+* The `is.na<-()` method for `vctrs_vctr` now supports numeric and
+  character subscripts to indicate where to insert missing values (#947). 
+
 * `vec_rbind()` and `vec_c()` with data frame inputs now consistently
   preserve the names of list-columns, df-columns, and matrix-columns
   (#689).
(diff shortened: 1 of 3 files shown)


DavisVaughan:
@lionel-  👍 it was there already! https://github.com/r-lib/vctrs/pull/954/files#diff-f5bef5a3e244ba1f60d5ac5f45d5cf0bR132

PR #955, opened 2020-03-26
What is a scalar?

krlmlr:
Follow-up to https://github.com/tidyverse/tibble/pull/747#issuecomment-603725815.

### DIFF
--- a/R/assert.R
+++ b/R/assert.R
@@ -16,7 +16,8 @@
 #'
 #' - [typeof()] is atomic: `"logical"`, `"integer"`, `"double"`,
 #'   `"complex"`, `"character"`, `"raw"`
-#' - the object is a [data.frame]
+#' - the object is a [data.frame], the `"class"` attribute must contain
+#'   `"data.frame"` as the *last* element
 #' - `typeof()` is `"list"`, and one of:
 #'     - the object is a bare `"list"` without a `"class"` attribute,
 #'       see [is_bare_list()]
@@ -32,6 +33,9 @@
 #'   treated as scalars.
 #' - Support for S4 vectors is currently limited to objects that inherit from
 #'   an atomic type.
+#' - Subclasses of [data.frame] that *append* their class to the `"class"`
+#'   attribute are not treated as vectors. If you inherit from an S3 class,
+#'   always prepend your class to the `"class"` attribute for correct dispatch.
 #'
 #' @section Error types:
 #'
--- a/R/assert.R
+++ b/R/assert.R
@@ -10,29 +10,32 @@
 #'
 #' @section Scalars and vectors:
 #'
-#' Informally, a vector is a collection that makes sense to use as column
-#' in a data frame.
-#' An object is a vector if one of the following conditions hold:
+#' Informally, a vector is a collection that makes sense to use as
+#' column in a data frame. An object is a vector if one of the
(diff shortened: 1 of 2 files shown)


R/assert.R
```diff
@@ -2,13 +2,41 @@
...
+#'
+#' - [typeof()] is atomic: `"logical"`, `"integer"`, `"double"`,
+#'   `"complex"`, `"character"`, `"raw"`
+#' - `typeof()` is `"list"`, and one of:
+#'     - the object is a bare `"list"` without a `"class"` attribute,
+#'       see [is_bare_list()]
+#'     - the object is a [data.frame]
+#'     - the `"class"` attribute contains `"list"`, i.e., `inherits(x, "list")`
+#'       is `TRUE`
+#'     - the class implements a [vec_proxy()] method
```
lionel-: This should be moved one level. You can have S3 environments implementing proxy for instance. Non-vector S4 objects will follow the same approach.

R/assert.R
```diff
@@ -2,13 +2,41 @@
...
+#' @section Scalars and vectors:
+# From vec_base_typeof():
+#' An object is a vector if one of the following conditions hold:
+#'
+#' - [typeof()] is atomic: `"logical"`, `"integer"`, `"double"`,
+#'   `"complex"`, `"character"`, `"raw"`
+#' - `typeof()` is `"list"`, and one of:
+#'     - the object is a bare `"list"` without a `"class"` attribute,
+#'       see [is_bare_list()]
+#'     - the object is a [data.frame]
```
lionel-: I think it's confusing to have data frames as a sub-bullet of list. Data frames are their own type in vctrs.

R/assert.R
```diff
@@ -2,13 +2,41 @@
...
+#'     - the `"class"` attribute contains `"list"`, i.e., `inherits(x, "list")`
+#'       is `TRUE`
+#'     - the class implements a [vec_proxy()] method
+#'
+#' Otherwise an object is inherently *scalar* and cannot be used as a vector.
```
lionel-: Should mention the main criterion of being a vector is to be a collection that makes sense to use as column in a data frame.

R/assert.R
```diff
@@ -2,13 +2,41 @@
...
+#' `NULL` is not a vector.
+#' Objects of mode [expression] have vector-like behavior but are currently
```
lionel-: I'd move these details to a separate paragraph since they are not important. Also paragraphs should be reflowed.

R/assert.R
```diff
@@ -2,13 +2,41 @@
...
-#'
-#' * `vec_assert()` throws an error when the input doesn't conform.
+#' * `vec_is()` is a predicate that checks if its input is a vector that
+#'   conforms to a prototype and/or a size.
+#'
+#' * `vec_assert()` throws an error when the input is not a vector or
+#'   doesn't conform.
+#'
+#' @section Scalars and vectors:
+# From vec_base_typeof():
```
lionel-: `vec_base_typeof()` is not quoted. But I don't think it should be mentioned here.

R/assert.R
krlmlr: Done above.

krlmlr:
I'm linking to `?vec_is` from the tibble blog post, would be great to have this merged by then.

PR #956, opened 2020-03-26
Improve user-friendliness of cast error messages

lionel-:
Part of #922.

* Cast errors mention "convert" instead of "cast".
* Lossy cast errors explain that the conversion loses precision.


R/conditions.R
```diff
@@ -288,7 +288,7 @@ conditionMessage.vctrs_error_cast_lossy <- function(c) {
 cnd_header.vctrs_error_cast_lossy <- function(cnd, ...) {
   x_label <- format_arg_label(vec_ptype_full(cnd$x), cnd$x_arg)
   to_label <- format_arg_label(vec_ptype_full(cnd$to), cnd$to_arg)
-  glue::glue("Lossy cast from {x_label} to {to_label}.")
+  glue::glue("Can't convert from {x_label} to {to_label} because this loses precision.")
```
hadley:
```suggestion
  glue::glue("Can't convert from {x_label} to {to_label} due to loss of precision.")
```
?

Issue #841, opened 2020-02-19
Better message for `vec_recycle()` when using `args_empty`


DavisVaughan:
```r
library(vctrs)

# default ptype2 message
vec_ptype2(1, "x")
#> Error: No common type for `x` <double> and `y` <character>.

# ptype2 has this nice message when args are empty
vec_ptype2(1, "x", x_arg = "", y_arg = "")
#> Error: No common type for <double> and <character>.

# default message
vec_recycle(c(1, 2), 3L)
#> Error: `x` can't be recycled to size 3.
#> x It must be size 3 or 1, not 2.

# I call vec_recycle() from C with `args_empty`. Should have a better message?
vec_recycle(c(1, 2), 3L, x_arg = "")
#> Error: `` can't be recycled to size 3.
#> x It must be size 3 or 1, not 2.
```

PR #957, opened 2020-03-26
Remove default args

lionel-:
I think the default `arg` values end up to be misleading for end users because they refer to internal details of internal functions.

This brings up the issue of how to refer to arguments in error messages. `vec_recycle()` now uses the same approach as in subscript errors and uses a general noun "Input".

Before:

```r
#> Error: `x` can't be recycled to size 4.
```

After:

```r
#> Error: Input can't be recycled to size 4.
```

If `arg = "foo"` is supplied, this is:

```r
#> Error: Input `foo` can't be recycled to size 4.
```

We should standardise on a user-friendly noun for referring to arguments. I thought "Input" is shorter and less jargony than "Argument".

cc @batpigandme @krlmlr

### DIFF
--- a/R/cast.R
+++ b/R/cast.R
@@ -119,14 +119,14 @@
 #'
 #' # Cast to common type
 #' vec_cast_common(factor("a"), factor(c("a", "b")))
-vec_cast <- function(x, to, ..., x_arg = "x", to_arg = "to") {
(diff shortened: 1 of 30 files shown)


DavisVaughan:
Does this close https://github.com/r-lib/vctrs/issues/841?

lionel-:
Looks like it!

PR #960, opened 2020-03-26
Remove unspecified methods

### LINKED ISSUE #952: vctrs_unspecified redundant method in type-integer64.R?
I have some methods to catch cases for `vctrs_unspecified` in my own code, for example something like:

```
#' @method vec_ptype2.myclass vctrs_unspecified
#' @export
vec_ptype2.myclass.vctrs_unspecified <- function(x, y, ...) x
```

I'm coming to the conclusion that it is impossible to get to these methods as unspecified cases are handled internally. If this assumption is correct (I'm not sure!) then, this represents a redundant line of code:

https://github.com/r-lib/vctrs/blob/71aedcf42d8fdedc2e3f344af18813d9cf71acff/R/type-integer64.R#L48

as evidenced by the fact that there's no [code coverage](https://codecov.io/gh/r-lib/vctrs/src/master/R/type-integer64.R#L48) for this line, even though tests should cover the case here:

https://github.com/r-lib/vctrs/blob/71aedcf42d8fdedc2e3f344af18813d9cf71acff/tests/testthat/test-type-integer64.R#L39

I'd like to know either way as it would clarify the case for unspecified for me.

lionel-:
Closes #952

This removes the unspecified method-generic without deprecation. rray is the only package that registers methods for it, and removing it doesn't seem to cause CMD check issues (rray already has other issues though).


hadley:
I guess we never mentioned this in the vignettes. Ooops!

lionel-:
IIRC we removed the unspecified methods from the vignette when we added internal handling, but we never removed the methods in our own code.

DavisVaughan:
``` r
library(vctrs)

vec_ptype2(array(1:5, c(1, 5)), unspecified())
#>      [,1] [,2] [,3] [,4] [,5]

vec_ptype2(unspecified(), array(1:5, c(1, 5)))
#> Error: No common type for `x` <vctrs_unspecified> and `y` <integer[,5]>.
```

<sup>Created on 2020-03-26 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

Something isn't quite right with shaped objects. When the RHS is `unspecified()`, `vec_default_ptype2()` catches it

lionel-:
Good catch, I'll add a unit test for this case and wait until the double dispatch update is merged.

Issue #794, opened 2020-02-05
Missing cast method: `vec_cast.character.ordered()`


DavisVaughan:
We have `vec_cast.character.factor()`

DavisVaughan:
I just realized that `vec_cast.character.factor()` should cover this through inheritance

lionel-:
Since cast should now mirror ptype2, and we're in the process of forbidding ptype2 inheritance, I think cast can't use inheritance either.

DavisVaughan:
Fixed as a part of #963 

```r
> vec_cast(factor(ordered = TRUE), "x")
character(0)
```

PR #963, opened 2020-03-27
Fix `vec_cast()` dispatch

### LINKED ISSUE #710: Avoid inheritance in double dispatch

### LINKED ISSUE #753: Prevent `vec_cast()` methods from being inherited

lionel-:
Branched from #951.

Closes #710 and closes #753.

Uses internal double dispatch in `vec_cast()`. This prevents inheritance and the need for default methods. All classes currently requiring special handling are special-cased in `vec_default_cast()`.


NEWS.md
```diff
@@ -17,16 +17,16 @@
...
-  implement. vctrs takes care of implementing the default and
-  unspecified methods. Methods no longer have to worry about order of
-  dispatch, for instance it is no longer needed to supply a second
+* Double-dispatch methods for `vec_ptype2()` and `vec_cast()` are no
+  longer inherited (#710, #753). Class implementers must implement one
+  set of methods for each compatible class. This requires a bit more
+  work but is safer. See the S3 dispatch section of `?vec_ptype2` for
+  more information.
+
+* Double-dispatch methods for `vec_ptype2()` and `vec_cast()` are now
```
hadley: If you have implemented, you can now delete these methods.

R/cast.R
```diff
@@ -165,11 +160,28 @@ vec_coercible_cast <- function(x, to, ..., x_arg = "", to_arg = "") {
 #' @inheritParams vec_cast
 #' @export
 vec_default_cast <- function(x, to, x_arg = "", to_arg = "") {
+  if (is_bare_list(to)) {
```
hadley: Does this have any performance implications?

R/type-vctr.R
```diff
@@ -115,25 +115,23 @@ vec_ptype2.vctrs_vctr <- function(x, y, ..., x_arg = "", y_arg = "") {
...
+vctr_cast <- function(x, to, ..., x_arg = "", to_arg = "") {
   # These are not strictly necessary, but make bootstrapping a new class
   # a bit simpler
   if (is.object(x)) {
-    attr_x <- utils::modifyList(attributes(x), list(names = NULL))
-    attr_y <- utils::modifyList(attributes(to), list(names = NULL))
-
-    if (identical(attr_x, attr_y)) {
-      return(x)
+    if (is_same_type(x, to)) {
```
hadley: This is a behaviour change, right? Is it worth a news bullet?

R/type-factor.R
```diff
@@ -146,17 +144,47 @@ vec_cast.factor.factor <- function(x, to, ..., x_arg = "", to_arg = "") {
...
+vec_cast.factor.character <-function(x, to, ...) {
+  fct_cast(x, to, ...)
+}
+
 #' @export
 #' @method vec_cast.character factor
-vec_cast.character.factor <- function(x, to, ...) as.character(x)
+vec_cast.character.factor <- function(x, to, ...) {
+  as.character(x)
+}
```
DavisVaughan: Might temporarily need `vec_cast.character.ordered()` to _technically_ be correct. But I think when we restrict `vec_cast()` that will get removed too? https://github.com/r-lib/vctrs/issues/794

DavisVaughan: And maybe `vec_cast.ordered.character`

R/type-vctr.R
```diff
@@ -290,7 +288,11 @@ as.character.vctrs_vctr <- function(x, ...) {
 
 #' @export
 as.list.vctrs_vctr <- function(x, ...) {
-  vec_cast(x, list())
+  if (vec_is_list(x)) {
+    x
```
DavisVaughan: So a classed list-like `vctrs_vctr` object would be returned as is, with the class still attached? I feel like I would expect a bare list to be returned, like `vec_chop()` gives?

R/cast.R
DavisVaughan: I imagine this will get removed when we restrict `vec_cast` to only common types?

Is that also true for the `vctr_cast()` and `vec_cast_to_list_of()` calls below?

R/type-vctr.R
lionel-: Definitely.

R/cast.R
lionel-: Yes this branch and the list-of branch are temporary.

I'd like to remove the `vctrs_vctr` branch but it's a bit more tricky. The class tries to be helpful in more cases in a way that I don't fully understand.

R/type-vctr.R
lionel-: This is a slight change, I don't think it's worth a bullet since this is rather esoteric.

hadley: Isn't it the root of a bunch of the "can't cast <x> to <x>" issues?

lionel-: I don't think so, these have been fixed already.

R/type-factor.R
lionel-: oh this works because this is handled at C level right? I was puzzled for a while. I'll add "never-called" indicative methods.

PR #965, opened 2020-03-27
Restrict `vec_cast()` methods to common types

### LINKED ISSUE #606: Make `vec_cast()` a directional coercion generic

### LINKED ISSUE #741: Remove `vec_cast.<datelike>.double()` methods
Part of 0.3.0, since it is breaking an existing cast method.

It was determined that casting dates/datetimes to and from integer/double is too relaxed here https://github.com/r-lib/vctrs/pull/728#issuecomment-570654648

- `vec_cast.Date.double()`
- `vec_cast.POSIXct.double()`
- `vec_cast.POSIXlt.double()`

Supersedes #728 and #715

lionel-:
Closes #606
Closes #741


R/cast.R
```diff
@@ -160,22 +160,14 @@ vec_coercible_cast <- function(x, to, ..., x_arg = "", to_arg = "") {
...
   if (is_asis(to)) {
     return(vec_cast_to_asis(x, to, x_arg = x_arg, to_arg = to_arg))
   }
 
-  if (inherits_only(to, c("vctrs_list_of", "vctrs_vctr"))) {
-    return(vec_cast_to_list_of(x, to, x_arg = x_arg, to_arg = to_arg))
-  }
-
-  if (inherits(to, "vctrs_vctr")) {
+  if (inherits(to, "vctrs_vctr") && !inherits(to, "vctrs_rcrd")) {
```
DavisVaughan: So `vctrs_list_of` goes through `vctr_cast()`?

tests/testthat/test-type-bare.R
```diff
@@ -268,44 +291,53 @@ test_that("complex is not coercible to logical", {
...
-  expect_equal(vec_cast(mat(list(NA)), to_mat), exp_mat)
+
+  # These used to be allowed
+  expect_error(vec_cast(mat(lgl(NA)), to_mat), class = "vctrs_error_incompatible_cast")
+  expect_error(vec_cast(mat(int(NA)), to_mat), class = "vctrs_error_incompatible_cast")
+  expect_error(vec_cast(mat(dbl(NA)), to_mat), class = "vctrs_error_incompatible_cast")
+  expect_error(vec_cast(mat(list(NA)), to_mat), class = "vctrs_error_incompatible_cast")
 })
 
 test_that("difftime gets special treatment", {
```
DavisVaughan: Maybe change the title here since it was the only thing in this test?

tests/testthat/error/test-rep.txt
```diff
@@ -3,8 +3,7 @@
 =============================
 
 > vec_rep(1, "x")
-Error: Can't convert from `times` <character> to <integer> due to loss of precision.
-* Locations: 1
+Error: Can't convert from `times` <character> to <integer>.
```
hadley: Not related to this PR but shouldn't it be "Can't convert `times` from <character> to <integer>."?

tests/testthat/test-type-date-time.R
```diff
@@ -173,21 +175,24 @@ test_that("safe casts work as expected", {
...
-  expect_equal(vec_cast("1970-02-01", datetime_l), datetime_l)
   expect_equal(vec_cast(datetime_c, datetime_l), datetime_l)
   expect_equal(vec_cast(datetime_l, datetime_l), datetime_l)
   expect_equal(vec_cast(as.Date(datetime_l), datetime_l), datetime_l)
-  expect_equal(vec_cast(list(datetime_l), datetime_l), datetime_l)
   expect_error(vec_cast(raw(), datetime_l), class = "vctrs_error_incompatible_cast")
+
+  # These used to be allowed
+  expect_error(vec_cast(2678400, datetime_c), class = "vctrs_error_incompatible_cast")
+  expect_error(vec_cast("1970-02-01", datetime_c), class = "vctrs_error_incompatible_cast")
```
hadley: This change makes me slightly nervous because `x < "2000-01-01"` is a common operation. Maybe we should allow ISO8601 dates-times in strings? That feels a bit similar to allowing "double" 1 to go into a integer.

lionel-: This doesn't work with CRAN vctrs because dates and characters don't have a common type. I think it'd make sense to add a common type, and only allow casting towards dates. I have opened an issue: https://github.com/r-lib/vctrs/issues/967

tests/testthat/error/test-rep.txt
lionel-: Noted in https://github.com/r-lib/vctrs/issues/922#issuecomment-605417304

R/cast.R
lionel-: yeah but don't ask me too much about what this means 😄 

I think we need to review the convenience implementations for `vctrs_vctr`. Opened issue #968.

PR #976, opened 2020-03-29
Fix ordered cast methods

DavisVaughan:
Part of #965 

Two ordered factor methods were missing. They normally would only get called if we have a lossy cast. We didn't get errors before because they just fell through to the default method, which errored for us.

Also fixes a registration problem with the `vec_cast.character.ordered` method


R/type-factor.R
```diff
@@ -180,10 +156,42 @@ vec_cast.ordered <- function(x, to, ...) {
...
+vec_cast.ordered.character <-function(x, to, ...) {
+  fct_cast(x, to, ...)
+}
+#' @export
+#' @method vec_cast.character ordered
+vec_cast.character.ordered <- function(x, to, ...) {
+  stop_native_implementation("vec_cast.character.ordered")
+}
+
+fct_cast <- function(x, to, ..., x_arg = "", to_arg = "") {
```
lionel-: I think the right place for this kind of function is just below the exported method-generic.

We might want to split this one into `fct_cast()` and `ord_cast()`. I think `fct_cast()` should mean "cast any subclasses of factors to factor". In the same way, `df_cast()` casts any subclasses of data frames to data frame.

DavisVaughan:
@lionel- FYI it looks like the fct->ordered and ord->factor fast path was missing from `vec_cast_dispatch()`, so I added it here

Issue #969, opened 2020-03-28
Switch to `vec_cast()`


lionel-:
- Throw incompatible type errors for unimplemented `vec_cast()` methods.
- Use `vec_cast()` in `vec_assign()`.

DavisVaughan:
* Use `vec_cast()` in `vec_as_location()` https://github.com/r-lib/vctrs/blob/1a06f1fe6aa3732b5629483b5ea045d8e518647c/src/subscript-loc.c#L393

lionel-:
Include `action` field in condition object so the generated message can be different if thrown from ptype2 (action = combine?) and cast (action = convert). Same approach as for subscript errors.

PR #983, opened 2020-04-01
Throw type errors from `vec_cast()`

DavisVaughan:
Part of #969 

- Removes all usage of `stop_incompatible_cast()` in favor of `stop_incompatible_type()`

- Adds an `action` argument to `stop_incompatible_type()` to customize the error depending on whether it is thrown from ptype2 or cast

   - In the process, this changed the current `vec_ptype2()` message

Questions:

- @lionel-, you mentioned throwing incompatible type errors for _unimplemented_ cast methods, but I feel like it makes sense to throw them for methods that are implemented as well if they just throw the error. That is what I did here.

- Related to the above point, we no longer use `stop_incompatible_cast()` anywhere. Should it be soft-deprecated?

- Right now this requires exporting this new `action` argument in `stop_incompatible_type()`. Would it be better to keep `stop_incompatible_type()` untouched and add internal `stop_incompatible_type_combine()` and `stop_incompatible_type_convert()` functions? `stop_incompatible_type()` would just call `stop_incompatible_type_combine()` but that way we don't export the new argument.


lionel-:
> you mentioned throwing incompatible type errors for unimplemented cast methods, but I feel like it makes sense to throw them for methods that are implemented as well if they just throw the error.

Agreed. Throwing manually is needed because a successful dispatch (based on classes) does not necessarily mean that the objects have a common type.

> we no longer use stop_incompatible_cast() anywhere. Should it be soft-deprecated?

> Right now this requires exporting this new action argument in stop_incompatible_type(). Would it be better to keep stop_incompatible_type() untouched and add internal stop_incompatible_type_combine() and stop_incompatible_type_convert() functions? 

My first inclination would be to use internal helpers, following the subscript mechanism where the `action` field is an interface for condition handlers. But maybe in this case it's better to expose it?

oh, what if `stop_incompatible_cast()` called the internal helper with `action = "convert"`? This way we don't change the UI, just the condition class, and we get the proper action by default. That would probably be clearer for class implementers.

DavisVaughan:
@lionel- this has gotten a little simpler now with your suggestions.

- `stop_incompatible_type()` calls `stop_incompatible_type_combine()`

- `stop_incompatible_cast()` calls `stop_incompatible_type_convert()`

As you said, this means the external UI doesn't change. It also means I don't have to change any internal uses of `stop_incompatible_type()` and `stop_incompatible_cast()`.

I'm slightly worried about this breaking packages that use tests expecting a `vctrs_error_incompatible_cast` error, which instead will be `vctrs_error_incompatible_type`. It would be an easy fix though.

hadley:
Could we make `vctrs_error_incompatible_cast` a subclass of `vctrs_error_incompatible_type`? (in the interests of reducing the number of breaking changes in this release)

lionel-:
A CRAN search reveals very few hits (glue, rray, hardhat, ipaddress). I think it's worth making the class hierarchy simpler.

DavisVaughan:
Noting that this PR removes `"from"` from the cast message as suggested here https://github.com/r-lib/vctrs/issues/922#issuecomment-605446861
And the ptype2 message is aligned with the suggestion of using "combine" here https://github.com/r-lib/vctrs/issues/922#issuecomment-605440214

hadley:
Yes, but those are mostly our packages, meaning that we need to scramble to get releases out; I'd prefer that we keep focussed on more important issues right now.

lionel-:
AFAIK rray and hardhat are already broken. So that leaves glue, and only the tests are broken, not the package, so there would be no rush for a release. And the release fix for glue would be easy.

I still recommend removing the class because I think keeping the error hierarchy clean is important, but I won't insist further. If we keep the class it should be a super class so that testthat users don't add new `class = ` for it.

DavisVaughan:
I just ran a full check on hardhat with dev vctrs and it doesn't break right now. But yet rray doesn't matter here.

DavisVaughan:
Can we please get a final say on the decision here? Summary is:

Breakage will occur in:
- [hardhat](https://github.com/tidymodels/hardhat/blob/0bed9ff404f76471a672d3e4161b1aaa39bbff41/tests/testthat/test-scream.R#L52)
- [glue](https://github.com/tidyverse/glue/blob/7131fc394cb36241d59f0ac46cccbaa726844f83/tests/testthat/test-vctrs.R#L47-L64)
- [ipaddress](https://github.com/davidchall/ipaddress/blob/4a271891b74e528257373ad71ea73d33b5159737/tests/testthat/test-ip_interface.R#L35)

Breakage is only in tests, not in functionality. A patch release would be required, but would not have to be rushed out.

hadley:
Aren't the tests going to cause R CMD check to fail? And hence require a CRAN release?

DavisVaughan:
Yes a release will be required because of that. I think what @lionel- was trying to say is that it won't break any user facing code, so we wouldn't have to rush a release for that

hadley:
Thinking on it more, we are 3-5 weeks out from a dplyr release, so I think it's fine to break this interface.

PR #986, opened 2020-04-02
Return immediately if restoring to the same reference

### LINKED ISSUE #984: vec_rbind() performance regression
``` r
library(vctrs)
library(tibble)
#> Warning: package 'tibble' was built under R version 3.6.2

x <- rep(list(data.frame(x = 1)), 10000)

fn <- function(x) {
  vec_rbind(!!!x)
}

# CRAN
bench::mark(fn(x), iterations = 10)
#> # A tibble: 1 x 6
#> 1 fn(x)        42.7ms   52.6ms      18.5     198KB     20.3

# Dev
bench::mark(fn(x), iterations = 10)
#> # A tibble: 1 x 6
#> 1 fn(x)         269ms    347ms      2.87     764MB     56.2
```

Not sure why yet

DavisVaughan:
Fixes #984 

This problem was really difficult to track down.

Right now, `vec_c()` and `vec_rbind()` do the right thing with row names because of this forced shallow duplication in `df_assign()`.

https://github.com/r-lib/vctrs/blob/fd249275c79ccd01910c947f63296ea9c5e6aa99/src/slice-assign.c#L341

The problem is that this is very slow, and should be unnecessary since `out` is a completely new object that we are assigning to, so we shouldn't have to duplicate the columns.

However, naively replacing that line with `r_maybe_duplicate()` or removing it entirely results in this problem where we lose row names:

```r
library(vctrs)
library(tibble)

df_x <- new_data_frame(list(x = 1:3), row.names = letters[1:3])
df_y <- new_data_frame(list(x = 4L), row.names = "d")

x <- tibble(dfcol = df_x)
y <- tibble(dfcol = df_y)

vec_rbind(x, y)$dfcol
#> New names:
#> * `` -> ...4
#> New names:
#> * `` -> ...1
#> * `` -> ...2
#> * `` -> ...3
#>   x
#> 1 1
#> 2 2
#> 3 3
#> 4 4
```

We lose them at the `vec_restore()` step here
https://github.com/r-lib/vctrs/blob/fd249275c79ccd01910c947f63296ea9c5e6aa99/src/slice-assign.c#L344

Note that in the call to `vec_restore()`, `assigned` and `out_elt` _are the same object_. This eventually passes through to `vec_restore_default()`. So when we get to this part of `vec_restore_default()` posted below we clear the attributes on `x` (`assigned`) by replacing them with `attrib`, which are the attributes of `to` (`out_elt`) but without a few of the core attributes, like the class. But because `x` and `to` are the same object, _this reset `to` as well_. So when we do the `is_data_frame(to)` check later on to decide if we should add row names on, it fails because `to` no longer has a class.

https://github.com/r-lib/vctrs/blob/fd249275c79ccd01910c947f63296ea9c5e6aa99/src/proxy-restore.c#L72-L82

I've fixed this for now by returning early from `vec_restore_default()` if we encounter identical objects. @lionel- does this seem like a reasonable fix?

### DIFF
--- a/tests/testthat/helper-memory.R
+++ b/tests/testthat/helper-memory.R
@@ -1,0 +1,11 @@
+maybe_referenced <- function(x) {
+  .Call(vctrs_maybe_referenced, x)
+}
+
+maybe_referenced_col <- function(x, i) {
+  .Call(vctrs_maybe_referenced_col, x, i)
+}
+
+new_df_unreferenced_col <- function() {
+  .Call(vctrs_new_df_unreferenced_col)
+}
--- a/tests/testthat/helper-memory.R
+++ b/tests/testthat/helper-memory.R
@@ -1,7 +1,3 @@
-maybe_referenced <- function(x) {
-  .Call(vctrs_maybe_referenced, x)
-}
-
 maybe_referenced_col <- function(x, i) {
   .Call(vctrs_maybe_referenced_col, x, i)
 }
--- a/src/init.c
+++ b/src/init.c
@@ -107,7 +107,6 @@ extern SEXP vctrs_assign_params(SEXP, SEXP, SEXP, SEXP);
 extern SEXP vctrs_has_dim(SEXP);
 extern SEXP vctrs_rep(SEXP, SEXP);
 extern SEXP vctrs_rep_each(SEXP, SEXP);
-extern SEXP vctrs_maybe_referenced(SEXP);
 extern SEXP vctrs_maybe_referenced_col(SEXP, SEXP);
 extern SEXP vctrs_new_df_unreferenced_col();
 
@@ -237,7 +236,6 @@ static const R_CallMethodDef CallEntries[] = {
   {"vctrs_has_dim",                    (DL_FUNC) &vctrs_has_dim, 1},
   {"vctrs_rep",                        (DL_FUNC) &vctrs_rep, 2},
   {"vctrs_rep_each",                   (DL_FUNC) &vctrs_rep_each, 2},
-  {"vctrs_maybe_referenced",           (DL_FUNC) &vctrs_maybe_referenced, 1},
   {"vctrs_maybe_referenced_col",       (DL_FUNC) &vctrs_maybe_referenced_col, 2},
   {"vctrs_new_df_unreferenced_col",    (DL_FUNC) &vctrs_new_df_unreferenced_col, 0},
   {NULL, NULL, 0}
--- a/src/init.c
+++ b/src/init.c
@@ -107,6 +107,9 @@ extern SEXP vctrs_assign_params(SEXP, SEXP, SEXP, SEXP);
 extern SEXP vctrs_has_dim(SEXP);
 extern SEXP vctrs_rep(SEXP, SEXP);
 extern SEXP vctrs_rep_each(SEXP, SEXP);
+extern SEXP vctrs_maybe_referenced(SEXP);
+extern SEXP vctrs_maybe_referenced_col(SEXP, SEXP);
+extern SEXP vctrs_new_df_unreferenced_col();
 
 // Maturing
 // In the public header
@@ -234,6 +237,9 @@ static const R_CallMethodDef CallEntries[] = {
   {"vctrs_has_dim",                    (DL_FUNC) &vctrs_has_dim, 1},
   {"vctrs_rep",                        (DL_FUNC) &vctrs_rep, 2},
   {"vctrs_rep_each",                   (DL_FUNC) &vctrs_rep_each, 2},
+  {"vctrs_maybe_referenced",           (DL_FUNC) &vctrs_maybe_referenced, 1},
+  {"vctrs_maybe_referenced_col",       (DL_FUNC) &vctrs_maybe_referenced_col, 2},
+  {"vctrs_new_df_unreferenced_col",    (DL_FUNC) &vctrs_new_df_unreferenced_col, 0},
   {NULL, NULL, 0}
 };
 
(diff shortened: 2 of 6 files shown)


lionel-:
It doesn't seem reasonable to just mutate the column proxies. Is there really a guarantee that we own the columns?

DavisVaughan:
I can add a `r_maybe_duplicate()` back in but we still need the other fix in `vec_restore_default()`

lionel-:
maybe-duplicate doesn't seem appropriate here because list components might not have their namedness bumped.

DavisVaughan:
Oh also `vec_proxy_assign_opts()` is guaranteed to `r_maybe_duplicate()` here already

lionel-:
Regarding the restore fix, it is rather ad hoc right?

DavisVaughan:
Yea, the other thing we could do is record `is_data_frame(to)` before we call `SET_ATTRIB(x, attrib)` to ensure we have all the required information about `to` before we start modifying `x`(which might modify `to` in the process as seen here). But even that doesn't feel great

lionel-:
Given the namedness limits, maybe we need some other way of signalling to df-assign that we recursively own the data frame columns.

DavisVaughan:
TBH I don't see how it can cause a problem here. Is there any way to construct an example we could use as a test?

lionel-:
We use `vec_assign()` in `vec_rbind()` and `vec_c()` etc with freshly initialised output vectors. That's not the case in general, especially with the C callable of vec-assign.

It should not be difficult to construct an example where calling `vec_assign()` causes mutation with this patch. `mtcars` is an example of a data frame with unNAMED columns.

lionel-:
Maybe `vec_proxy()` should always bump namedness? Though that wouldn't help the performance issue. Maybe we should (partly?) reconsider #842?

DavisVaughan:
I think we are ok actually. TIL that `Rf_shallow_duplicate()` loops over elements of lists and ensures that they are at max namedness. We call this on the data frame in `df_assign()`.

Here is `Rf_shallow_duplicate()`:
https://github.com/wch/r-source/blob/bac4cd3013ead1379e20127d056ee036278b47ff/src/main/duplicate.c#L151-L158

Which calls `duplicate1(s, FALSE)`. For lists, that loops over the list calling `duplicate_child(elt, FALSE)`
https://github.com/wch/r-source/blob/bac4cd3013ead1379e20127d056ee036278b47ff/src/main/duplicate.c#L333-L341

Which immediately calls `lazy_duplicate()`
https://github.com/wch/r-source/blob/bac4cd3013ead1379e20127d056ee036278b47ff/src/main/duplicate.c#L204-L208

And the job of `lazy_duplicate()` is to ensure max namedness.
https://github.com/wch/r-source/blob/bac4cd3013ead1379e20127d056ee036278b47ff/src/main/duplicate.c#L170-L196

DavisVaughan:
That is why this works, I think.

``` r
library(vctrs)

maybe_referenced <- cbuild::source_function("
  SEXP maybe_referenced(SEXP x) {
    return Rf_ScalarLogical(MAYBE_REFERENCED(x));
  }
")

maybe_referenced_col <- cbuild::source_function("
  SEXP maybe_referenced_col(SEXP x) {
    return Rf_ScalarLogical(MAYBE_REFERENCED(VECTOR_ELT(x, 0)));
  }
")

maybe_referenced(mtcars)
#> [1] TRUE
maybe_referenced_col(mtcars)
#> [1] FALSE

new_row <- new_data_frame(setNames(rep(list(1), ncol(mtcars)), colnames(mtcars)))
new_row
#>   mpg cyl disp hp drat wt qsec vs am gear carb
#> 1   1   1    1  1    1  1    1  1  1    1    1

assigned <- vec_assign(mtcars, 1, new_row)

head(assigned, 2)
#>               mpg cyl disp  hp drat    wt  qsec vs am gear carb
#> Mazda RX4       1   1    1   1  1.0 1.000  1.00  1  1    1    1
#> Mazda RX4 Wag  21   6  160 110  3.9 2.875 17.02  0  1    4    4

head(mtcars, 2)
#>               mpg cyl disp  hp drat    wt  qsec vs am gear carb
#> Mazda RX4      21   6  160 110  3.9 2.620 16.46  0  1    4    4
#> Mazda RX4 Wag  21   6  160 110  3.9 2.875 17.02  0  1    4    4

maybe_referenced(mtcars)
#> [1] TRUE
maybe_referenced_col(mtcars)
#> [1] TRUE
```

<sup>Created on 2020-04-02 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

DavisVaughan:
Even easier to see:

```r
maybe_referenced(mtcars)
#> [1] TRUE
maybe_referenced_col(mtcars)
#> [1] FALSE

x <- rlang::duplicate(mtcars, shallow = TRUE)

maybe_referenced_col(mtcars)
#> [1] TRUE
```

lionel-:
But then how does this solve the performance issue?

DavisVaughan:
Because in `vec_c()` and `vec_rbind()` we give `df_assign()` a fresh df so `r_maybe_duplicate()` never calls `Rf_shallow_duplicate()`

lionel-:
oh nice! Can you add a comment about this please. Also perhaps include some tests to check for mutation?

I guess the restore fix is good enough, though I'd prefer something more general. Maybe your last suggestion is better.

DavisVaughan:
Is there an easy way to construct a list with columns that aren't named?

lionel-:
I don't think so, the easiest way is to do it from C. This is where most of those data frames will come from, for example:

```r
df <- vctrs::vec_split(mtcars[1:4, 1:3], mtcars$cyl[1:4])
lobstr::sxp(df)
```

tests/testthat/helper-memory.R
```diff
@@ -0,0 +1,11 @@
+maybe_referenced <- function(x) {
+  .Call(vctrs_maybe_referenced, x)
```
lionel-, line 2: This will always be referenced because the input is binded in the execution env of the closure wrapper. To work around this you need to defuse `x` with `ensym()` and pass the caller env along with the defused symbol to the native routine, and then use `r_env_get()`.

PR #991, opened 2020-04-07
Move ptype2 shape handling to C

DavisVaughan:
This PR moves the common shape handling in `vec_ptype2()` down to C. This has the nice benefit of allowing us to mark the R level ptype2 methods for bare types with `stop_native_implementation()`.

- New `dim.c` and `dim.h` to hold dim specific functions, and make a few of the performance critical ones inline. This aligns better with `dim.R` too.

- New `shape.c` and `shape.h` to hold the common shape functionality

   - The idea of "shape" is tweaked slightly. It is now the dimensions vector with the first axis set to 0. It used to be all of the axes except the first, but this was a little harder to build on.

   - `vec_shape2()` is the main function here, it takes two sets of dimensions to find a common shape for and returns that common shape (as defined above). It replaces `shape_common()`.

   - `vec_shaped_ptype()` replaces `shape_match()` and is the entry point that `vec_ptype2()` uses. It computes the common shape from `x` and `y` and applies it to the `ptype`.

- When I added in the `stop_native_implementation()`, I learned that the complex methods were missing from `vec_ptype2_switch_native()`. I added them.

I detect essentially 0 loss of performance for non-shaped inputs.

```r
library(vctrs)

x <- 1
y <- 1

# Master
bench::mark(vec_ptype2(x, y), iterations = 1000000)
#> # A tibble: 1 x 6
#> 1 vec_ptype2(x, y)    673ns    866ns  1076072.        0B     107.

# This PR
bench::mark(vec_ptype2(x, y), iterations = 1000000)
#> # A tibble: 1 x 6
#> 1 vec_ptype2(x, y)    699ns    875ns  1071231.        0B     107.
```

Really nice performance now for array common types

```r
library(vctrs)

x <- array(1, dim = c(2, 1, 4, 6))
y <- array(1, dim = c(3, 5, 1, 6))

# Master
bench::mark(vec_ptype2(x, y), iterations = 100000)
#> # A tibble: 1 x 6
#> 1 vec_ptype2(x, y)   19.9µs   23.4µs    39464.    99.1KB     54.5

# This PR
bench::mark(vec_ptype2(x, y), iterations = 100000)
#> # A tibble: 1 x 6
#> 1 vec_ptype2(x, y)    935ns   1.31µs   744330.        0B     89.3
```


src/shape.c
```diff
@@ -0,0 +1,140 @@
...
+  }
+
+  UNPROTECT(1);
+  return out;
+}
+
+// -----------------------------------------------------------------------------
+
+// Sets the first axis to zero
+static inline SEXP vec_shape(SEXP dimensions) {
```
lionel-, line 109: Likely not an issue here, but as a matter of style I'd probably avoid marking such routines with more than 4 instructions inline. They can slow things down by saturating the instruction cache.

src/shape.c
```diff
@@ -0,0 +1,140 @@
...
+
+static inline int vec_dimension2(int axis, int x_dimension, int y_dimension) {
+  if (x_dimension == y_dimension) {
+    return x_dimension;
+  } else if (x_dimension == 1) {
+    return y_dimension;
+  } else if (y_dimension == 1) {
+    return x_dimension;
+  } else {
+    Rf_errorcall(R_NilValue, "Incompatible lengths along axis %i: %i, %i.", axis, x_dimension, y_dimension);
```
lionel-, line 138: Maybe something along the lines of:

> Can't combine shaped vectors.
> x Incompatible sizes {foo} and {bar} along axis {i}.

DavisVaughan:
@lionel- (and FYI for @hadley) we now throw `vctrs_error_incompatible_type` errors in `vec_ptype2()` when two shapes are incompatible. The error message is much better too.

``` r
library(vctrs)

x <- array(integer(), dim = c(0, 2, 3))
y <- array(integer(), dim = c(0, 1, 4))
```

```r
# Master
vec_ptype2(x, y)
#> Error: Incompatible lengths: 3, 4.

# This PR
vec_ptype2(x, y)
#> Error: Can't combine <integer[,2,3]> and <integer[,1,4]>.
#> x Incompatible sizes 3 and 4 along axis 3.
```

```r
# Master
vec_ptype2(x, y, x_arg = "foo", y_arg = "bar")
#> Error: Incompatible lengths: 3, 4.

# This PR
vec_ptype2(x, y, x_arg = "foo", y_arg = "bar")
#> Error: Can't combine `foo` <integer[,2,3]> and `bar` <integer[,1,4]>.
#> x Incompatible sizes 3 and 4 along axis 3.
```

@lionel- I rebased to bring in utilities from the type-error changes, but if you look at just the most recent commit here then that is where the changes are. Essentially I'm now threading all of the info required to give a good error message through `vec_shape2()`.

lionel-:
nice!

hadley:
The error messages are a lot better, but I'm a little worried about using "axis" here; I think dimension would be clearer to most R users.

DavisVaughan:
How strongly do you feel about that? I'm fairly confident that axis is the right technical term here. If we changed it, the error message would be different from the terminology used in the underlying code.

I like that axis specifies the direction along the array, and dimension specifies the measure along that axis (like the length). i.e. "the second axis has dimension 3". To me, that terminology also makes the code much easier to follow.

hadley:
I don't feel too strongly about it, given that mostly people won't see these messages. But I think most people would describe a matrix as having two dimensions, not having two axes.

PR #994, opened 2020-04-07
Add `x_arg` to `vec_ptype()`

DavisVaughan:
This PR follows up on the notes in #823 and gives `vec_ptype()` an `x_arg` argument.

I also took the chance to finally switch `vec_type()` to `vec_ptype()` internally.

The nicest thing about this is a small simplification to `vec_ptype2()`, where we no longer have to do a `vec_assert()` that is guarded by a `!vec_is_partial()` before calling `vec_ptype()`. Because of #823, `vec_ptype()` handles partial types, base R scalars (like `quote(x)`), and classed scalars (like `foobar()`) correctly. Now that we can pass the `x_arg` through to `vec_ptype()`, we don't need this check anymore.

### DIFF
--- a/src/type.c
+++ b/src/type.c
@@ -8,10 +8,18 @@ static SEXP fns_vec_ptype_finalise_dispatch = NULL;
 
 
 static SEXP vec_type_slice(SEXP x, SEXP empty);
-static SEXP s3_type(SEXP x);
+static SEXP s3_type(SEXP x, struct vctrs_arg* x_arg);
 
-// [[ include("vctrs.h"); register() ]]
-SEXP vec_ptype(SEXP x) {
+// [[ register() ]]
+SEXP vctrs_ptype(SEXP x, SEXP x_arg) {
+  struct vctrs_arg x_arg_ = vec_as_arg(x_arg);
+  return vec_ptype(x, &x_arg_);
+}
+
+static SEXP col_ptype(SEXP x);
+
+// [[ include("vctrs.h") ]]
+SEXP vec_ptype(SEXP x, struct vctrs_arg* x_arg) {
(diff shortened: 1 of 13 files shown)


src/type.c
```diff
@@ -143,17 +155,15 @@ SEXP vctrs_type_common(SEXP call, SEXP op, SEXP args, SEXP env) {
...
 
   if (r_is_true(r_peek_option("vctrs.no_guessing"))) {
     Rf_errorcall(R_NilValue, "strict mode is activated; you must supply complete `.ptype`.");
   }
 
   // Start reduction with the `.ptype` argument
-  struct vctrs_arg ptype_arg = new_wrapper_arg(NULL, ".ptype");
-
-  SEXP type = PROTECT(reduce(ptype, &ptype_arg, dots, &vctrs_type2_common));
+  SEXP type = PROTECT(reduce(ptype, args_dot_ptype, dots, &vctrs_type2_common));
```
lionel-, line 166: (That's a case where the `p_` convention would be helpful, e.g. `p_args_dot_ptype`.)

tests/testthat/error/test-type2.txt
```diff
@@ -0,0 +1,20 @@
+
+can override scalar vector error message for base scalar types
+==============================================================
+
+> vec_ptype2(NULL, quote(x), y_arg = "foo")
+Error: `foo` must be a vector, not a symbol.
```
lionel-, line 6: 👍

src/type.c
DavisVaughan: Yea normally I would, but I went for matching `args_empty` here

lionel-: This was more of a comment about the `args_` prefix generally.

Issue #968, opened 2020-03-28
Review convenience implementations for `vctrs_vctr`


lionel-:
Currently unclear what are the consequences of the special handling of `vctrs_vctr` classes. It makes me nervous that rcrds and list-ofs inherit from the current behaviour that I don't understand well. I think we need a better and clearer foundation for the convenience implementations.

Should `vctrs_vctr` classes inherit from the base coercion diagram, i.e. if the base type is double it's coercible to integer and logical automatically? Should this be by default or opt-in? Should this be only for classes that don't implement any `vec_ptype2()` methods?

lionel-:
I also wondered if explicit inheritance from a base type would be the way to inherit from the base coercion dag. With this approach the UI is a bit weird since classes normally don't inherit from coercion methods. The base class would be a confusing exception.

I think my preference goes to `vctrs_vctr` classes inheriting the coercion dag as long as they don't implement a ptype2 method. Once implemented, the full set of ptype2 and cast methods needs to be implemented.

DavisVaughan:
I'm starting to think that it would be nice to remove all special casing of `vctrs_vctr`, `vctrs_rcrd`, and `vctrs_list_of`, and even remove any mentioning of them at the C level (except `fields.c`).

I think it should be possible for these classes to be implemented in a separate package outside of vctrs (I'm not saying we should, I just think that clear separation would be valuable).

I looked at the codebase, and I don't think it would be that hard. It would force us to make sure that we are providing all the correct customization points.

---

Here is what would have to be tweaked (it isn't much!):

- C level `new_list_of()` and `init_list_of()` are no longer used and can be removed.

   - This also means we can remove `classes_list_of`.

- The C type `vctrs_class_list_of` is only used in `vec_is_list()`, but if we remove it we will still get the right results because of `vec_is_vector()` (i.e. `vec_proxy.vctrs_list_of` exists, proof that we have the right customization point here).

   - That would let us remove `strings_vctrs_list_of`

- The C type `vctrs_class_rcrd` is used in `vec_is_list()` as a placeholder, but removing it still results in correct fallthrough behavior. It is also used in the C level `is_record()`, which is not used anywhere.

   - This would let us remove `vctrs_class_posixlt` too, but not `vctrs_class_bare_posixlt`.

   - This would let us remove `strings_vctrs_rcrd`

- The above two would let us remove `strings_vctrs_vctr`

And I think that is all of the C level usage of these vctrs types.

---

[...]

PR #1003, opened 2020-04-10
Remove internal usage of vctrs_vctr/rcrd/list_of

DavisVaughan:
Related to https://github.com/r-lib/vctrs/issues/968#issuecomment-611092692

This PR does two things:

- C level special casing of `vctrs_vctr`, `vctrs_list_of` and `vctrs_rcrd` has been removed. All handling of these classes is done at the R level in such a way that they could now live in an entirely different package if they wanted to (with the exception of the 1 call to `vctrs_cast()` in `vec_default_cast()`, which we will revisit elsewhere)

- In the process, I learned that `vec_is_list()` is not implemented quite right. I had to do something here, so I've changed it to first take the proxy, then check if that proxy either directly inherits from `"list"` or is a bare list.

---

The above point about `vec_is_list()` was discussed on slack, and we didn't come to a complete agreement on what the solution should be, so it is open to discussion.

The problem is that one of the ways an object can be considered a list is if:

“x is an S3 list that vec_is() returns TRUE for. For this to return TRUE, the class must implement a vec_proxy() method.”

This works nicely for `vctrs_list_of,` but it implies that `vctrs_rcrd` and `POSIXlt` should both return `TRUE` from `vec_is_list()` because they both have `vec_proxy()` methods, but I don’t think that we want that.

The proposed change uses the following rules for `vec_is_list()`:

- x is a bare list with no class
- x is a list explicitly inheriting from `"list"`
- x is an S3 object that implements a `vec_proxy()` method, and the proxy satisfies one of the above two rules

A few more notes:

- Using the proxy means that S3 types that aren't internally a list can still be considered as lists if their proxy returns a list.
- This means that S3 lists that want to be considered lists should `unclass()` in the proxy
- One alternative suggested by @lionel- is to make `vec_is_list()` generic. One problem with that that he noted is that we might end up in a scenario where an object declares itself as a list even if internally it isn't. This could cause problems in vctrs methods like `vec_unchop()`. The proxy implementation guards against this by checking the `TYPEOF(proxy)` first, so even if the user lies and attaches an explicit `"list"` class on their non-list proxy, it will return `FALSE`

### DIFF
--- a/R/type-vctr.R
+++ b/R/type-vctr.R
@@ -75,7 +75,7 @@ new_vctr <- function(.data,
 
   if (is_list(.data)) {
     if (is.data.frame(.data)) {
-      abort("`.data` must not be a data frame.")
+      abort("`.data` can't be a data frame.")
     }
     class <- c(class, "vctrs_vctr", "list")
   } else {
--- a/R/type-vctr.R
+++ b/R/type-vctr.R
@@ -1,9 +1,16 @@
 #' vctr (vector) S3 class
 #'
+#' @description
 #' This abstract class provides a set of useful default methods that makes it
 #' considerably easier to get started with a new S3 vector class. See
 #' `vignette("s3-vector")` to learn how to use it to create your own S3
 #' vector classes.
+#'
+#' @details
+#' List vctrs are special cases. When created through `new_vctr()`, the
+#' resulting list vctr should always be recognized as a list by
+#' `vec_is_list()`. Because of this, the `inherit_base_type` argument is
+#' ignored for lists, and is always set to `TRUE`.
 #'
 #' @section Base methods:
 #' The vctr class provides methods for many base generics using a smaller
@@ -51,7 +58,8 @@
 #' @param class Name of subclass.
 #' @param inherit_base_type \Sexpr[results=rd, stage=render]{vctrs:::lifecycle("experimental")}
 #'   Does this class extend the base type of `.data`?  i.e. does the
-#'   resulting object extend the behaviour the underlying type?
+#'   resulting object extend the behaviour of the underlying type? For list
+#'   `.data`, this argument is ignored and is always set to `TRUE`.
 #' @export
 #' @keywords internal
 #' @aliases vctr
@@ -65,7 +73,15 @@ new_vctr <- function(.data,
 
   nms <- validate_names(.data)
 
-  class <- c(class, "vctrs_vctr", if (inherit_base_type) typeof(.data))
+  if (is_list(.data)) {
+    if (is.data.frame(.data)) {
+      abort("`.data` must not be a data frame.")
+    }
+    class <- c(class, "vctrs_vctr", "list")
+  } else {
+    class <- c(class, "vctrs_vctr", if (inherit_base_type) typeof(.data))
+  }
+
   attrib <- list(names = nms, ..., class = class)
 
   vec_set_attributes(.data, attrib)
--- a/R/type-vctr.R
+++ b/R/type-vctr.R
@@ -9,8 +9,8 @@
 #' @details
 #' List vctrs are special cases. When created through `new_vctr()`, the
 #' resulting list vctr should always be recognized as a list by
-#' `vec_is_list()`. Because of this, the `inherit_base_type` argument is
-#' ignored for lists, and is always set to `TRUE`.
+#' `vec_is_list()`. Because of this, if `inherit_base_type` is `FALSE`
+#' an error is thrown.
 #'
 #' @section Base methods:
 #' The vctr class provides methods for many base generics using a smaller
@@ -57,16 +57,17 @@
 #' @param ... Name-value pairs defining attributes
 #' @param class Name of subclass.
 #' @param inherit_base_type \Sexpr[results=rd, stage=render]{vctrs:::lifecycle("experimental")}
-#'   Does this class extend the base type of `.data`?  i.e. does the
-#'   resulting object extend the behaviour of the underlying type? For list
-#'   `.data`, this argument is ignored and is always set to `TRUE`.
+#'   A single logical, or `NULL`. Does this class extend the base type of
+#'   `.data`? i.e. does the resulting object extend the behaviour of the
+#'   underlying type? Defaults to `FALSE` for all types except lists, which
+#'   are required to inherit from the base type.
 #' @export
 #' @keywords internal
 #' @aliases vctr
 new_vctr <- function(.data,
                      ...,
                      class = character(),
-                     inherit_base_type = FALSE) {
+                     inherit_base_type = NULL) {
   if (!is_vector(.data)) {
     abort("`.data` must be a vector type.")
   }
@@ -77,11 +78,20 @@ new_vctr <- function(.data,
     if (is.data.frame(.data)) {
       abort("`.data` can't be a data frame.")
     }
-    class <- c(class, "vctrs_vctr", "list")
-  } else {
-    class <- c(class, "vctrs_vctr", if (inherit_base_type) typeof(.data))
+
+    if (is.null(inherit_base_type)) {
+      inherit_base_type <- TRUE
+    } else if (is_false(inherit_base_type)) {
+      abort("List `.data` must inherit from the base type.")
+    }
   }
 
+  # Default to `FALSE` in all cases except lists
+  if (is.null(inherit_base_type)) {
+    inherit_base_type <- FALSE
+  }
+
+  class <- c(class, "vctrs_vctr", if (inherit_base_type) typeof(.data))
   attrib <- list(names = nms, ..., class = class)
 
   vec_set_attributes(.data, attrib)
--- a/src/type-info.c
+++ b/src/type-info.c
@@ -109,9 +109,8 @@ bool vec_is_list(SEXP x) {
   case vctrs_class_none:
     return true;
 
-  // Explicit lists
+  // Explicit list
   case vctrs_class_list:
-  case vctrs_class_list_of:
     return true;
 
   // Non-explicit S3 lists
--- a/src/type-info.c
+++ b/src/type-info.c
@@ -117,10 +117,6 @@ bool vec_is_list(SEXP x) {
   case vctrs_class_unknown:
     return vec_is_vector(x);
 
-  // TODO: Can this ever be considered a list?
-  case vctrs_class_rcrd:
-    return false;
-
   // List-like classes known by `class_type()`.
   // All data frame classes, posixlt.
   default:
--- a/src/type-info.c
+++ b/src/type-info.c
@@ -98,37 +98,67 @@ enum vctrs_type vec_proxy_typeof(SEXP x) {
   return vec_base_typeof(x, true);
 }
 
-static bool vec_is_list_impl(SEXP proxy);
 
-// [[ include("vctrs.h") ]]
-bool vec_is_list(SEXP x) {
-  SEXP proxy = PROTECT(vec_proxy(x));
-
-  bool out = vec_is_list_impl(proxy);
-
-  UNPROTECT(1);
-  return out;
-}
-
-static bool vec_is_list_impl(SEXP proxy) {
-  // Guarantee that the `proxy` is internally a list
-  if (TYPEOF(proxy) != VECSXP) {
-    return false;
-  }
-
-  if (OBJECT(proxy)) {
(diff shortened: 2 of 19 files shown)


src/type-info.c
```diff
@@ -98,36 +98,33 @@ enum vctrs_type vec_proxy_typeof(SEXP x) {
...
-  // Non-explicit S3 lists
-  case vctrs_class_unknown:
-    return vec_is_vector(x);
+  UNPROTECT(1);
+  return out;
+}
 
-  // TODO: Can this ever be considered a list?
-  case vctrs_class_rcrd:
+static bool vec_is_list_impl(SEXP proxy) {
```
lionel-, line 113: `vec_proxy_is_list()`? Not sure it makes sense to extract the implementation in a different funtion though.

lionel-:
IIUC it is now possible to create a record vector that is considered a list by having a data frame proxy that explicitly inherits from list as base class. This might throw off the rest of the vctrs internals. Do we need to worry about testing for this case?

Anyway, if we were to support record lists via explicit inheritance, it should be the main class that inherits from list, not the proxy. That means you can't know if a class is a list based on the proxy, but that is generally consistent with the idea of a proxy. For example a data frame proxy does not indicate the main class is a data frame.

Though explicit inheritance would be problematic for `vec_slice2()`. How do we find the data? So perhaps we need the combination of inheritance of the main class to signal lists, and a special column `.list` to store the data in a place where vctrs can find it? Or better, require the first column of the proxy of a list record to be the list data for efficiency. If the first column is not a list, that would be an error.

(Tangent: Are data frames that inherit from list generally something that we'd support in vctrs to signal colwise dfs? I wondered about that when @hadley suggested to change the return type of `safely()` to a data frame to help with auto-splicing. This would be mostly compatible with base primitives but not with vctrs ones. Explicit list inheritance to make them vectors of columns might help but also might have unpredicted consequences.)

DavisVaughan:
> IIUC it is now possible to create a record vector that is considered a list by having a data frame proxy that explicitly inherits from list as base class. This might throw off the rest of the vctrs internals. Do we need to worry about testing for this case?

I don't think this is a big issue because if the class hierarchy looks like `c("data.frame", "list")` then vctrs generally won't treat this as a data frame anyways. We explicitly look to see if the base class is `"data.frame"` to determine if something is a data frame.

src/type-info.c
DavisVaughan: This was mainly to keep me from having to `UNPROTECT(1)` 3 times in `vec_is_list()` since we exit this function at different places. I don't think the name is too important here

lionel-: A standalone function conveys meaning. The name is important for this reason. I don't think we should extract logic just to avoid unprotect statements.

DavisVaughan:
In light of this point:

> That means you can't know if a class is a list based on the proxy, but that is generally consistent with the idea of a proxy. For example a data frame proxy does not indicate the main class is a data frame.

I'm somewhat convinced we shouldn't be operating on the proxy. Maybe the rule has nothing to do with the proxy and is simply:

- `x` is a bare list
- `x` inherits explicitly from `"list"` as the base class

Two thoughts:

- I think this would imply `vctrs_list_of` should inherit explicitly from `"list"`

- I thought this might be problematic in cases where `x` is a list but the `proxy` is not, but I'm not sure it actually is?

lionel-:
> I'm somewhat convinced we shouldn't be operating on the proxy. Maybe the rule has nothing to do with the proxy and is simply:

I agree that list inheritance seems to be the main criterion. The proxy is also important in a way because we need to find the elements of a list, so the proxy has to be a list (with the added rule that it can be wrapped in a data frame, as first column).

> I think this would imply vctrs_list_of should inherit explicitly from "list"

Yes I think the idea was to inherit from list eventually, we didn't do it immediately because of the inheritance of ptype2/cast methods, which is now fixed.

lionel-:
Since requiring explicit list inheritance is intrusive for foreign classes, I think this suggests:

- Inherit explicitly from list _or_ implements `vec_is_list()` method.
- Proxy (by default identity) must be a list.

We'd no longer treat a list proxy as a criterion, bus as a requirement. This would be a breaking change but I think it would bring clarity to the API.

DavisVaughan:
So the `vec_is_list()` generic would drop into C and check if there is a `vec_is_list()` method available for `x`? Would it even call it, or is existence of the method enough?

How do we check list-ness of the proxy? It would have to either be a bare list or explicitly inherit from `"list"` right?

With that in mind, I came up with this:

```c
static SEXP vec_is_list_method(SEXP x);
static bool vec_proxy_is_list(SEXP x);

// [[ register() ]]
SEXP vctrs_is_list(SEXP x) {
  return Rf_ScalarLogical(vec_is_list(x));
}

// [[ include("vctrs.h") ]]
bool vec_is_list(SEXP x) {
  // Require `x` to be a list internally?
  if (TYPEOF(x) != VECSXP) {
    return false;
  }

  // Implicit list
  if (!OBJECT(x)) {
    return true;
  }

  // Explicit list, require proxy to be a list
  if (class_type(x) == vctrs_class_list) {
    return vec_proxy_is_list(x);
  }

  // Implements `vec_is_list()` method
  if (vec_is_list_method(x) != R_NilValue) {
    return vec_proxy_is_list(x);
  }

  return false;
}

static SEXP vec_is_list_method(SEXP x) {
  return s3_find_method("vec_is_list", x, vctrs_method_table);
}

static bool vec_proxy_is_list(SEXP x) {
  SEXP proxy = PROTECT(vec_proxy(x));

  // Require `proxy` to be a list internally
  if (TYPEOF(proxy) != VECSXP) {
    UNPROTECT(1);
    return false;
  }

  // Implicit list
  if (!OBJECT(proxy)) {
    UNPROTECT(1);
    return true;
  }

  bool out = class_type(proxy) == vctrs_class_list;

  UNPROTECT(1);
  return out;
}
```

lionel-:
> So the vec_is_list() generic would drop into C and check if there is a vec_is_list() method available for x? Would it even call it, or is existence of the method enough?

I think we'd call it to avoid weirdness/magic?

> How do we check list-ness of the proxy? It would have to either be a bare list or explicitly inherit from "list" right?

I think we just require it to be a non-data frame VECSXP, no need for explicit inheritance at the proxy level.

DavisVaughan:
> I think we'd call it to avoid weirdness/magic?

Do we still have to check the proxy though? So we'd call it, check that the result is `TRUE` and then continue on to check the type of the proxy as well? That felt magical too because a user might have a `vec_is_list()` method that returns `TRUE`, but then `vec_is_list()` returns `FALSE` because their proxy isn't a list

lionel-:
Yes I was thinking an error if the proxy isn't a list and vec-is-list or inherits-list are true.

This requires two dispatches/lookup (vec-is-list and vec-proxy) but that seems ok because explicit inheritance can be used to avoid one of these.

hadley:
This discussion is starting to feel rather abstract to me. Could we bring it back to concreteness with some examples using existing R classes?

DavisVaughan:
Sure!

The automatic behavior is straightforward:

```r
# types we want to be considered lists automatically

list()

structure(list(), class = c("foobar", "list"))

# ^ Similar to this:
# - `vctrs_list_of` if we make it inherit from `"list"`
# - other package developers extending lists
```

```r
# types we don't want to be considered lists automatically

structure(list(), class = "foobar")

# ^ Similar to this:
# - POSIXlt
# - lm
# - rcrd
# - data.frame
```

We could give package developers a non-invasive way to register their existing class as a list without explicitly adding a `"list"` base class. i.e. do we need this escape hatch?

```r
x <- structure(list(), class = "foobar")

vec_is_list.foobar <- function(x) {
  TRUE
}
```

@lionel- mentioned this, but now that I think about it, `vec_is_vector()` doesn't have this behavior with lists, so I'm not sure that `vec_is_list()` should either (we could end up declaring something as a list because it implements this method, but not a vector, which feels inconsistent).

The other thing we are thinking about guarding against is what happens when something we automatically consider a list doesn't have a list proxy. In `vec_is_list()` we _could_ do an additional check that the proxy is a list (in a looser sense, we'd check that it is a VECSXP that isn't a data frame). We would error if the proxy was not a list, like here:

```r
# will this cause problems?

# automatically considered a list
x <- structure(list(), class = c("foobar", "list"))

# non list proxy
vec_proxy.foobar <- function(x) {
  1
}
```

lionel-:
> vec_is_vector() doesn't have this behavior with lists,

This would be taken care of in the fallback behaviour. If vec-is-list returns TRUE, then the class is a vector. The default method for vec-is-list would check for explicit inheritance.

hadley:
Do we have any existing examples of classes that should be considered as lists that don't have class "list"? (apart from `list_of()`, which we obviously need to fix).

lionel-:
> Do we have any existing examples of classes that should be considered as lists that don't have class "list"?

I can't think of any because I'm not familiar with many foreign classes, but I would guess many of the S3 lists that need to declare themselves as vectors will also need to declare themselves as lists, once we add slice2, assign2, etc.

We have an escape hatch for is-vector (implement proxy), and I think we need one for is-list. Part of this is that changing the class vector is quite an invasive change to require from maintainers. The other part is that explicit inheritance is a maturing concept in which I don't have full confidence, so it feels weird to impose it as if it were stable.

If this unblocks the situation, we can go ahead with inherits-list and add S3 dispatch later on, it would be an easy change.

DavisVaughan:
I've left out the S3 dispatch part for now.

One thing that does fall out from this that is a little strange is that a `vctrs_vctr` built from a list is automatically considered a vector but not a list.

``` r
library(vctrs)

x <- new_vctr(list(1))

vec_is(x)
#> [1] TRUE

vec_is_list(x)
#> [1] FALSE
```

<sup>Created on 2020-04-14 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

lionel-:
I think `new_vctr()` should by default add `"list"` to the class vector of lists.

DavisVaughan:
> I think new_vctr() should by default add "list" to the class vector of lists.

I had to think about this a little, but I think I agree.

If we added this, I was wondering if there was a way to "turn off" the explicit `"list"` inheritance that you'd get automatically with `new_vctr()`, but I don't think it would ever make sense to do so, right? By turning it off, you would be declaring yourself as a scalar with our new rules, which `new_vctr()` should never create.

hadley:
@DavisVaughan agreed

R/type-vctr.R
```diff
@@ -65,7 +73,15 @@ new_vctr <- function(.data,
 
   nms <- validate_names(.data)
 
-  class <- c(class, "vctrs_vctr", if (inherit_base_type) typeof(.data))
+  if (is_list(.data)) {
+    if (is.data.frame(.data)) {
+      abort("`.data` must not be a data frame.")
```
lionel-, line 78:
```suggestion
      abort("`.data` can't be a data frame.")
```

tests/testthat/test-type-vctr.R
```diff
@@ -33,6 +33,14 @@ test_that("Can opt out of base type", {
   expect_s3_class(x, c("x", "vctrs_vctr"), exact = TRUE)
 })
 
+test_that("base type is always set for lists", {
+  expect_s3_class(new_vctr(list(), inherit_base_type = FALSE), "list")
+})
+
+test_that("data frames are not allowed", {
+  expect_error(new_vctr(mtcars), "must not be a data frame")
```
lionel-, line 41:
```suggestion
  expect_error(new_vctr(mtcars), "can't be a data frame")
```

R/type-vctr.R
```diff
@@ -1,10 +1,17 @@
...
 #' This abstract class provides a set of useful default methods that makes it
 #' considerably easier to get started with a new S3 vector class. See
 #' `vignette("s3-vector")` to learn how to use it to create your own S3
 #' vector classes.
 #'
+#' @details
+#' List vctrs are special cases. When created through `new_vctr()`, the
+#' resulting list vctr should always be recognized as a list by
+#' `vec_is_list()`. Because of this, the `inherit_base_type` argument is
+#' ignored for lists, and is always set to `TRUE`.
```
lionel-, line 13: Since the default value of `inherit_base_type` is not constant, it should be `NULL` instead of `FALSE`.

DavisVaughan: Oops we talked about this, my bad

PR #1012, opened 2020-04-14
Run document to update the URL

DavisVaughan:


lionel-:
Can you push this kind of changes directly to master please?

Issue #1016, opened 2020-04-17
Improve vctrs documentation


lionel-:
* [ ] FAQ about how to implement common types for data frame class and include message in fallback warning implemented in #981.

* [ ] FAQ about lossy cast errors (see also #922).

* [ ] Discussion of promotion vs class hierarchy #980.

lionel-:
The advice for `<foo>` to `<foo>` coercion errors should be different than for `<foo>` to `<bar>` errors.

* The former means the same-class fallback failed because of diverging attribtutes. It always makes sense to implement ptype2 and cast methods for a single class. We should point to a relevant FAQ.

* In the latter case, we should be careful not to encourage implementation of coercion methods too strongly. Instead, the advice should focus on recommending users to normalise their inputs.

PR #1017, opened 2020-04-17
Add fallback for foreign data frame classes

### LINKED ISSUE #981: data.frame cast fallback
I think we may need a some data.frame fallbacks so that we get something like this:

```
vec_rbind(tibble(x = 1), data.table(x = 1))
# Warning: combining <tibble> and <data.table>; falling back to <data.frame>
```

(This is not a general regret that we're moving away from inheritance; it's a specific pragmatic fallback + warning for people working with multiple data frame subclasses)

lionel-:
Closes #981 

The warning is thrown during the ptype2 reduction. A sentinel data frame class avoids repeat warnings. I haven't implemented a fallback in `vec_cast()` because it isn't necessary for `vec_cbind()` and `vec_rbind()`.

```r
foo <- structure(mtcars[1:3], class = c("foo", "data.frame"))
bar <- structure(mtcars[9:11], class = c("bar", "data.frame"))

vec_ptype_common(foo, bar)
#> [1] mpg  cyl  disp am   gear carb
#> Warning: Can't combine <foo> and <bar>.
#> i vctrs coercion methods should be implemented for these classes.
#> i Falling back to <data.frame>.
```

The warning should eventually refer to a FAQ about normalising inputs to `data.frame`. This is tracked in #1016.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -30,6 +30,12 @@
     in `vec_assign()`), two double dispatch were needed. Now it can be
     done with one double dispatch by calling `vec_cast()` directly.
 
+* `vec_ptype2()` is now more permissive with subclasses of data frames
+  and falls back to a bare data frame when the classes do not have a
+  common type (#981). This is for convenience, users should normalise
+  their inputs to a common data frame class manually to avoid the
+  warning (if applicable, the classes could also implement a common type).
+
 * `vec_cbind()` now calls `vec_restore()` on inputs emptied of their
   columns before computing the common type. This has
   consequences for data frame classes with special columns that
--- a/R/type-data-frame.R
+++ b/R/type-data-frame.R
@@ -89,6 +89,37 @@ df_ptype2 <- function(x, y, ..., x_arg = "", y_arg = "") {
   .Call(vctrs_df_ptype2, x, y, x_arg, y_arg)
 }
 
+# Fallback for data frame subclasses (#981)
+vec_ptype2_df_fallback <- function(x, y, x_arg = "", y_arg = "") {
+  ptype <- vec_ptype2(
+    new_data_frame(x),
+    new_data_frame(y)
+  )
+
+  if (!is_df_fallback(x) && !is_df_fallback(y)) {
+    msg <- cnd_type_message(x, y, x_arg, y_arg, NULL, "combine", NULL)
+    warn(c(
+      msg,
+      i = "vctrs coercion methods should be implemented for these classes.",
+      i = "Falling back to <data.frame>."
+    ))
+  }
+
+  # Return a fallback class so we don't warn multiple times. This
+  # fallback class is stripped in `vec_ptype_finalise()`.
+  new_fallback_df(ptype)
+}
+
+is_df_subclass <- function(x) {
+  inherits(x, "data.frame") && !identical(class(x), "data.frame")
+}
+is_df_fallback <- function(x) {
+  inherits(x, "vctrs:::df_fallback")
+}
+new_fallback_df <- function(x, n = nrow(x)) {
+  new_data_frame(x, n = n, class = "vctrs:::df_fallback")
+}
+
 
 # Cast --------------------------------------------------------------------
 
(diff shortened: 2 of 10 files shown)


R/type-data-frame.R
```diff
@@ -89,6 +89,38 @@ df_ptype2 <- function(x, y, ..., x_arg = "", y_arg = "") {
...
+    new_data_frame(x),
+    new_data_frame(y)
+  )
+
+  if (!is_df_fallback(x) && !is_df_fallback(y)) {
+    types <- c(class(x)[[1]], class(y)[[1]])
+    msg <- cnd_type_message(x, y, x_arg, y_arg, NULL, "combine", NULL, types = types)
+    warn(c(
+      msg,
+      i = "vctrs coercion methods should be implemented for these classes.",
```
hadley, line 104: I think it would be better to write this for the end user. Something like: "Convert all inputs to the same to avoid this warning"

tests/testthat/error/test-type-data-frame.txt
```diff
@@ -0,0 +1,14 @@
+
+combining data frames with foreign classes uses fallback
+========================================================
+
+> foo <- structure(mtcars[1:3], class = c("foo", "data.frame"))
+> bar <- structure(mtcars[9:11], class = c("bar", "data.frame"))
+> vec_ptype_common(foo, bar)
```
hadley, line 7: Do you want to have a `vec_ptype_common()` here just to check that the warning only occurs once?

NEWS.md
```diff
@@ -30,6 +30,12 @@
     in `vec_assign()`), two double dispatch were needed. Now it can be
     done with one double dispatch by calling `vec_cast()` directly.
 
+* `vec_ptype2()` is now more permissive with subclasses of data frames
+  and fall back to a bare data frame when the classes do not have a
```
DavisVaughan, line 34:
```suggestion
  and falls back to a bare data frame when the classes do not have a
```

R/type-data-frame.R
lionel-: oh yeah I forgot to change this.

NEWS.md
```diff
@@ -30,6 +30,12 @@
...
+  common type (#981). This is for convenience, users should normalise
+  their inputs to a common data frame class manually to avoid the
+  warning (if applicable, the classes could also implement a common type).
```
DavisVaughan, line 37: Can we be a bit more encouraging here that really we want package developers to define the behavior?

```suggestion
  warning. If applicable, package developers should implement a common type to define the behavior precisely.
```

R/type-data-frame.R
```diff
@@ -89,6 +89,38 @@ df_ptype2 <- function(x, y, ..., x_arg = "", y_arg = "") {
   .Call(vctrs_df_ptype2, x, y, x_arg, y_arg)
 }
 
+# Fallback for data frame subclasses (#981)
+vec_ptype2_df_fallback <- function(x, y, x_arg = "", y_arg = "") {
+  ptype <- vec_ptype2(
+    new_data_frame(x),
+    new_data_frame(y)
```
DavisVaughan, line 96: I really like that constructors zap input attributes so we can do this

R/type-data-frame.R
```diff
@@ -89,6 +89,38 @@ df_ptype2 <- function(x, y, ..., x_arg = "", y_arg = "") {
...
+    ))
+  }
+
+  # Return a fallback class so we don't warn multiple times. This
+  # fallback class is stripped in `vec_ptype_finalise()`.
+  new_fallback_df(ptype)
+}
+
+is_df_subclass <- function(x) {
+  inherits(x, "data.frame") && !identical(class(x), "data.frame")
```
DavisVaughan, line 115: You might could use `.Call(vctrs_class_type, x) == "data_frame"`. It also returns `bare_data_frame` and `bare_tibble` but those cases should be covered elsewhere

NEWS.md
DavisVaughan: (after talking more with Lionel, maybe ignore this if the message is mainly for users)

tests/testthat/error/test-type-data-frame.txt
lionel-: yup, also improved the logic so that repeat warnings for new classes that we haven't seen before are issued.

R/type-data-frame.R
lionel-: Good to know, but probably better to keep it simple here.

Issue #945, opened 2020-03-24
Figure out composition of restore methods


lionel-:
Data frame restoration is currently weird because we first call restore-default, then dispatch on `vec_restore()`, which ends up calling restore-default again.

I suspect it makes sense to automatically call restore methods for all classes in reverse order without relying on S3 dispatch. We'd first restore core attributes, then data frame ones, etc. For consistency the proxy methods would also be called (in order this time) automatically. I worry about the efficiency of this scheme though.

Alternatively we could disallow inheritance and force classes to explicitly call their parent methods. It seems like that would introduce too much friction.

lionel-:
Another example whene composition of proxy and restore would be needed: If we implement names support for record vectors through a special rcrd field.

lionel-:
Restoring the classes that don't implement `vec_restore()` would prevent accidentally inheriting ptype2 methods via `vec_ptype()`:

```r
vec_ptype2.vctrs_foobar.vctrs_foobar <- function(x, y, ...) x

vec_ptype2(foobar(1), foobar(1, class = "quux"))
#> Error: Can't combine <vctrs_foobar> and <quux>.

vec_restore.vctrs_foobar <- function(x, to, ...) foobar(x)

vec_ptype2(foobar(1), foobar(1, class = "quux"))
#> numeric(0)
#> attr(,"class")
#> [1] "vctrs_foobar"
```

PR #1019, opened 2020-04-17
Allow factor/ordered `vec_restore()` method to be inherited

### LINKED ISSUE #1015: Factors subclass triggers error
``` r
library(vctrs)

a <- factor("a")
b <- factor("b")

class(a) <- c("myclass", "factor")
class(b) <- c("myclass", "factor")

vec_c(a, b)
#> Error: `vec_restore.factor()` is implemented at C level.
#> This R function is purely indicative and should never be called.
```

<sup>Created on 2020-04-17 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

DavisVaughan:
Closes #1015 

I got a little carried away and forgot that `vec_restore()` is still inherited.

This sort of begs the question, should it be? Related to https://github.com/r-lib/vctrs/issues/945

### DIFF
--- a/R/type-date-time.R
+++ b/R/type-date-time.R
@@ -73,17 +73,17 @@ vec_proxy_compare.POSIXlt <- function(x, ...) {
 
 #' @export
 vec_restore.Date <- function(x, to, ...) {
-  stop_native_implementation("vec_restore.Date")
+  NextMethod()
 }
 
 #' @export
 vec_restore.POSIXct <- function(x, to, ...) {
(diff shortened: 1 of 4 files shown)


lionel-:
I think it makes sense to inherit proxy and restore methods. Ideally we'd call all the methods in the class hierarchy automatically and in the right order instead of relying on the implementors calling next-method.

The reason inheritance makes sense for proxy / restore is that it's about the _subclassing_ hierarchy, i.e. the strict addition of attributes and classes.

Issue #623, opened 2019-10-17
`vec_cast()` should preserve names


lionel-:
```r
vec_cast(c(a = 1, b = 3), int())
#> [1] 1 3
```

DavisVaughan:
Two notes:

First, when dimensions (that aren't the 1st dim) are broadcast, should we recycle existing names or drop them? Current behavior is recycling (we should definitely test these expectations if we don't):

``` r
vctrs::vec_cast(
  matrix(1, dimnames = list("r1", "c1")),
  matrix(1, ncol = 2)
)
#>    c1 c1
#> r1  1  1
```

Second, it seems that `shape_broadcast()` drops dim names when the dimensionality changes due to the `dim(x) <- ` step here https://github.com/r-lib/vctrs/blob/ba92613dc230beea237aa0636b31f534b6900c5c/R/shape.R#L62

``` r
vctrs::vec_cast(
  c(x = 1),
  matrix(1, dimnames = list("r1", "c1"))
)
#>      [,1]
#> [1,]    1
```

lionel-:
If `x` is unnamed and `to` is named, should `x` gain a vector of empty names? This doesn't seem consistent with the idea of preserving the names of `x`.

This is related to the question of whether namedness is part of prototypes, and whether `vec_ptype()` should strip the names.

DavisVaughan:
> If x is unnamed and to is named, should x gain a vector of empty names?

I would say no, I don't think the names of `to` are considered at all

lionel-:
If we preserve names, shouldn't we preserve dimensions?

lionel-:
oh no because broadcasting is part of the common type determination.

DavisVaughan:
This is an issue with `vec_c()` and `vec_unchop()` when named inputs are cast

``` r
library(vctrs)

vec_c(c(x = 1), 1)
#> x   
#> 1 1

# casting dropped names
vec_c(c(x = 1L), 1)
#> [1] 1 1

args <- list(c(x = 1), 1)
vec_unchop(args, indices = list(1, 2))
#> x   
#> 1 1

# casting dropped names
args <- list(c(x = 1L), 1)
vec_unchop(args, indices = list(1, 2))
#> [1] 1 1
```

Found while investigating https://github.com/r-lib/vctrs/issues/1263

lionel-:
Currently native cast methods don't preserve shapes:

```r
x <- 1:4
dim(x) <- c(2, 2)

[...]

PR #1020, opened 2020-04-17
Improve consistency of names handling in prototype functions

lionel-:
* Consistently return unnamed vectors from `vec_ptype2()`. Record vectors gain a `names<-` method to support this. It currently only allows setting names to `NULL`, but should be extended in the future by storing names in a special field.

* Consistently preserve names in `vec_ptype()`, including row names. I still wonder if `vec_ptype()` should unname though.

* In the same spirit of unnaming the results of ptype2 methods, we also empty them. This simplifies the implementation of methods which can now just return `x` or `y`.


R/partial.R
```diff
@@ -18,6 +18,18 @@ new_partial <- function(..., class = character()) {
   new_sclr(..., class = c(class, "vctrs_partial"))
 }
 
+## Needed because partial classes inherit from `vctrs_sclr` which
+## can't be renamed. And `vec_ptype2()` etc zap the names.
+#' @export
+`names<-.vctrs_partial` <- function(x, value) {
```
hadley, line 24: This doesn't feel quite right to me — I don't think we should be systematically stripping names, and then providing `names<-` methods that ignore the stripping. Why not just strip names on atomic vectors?

lionel-: The partial vectors are not vectors, which makes them weird in vctrs. This is just a stopgap. I don't think the hacks around partial vectors should make us pause.

As for record vectors, they should support `names<-` at some point, like other vectors. I would really prefer a general solution than ignoring S3 vectors.

lionel-: More generally, maybe there really exists some vectors that do not support names. It doesn't seem ill-founded to require of these vectors to implement set-to-null as a no-op in their `names<-` method.

src/type.c
```diff
@@ -30,13 +32,21 @@ SEXP vec_ptype(SEXP x, struct vctrs_arg* x_arg) {
...
+// [[ include("vctrs.h") ]]
+SEXP vec_ptype_unnamed(SEXP x, struct vctrs_arg* x_arg) {
+  SEXP out = PROTECT(vec_ptype(x, x_arg));
+  out = vec_set_names(out, R_NilValue);
+  UNPROTECT(1);
+  return out;
+}
+
 static SEXP col_ptype(SEXP x) {
   return vec_ptype(x, args_empty);
```
DavisVaughan, line 51: Should the columns of the data frame be named? Or should names be stripped from them too?

src/type.c
```diff
@@ -30,13 +32,21 @@ SEXP vec_ptype(SEXP x, struct vctrs_arg* x_arg) {
...
-  case vctrs_type_dataframe:   return bare_df_map(x, &col_ptype);
+  case vctrs_type_dataframe:   return df_ptype(x, true);
   case vctrs_type_s3:          return s3_type(x, x_arg);
   case vctrs_type_scalar:      stop_scalar_type(x, x_arg);
   }
   never_reached("vec_ptype");
 }
 
+// [[ include("vctrs.h") ]]
+SEXP vec_ptype_unnamed(SEXP x, struct vctrs_arg* x_arg) {
```
DavisVaughan, line 43: It does feel a bit awkward that we have to use `vec_ptype_unnamed()` in `vec_ptype2()` to get the unname behavior, but `vec_ptype()` retains names. I am also leaning towards letting `vec_ptype()` unname, which would remove the need for the specialized `vec_ptype_unnamed()`

src/type.c
```diff
@@ -75,6 +85,24 @@ static SEXP s3_type(SEXP x, struct vctrs_arg* x_arg) {
   return vec_slice(x, R_NilValue);
```
DavisVaughan, line 85: It would be interesting to consider a version of `vec_slice()` that doesn't preserve names so we don't have to do as much work to unname after we slice. This feels somewhat similar to having a version of `vec_c()` that doesn't retain names

src/type.c
lionel-: I feel like `vec_ptype()` always returning unnamed vectors would simplify the notion of a prototype in vctrs.

src/type.c
DavisVaughan: If we did this, we'd have to reconsider how names are tracked in the lubridate PR. When we proxy, the names are added as a new column. This would get sliced and then restored even if we turned on the option to not preserve names.
https://github.com/tidyverse/lubridate/pull/871/files#diff-29798887c998d8e58008ea67fdf141d3R32

On the other hand, POSIXlt stores names directly on the `$year` field. So when we proxy we get a column that is named. If we turn off name preservation that should strip these names

src/type.c
lionel-: Good point. I think they should be unnamed for consistency with vec-assign.

lionel-:
Related issue: https://github.com/r-lib/vctrs/issues/623. `vec_cast()` should preserve the names of the input

src/type.c
DavisVaughan: It seems like if we did this then we wouldn't even need a `vec_set_names()` call

hadley:
I am uncomfortable with this change. I don't think it's needed for dplyr 1.0.0, so I'd prefer to put off any discussion until we have more time.

src/type.c
lionel-: We may still need the set-names call on S3 objects (or at least on S3 data frames) because the names might be encoded in a field, as in POSIXlt vectors. Or maybe we should enforce names as a special `vctrs::rcrd_names` field in data frame proxies for performance and simplicity. Then vec-slice could rely on either the names attribute or the record field.

DavisVaughan: > because the names might be encoded in a field, as in POSIXlt vectors

I think I'm arguing above that a `vec_slice(x, R_NilValue, preserve_names = false)` would already work correctly in this case.

lionel-:
Interestingly we have this in `vec_is()`, which also goes in the direction of removing names from prototypes:

```r
  x <- vec_slice(x, integer())
  ptype <- vec_slice(ptype, integer())

  # FIXME: Remove row names for matrices and arrays, and handle empty
  # but existing dimnames
  x <- vec_set_names(x, NULL)
  ptype <- vec_set_names(ptype, NULL)
```

Worth noting that rcrd vectors are currently not doing the right thing with this implementation, since vec-set-names currently goes through `names<-.vctrs_vctr`. This overwrites the field names.

src/type.c
lionel-: Maybe POSIXlt is not a good example. I was thinking the difference between named and unnamed record vectors of size zero would be the presence of a names field. This is equivalent to the presence or absence of a names attributes for normal vectors of size zero.

Issue #989, opened 2020-04-03
Support for sfc lists


lionel-:
From https://github.com/r-lib/vctrs/issues/985#issuecomment-608260268.

Since we're moving away from unstructured dispatch, we need to implement special support for other mechanisms like `AsIs` wrappers. sfc lists might need special support, but the general mechanism that would support them isn't clear.

sfc is a bit special because combination of disparate geometry types like `sfc<point>` and `sfc<line>` falls back to an abstract geometry type `sfc<ANY>`, something we generally don't do in vctrs (like falling back to deparsing or chopping). Combinations with types outside the sfc hierarchy should still be an error.

A similar type to sfc is `list_of()`, and the abstract geometry list is like a `list()`. Maybe a first step towards supporting sfc would be to implement a variant of `list_of()` that falls back to list as soon as one element doesn't have a common type?

(Ideally combining a `sfc<polygon>` and a `sfc<multipolygon>` would return the latter. Combining with something that isn't compatible falls back to the abstract type `sfc<ANY>`. These more refined rules aren't implemented yet though.)

lionel-:
If we can't solve this for vctrs 0.3.0, we could special-case sfc until we figure this out.

lionel-:
I'm a bit wary about providing general hooks in the coercion mechanism because I think it needs to be clearly structured to keep the system transparent. But maybe the way of supporting these special classes (AsIs and sfc) would be a second layer of double dispatch called from `vec_default_ptype2()` and that allows inheritance. We wouldn't document this very prominently.

krlmlr:
Is this similar to templates in C++? sfc_POINT is more like sfc<POINT>.

I'd like to propose `vec_stype(x, ...)` with implementations: POSIXct -> POSIXt, POSIXlt -> POSIXt, sfc_* -> sfc, AsIs -> `NextMethod()` . `vec_ptype2()` would dispatch over `vec_stype(x)` and `vec_stype(y)` without inheritance, `vec_stype()` is a regular S3 method with full S3 dispatch (perhaps with inheritance prohibited?) that needs to be registered in the vctrs namespace.

lionel-:
@krlmlr Left and right would be needed for cast and I'd prefer ptype2 to be symmetric. It also simplifies implementations because you know exactly which argument was dispatched on.  I don't think these generics should have short names because they would be experimental stopgaps. Finally POSIXlt is not really a concern.

lionel-:
[...]

PR #1024, opened 2020-04-20
Fix compatibility with `sfc` classes

lionel-:
Part of #989

The sfc classes rely on inheritance so we don't have a great alternative to offer yet.

Also import sf unit tests locally to make sure they keep working.

### DIFF
--- a/tests/testthat/test-type-sf.R
+++ b/tests/testthat/test-type-sf.R
@@ -1,0 +1,84 @@
+
+skip_if_not_installed("sf")
+
+# Avoids adding `sf` to Suggests
+import_from("sf", c(
+  "st_sfc",
+  "st_point",
+  "st_bbox",
+  "st_precision",
+  "st_crs",
+  "st_linestring"
+))
+
+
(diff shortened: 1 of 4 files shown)


tests/testthat/test-type-sf.R
```diff
@@ -0,0 +1,84 @@
...
+
+# Keep these tests in sync with
+# https://github.com/r-spatial/sf/blob/master/tests/testthat/test_vctrs.R
+
+test_that("`sfc` vectors are treated as vectors", {
+	expect_true(vctrs::vec_is(st_sfc(st_point())))
+})
+
+test_that("`n_empty` attribute of `sfc` vectors is restored", {
+	pt1 = st_sfc(st_point(c(NA_real_, NA_real_)))
```
hadley, line 32: Looks like the indenting is off here?

lionel-: yup the sf codebase uses tabs. Since this is imported from upstream it's better not to make any changes to keep the diffs clean.

PR #1026, opened 2020-04-21
Explain why classes are incompatible

lionel-:
We currently have this confusing error message when the is-same-class fallback fails because of diverging attribtues.

```r
vec_cast(foobar(1, bar = TRUE), foobar(2, baz = TRUE))
#> Error: Can't convert <vctrs_foobar> to <vctrs_foobar>.

vec_ptype2(foobar(1, bar = TRUE), foobar(2, baz = TRUE))
#> Error: Can't combine <vctrs_foobar> and <vctrs_foobar>.
```

With this PR, we explain the attributes are incompatible and refer to a FAQ.

```r
vec_cast(foobar(1, bar = TRUE), foobar(2, baz = TRUE))
#> Error: Can't convert <vctrs_foobar> to <vctrs_foobar>.
#> ✖ Some attributes are incompatible.
#> ℹ The author of the class should implement vctrs methods.
#> ℹ See <https://vctrs.r-lib.org/reference/faq-error-incompatible-attributes.html>.

vec_ptype2(foobar(1, bar = TRUE), foobar(2, baz = TRUE))
#> Error: Can't combine <vctrs_foobar> and <vctrs_foobar>.
#> ✖ Some attributes are incompatible.
#> ℹ The author of the class should implement vctrs methods.
#> ℹ See <https://vctrs.r-lib.org/reference/faq-error-incompatible-attributes.html>.
```

The FAQ is not written yet. This is tracked in #1016.


R/conditions.R
```diff
@@ -232,6 +232,15 @@ cnd_type_message <- function(x,
   if (is_null(types)) {
     x_type <- vec_ptype_full(x)
     y_type <- vec_ptype_full(y)
+
+    if (identical(x_type, y_type)) {
```
DavisVaughan, line 236: I know we don't have a great method for determining when to add this extra message, but I'm a bit worried about using `vec_ptype_full()`. Some types use attribute data when constructing the full ptype string, which would result in this message not being applied.

```r
> vctrs::vec_ptype_full(new_datetime(tzone = "America/New_York"))
[1] "datetime<America/New_York>"
> vctrs::vec_ptype_full(new_datetime(tzone = "America/Pacific"))
[1] "datetime<America/Pacific>"
```

What if we just strip all attributes and compare the underlying raw data?

lionel-: Actually I think I meant to check for `class(x)[[1]]` originally, to determine whether to show the warning.

But now I realise the whole thing is not well founded. My reasoning was that we would only end up here if the is-same-class fallback failed, but we could end up here from an actual ptype2 method if the common type determination failed.

lionel-: We now check for `class()` instead of the type label. And we detect when we are called from dispatch. In that case we don't display we don't display the attributes advice.

PR #1027, opened 2020-04-21
Improve data frame coercion errors and warnings

lionel-:
Branched from #1026.

When two data frame classes are incompatible because of their class but not because of their columns, the structure of columns no longer appear in the error message.

```r
df <- data.frame(x = 1, y = "")
foo <- structure(df, class = c("vctrs_foo", "data.frame"))
bar <- structure(df, class = c("vctrs_bar", "data.frame"))
```

Before:

```r
vec_cast(foo, bar)
#> Error: Can't convert <vctrs_foo<
#>   x: double
#>   y: character
#> >> to <vctrs_bar<
#>   x: double
#>   y: character
#> >>.
```

After:

```r
vec_cast(foo, bar)
#> Error: Can't convert <vctrs_foo[,2]> to <vctrs_bar[,2]>.
```


tests/testthat/error/test-type-data-frame.txt
```diff
@@ -6,23 +6,23 @@ combining data frames with foreign classes uses fallback
 > bar <- structure(mtcars[4:6], class = c("bar", "data.frame"))
 > baz <- structure(mtcars[7:9], class = c("baz", "data.frame"))
 > vec_ptype_common(foo, bar, baz)
-Warning: Can't combine <foo> and <bar>.
+Warning: Can't combine <foo[,3]> and <bar[,3]>.
```
hadley, line 9: What does `[,3]` mean?

hadley: Oh it's the number of columns? I find this super confusing so I think you should just drop it, and return to the previous `vec_ptype_abbr()`.

R/type-data-frame.R
```diff
@@ -60,7 +60,12 @@ vec_ptype_full.data.frame <- function(x, ...) {
 
 #' @export
 vec_ptype_abbr.data.frame <- function(x, ...) {
-  paste0("df", vec_ptype_shape(x))
+  if (inherits_only(x, "data.frame")) {
+    abbr <- "df"
```
hadley, line 64: Maybe this should be `data.frame`?

lionel-: Is this a bit too long for an abbreviation? I think the idea for this generic was to be the vctrs version of pillar's `type_sum()`.

tests/testthat/error/test-type-data-frame.txt
lionel-: Should we drop it from errors too? Right now I've just made it consistent with errors.

lionel-: Though I guess errors returned the full column spec until recently.

hadley: Yeah, exactly.

R/type-data-frame.R
hadley: Oh yeah, leave as is and we'll reconsider when we think about what these functions really should be called.

lionel-:
@hadley Dimensions omitted from messages.

Also we now distinguish between the incompatible attributes and incompatible classes cases.

The warning for incompatible attributes recommends contacting the author of the class:

```r
foo_bar <- foobar(mtcars, bar = TRUE)
foo_boo <- foobar(mtcars, boo = TRUE)

vec_ptype2(foo_bar, foo_boo)
#>  [1] mpg  cyl  disp hp   drat wt   qsec vs   am   gear carb
#> Warning message:
#> Can't combine <vctrs_foobar> and <vctrs_foobar>.
#> ✖ Some attributes are incompatible.
#> ℹ The author of the class should implement vctrs methods.
#> ℹ See <https://vctrs.r-lib.org/reference/faq-error-incompatible-attributes.html>.
#> ℹ Falling back to <data.frame>.
```

The warning for incompatible classes recommends normalising the inputs. This warning now references (yet to be written) user FAQ:

```r
foobar <- foobar(mtcars)
baz <- structure(mtcars, class = c("vctrs_baz", "data.frame"))
vec_ptype2(foobar, baz)
#>  [1] mpg  cyl  disp hp   drat wt   qsec vs   am   gear carb
#> Warning message:
#> Can't combine <vctrs_foobar> and <vctrs_baz>.
#> ℹ Convert all inputs to the same class to avoid this warning.
#> ℹ See <https://vctrs.r-lib.org/reference/faq-warning-convert-inputs.html>.
#> ℹ Falling back to <data.frame>.
```

PR #1031, opened 2020-04-21
Remove `"list"` base class from list-of

DavisVaughan:
Reverts part of #1003 

After speaking with @lionel-, it seems that adding a base `"list"` class to list-of might not be the best idea in the long run. In the future, we might allow classes that explicitly inherit from a base type to be coercible to those base types automatically, but we probably wouldn't want that with list-ofs (we want them to be strict and not lose information easily).

This required tweaking `vec_is_list()` to know about `vctrs_list_of` and allow explicit inheritance from `"vctrs_list_of"` to be a sufficient condition for being a list.

This additionally required that `new_vctr()` be less strict about `inherit_base_type = FALSE` with list input. We need to do this in `new_list_of()`. Separately, this is also how we would create list-vctrs that are treated as atomic vectors. `numeric_version` is one such example of this. It is implemented internally as a list, but it should not be treated as one. So `vec_is()` should return `TRUE`, but `vec_is_list()` should return `FALSE`.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,13 +1,13 @@
 
 # vctrs (development version)
 
-* `new_vctr()` now always appends a base `"list"` class to list `.data` to
-  be compatible with changes to `vec_is_list()`. This affects `new_list_of()`,
-  which now returns an object with a base class of `"list"`.
+* `new_vctr()` now defaults to `inherit_base_type = TRUE` for list `.data`, as
+  generally they should be considered lists by `vec_is_list()`.
 
 * `vec_is_list()` no longer allows S3 lists that implement a `vec_proxy()`
   method to automatically be considered lists. A S3 list must explicitly
-  inherit from `"list"` in the base class to be considered a list.
+  inherit from `"list"` or `"vctrs_list_of"` in the base class to be considered
+  a list.
 
 * `vec_rbind()` gains option to treat input names as row names. This
   is disabled by default (#966).
--- a/R/assert.R
+++ b/R/assert.R
@@ -154,7 +154,7 @@ vec_is_vector <- function(x) {
 #' returns `TRUE` if:
 #'
 #' * `x` is a bare list with no class.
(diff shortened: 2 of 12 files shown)


hadley:
I strongly believe that `list_of` should inherit from list — I don't think we should be adding to the base of our inheritance hierarchy without extreme need.

lionel-:
> I strongly believe that list_of should inherit from list

This is not about making the decision that list-of should not inherit from list. This is about not making the decision right away that it should inherit from list, for the reasons outlined by Davis. I think we need more time.

lionel-:
@jennybc made the good point on Slack that not inheriting from list prevents the list-of vectors created by tidyr to dispatch on list methods. It seems that this is reason enough to go forward with the list inheritance to avoid friction for tidyr users.

@DavisVaughan This makes me think that even if we do end up allowing cast-to-base-class, maybe we could just add a strict parameter if needed. The stricter list-of would not be castable to list, and would not allow list dispatch? The default lenient version of list-of would be easier to use for tidyr users.

NEWS.md
```diff
@@ -1,13 +1,13 @@
...
-* `new_vctr()` now always appends a base `"list"` class to list `.data` to
-  be compatible with changes to `vec_is_list()`. This affects `new_list_of()`,
-  which now returns an object with a base class of `"list"`.
+* `new_vctr()` now defaults to `inherit_base_type = TRUE` for list `.data`, as
+  generally they should be considered lists by `vec_is_list()`.
 
 * `vec_is_list()` no longer allows S3 lists that implement a `vec_proxy()`
   method to automatically be considered lists. A S3 list must explicitly
-  inherit from `"list"` in the base class to be considered a list.
+  inherit from `"list"` or `"vctrs_list_of"` in the base class to be considered
```
hadley, line 9: If this is just a stopgap, I don't think we should advertise the `vctrs_list_of` special case

PR #1034, opened 2020-04-24
Fix data frame fallback

### LINKED ISSUE #1028: vec_cast() needs fallback support for data frames too
``` r

xy <- readr::read_csv("x,y\n1,2\n3,4", col_types = "dc")
xy2 <- tibble(x = 3.4, y = 'd')

bind_rows(xy, xy2)
#> Warning: Can't combine <spec_tbl_df> and <tbl_df>.
#> ℹ Convert all inputs to the same class to avoid this warning.
#> ℹ Falling back to <data.frame>.
#> Error: Can't convert <spec_tbl_df<
#>   x: double
#>   y: character
#> >> to <data.frame<
#>   x: double
#>   y: character
#> >>.
```

<sup>Created on 2020-04-21 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

lionel-:
* Don't fall back to data frame by default, only from `vec_rbind()` and `vec_cbind()`.

* Add `vec_cast()` fallback. The binding functions would previously only work if the is-same-type fallback kicked in, which is not the case when attributes differ. Closes #1028.

### DIFF
--- a/R/type2.R
+++ b/R/type2.R
@@ -84,24 +84,47 @@ vec_default_ptype2 <- function(x, y, ..., x_arg = "", y_arg = "") {
     return(vec_ptype(x, x_arg = x_arg))
   }
 
-  if (is_df_subclass(x) && is.data.frame(y)) {
-    return(vec_ptype2_df_fallback(x, y))
-  }
-  if (is_df_subclass(y) && is.data.frame(x)) {
-    return(vec_ptype2_df_fallback(x, y))
+  internal <- ptype2_params(...)
+
+  if (internal$df_fallback) {
+    if (is_df_subclass(x) && is.data.frame(y)) {
+      return(vec_ptype2_df_fallback(x, y))
+    }
+    if (is_df_subclass(y) && is.data.frame(x)) {
+      return(vec_ptype2_df_fallback(x, y))
+    }
   }
 
+  # The from-dispatch parameter is set only when called from our S3
+  # dispatch mechanism, when there is no methods. It indicates whether
+  # the error message should provide advice about diverging attributes.
   stop_incompatible_type(
     x,
     y,
     x_arg = x_arg,
     y_arg = y_arg,
-    `vctrs:::from_dispatch` = from_dispatch(...)
+    `vctrs:::from_dispatch` = internal$from_dispatch
   )
 }
 
+ptype2_params <- function(...,
+                          `vctrs:::df_fallback` = FALSE) {
+  list(
+    from_dispatch = from_dispatch(...),
+    df_fallback = `vctrs:::df_fallback`
+  )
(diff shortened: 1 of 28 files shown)


tests/testthat/error/test-type-data-frame.txt
```diff
@@ -33,3 +33,37 @@ i Falling back to <data.frame>.
 [1] mpg  cyl  disp qsec vs   am   hp   drat wt  
 <0 rows> (or 0-length row.names)
 
+> invisible(vec_rbind(foo, data.frame(), foo))
+Warning: Can't combine <foo> and <data.frame>.
+i Convert all inputs to the same class to avoid this warning.
+i See <https://vctrs.r-lib.org/reference/faq-warning-convert-inputs.html>.
+i Falling back to <data.frame>.
```
hadley, line 40: I think this bullet should come first

hadley: Or maybe the first sentence should be `Can't combine <foo> and <data.frame>; falling back to <data.frame>`.

lionel-: The fallback is now mentioned in the header.

R/type2.R
```diff
@@ -84,26 +84,51 @@ vec_default_ptype2 <- function(x, y, ..., x_arg = "", y_arg = "") {
...
+    if (is_df_subclass(x) && is.data.frame(y)) {
+      return(vec_ptype2_df_fallback(x, y))
+    }
+    if (is_df_subclass(y) && is.data.frame(x)) {
+      return(vec_ptype2_df_fallback(x, y))
+    }
   }
 
+  # The from-dispatch parameter is set only when called from our S3
+  # dispatch mechanism, when there is no methods. It indicates whether
```
DavisVaughan, line 99:
```suggestion
  # dispatch mechanism, when no method is found to dispatch to. It indicates whether
```

src/vctrs.h
```diff
@@ -383,6 +382,11 @@ SEXP vec_group_loc(SEXP x);
 SEXP vec_match_params(SEXP needles, SEXP haystack, bool na_equal,
                       struct vctrs_arg* needles_arg, struct vctrs_arg* haystack_arg);
 
+#include "cast.h"
+static inline SEXP vec_cast(SEXP x, SEXP to, struct vctrs_arg* x_arg, struct vctrs_arg* to_arg) {
```
DavisVaughan, line 386: Is it worth it to move this and `vec_cast_common` and maybe `vec_cast_e` to `cast.h`?

lionel-: Yes, I think we should clean the header files in one go after the release.

PR #1036, opened 2020-04-24
Alter `new_data_frame()` signature

### LINKED ISSUE #949: `new_data_frame()` signature collisions
See https://github.com/tidyverse/dplyr/issues/5042

The `n` argument in `new_data_frame()` is causing issues because you can't set an attribute named `n`. The arguments should all probably be prefixed with a `.` to prevent other collisions.

@lionel- has mentioned in the past that `n` should probably be `size`, so this seems as good a time as any to change it to:

```
new_data_frame(.x = list(), .size = NULL, ..., .class = NULL)
```

DavisVaughan:
Closes #949

This is a breaking change, but I think it will be good in the long run. The signature is now:

```r
new_data_frame(x = list(), n = NULL, ..., class = NULL)

|
v

new_data_frame(.x = list(), ..., .size = NULL, .class = NULL)
```

This should avoid most collisions.

While I was there, I standardized the internal code to use `size` where we used to use `n`.

Two tests that used to splice in attributes were removed, since they splice in a `class` attribute and the signature now expects `.class`.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,6 +1,10 @@
 
 # vctrs (development version)
 
+* The signature of `new_data_frame()` has been changed to avoid collisions
+  when setting attributes with names that are identical to the argument
+  names (#949).
+
 * `new_vctr()` now always appends a base `"list"` class to list `.data` to
   be compatible with changes to `vec_is_list()`. This affects `new_list_of()`,
   which now returns an object with a base class of `"list"`.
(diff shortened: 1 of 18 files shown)


hadley:
I think we should wait until the next major release before merging this.

DavisVaughan:
We have ultimately decided that this is just Not Worth It™️ 

The current signature isn't that bad, conflicts should be rather rare. It came up here but this seems not very common https://github.com/tidyverse/dplyr/issues/5042#issuecomment-603805249

PR #1037, opened 2020-04-24
Add direct double dispatch

### LINKED ISSUE #985: Simplify double dispatch registration and lookup
Regarding #982, I wonder if we could simplify the registration by registering all methods for `vec_ptype2()` and `vec_cast()` in the vctrs namespace.

For instance you'd register `vec_ptype2.character.foobar` as a `character.foobar` method for `vec_ptype2` instead of a `foobar` method for `vec_ptype2.character`. It feels like this should work?

That would simplify the registration process because the left methods wouldn't need to be exported or registered at all. Since they are no longer evaluated, it also feels cleaner that they don't need to exist. There wouldn't be multiple namespaces involved in the dispatch or registration, making the whole system more transparent and easier to document and use.

lionel-:
Closes #985.

ptype2 and cast methods are now as simple to implement as regular S3 methods.


NEWS.md
```diff
@@ -75,17 +75,22 @@
...
-  instance it is no longer needed to supply a second argument as in
-  `UseMethod("vec_ptype2.mytype", y)`.
+  ```
+  #' @export
+  vec_ptype2.foo.bar <- function(x, y, ...) new_foo()
+  ```
+
+  vctrs also takes care of implementing the default and unspecified
+  methods. If you have implemented these methods, they are no longer
+  called and can now be removed.
```
hadley, line 88: But if you do have them, everything will continue to work, right?

lionel-: Right.

PR #1043, opened 2020-04-25
Fix normalisation of data frames

lionel-:
Don't empty the data frame before assignment as that may remove the class. Patch the names after assignment instead.

Fixes three revdep failures.


hadley:
That's what I was trying to warn about in my comment 😄

lionel-:
oh I must have missed your comment?

PR #1055, opened 2020-04-27
Update to latest GitHub Actions

DavisVaughan:


DavisVaughan:
Documenting some new R 4.0.0 behavior that has to do with these failures.

Previously, `SET_VECTOR_ELT()` would not touch the `NAMED()`ness of a SEXP. So if you assign a fresh vector into a list, then that vector would still not be referenced, i.e. `MAYBE_REFERENCED(VECTOR_ELT(x, i))` would return false.

Now, it seems like reference counting tracks this, and `SET_VECTOR_ELT()` bumps the refcount. You can see that here
https://github.com/wch/r-source/blob/bf4cf8912596162a3ae5a2b7da1a74220105b172/src/main/memory.c#L3996

With `FIX_REFCNT()` being a new part of R 4.0.0 that does this
https://github.com/wch/r-source/blob/bf4cf8912596162a3ae5a2b7da1a74220105b172/src/main/memory.c#L1228-L1245

We might need to revise our usage of `MAYBE_REFERENCED()` in light of this, instead moving to using `MAYBE_SHARED()`

PR #1056, opened 2020-04-27
Support `ptype` in `vec_c()` fallback

lionel-:
Fixes https://github.com/r-quantities/units/issues/239

### DIFF
--- a/src/c.c
+++ b/src/c.c
@@ -29,8 +29,8 @@ SEXP vec_c(SEXP xs,
            const struct name_repair_opts* name_repair) {
   R_len_t n = Rf_length(xs);
 
-  if (needs_vec_c_fallback(xs)) {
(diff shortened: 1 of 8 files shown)


src/c.c
```diff
@@ -146,6 +146,10 @@ bool needs_vec_c_fallback(SEXP xs) {
     return false;
   }
 
+  if (ptype != R_NilValue && !equal_object(r_class(x), r_class(ptype))) {
```
DavisVaughan, line 149: I guess the class check is good enough? I think thats the only thing that was mildly surprising here

lionel-: hmm I think we don't want to make that check when `ptype` is `NULL`, otherwise we'd return false without continuing the fallback.

PR #1057, opened 2020-04-27
Fix performance in R 4.0.0

### LINKED ISSUE #838: Switch to `MAYBE_SHARED()` in `r_maybe_duplicate()`

DavisVaughan:
Closes #838 

Requires https://github.com/r-lib/rlang/pull/963 to ensure that the NAMEDNESS of captured `...` is always set to the max value.

Unfortunately R 4.0.0 is going to cause a huge performance regression to `vec_rbind()`. It is very similar to the problem in https://github.com/r-lib/vctrs/issues/984

Here is current master, with R 4.0.0. This is supposed to take ~20ms.

```r
library(vctrs)
library(tibble)

x <- rep(list(data.frame(x = 1)), 10000)

fn <- function(x) {
  vec_rbind(!!!x)
}

# master
bench::mark(fn(x), iterations = 10)
#> # A tibble: 1 x 6
#> 1 fn(x)         285ms    311ms      3.12     764MB     54.2

# this PR
bench::mark(fn(x), iterations = 10)
#> # A tibble: 1 x 6
#> 1 fn(x)        19.8ms   21.7ms      43.0     200KB     18.4
```

The performance regression here is basically the same as in https://github.com/r-lib/vctrs/issues/984, in `df_assign()` we are making a copy of the final output container every time we assign one of the 10000 data frames into it.

The main issue is that `SET_VECTOR_ELT()` now bumps the refcnt of the `value` you are assigning, where it didn't used to do that. This makes `MAYBE_REFERENCED()` return true more often, in cases where we really don't need to make a duplicate. The fix is to switch to `MAYBE_SHARED()`, but also requires a patch to rlang on R < 4.0.0 to ensure that the result of rlang's `rlang_env_dots_list()` C callable returns the captured `...` with max namedness. This is important because it forces us to always copy the dots so we don't accidentally modify by reference (the copying was already happening when we were using `MAYBE_REFERENCED()`).

I've left very detailed notes about this in one of the tests
https://github.com/r-lib/vctrs/pull/1057/files#diff-51c944df297abe15044974191b4f4ebbR593-R627

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -45,8 +45,7 @@ Suggests:
     xml2,
     zeallot
 Remotes: 
-    r-lib/pillar,
-    r-lib/rlang#963
+    r-lib/pillar
 VignetteBuilder: 
     knitr
 Encoding: UTF-8
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -45,7 +45,8 @@ Suggests:
     xml2,
     zeallot
 Remotes: 
-    r-lib/pillar
+    r-lib/pillar,
+    r-lib/rlang#963
 VignetteBuilder: 
     knitr
 Encoding: UTF-8
--- a/R/bind.R
+++ b/R/bind.R
@@ -162,7 +162,7 @@ vec_cbind <- function(...,
                       .ptype = NULL,
                       .size = NULL,
                       .name_repair = c("unique", "universal", "check_unique", "minimal")) {
-  .External2(vctrs_cbind, .ptype, .size, .name_repair)
+  .External(vctrs_cbind, list2(...), .ptype, .size, .name_repair)
 }
 vec_cbind <- fn_inline_formals(vec_cbind, ".name_repair")
 
--- a/R/bind.R
+++ b/R/bind.R
@@ -147,7 +147,7 @@ vec_rbind <- function(...,
                       .ptype = NULL,
                       .names_to = rlang::zap(),
                       .name_repair = c("unique", "universal", "check_unique")) {
-  .External2(vctrs_rbind, .ptype, .names_to, .name_repair)
+  .External(vctrs_rbind, list2(...), .ptype, .names_to, .name_repair)
 }
 vec_rbind <- fn_inline_formals(vec_rbind, ".name_repair")
 
--- a/R/c.R
+++ b/R/c.R
@@ -47,6 +47,6 @@ vec_c <- function(...,
                   .ptype = NULL,
                   .name_spec = NULL,
                   .name_repair = c("minimal", "unique", "check_unique", "universal")) {
-  .External2(vctrs_c, .ptype, .name_spec, .name_repair)
+  .External(vctrs_c, list2(...), .ptype, .name_spec, .name_repair)
 }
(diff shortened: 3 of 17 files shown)


lionel-:
I'd rather not rush the ownership fixes in rlang. For reference here are the perf diffs when we change to taking dots with `list2()` at the R level:

```r
bench::mark(
  .External2 = vec_c(1, 1),
  .External = vec_c2(1, 1),
  .Extlist2 = vec_c3(1, 1),
  list2 = list2(1, 1),
  base = c(1, 1),
  iterations = 100000,
  check = FALSE
)[1:8]
#> # A tibble: 5 x 8
#>   expression      min   median `itr/sec` mem_alloc `gc/sec`  n_itr  n_gc
#>   <bch:expr> <bch:tm> <bch:tm>     <dbl> <bch:byt>    <dbl>  <int> <dbl>
#> 1 .External2   3.69µs   4.72µs   198184.        0B     5.95  99997     3
#> 2 .External    4.09µs   5.17µs   180357.        0B     5.41  99997     3
#> 3 .Extlist2    5.19µs   6.38µs   146162.        0B     7.31  99995     5
#> 4 list2        1.73µs   2.43µs   381401.        0B    11.4   99997     3
#> 5 base          163ns    193ns  4470507.        0B     0    100000     0
```

In the second line we change `.External2()` by `.External()`. In the third line we use `.External()` with dots captured at the R level. The fourth and fifth lines show `list2()` and `c()` for reference. In any case we're still much much slower than `c()`, but if possible I'd prefer to keep collecting dots at C level. Also it's probably worth keeping `.External2()` for performance-critical functions.

That said, this isn't very important for dplyr 1.0. It will become more important once we export `v()` from funs and people start using `vec_c()` in their scripts, including in loops. In that case we'll be able to improve performance by:

* Collecting dots faster in simple cases: r-lib/rlang#937.
* Bypassing all the coercion stuff for non-S3 base types.

Since that's not important right now, and since the performance difference with `c()` is still so big, I suggest we just collect dots at the R level, which should fix the refcounts.

DavisVaughan:
@lionel- ready for another round of review. Dropped the rlang Remote in favor of using `list2()` and `dots_values()`, as you suggested.

lionel-:
Could you add other more specific tests to make sure the refcounts are correct and we're not mutating inputs please? This repair-names test seems a bit indirect and it might be refactored away in the future.

PR #1062, opened 2020-04-29
Fix performance issues in R 4.0.0 with an `owned` argument for assignment

DavisVaughan:
Closes #1057 

After speaking with @lionel- about this, it seems like this is an alternate approach to fixing the performance issues seen in R 4.0.0.

@lionel- also reminded me that this is very similar to the old way `vec_assign()` used to work, with a `clone` parameter.

This approach uses an `owned` argument to decide how to _potentially_ duplicate the proxy. The core change is:

```c
if (owned) {
  out = r_maybe_duplicate_shared(x);
} else {
  out = r_maybe_duplicate(x);
}
```

Where:
- `r_maybe_duplicate_shared()` duplicates if `refcount(x) > 1`
- `r_maybe_duplicate()` duplicates if `refcount(x) > 0`

This distinction matters on R 4.0.0 now, because freshly created data frames created through `vec_init()` will have columns that all have a refcount of 1. The increment of the refcount happens when the columns are assigned into the data frame with `SET_VECTOR_ELT()`.

Practically this makes a difference in `vec_rbind()` where we assign into a fresh data frame in a loop, repeatedly calling `df_assign()`. Since each column is already _referenced_, the old scheme with `r_maybe_duplicate()` would duplicate the column at every iteration of the rbind loop. The new scheme allows us to take advantage of `owned = true` here because each column is referenced _but not shared_.

The usage of `owned = true` should be done incredibly sparingly. It should only be done when you recursively own the data structure you just created, and are filling it in some way.

### DIFF
--- a/tests/testthat/test-bind.R
+++ b/tests/testthat/test-bind.R
@@ -257,6 +257,16 @@ test_that("can assign row names in vec_rbind()", {
   expect_identical(out, exp)
 })
 
+test_that("monitoring: name repair while rbinding doesn't modify in place", {
+  df <- new_data_frame(list(x = 1, x = 1))
+  expect <- new_data_frame(list(x = 1, x = 1))
+
+  # Name repair occurs
+  expect_named(vec_rbind(df), c("x...1", "x...2"))
+
+  # No changes to `df`
+  expect_identical(df, expect)
+})
 
 # cols --------------------------------------------------------------------
 
@@ -441,6 +451,17 @@ test_that("vec_cbind() fails with arrays of dimensionality > 3", {
   a <- array(NA, c(1, 1, 1))
   expect_error(vec_cbind(a), "Can't bind arrays")
   expect_error(vec_cbind(x = a), "Can't bind arrays")
+})
+
+test_that("monitoring: name repair while cbinding doesn't modify in place", {
+  df <- new_data_frame(list(x = 1, x = 1))
+  expect <- new_data_frame(list(x = 1, x = 1))
+
+  # Name repair occurs
+  expect_named(vec_cbind(df), c("x...1", "x...2"))
+
+  # No changes to `df`
+  expect_identical(df, expect)
 })
 
 test_that("vec_rbind() consistently handles unnamed outputs", {
--- a/R/slice.R
+++ b/R/slice.R
@@ -183,8 +183,8 @@ vec_assign_seq <- function(x, value, start, size, increasing = TRUE) {
   .Call(vctrs_assign_seq, x, value, start, size, increasing)
 }
 
-vec_assign_params <- function(x, i, value, assign_names = FALSE) {
-  .Call(vctrs_assign_params, x, i, value, assign_names)
+vec_assign_params <- function(x, i, value, assign_names = FALSE, owned = FALSE) {
+  .Call(vctrs_assign_params, x, i, value, assign_names, owned)
 }
 
 vec_remove <- function(x, i) {
--- a/src/bind.c
+++ b/src/bind.c
@@ -150,7 +150,8 @@ static SEXP vec_rbind(SEXP xs, SEXP ptype, SEXP names_to, struct name_repair_opt
   R_len_t counter = 0;
 
   const struct vec_assign_opts bind_assign_opts = {
-    .assign_names = true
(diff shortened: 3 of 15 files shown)


DavisVaughan:
I do believe we will have to expose the `owned = true` case in the public API in some way. In slider, I use `vec_proxy_assign()` in a loop to fill an `out` object that I internally create with `vec_init()`. This is exactly the same kind of case as `vec_rbind()`, where performance can be crippled if `out` is a data frame.

For an example, when `slide_vec(.ptype = <data.frame>)` is used, performance suffers a lot from this on R 4.0.0.

```r
library(slider)
library(purrr)

df <- data.frame(x = 1:5000, y = 1:5000)
```

```r
# Master / R 3.6
bench::mark(slide_vec(df, identity, .ptype = df))
#> # A tibble: 1 x 6
#>   expression                              min median `itr/sec` mem_alloc
#>   <bch:expr>                           <bch:> <bch:>     <dbl> <bch:byt>
#> 1 slide_vec(df, identity, .ptype = df) 14.3ms 17.4ms      58.4     291KB

# This PR + R 4.0.0
bench::mark(slide_vec(df, identity, .ptype = df))
#> # A tibble: 1 x 6
#>   expression                             min median `itr/sec` mem_alloc `gc/sec`
#>   <bch:expr>                           <bch> <bch:>     <dbl> <bch:byt>    <dbl>
#> 1 slide_vec(df, identity, .ptype = df)  79ms   84ms      11.2     191MB     44.8
```

Performance doesn't suffer with non-data frame objects because their reference count is `0` when they are created with `vec_init()`. Only the columns of a freshly created data frame have the issue.

```r
# Master / R 3.6
bench::mark(slide_int(df$x, identity), map_int(df$x, identity), iterations = 200)
#> # A tibble: 2 x 6
#> 1 slide_int(df$x, identity)   2.21ms    2.8ms      353.    29.5KB     28.6
#> 2 map_int(df$x, identity)     1.97ms   2.22ms      445.    33.7KB     30.9

# This PR + R 4.0.0
bench::mark(slide_int(df$x, identity), map_int(df$x, identity), iterations = 200)
#> # A tibble: 2 x 6
#> 1 slide_int(df$x, identity)   2.57ms   3.08ms      331.    29.5KB     24.9
#> 2 map_int(df$x, identity)     2.16ms   2.43ms      409.    33.6KB     33.2
```

DavisVaughan:
I should also add that the original reason we went away from the `clone` parameter was outlined in this PR https://github.com/r-lib/vctrs/pull/842

It had to do with ALTREP, where `vec_init()` would construct a data frame with ALTREP compact-rep columns. We previously had `x = clone ? r_maybe_duplicate(x) : x;` which was a problem with this approach because in `vec_rbind()` we would set `clone = false`. This means that we would be incorrectly modifying ALTREP columns directly. They are marked as mutable so they should have been duplicated. This approach avoids that problem because we use `r_maybe_duplicate_shared(x)` even when we own `x`. ALTREP columns _are_ considered shared because by marking them as mutable we set their refcount to the max value, so they are correctly duplicated here.

DavisVaughan:
An additional note that I think continuing to use `MAYBE_REFERENCED()` (i.e. refcount > 0) to determine if we need to duplicate is the correct thing to do in 99% of the cases. It matches up with this note from Luke in https://developer.r-project.org/Refcnt.html

"This means that, except in very special and well understood circumstances, an argument passed down to C code should not be modified if it has a positive reference count, even if that count is equal to one."

So I think https://github.com/r-lib/vctrs/issues/838 could be closed because of this

tests/testthat/test-bind.R
```diff
@@ -257,6 +257,16 @@ test_that("can assign row names in vec_rbind()", {
   expect_identical(out, exp)
 })
 
+test_that("monitoring: name repair while rbinding doesn't modify in place", {
```
lionel-, line 260: What is this monitoring prefix?

DavisVaughan: I've used it in a few places. I don't have a better way to describe it other than "this tests very internal details that aren't really specific to functionality" https://github.com/r-lib/vctrs/search?q=monitoring&unscoped_q=monitoring

PR #1066, opened 2020-05-01
Mention loss of generality instead of precision when relevant

lionel-:
`maybe_lossy_cast()` gains a `loss_type` argument that can take values `c("precision", "generality")` to indicate the kind of lossy cast in the error message. We now use "generality" for character -> factor conversions.

Before:

```r
vec_cast("a", factor("b"))
#> Error: Can't convert from <character> to <factor<ddf10>> due to loss of precision.
#> * Locations: 1
```

After:

```r
vec_cast("a", factor("b"))
#> Error: Can't convert from <character> to <factor<ddf10>> due to loss of generality.
#> * Locations: 1
```

### DIFF
--- a/R/conditions.R
+++ b/R/conditions.R
@@ -413,6 +413,9 @@ stop_incompatible_size <- function(x,
 #'   Can also be a single `TRUE`, but note that `locations` picks up
 #'   locations from this vector by default. In this case, supply your
 #'   own location vector, possibly empty.
+#' @param loss_type The kind of lossy cast to be mentioned in error
+#'   messages. Can be loss of precision (for instance from double to
+#'   integer) or loss of generality (from character to factor).
 #' @param locations An optional integer vector giving the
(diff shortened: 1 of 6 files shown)


R/conditions.R
```diff
@@ -424,6 +427,7 @@ maybe_lossy_cast <- function(result, x, to,
                              lossy = NULL,
                              locations = NULL,
                              ...,
+                             loss_type = c("precision", "generality"),
```
hadley, line 430: Maybe just make it a string so it's a bit more flexible in the future? I'm not sure "loss of generality" will mean much to most people, and there doesn't seem to be a strong reason to limit the interface, IMO.

lionel-: I think the strong reason is that generation of messages based on arbitrary nouns is very difficult, and even more so when translations are involved. I think we should prefer the stricter interface by default.

PR #1067, opened 2020-05-01
Fix ordered combinations

lionel-:
Combining ordered factors of different levels in now an error:

```r
# Before
vec_ptype2(ordered("a"), ordered("b"))
#> ordered(0)
#> Levels: a < b

# After
vec_ptype2(ordered("a"), ordered("b"))
#> Error: Can't combine <ordered<127a2>> and <ordered<ddf10>>.
```

Casting factors and ordered is now an error, consistently with the new cast restrictions:

```r
# Before
vec_cast(ordered("a"), factor("a"))
#> [1] a
#> Levels: a

# After
vec_cast(ordered("a"), factor("a"))
#> Error: Can't convert <ordered<127a2>> to <factor<127a2>>.
```


hadley:
This is technically a breaking change, but it seems unlikely to affect much code in the wild.

R/type-factor.R
```diff
@@ -82,46 +82,50 @@ vec_ptype_abbr.ordered <- function(x, ...) {
...
 #' @export
-vec_ptype2.ordered <- function(x, y, ...) UseMethod("vec_ptype2.ordered")
-#' @method vec_ptype2.ordered character
+vec_ptype2.ordered <- function(x, y, ...) {
+  UseMethod("vec_ptype2.ordered")
+}
 #' @export
-vec_ptype2.ordered.character <- function(x, y, ...) character()
-#' @method vec_ptype2.character ordered
+vec_ptype2.ordered.ordered <- function(x, y, ..., x_arg = "", y_arg = "") {
```
DavisVaughan, line 109: `x_arg` and `y_arg` could be removed here, just for consistency

R/type-factor.R
```diff
@@ -188,15 +188,6 @@ vec_cast.ordered.ordered <- function(x, to, ...) {
   ord_cast(x, to, ...)
 }
 
-#' @export
-vec_cast.factor.ordered <- function(x, to, ...) {
```
DavisVaughan: Is it better to remove these altogether, or use `stop_incompatible_type()` as a reminder to ourselves that we have thought about this, and it wasn't just an oversight? It would also keep other packages from trying to add a factor.ordered method themselves

tests/testthat/test-type-factor.R
```diff
@@ -105,6 +105,37 @@ test_that("vec_ptype2() errors with malformed ordered factors", {
...
+test_that("ordered factors with different levels are not compatible", {
+  expect_error(
+    vec_ptype2(ordered("a"), ordered("b")),
+    class = "vctrs_error_incompatible_type"
+  )
+  expect_error(
+    vec_ptype2(ordered("a"), ordered(c("a", "b"))),
+    class = "vctrs_error_incompatible_type"
+  )
+})
```
DavisVaughan, line 117: Maybe also a `vec_cast(ordered("a"), ordered(c("a", "b")))` check to ensure that `vec_cast()` is doing the right thing now too?

R/type-factor.R
lionel-: I think for consistency it's better to remove them.

PR #1068, opened 2020-05-01
Review `?vec_ptype2`, `?vec_cast`, and `?vec_proxy`

### LINKED ISSUE #1005: cast.png is out of date
Since https://github.com/r-lib/vctrs/commit/67c49a1cd0be1c1cd8d8d8f7b806baa9de8569b8 (https://github.com/r-lib/vctrs/issues/606) the [image](https://github.com/r-lib/vctrs/blob/master/man/figures/cast.png) in the vec_cast documentation is out of date, and it seems [coerce.png](https://github.com/r-lib/vctrs/blob/master/man/figures/coerce.png) should replace it.

lionel-:
* Document `vec_restore()` with `vec_proxy()` instead of `vec_cast()`.

* Refer to FAQ pages in `vec_ptype2()` and `vec_cast()` documentation. Remove redundant and outdated contents from the `?` pages.

* Document `vec_default_ptype2()` and `vec_default_cast()` together and mark them as internal. On the other hand, `vec_ptype2()` and `vec_cast()` are no longer internal.

Closes #1005.


R/cast.R
```diff
@@ -1,95 +1,38 @@
-#' Cast a vector to specified type
+#' Cast a vector to the type of another vector
```
hadley, line 1: This title seems less clear to me

R/cast.R
```diff
@@ -1,95 +1,38 @@
...
 #'   in error messages to inform the user about the locations of
 #'   incompatible types (see [stop_incompatible_type()]).
 #' @return A vector the same length as `x` with the same type as `to`,
 #'   or an error if the cast is not possible. An error is generated if
 #'   information is lost when casting between compatible types (i.e. when
 #'   there is no 1-to-1 mapping for a specific value).
+#'
+#' @section Implementing `vec_cast()` methods:
+#'
+#' - For an overview of how these generics work and their roles in
```
hadley, line 25: I wonder if we could make`[?theory-faq-coercion]` generate these links?

R/proxy.R
```diff
@@ -1,28 +1,40 @@
-#' Extract underlying data
+#' Proxy and restore
 #'
 #' @description
 #'
-#' Extract the data underlying an S3 vector object, i.e. the underlying
-#' (named) atomic vector or list.
+#' `vec_proxy()` returns the memory representation of a vector. You
```
hadley, line 5: I don't know what "memory representation" means

R/proxy.R
```diff
@@ -1,28 +1,40 @@
...
+#' normally don't need to implement it, except in these cases:
 #'
-#' * `vec_data()` returns unstructured data. The only attributes
-#'   preserved are names, dims, and dimnames.
+#' - S3 lists are considered scalars by default. This is the safe
```
hadley, line 8: Wouldn't it be better to advertise using `list` in the class vector?

R/proxy.R
```diff
@@ -1,28 +1,40 @@
...
+#' - It restores attributes and classes. These may be lost when the
+#'   memory values are manipulated. For example slicing a subset of a
+#'   vector's proxy causes a new proxy to be allocated.
+#'
+#' By default vctrs restores all attributes and classes
+#' automatically. You only need to implement a `vec_restore()` method
+#' if your class has attributes that depend on the data.
+#'
+#' @param x A vector.
+#' @inheritParams ellipsis::dots_empty
```
hadley, line 35: I think this should be `@keywords` internal and mention that this mechanism is still under development.

R/type2.R
```diff
@@ -1,62 +1,35 @@
...
-#' does not work inside `vec_ptype2()` methods, and `default` methods
-#' are never called.
-#'
-#' See `vignette("s3-vector")` for full details.
-#' @keywords internal
+#' Find the common type for a pair of vectors
+#'
+#' @description
+#'
+#' `vec_ptype2()` provides directional conversions from one type of
```
hadley, line 5: It doesn't do conversion?

R/cast.R
lionel-: For now I have extracted the links into an Rmd file that can be included as needed. Because of the limitations of includeRmd, it seems that these links must share the same section title everywhere they are included though. But this works here.

R/proxy.R
lionel-: For factors it's the C array of `int`, and by extension the `INTSXP` that wraps it.

I have rephrased this way:

```
#' `vec_proxy()` returns the data structure containing the values of a
#' vector. This data structure is usually the vector itself. In this
#' case the proxy is the [identity function][base::identity], which is
#' the default `vec_proxy()` method.
#'
#' You normally don't need to implement `vec_proxy()`, except in these
#' cases:
```

R/proxy.R
lionel-: Good point. I have moved this bullet to last position, and I mention how this is only an alternative to adding `"list"` to the class vector.

```
#' - S3 lists are considered scalars by default. This is the safe
#'   choice for list objects such as returned by `stats::lm()`. To
#'   declare that your S3 list class is a vector, you normally add
#'   `"list"` to the right of your class vector. Explicit inheritance
#'   from list is generally the preferred way to declare an S3 list in
#'   R, for instance it makes it possible to dispatch on
#'   `generic.list` S3 methods.
#'
#'   If you can't modify your class vector, you can implement an
#'   identity proxy (i.e. a proxy method that just returns its input)
#'   to let vctrs know this is a vector list and not a scalar.
```

R/proxy.R
lionel-: Now marked as experimental.

R/type2.R
lionel-: This is now:

```
#' `vec_ptype2()` defines the coercion hierarchy for a set of related
#' vector types. Along with [vec_cast()], this generic forms the
#' foundation of type coercions in vctrs.
```

R/proxy.R
```diff
@@ -1,28 +1,54 @@
...
-#' * `vec_data()` returns unstructured data. The only attributes
-#'   preserved are names, dims, and dimnames.
+#' `vec_proxy()` returns the data structure containing the values of a
+#' vector. This data structure is usually the vector itself. In this
+#' case the proxy is the [identity function][base::identity], which is
+#' the default `vec_proxy()` method.
 #'
-#'   Currently, due to the underlying memory architecture of R, this
-#'   creates a full copy of the data.
+#' You normally don't need to implement `vec_proxy()`, except in these
```
hadley, line 12:
```suggestion
#' You only need to implement `vec_proxy()` if:
```

R/proxy.R
```diff
@@ -1,28 +1,54 @@
...
+#' cases:
 #'
-#' * `vec_proxy()` may return structured data. This generic is the
-#'   main customisation point in vctrs, along with [vec_restore()].
-#'   See the section below to learn when you should implement
-#'   `vec_proxy()`.
+#' - Some vector types have vectorised attributes, i.e. metadata for
```
hadley, line 15:
```suggestion
#' - A vector has vectorised attributes, i.e. metadata for
```

hadley: But maybe this bullet should be de-emphasised too, since it's only us that needs to implement this method?

R/proxy.R
```diff
@@ -121,7 +163,48 @@ vec_restore_dispatch <- function(x, to, ..., n = NULL) {
...
+#' Extract underlying data
+#'
+#' @description
+#'
+#' \Sexpr[results=rd, stage=render]{vctrs:::lifecycle("experimental")}
+#'
+#' Extract the data underlying an S3 vector object, i.e. the underlying
+#' (named) atomic vector or list.
+#'
+#' * `vec_data()` returns unstructured data. The only attributes
```
hadley, line 179: Maybe put this in a separate section? Like `vec_data()` vs `vec_proxy()`?

R/proxy.R
lionel-: What do you mean it's only us? What if you have a POSIXlt class?

hadley: Doesn’t the rcrd subclass take care of that for you?

lionel-: If you have a POSIXlt-like class you don't have a rcrd subclass.

The rcrd class is higher level than this documentation. In principle it can live in another package.

lionel-: I'll add a note that for classes started from scratch the rcrd class is the easiest way.

hadley: Yeah, just make it clear that this is for experts only.

lionel-: By the way this made me wonder if we should make `fields()` more generic by defining it on any vector whose proxy is a data frame. Though overall I think these functions are not really needed and it feels a bit wrong to use up unnamespaced short names like this, especially since vctrs is meant to be imported.

PR #1072, opened 2020-05-01
Refine `vec_assign()` ownership semantics

DavisVaughan:
Follow up to #1062 

This PR refines the ownership model of `vec_assign()`. It changes from a boolean ownership model to one with 3 states: owned, shared, or unknown.

It works similarly to how it was set up before, except for calls to `vec_proxy_assign_opts()`. Now, if `vec_proxy_assign_opts()` is called with `vctrs_ownership_unknown` then it will attempt to determine the ownership for the data structure by a call to `NO_REFERENCES()` on the proxy. If there are no references, it is owned.

This change helps with slider. If slider creates a data frame at the C level and then calls `vec_proxy_assign()` to fill it, a default of `vctrs_ownership_unknown` is assumed. But `NO_REFERENCES(proxy)` will be true on the fresh data frame, so the ownership is updated to `vctrs_ownership_owned`. The columns of the data frame are all passed this updated ownership value, so no copies are made.

In the obscure case where we create a data frame at the C level, and then set one of the columns in this freshly created data frame to be a vector that is already referenced somewhere else, we are still safe. Assigning into that fresh data frame will use `vctrs_ownership_owned`, which will be passed on to each column, but columns that are also referenced elsewhere will have a refcount of at least 2. One from being referenced elsewhere, and one from being set in this fresh data frame. So `r_maybe_duplicate_shared()` will copy it as needed to not modify it by reference accidentally. But again, I don't see this case ever arising.

### DIFF
--- a/src/c.c
+++ b/src/c.c
@@ -75,7 +75,7 @@ SEXP vec_c(SEXP xs,
 
   const struct vec_assign_opts c_assign_opts = {
     .assign_names = true,
-    .owned = true
+    .ownership = vctrs_ownership_owned
   };
 
   for (R_len_t i = 0; i < n; ++i) {
@@ -99,7 +99,7 @@ SEXP vec_c(SEXP xs,
       SEXP x_nms = PROTECT(apply_name_spec(name_spec, outer, inner, size));
 
       if (x_nms != R_NilValue) {
-        out_names = chr_assign(out_names, idx, x_nms, true);
+        out_names = chr_assign(out_names, idx, x_nms, vctrs_ownership_owned);
         REPROTECT(out_names, out_names_pi);
       }
 
--- a/src/c.c
+++ b/src/c.c
@@ -74,8 +74,7 @@ SEXP vec_c(SEXP xs,
   R_len_t counter = 0;
 
   const struct vec_assign_opts c_assign_opts = {
-    .assign_names = true,
-    .ownership = vctrs_ownership_owned
+    .assign_names = true
   };
 
   for (R_len_t i = 0; i < n; ++i) {
@@ -90,7 +89,7 @@ SEXP vec_c(SEXP xs,
 
     init_compact_seq(idx_ptr, counter, size, true);
 
-    out = vec_proxy_assign_opts(out, idx, elt, &c_assign_opts);
+    out = vec_proxy_assign_opts(out, idx, elt, vctrs_ownership_owned, &c_assign_opts);
     REPROTECT(out, out_pi);
 
     if (has_names) {
--- a/src/c.c
+++ b/src/c.c
@@ -89,6 +89,7 @@ SEXP vec_c(SEXP xs,
 
     init_compact_seq(idx_ptr, counter, size, true);
 
+    // Total ownership of `out` because it was freshly created with `vec_init()`
     out = vec_proxy_assign_opts(out, idx, elt, vctrs_ownership_total, &c_assign_opts);
     REPROTECT(out, out_pi);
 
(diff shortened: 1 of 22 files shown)


src/slice-assign.h
```diff
@@ -1,24 +1,48 @@
...
 #define VCTRS_SLICE_ASSIGN_H
 
+enum vctrs_ownership {
+  vctrs_ownership_owned,
+  vctrs_ownership_shared,
+  vctrs_ownership_unknown
+};
+
+static inline enum vctrs_ownership parse_ownership(SEXP ownership) {
+  if (TYPEOF(ownership) != STRSXP || Rf_length(ownership) != 1) {
```
lionel-, line 11:
```suggestion
  if (!r_is_string(ownership)) {
```

src/slice-assign.h
```diff
@@ -1,24 +1,48 @@
 #ifndef VCTRS_SLICE_ASSIGN_H
 #define VCTRS_SLICE_ASSIGN_H
 
+enum vctrs_ownership {
+  vctrs_ownership_owned,
+  vctrs_ownership_shared,
+  vctrs_ownership_unknown
+};
+
+static inline enum vctrs_ownership parse_ownership(SEXP ownership) {
```
lionel-, line 10: Move to .c? A header doesn't seem like a good place for this helper, and it doesn't seem important to have it inlinable.

src/c.c
```diff
@@ -90,7 +89,7 @@ SEXP vec_c(SEXP xs,
 
     init_compact_seq(idx_ptr, counter, size, true);
 
-    out = vec_proxy_assign_opts(out, idx, elt, &c_assign_opts);
+    out = vec_proxy_assign_opts(out, idx, elt, vctrs_ownership_owned, &c_assign_opts);
```
lionel-, line 92: Needs a comment that we own the `out` proxy because it was taken of a fresh `vec_init()`. Same for bind ops etc.

src/slice-assign.h
```diff
@@ -1,24 +1,48 @@
 #ifndef VCTRS_SLICE_ASSIGN_H
 #define VCTRS_SLICE_ASSIGN_H
 
+enum vctrs_ownership {
```
lionel-, line 4: Needs a comment that this is recursive ownership.

src/slice-assign.h
```diff
@@ -1,24 +1,48 @@
...
+  vctrs_ownership_owned,
```
lionel-, line 5: Should it be `vctrs_ownership_total`?

The first case is also the default case (= 0), it should be unknown.

DavisVaughan:
Alright @lionel-, I've made a few more simplifications.

We are now back to a binary ownership model: `vctrs_ownership_total` and `vctrs_ownership_shared`. I realized that there aren't actually any times where ownership should be "unknown".

- When we get input from the R side with `vec_assign()`, we proxy that input and then check the ownership of the proxy with `proxy_ownership()`. Because of this, ownership isn't unknown here.

- When `vec_proxy_assign()` is called from slider, we check the ownership of the provided proxy with `proxy_ownership()`. Again, this makes ownership known up front.

Other than those two cases, our internal usage of `ownership` is done in a way where we always know if we own the object or not.

This allowed me to make `ownership` `const` everywhere, which makes me much happier about this approach. That made me a little uncomfortable before.

Additionally, `ownership` is now strictly related to the `proxy`. This means it isn't an input to functions like `vec_assign_opts()` which take `x` and then create the `proxy` to pass on to `vec_proxy_assign_opts()`. I like this separation.

I've also removed the C level `vec_assign()`, which wasn't being used right now.

PR #1085, opened 2020-05-05
Make tibble subclass example more realistic

### LINKED ISSUE #1065: Coercion documentation
When reading this https://vctrs.r-lib.org/reference/theory-faq-coercion.html, found "Type compatibility" and "Richness of type" sections confusing. Primarily, the notion of "richer"/"rich" is ambiguous/changing when referring to 1d vectors and 2d data frames (acknowledged in the FAQ).

The `factor` vs `character` 1d example suggests that `character` is richer (general) than `factor` (specific), and hence a combination of two results in `character` (general). However, this  rationale doesn't hold for the following 2d example `data.table` (specific) vs `data.frame` (general), because we should choose richer behaviours instead of richer data. I could probably argue `factor` is richer in behaviours than `character`, since `factor` owns the `level()`. 

I'm not sure `data.table` vs `data.frame` is a good example (neither `tbl_df` vs `data.frame`). Would be `grouped_df` vs `data.frame` a better alternative? Any reason on `vec_ptype2(<grouped_df>, <data.frame>)` returns `<grouped_df>`? I think `data.frame` is richer than `grouped_df`, because it's more flexible with fewer attributes/constraints.

For 1d, we should choose a more general type, but 2d goes with a more specific type?

lionel-:
Closes #1065

### DIFF
--- a/man/faq/developer/howto-coercion-data-frame.Rmd
+++ b/man/faq/developer/howto-coercion-data-frame.Rmd
@@ -112,22 +112,35 @@ vec_cbind(data.frame(x = 1:3), data.table(y = "foo"))
 
 ## A tibble example
 
-In this example we implement coercion methods for a tibble subclass:
+In this example we implement coercion methods for a tibble subclass that carries a colour as metadata:
 
 ```{r}
 # User constructor
-my_tibble <- function(...) {
-  new_my_tibble(tibble::tibble(...))
+my_tibble <- function(colour = NULL, ...) {
+  new_my_tibble(tibble::tibble(...), colour = colour)
 }
 # Developer constructor
-new_my_tibble <- function(x) {
+new_my_tibble <- function(x, colour = NULL) {
   stopifnot(is.data.frame(x))
-  tibble::new_tibble(x, class = "my_tibble", nrow = nrow(x))
+  tibble::new_tibble(
+    x,
+    colour = colour,
+    class = "my_tibble",
+    nrow = nrow(x)
+  )
+}
+
+df_colour <- function(x) {
+  if (inherits(x, "my_tibble")) {
+    attr(x, "colour")
+  } else {
+    NULL
+  }
 }
 
 #'@export
 print.my_tibble <- function(x, ...) {
-  cat("<my_very_own_tibble_class>\n")
+  cat(sprintf("<my_tibble: %s>\n", df_colour(x)))
   cli::cat_line(format(x)[-1])
 }
 ```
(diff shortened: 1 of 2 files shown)


man/faq/developer/howto-coercion-data-frame.Rmd
```diff
@@ -112,22 +112,35 @@ vec_cbind(data.frame(x = 1:3), data.table(y = "foo"))
 
 ## A tibble example
 
-In this example we implement coercion methods for a tibble subclass:
+In this example we implement coercion methods for a tibble subclass that carries a colour as metadata:
```
hadley, line 115: Is this is a scalar or vector?

man/faq/developer/howto-coercion-data-frame.Rmd
```diff
@@ -112,22 +112,35 @@ vec_cbind(data.frame(x = 1:3), data.table(y = "foo"))
...
+    attr(x, "colour")
+  } else {
+    NULL
+  }
 }
 
 #'@export
 print.my_tibble <- function(x, ...) {
-  cat("<my_very_own_tibble_class>\n")
+  cat(sprintf("<my_tibble: %s>\n", df_colour(x)))
```
hadley, line 143: Should this use `class(x)[[1]]`? (so if you extend it the parent print method is still moderately useful? (tibble doesn't do this and it confused me a couple of weeks ago)

man/faq/developer/howto-coercion-data-frame.Rmd
```diff
@@ -112,22 +112,35 @@ vec_cbind(data.frame(x = 1:3), data.table(y = "foo"))
...
+  tibble::new_tibble(
+    x,
+    colour = colour,
+    class = "my_tibble",
+    nrow = nrow(x)
+  )
+}
+
+df_colour <- function(x) {
+  if (inherits(x, "my_tibble")) {
```
hadley, line 134: Do we need this for a simple example?

man/faq/developer/howto-coercion-data-frame.Rmd
```diff
@@ -141,47 +154,103 @@ knitr_local_registration("base::print", "my_tibble")
...
-vec_rbind(df, tibble::tibble(x = 10:12))
+vec_rbind(red, tibble::tibble(x = 10:12))
 ```
 
-Instead of falling back to a data frame, we would like to return a `<my_tibble>` when combined with a data frame or a tibble. This subclass of tibble is a supertype of tibble and data frame, i.e. it is the richer type.
+Instead of falling back to a data frame, we would like to return a `<my_tibble>` when combined with a data frame or a tibble. Because this subclass has more metadata than normal data frames (it has a colour), it is a _supertype_ of tibble and data frame, i.e. it is the richer type. This is similar to how a grouped tibble is a more general type than a tibble or a data frame. Conceptually, the latter are pinned to a single constant group.
+
+The coercion methods for data frames operate in two steps:
+
+- They check for compatible subclass attributes. In our case the tibble colour has to be the same, or be undefined.
```
hadley, line 176: Wouldn't we usually take the colour from the lhs? (like the tz?)

lionel-: In this case this is a data attribute, not a detail attribute, so the colour has to match. I should add a discussion about this.

man/faq/developer/howto-coercion-data-frame.Rmd
lionel-: I think so, it's used later in the implementations. It also implies that there is no additional behaviour for colour tibbles, in the sense that the colour interface is defined on data frames as well.

man/faq/developer/howto-coercion-data-frame.Rmd
lionel-: Good idea.

Issue #971, opened 2020-03-28
Missing vec_arith.POSIXlt()


DavisVaughan:
Should probably be similar to `vec_arith.POSIXct()`

Issue #997, opened 2020-04-08
Should `vec_arith.numeric` be split into `vec_arith.double` and `vec_arith.integer`?


DavisVaughan:
I feel like this would provide finer control, and would be a little less confusing

c-marc:
I'm not sure what's the roadmap for this but I find it quite confusing indeed.

It's even more confusing in the context of `vctrs` purpose, where it's emphasized everywhere that type should be strictly dealt with, and coercing/casting requires avoiding `numeric` and explicitly dealing with `integer` or `double`...

The documentation <https://vctrs.r-lib.org/articles/s3-vector.html#arithmetic-1> is misleading:

- "vctrs provides the hybrid S3 generics/methods for most of the base R types, like vec_arith.integer()" and states that the setup instructions of full import `@import vctrs` is enough. But `vec_arith.integer` is not exported from `vctrs`
- the example is not working and confusingly requires to tweak to `vec_arith.numeric integer` (see comments in code) to not throw and error. At least from my up-to-date CRAN install of `vctrs 0.4.2`:

```
#' @export
#' @method vec_arith.my_type my_type
vec_arith.my_type.my_type <- function(op, x, y, ...) {
  # implementation here
}

#' @export
#' @method vec_arith.my_type integer
vec_arith.my_type.integer <- function(op, x, y, ...) {
  # implementation here
}

#' @export
#' @method vec_arith.numeric my_type #CHANGE HERE
vec_arith.numeric.my_type <- function(op, x, y, ...) { #CHANGE HERE
  # implementation here
}
```

PR #1086, opened 2020-05-05
Implement `vec_arith()` and `vec_math()` for POSIXlt

DavisVaughan:
Closes https://github.com/r-lib/vctrs/issues/971

Required for https://github.com/tidyverse/lubridate/pull/871

I've also revised the tests a little bit so it is clearer to see that all of the methods are being tested symmetrically


lionel-:
Should we wait until next version before adding arith methods for lubridate? We haven't updated the double dispatch mechanism yet.

Alternatively we could switch to the new mechanism now, I don't expect it should have much impact on revdeps.

DavisVaughan:
If it was eventually implemented with the "legacy double dispatch" fallback mechanism then I didn't think it would matter too match to implement it in vctrs right now. lubridate could be updated later as required. The main reason I wanted to do it was to use it in slider, dplyr won't need it, so it can definitely wait.

lionel-:
I meant that the semantics would change (inheritance).

DavisVaughan:
Oh right. I hadn't thought about that too much for lubridate. I wasn't even planning on guarding against it.

I haven't seen many people subclassing Date, POSIXct, or lubridate Period objects. The only thing to potentially worry about with subclasses would be `<Period> * <numeric-subclass>` but that seems pretty rare too.

I guess the bigger problem would be that we currently export `vctrs::vec_arith.numeric` rather than individual methods for integer and double. This is https://github.com/r-lib/vctrs/issues/997. With the no-inheritance model I think we would switch to explicit integer/double methods, which would require changes in lubridate.

DavisVaughan:
I'm happy to wait on lubridate vec-arith methods until the next release of vctrs.

If there are no objections to adding POSIXlt support, I do think we could still do that now.

hadley:
I think we could change the dispatch mechanism if you wanted — would just need to submit PRs to the affected packages (https://github.com/search?q=org%3Acran+vec_arith&type=Code)

R/type-date-time.R
```diff
@@ -427,6 +439,37 @@ vec_arith.difftime.Date <- function(op, x, y, ...) {
     stop_incompatible_op(op, x, y)
   )
 }
+#' @method vec_arith.POSIXct difftime
+#' @export
+vec_arith.POSIXct.difftime <- function(op, x, y, ...) {
+  y <- vec_cast(y, new_duration(units = "secs"))
+
+  switch(op,
+         `+` = vec_restore(vec_arith_base(op, x, y), x),
```
hadley, line 448: Indenting

DavisVaughan: I hate that RStudio does this automatically

PR #1087, opened 2020-05-05
Improve consistency of recycling errors

lionel-:


tests/testthat/test-type-vec-size-common-error.txt
```diff
@@ -1,10 +1,10 @@
 
 vec_size_common(1:2, 1, 1:4):
 
-No common size for `..1`, size 2, and `..3`, size 4. 
+Can't recycle `..1` (size 2) and `..3` (size 4).
```
DavisVaughan: This seems like a strange error to me. I think recycling and finding a common size are different enough that this shouldn't say anything about "recycle". I view recycling as taking `x` and extending it to have size `size`. It isn't an operation between `x` and `y` directly.

Maybe:

```
Can't compute a common size among inputs.
* `..1` has size 2.
* `..3` has size 4. 
```

If `x_arg = ""` I'm not quite sure what you'd do here.

`vec_recycle_common()` would use the same error because an error would only occur there when computing a common size, but `vec_recycle()` would be different:

```
Can't recycle input `x` to size 4.
* It must have size 1 or 4.
* It has size 3.
```

lionel-: The goal is to make it consistent with vec-cast-common and vec-cast. vec-recycle would eventually match the same form.

I originally had "combine" instead of "recycle" but @hadley preferred to mention recycling.

lionel-:
The recycling and size combination errors now share the same code and the same class. We could add an `action = c("combine", "recycle")` condition field later on to distinguish the two if needed. This is similar to our new setup for ptype2 and cast errors.

I couldn't remove the recycle class because it is used in tibble, we'll need to update tibble first.

tests/testthat/error/test-conditions.txt
```diff
@@ -111,3 +111,19 @@ ordered cast failures mention conversion
...
+========================
+
+> stop_incompatible_size(1:2, 3:5, 2L, 3L, x_arg = "", y_arg = "")
+Error: Can't recycle size 2 to match size 3.
+
+> stop_incompatible_size(1:2, 3:5, 2L, 3L, x_arg = quote(foo), y_arg = "")
+Error: Can't recycle `foo` (size 2) to match size 3.
+
+> stop_incompatible_size(1:2, 3:5, 2L, 3L, x_arg = "", y_arg = "bar")
+Error: Can't recycle size 2 to match `bar` (size 3).
```
hadley, line 125: I think match should only be used in this case — if we don't have the name of RHS, it should just be "to size 3"

tests/testthat/error/test-conditions.txt
```diff
@@ -111,3 +111,19 @@ ordered cast failures mention conversion
 > vec_cast(ordered("x"), ordered("y"))
 Error: Can't convert <ordered<5a425>> to <ordered<df698>>.
 
+
+incompatible size errors
+========================
+
+> stop_incompatible_size(1:2, 3:5, 2L, 3L, x_arg = "", y_arg = "")
+Error: Can't recycle size 2 to match size 3.
```
hadley, line 119: And if there's no name of the LHS, maybe this should be: "Can't recycle vector of size 2"?

lionel-: Did you mean to not mention the target size?

DavisVaughan: That might imply that when you do have a LHS name, you don't need parenthesis

```
Can't recycle vector of size 2 to size 3.

Can't recycle `foo` of size 2 to size 3.

Can't recycle vector of size 2 to match `foo` of size 3.

Can't recycle `foo` of size 2 to match `bar` of size 3.
```

Then you generally just switch "vector" for the name, and maybe add "match"

lionel-: I like it.

DavisVaughan: Oh, maybe a slightly different message depending on if you are using `vec_recycle()` VS `vec_recycle_common()`/`vec_size_common()`. `vec_recycle()` just wouldn't mention the "to match" part since you are just recycling to a specific size.

```r
vec_recycle(x, 3)
#> Can't recycle vector of size 2 to size 3.

vec_recycle(x, 3, x_arg = "foo")
#> Can't recycle `foo` of size 2 to size 3.

# hypothetical, see #1000 
vec_size2(1:2, 1:3)
vec_recycle2(1:2, 1:3)
#> Can't recycle vector of size 2 to match vector of size 3.

vec_size2(1:2, 1:3, x_arg = "foo")
vec_recycle2(1:2, 1:3, x_arg = "foo")
#> Can't recycle `foo` of size 2 to match vector of size 3.

vec_size_common(1:2, 1:3)
vec_recycle_common(1:2, 1:3)
#> Can't recycle `..1` of size 2 to match `..2` of size 3.

vec_size_common(foo = 1:2, 1:3)
vec_recycle_common(foo = 1:2, 1:3)
#> Can't recycle `foo` of size 2 to match `..2` of size 3.

vec_size_common(foo = 1:2, bar = 1:3)
vec_recycle_common(foo = 1:2, bar = 1:3)
#> Can't recycle `foo` of size 2 to match `bar` of size 3.
```

hadley: I don't love "`foo` of size 2". If you don't want parentheses, I think "with" would be better than "of".

PR #1089, opened 2020-05-06
Silently fall back to base data frame type in comparison operators

lionel-:
`vec_match()`, `vec_in()`, `vec_equal()`, and `vec_compare()` now will compare any data frame subclasses. Currently uses the inefficient fallback path, this will need to be improved later.

Eventually we'll probably want to do this for any class that explictly inherits from its base type (i.e. one of the supported proxy type).

Also puts in place some infrastructure for falling back without warning, which we'll need for cbind and rbind prior to release.

### DIFF
--- a/tests/testthat/test-equal.R
+++ b/tests/testthat/test-equal.R
@@ -221,6 +221,13 @@ test_that("equal_scalar() compares", {
   expect_error(test_equal_scalar(expression(0), 0, expression(1), 1), "Unsupported")
 })
 
+test_that("vec_equal() silently falls back to base data frame", {
+  expect_silent(expect_identical(
+    vec_equal(foobar(mtcars), foobar(tibble::as_tibble(mtcars))),
+    rep(TRUE, 32)
+  ))
+})
+
 
 # object ------------------------------------------------------------------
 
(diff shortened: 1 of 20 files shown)


tests/testthat/test-equal.R
```diff
@@ -221,6 +221,13 @@ test_that("equal_scalar() compares", {
   expect_error(test_equal_scalar(expression(0), 0, expression(1), 1), "Unsupported")
 })
 
+test_that("vec_equal() silently falls back to base data frame", {
+  expect_silent(expect_identical(
+    vec_equal(foobar(mtcars), foobar(tibble::as_tibble(mtcars))),
```
DavisVaughan, line 226: What happens with df-cols where one df-col is a data frame and one is a tibble? i.e. is this recursive?

lionel-: Yes the fallback parameter is passed recursively. This parameter passing isn't implemented yet when the ptype2 method is implemented at R level. Once we pass it, this should work if the method passes the `...` back to `df_ptype2()` as we have documented.

lionel-: I think we should wait for the next release cycle before passing extra parameters to ptype2 methods.

PR #1095, opened 2020-05-06
Enable fallbacks globally and only warn in dev versions

lionel-:
* Falling back to data frame is now enabled globally.

* We only warn in dev versions, not in CRAN versions, as determined at load time from the version number.

* `vec_match()` etc never warn, as before.

This way we get consistent coercion semantics in vctrs. We don't annoy CRAN users with warnings. And we keep emitting warnings with the dev versions to encourage class authors to write coercion methods. In the future we may want to extend these fallbacks to other base types when they are an explicit part of the class vector.


lionel-:
Good point, I meant to do that but forgot.

PR #1097, opened 2020-05-09
Fix UBSAN error

lionel-:
```
vctrs.Rcheck/tests/testthat.Rout:subscript.c:168:19: runtime error: 2.14748e+09 is outside the range of representable values of type 'int'
    #0 0x7f21c83e0d08 in dbl_cast_subscript /data/gannet/ripley/R/packages/incoming/vctrs.Rcheck/00_pkg_src/vctrs/src/subscript.c:168:19
    #1 0x7f21c83e0d08 in vec_as_subscript_opts /data/gannet/ripley/R/packages/incoming/vctrs.Rcheck/00_pkg_src/vctrs/src/subscript.c:61:17
    #2 0x7f21c83dc678 in vec_as_location_opts /data/gannet/ripley/R/packages/incoming/vctrs.Rcheck/00_pkg_src/vctrs/src/subscript-loc.c:292:15
    #3 0x7f21c83deba1 in vctrs_as_location /data/gannet/ripley/R/packages/incoming/vctrs.Rcheck/00_pkg_src/vctrs/src/subscript-loc.c:418:10 
```

Can't reproduce these warnings, but hopefully this fixes it:

* Check for all non-finite values.
* Check for range before coercion.

### DIFF
--- a/src/subscript.c
+++ b/src/subscript.c
@@ -165,17 +165,21 @@ static SEXP dbl_cast_subscript(SEXP subscript,
       continue;
     }
 
-    int out_elt = (int) elt;
-
-    // Detect out-of-bounds and fractional numbers
-    if (elt > INT_MAX || out_elt != elt) {
+    if (!isfinite(elt) || elt > INT_MAX) {
       // Once we throw lazy errors from the cast method, we should
       // throw the error here as well
       UNPROTECT(1);
       return dbl_cast_subscript_fallback(subscript, opts, err);
     }
 
-    out_p[i] = out_elt;
+    int elt_int = (int) elt;
+
+    if (elt != elt_int) {
+      UNPROTECT(1);
+      return dbl_cast_subscript_fallback(subscript, opts, err);
+    }
+
+    out_p[i] = elt_int;
   }
 
   UNPROTECT(1);
--- a/src/subscript.c
+++ b/src/subscript.c
@@ -165,7 +165,7 @@ static SEXP dbl_cast_subscript(SEXP subscript,
       continue;
     }
 
-    if (!isfinite(elt) || elt > INT_MAX) {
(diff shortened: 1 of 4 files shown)


src/subscript.c
```diff
@@ -165,17 +165,21 @@ static SEXP dbl_cast_subscript(SEXP subscript,
       continue;
     }
 
-    int out_elt = (int) elt;
-
-    // Detect out-of-bounds and fractional numbers
-    if (elt > INT_MAX || out_elt != elt) {
+    if (!isfinite(elt) || elt > INT_MAX) {
```
DavisVaughan, line 168: So we were already capturing `NA_real_` and `NaN` above with `isnan()`, and the existing `elt > INT_MAX` check should have been capturing any large double like `Inf` or `INT_MAX+1`, so I imagine the problem came from some kind of double value that was too small?

I think this check would still improperly allow a double of `INT_MIN - 1` through, which would still fail?

I think I would use `if (elt > INT_MAX || elt < INT_MIN + 1)`

This follows what I think base R does when converting a double to integer (see link below). `elt < INT_MIN + 1` should capture any doubles that are too small to fit in an integer, and also doesn't allow a double that is identical to `INT_MIN`, because that is equivalent to `NA_integer_`.

https://github.com/wch/r-source/blob/5a156a0865362bb8381dcd69ac335f5174a4f60c/src/main/cum.c#L47-L51

jimhester:
I was able confirm the observed UBSAN error with the current master and this PR resolves it, so modulo Davis' suggestion to check for INT_MIN+1 I think this is good.

src/subscript.c
lionel-: Good point

lionel-: The existing `>` check occurred after conversion by the way, which would cause the Inf cast UB. So we still need to check for finiteness.

PR #1104, opened 2020-05-15
Draft reference on vctrs primitives and fallbacks

lionel-:
Branched from #1103.


man/faq/developer/reference-compatibility.Rmd
```diff
@@ -0,0 +1,42 @@
+
+vctrs provides a framework for working with vector classes in a generic way. However, it implements several compatibility fallbacks to base R methods. In this reference you will find how vctrs tries to be compatible with your vector class, and what base methods you need to implement for compatibility.
```
hadley, line 2: (If you're starting from scratch, we think you'll find it easier to start using `new_vctr()`; this guide is aimed developers with existing vector classes)

man/faq/developer/reference-compatibility.Rmd
```diff
@@ -0,0 +1,42 @@
...
+
+
+## The vctrs primitives
+
+Most functions in vctrs are aggregate operations: they call other vctrs functions which themselves call other vctrs functions. The dependencies of a vctrs functions are listed in the Dependencies section of its documentation page. Take a look at [vec_count()] for an example.
+
+These dependencies form a tree whose roots are the four vctrs primitives.
```
hadley, line 9: Could you draw the tree for `vec_count()` here? (Or if you spell it out for me, I can draw it in omnigraffle)

man/faq/developer/reference-compatibility.Rmd
```diff
@@ -0,0 +1,42 @@
...
+
+
+### The proxy and restoration generics
```
hadley, line 12: These seem a weird place to start given that they're rarely needed, and we're not 100% sure about them.

man/faq/developer/reference-compatibility.Rmd
```diff
@@ -0,0 +1,42 @@
...
+- [vec_cast()]
+
+See the [theory overview][theory-faq-coercion].
+
+Two objects with the same class and the same attributes are always considered compatible by ptype2 and cast. If the attributes or classes differ, they throw an incompatible type error.
+
+Coercion errors are the main source of incompatibility with vctrs. See the [howto guide][howto-faq-coercion] if you need to implement methods for these generics.
+
+
+## Aggregate operations with fallbacks
```
hadley, line 36: I think this should come earlier; this is the main point of this doc.

man/faq/developer/reference-compatibility.Rmd
```diff
@@ -0,0 +1,42 @@
...
+
+vctrs can achieve all necessary operations on vectors with the four primitives listed boved. However there are many higher level operations. The most important ones implement fallbacks to base generics for maximum compatibility with existing classes.
+
+- [vec_slice()] falls back to the base `[` generic if no [vec_proxy()] method is implemented. This way foreign classes  that do not implement [vec_restore()] can restore attributes based on the new subsetted contents.
+
+- [vec_c()] falls back to `base::c()` if all inputs have the same class for which a `c()` method is implemented but no self-to-self [vec_ptype2()] method is implemented.
```
hadley, line 42: I think you should list the downsides of these methods (i.e. where will they fail)

lionel-: The only downside I can think of is that while you can combine vectors that are part of the same class hierarchy, you can't combine different vectors of different classes, since there is no extensible mechanism of coercion that you can use with `c()`.

hadley: I just mean that this needs a small justification as to why if we can make `c()` work, we invented all the machinery of `vec_ptype2()` etc

Issue #1107, opened 2020-05-19
Proxy and restore recursively before assignment


lionel-:
To fix premature restoration. Will help with #1098.

lionel-:
Proxy and restore at first level in `vec_rbind()` as a first step:

```r
evalq({
  local_methods(
    vec_restore.vctrs_foobar = function(x, to, ...) {
      if (anyDuplicated(x$x, incomparables = NA)) {
        abort("tilt")
      }
      foobar(x)
    }
  )

  df <- foobar(data.frame(x = 1))
  vec_rbind(df, df)
})
#>   x
#> 1 1
#> 2 1
```

lionel-:
Will solve the proxy-assign conundrum where input is proxied but value is not.

lionel-:
Reconsider whether data frames can be proxied (without the vcols implemented in #730).

lionel-:
Here is how we could support all of these features simultaneously:

- List proxying into data frames to support vectorised attributes

- Data frames proxying into different data frames (typically extra columns for vectorised attributes)

- Recursive proxying and restoration

- Restore methods that inspect the types and/or the contents of their columns (which seems common in tidymodels)

- Backward compatibility with existing `vec_restore()` methods


What we gain:

1. Supporting extra columns in df-proxies will be useful to implement vectorised attributes.

2. Supporting recursive proxying and restoration will be useful to simplify the internals of vctrs, fix the semantics of restoration (restore methods are currently called on partially assembled data when there are df-cols), and improve performance (no excessive restoration, row-binding in parallel).


How to do this:

1. We proxy recursively over data frame columns. Proxying might create more df-cols in the proxy if there are any record vectors (POSIXlt, vctrs_vctr, ...). For this reason the recursive proxying goes inside out, from innermost df-cols to the outermost data frame. This way we don't proxy again newly created columns.

2. When the transformation is done and it's time to restore, we first "unproxy" all data frames, following df-cols recursively from the outermost data frame to the innermost ones. For record vectors, the unproxying eliminates the df-col and recursion ends there. For data frames, the unproxying restores its overall structure (names and length, i.e. the identity of each column).

3. When rolling back the recursion, we call `vec_restore()` on each node, restoring from innermost columns to the outermost data frame. Because we restore inside out, the restore methods are called with fully restored columns. Their types and contents can be checked during restoration, if needed.

[...]

PR #1109, opened 2020-05-19
Proxy and restore in `vec_rbind()`

DavisVaughan:
Part of #1107 

This PR ensures that `vec_rbind()` proxies the data frame ptype before initializing it, and then restores after filling it.

The tests are a little long, but I think they capture two important use cases:

1) A data frame subclass might fallback to a bare data frame in `vec_restore()` if certain conditions aren't met. (the dials case)

2) A data frame subclass might error in `vec_restore()` if certain conditions aren't met

### DIFF
--- a/src/bind.c
+++ b/src/bind.c
@@ -113,8 +113,13 @@ static SEXP vec_rbind(SEXP xs,
     }
   }
 
+  SEXP proxy = PROTECT(vec_proxy(ptype));
+  if (!is_data_frame(proxy)) {
+    Rf_errorcall(R_NilValue, "Can't fill a data frame that doesn't have a data frame proxy.");
+  }
+
   PROTECT_INDEX out_pi;
-  SEXP out = vec_init(ptype, nrow);
+  SEXP out = vec_init(proxy, nrow);
   PROTECT_WITH_INDEX(out, &out_pi);
 
   SEXP idx = PROTECT_N(compact_seq(0, 0, true), &nprot);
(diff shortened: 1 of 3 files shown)


src/bind.c
```diff
@@ -113,8 +113,13 @@ static SEXP vec_rbind(SEXP xs,
     }
   }
 
+  SEXP proxy = PROTECT(vec_proxy(ptype));
+  if (!is_data_frame(proxy)) {
```
lionel-, line 117: Probably this check should be done in `vec_proxy()`?

tests/testthat/test-bind.R
```diff
@@ -267,6 +267,73 @@ test_that("can assign row names in vec_rbind()", {
...
+      x
+    },
+    vec_proxy.vctrs_foobar = function(x, ...) {
+      x
+    },
+    vec_restore.vctrs_foobar = function(x, to, ...) {
+      if (any(is.na(x$x))) {
+        new_data_frame(x)
+      } else {
+        NextMethod()
```
lionel-, line 288: Can you make these tests work without calling next-method? The semantics of default methods are likely to change in 0.4.0.

DavisVaughan: I could directly call `vec_restore_default()`?

src/bind.c
DavisVaughan: I'm not sure the best way to approach this right now. The check would have to be pretty low level, probably in `vec_proxy_invoke()`, with something like:

```c
if (is_data_frame(x) && !is_data_frame(proxy)) {
  error();
}
```

This also feels related to `vec_is_list()` where we currently validate that the proxy is also a list, so the check might not just be specific to data frames

lionel-: I think it is generally related to inheriting from a base type. For instance if you inherit from `"integer"` it should be illegal to return a different type from the proxy. We can figure this out later and I guess an extra check doesn't hurt.

PR #1111, opened 2020-05-20
Fix fallback rbind

lionel-:
Branched from #1104.

* Add a new common class fallback option in the coercion mechanism. The options are wrapped in a struct and a structured `list()` which can be passed back and forth between the R and C sides.

* When this fallback is turned on, the common class suffix is collected by `vec_ptype_common()`, and stored in a sentinel ptype. This sentinel is not meant to be leaked to users. With the current design, the fallback should only be used through higher level operations like `vec_cast_common()` or `vec_rbind()`. We may be able to find a more general approach for 0.4.0.

* `vec_cast_common()` leaves untouched the inputs which have a fallback.

* `vec_c()` and `vec_rbind()` call `base::c()` if inputs have a fallback common class and implement a `c()` method. There are still several cases which are not handled well, and the rbind code needs to be refactored. We'll do this for 0.4.0 as part of the switch to recursive proxying and restoration.

* The special-casing of sfc lists has been removed, and we fall back to `c.sfc()` instead. Fixes r-spatial/sf#1390.

Adapting `vec_match()` to use this fallback will be a simple matter of using a fallback-enabled variant of `vec_cast_common()`.

`vec_unchop()` will be adapted to the new fallback in a different PR.


lionel-:
> vec_cast_common() leaves untouched the inputs which have a fallback.

> Adapting vec_match() to use this fallback will be a simple matter of using a fallback-enabled variant of vec_cast_common().

I don't know what I was thinking. It is trivially easy to show this can't work. The type is defined by both the class and the attributes and so the values of different objects are not in general comparable if they don't have exactly the same class and attributes. Simple example: the values of `units::set_units(1:2, "m/s")` and `units::set_units(3L, "km/s")` are not comparable without coercion, even though they have the same class and the same attribute names.

I think we'll need something like this:

```r
library(rlang)
library(purrr)

# Obtain ptypes with `x[0]` and concatenate them to find the common ptype
base_ptype_common <- function(...) {
  # Would first check for common class and throw incompatible type
  # error otherwise
  ptypes <- map(list2(...), `[`, 0L)
  exec("c", !!!ptypes)
}

# Concatenate inputs with ptype in first location. This causes a
# coercion of the inputs to a common type. Slice the result and return
# list of coerced vectors of the same sizes as the inputs.
base_cast_common <- function(..., .ptype = NULL) {
  dots <- list2(...)

  if (is.null(.ptype)) {
    .ptype <- base_ptype_common(!!!dots)
  }
  common <- exec("c", .ptype, !!!dots)

  sizes <- lengths(dots)
  locs <- reduce(sizes, .init = list(0), function(output, input) {
    n <- last(last(output))
    c(output, list(seq(n + 1, n + input)))
  })
  locs <- locs[-1]

  map(locs, function(loc) common[loc])
}
last <- function(x) {
  x[[length(x)]]
}
```

Works well with units:

```r
(x <- units::set_units(1:2, "m/s"))
#> Units: [m/s]
#> [1] 1 2
(y <- units::set_units(3, "km/s"))
#> 3 [km/s]

base_ptype_common(x, y)
#>  [m/s]

base_cast_common(x, y)
#> [[1]]
#> Units: [m/s]
#> [1] 1 2
#>
#> [[2]]
#> 3000 [m/s]
```

For sf this is a bit funny though, because `[` isn't type-stable. The result is downcasted to a more specific class if possible. The common type of different sfc subclasses is normally an abstract `sfc_GEOMETRY`, :

```r
sfc_x <- sf::st_sfc(sf::st_point())
sfc_y <- sf::st_sfc(sf::st_multipoint(), sf::st_multipoint())

base_ptype_common(sfc_x, sfc_y)
#> Geometry set for 0 features
#> bbox:           xmin: NA ymin: NA xmax: NA ymax: NA
#> CRS:            NA

class(base_ptype_common(sfc_x, sfc_y))
#> [1] "sfc_GEOMETRY" "sfc"
```

However the inputs are being coerced back and forth and end up unchanged:

```r
base_cast_common(sfc_x, sfc_y)
#> [[1]]
#> Geometry set for 1 feature  (with 1 geometry empty)
#> geometry type:  POINT
#> dimension:      XY
#> bbox:           xmin: NA ymin: NA xmax: NA ymax: NA
#> CRS:            NA
#> POINT EMPTY
#>
#> [[2]]
#> Geometry set for 2 features  (with 2 geometries empty)
#> geometry type:  MULTIPOINT
#> dimension:      XY
#> bbox:           xmin: NA ymin: NA xmax: NA ymax: NA
#> CRS:            NA
#> MULTIPOINT EMPTY
#> MULTIPOINT EMPTY
```

At this point it's unclear that we can find a general fallback for vec-cast-common. Maybe we should first scrape CRAN packages ordered by popularity for implementations of `[` and `c()`, and evaluate the generality of the approach drafted above. This will have to wait after dplyr 1.0.

This shouldn't affect much the current fallback to `base::c()` implemented in this PR, so it is still good to go.

R/utils.R
```diff
@@ -161,3 +161,33 @@ df_has_base_subset <- function(x) {
   method <- .Call(vctrs_s3_find_method, "[", x, table)
   is_null(method) || identical(method, `[.data.frame`)
 }
+
+last <- function(x) {
+  x[[length(x)]]
+}
+
+vec_common_suffix <- function(x, y) {
```
DavisVaughan, line 169: A comment about how this works might be useful, I spent a few minutes reading it to be able to figure it out.

man/faq/developer/reference-compatibility.Rmd
```diff
@@ -37,6 +37,10 @@ Coercion errors are the main source of incompatibility with vctrs. See the [howt
 
 vctrs can achieve all necessary operations on vectors with the four primitives listed boved. However there are many higher level operations. The most important ones implement fallbacks to base generics for maximum compatibility with existing classes.
 
+- We have started to implement a fallback mechanism in [vec_cast_common()]. This function normally calls [vec_ptype2()] and [vec_cast()] to normalise a set of inputs to the same type. When the inputs do not implement the coercion generics but they all inherit from the same base class, we do not attempt any coercion and leave the inputs as is.
+
+  This fallback should be safe to use before operations like [vec_match()] or [vec_rbind()]. It is currently internal.
```
DavisVaughan, line 42: Do you want to update these comments in light of your github comment that states this this won't work?

src/cast.c
```diff
@@ -181,7 +183,7 @@ SEXP vec_cast_dispatch_s3(const struct cast_opts* opts) {
   PROTECT(method);
 
   if (method == R_NilValue) {
-    SEXP out = vec_cast_default(x, to, x_arg_obj, to_arg_obj, opts->df_fallback);
+    SEXP out = vec_cast_default(x, to, x_arg_obj, to_arg_obj, &opts->fallback);
```
DavisVaughan, line 186:
```suggestion
    SEXP out = vec_cast_default(x, to, x_arg_obj, to_arg_obj, &(opts->fallback));
```

src/c.c
```diff
@@ -123,14 +165,33 @@ SEXP vec_c(SEXP xs,
...
   return out;
 }
 
 static inline bool vec_implements_base_c(SEXP x);
 
-// [[ include("vctrs.h") ]]
-bool needs_vec_c_fallback(SEXP xs, SEXP ptype) {
+// [[ include("c.h") ]]
+bool needs_vec_c_fallback(SEXP ptype) {
+  if (!Rf_inherits(ptype, c_strs_vctrs_common_class_fallback)) {
```
DavisVaughan, line 176:
```suggestion
  if (!vec_is_common_class_fallback(ptype)) {
```

?

src/type-data-frame.c
```diff
@@ -444,20 +435,13 @@ SEXP df_ptype2_loop(const struct ptype2_opts* opts,
 }
 
 // [[ register() ]]
-SEXP vctrs_df_cast_params(SEXP x, SEXP to, SEXP x_arg_, SEXP to_arg_, SEXP df_fallback_) {
-  struct vctrs_arg x_arg = vec_as_arg(x_arg_);
-  struct vctrs_arg to_arg = vec_as_arg(to_arg_);;
-  enum df_fallback df_fallback = r_int_get(df_fallback_, 0);
+SEXP vctrs_df_cast_opts(SEXP x, SEXP to, SEXP opts, SEXP x_arg, SEXP to_arg) {
+  struct vctrs_arg c_x_arg = vec_as_arg(x_arg);
+  struct vctrs_arg c_to_arg = vec_as_arg(to_arg);;
```
DavisVaughan, line 440:
```suggestion
  struct vctrs_arg c_to_arg = vec_as_arg(to_arg);
```

src/type2.c
```diff
@@ -158,21 +152,18 @@ bool vec_is_coercible(const struct ptype2_opts* opts,
...
-  struct vctrs_arg x_arg = vec_as_arg(x_arg_);
-  struct vctrs_arg y_arg = vec_as_arg(y_arg_);;
-  enum df_fallback df_fallback = r_int_get(df_fallback_, 0);
+SEXP vctrs_is_coercible(SEXP x,
+                        SEXP y,
+                        SEXP opts,
+                        SEXP x_arg,
+                        SEXP y_arg) {
+  struct vctrs_arg c_x_arg = vec_as_arg(x_arg);
+  struct vctrs_arg c_y_arg = vec_as_arg(y_arg);;
```
DavisVaughan, line 161:
```suggestion
  struct vctrs_arg c_y_arg = vec_as_arg(y_arg);
```

tests/testthat/test-cast.R
```diff
@@ -135,3 +135,93 @@ test_that("can signal deprecation warnings for lossy casts", {
...
+  expect_identical(fallback_class(common_sentinel$x), "vctrs_foobar")
+
+  common <- vec_cast_common_opts(x = x, y = y, .opts = fallback_ptype2_opts())
+  expect_identical(common, list(x = x, y = y))
+
+  common <- vec_cast_common_opts(x = x_df, y = y_df, .opts = fallback_ptype2_opts())
+  expect_identical(common, list(x = x_df, y = y_df))
+})
+
+test_that("vec_ptype_common() collects common type", {
```
DavisVaughan, line 185:
```suggestion
test_that("vec_ptype_common_fallback() collects common type", {
```

?

R/bind.R
```diff
@@ -82,8 +82,9 @@
 #'
 #' - [base::c()]
 #'
-#' If columns to combine have the same class, `vec_rbind()` falls back
-#' to `base::c()` if there exists a `c()` method for this class.
+#' If columns to combine inherit from a common class hierarchy,
```
hadley, line 85:
```suggestion
#' If columns to combine inherit from a common class,
```

NEWS.md
```diff
@@ -1,6 +1,11 @@
 
 # vctrs (development version)
 
+* `vec_c()` and `vec_rbind()` now fall back to `base::c()` if the
+  inputs have a common class hierarchy for which a `c()` method is
+  implemented but no self-to-self `vec_ptype2()` method is
+  implemented.
+
```
hadley, line 8:
```suggestion
* `vec_c()` and `vec_rbind()` now fall back to `base::c()` if the
    inputs have a common base class with a `c()` method
    (and no self-to-self `vec_ptype2()` method)

```

tests/testthat/error/test-bind.txt
```diff
@@ -5,7 +5,7 @@ vec_rbind() fails with complex foreign S3 classes
 > x <- structure(foobar(1), attr_foo = "foo")
 > y <- structure(foobar(2), attr_bar = "bar")
 > vec_rbind(set_names(x, "x"), set_names(y, "x"))
-Error: Can't combine `..1$x` <vctrs_foobar> and `..2$x` <vctrs_foobar>.
+Error: Can't combine `..1` <vctrs_foobar> and `..2` <vctrs_foobar>.
```
hadley, line 8: I think you could now remove the `set_names()` in the previous line since it's no longer used

tests/testthat/test-bind.R
```diff
@@ -712,11 +712,63 @@ test_that("vec_rbind() fails with complex foreign S4 classes", {
...
+  foo_df <- foobaz(x_df)
+  bar_df <- foobaz(y_df)
+
+  out <- with_methods(
+    c.vctrs_foobar = function(...) quux(NextMethod()),
+    vec_rbind(foo_df, bar_df)
+  )
+  expect_identical(out, foobaz(data_frame(x = quux(c(1, 2)))))
+
+  out <- with_methods(
```
hadley, line 738: What's this testing? From the methods I expected it to be checking the `vec_ptype2` beats `c`, but methods are for different classes?

tests/testthat/test-bind.R
```diff
@@ -712,11 +712,63 @@ test_that("vec_rbind() fails with complex foreign S4 classes", {
...
 
 test_that("vec_rbind() falls back to c() if S4 method is available", {
-  skip("TODO")
+  joe <- data_frame(x = .Counts(c(1L, 2L), name = "Joe"))
+  jane <- data_frame(x = .Counts(3L, name = "Jane"))
+
+  expect_error(vec_rbind(joe, jane), class = "vctrs_error_incompatible_type")
+
+  out <- with_methods(
+    c.vctrs_Counts = function(...) .Counts(NextMethod(), name = "dispatched"),
```
hadley, line 768: This is a S3 method?

R/type-data-frame.R
```diff
@@ -107,29 +107,38 @@ df_is_coercible <- function(x, y, df_fallback = FALSE) {
...
 #' @export
 df_ptype2 <- function(x, y, ..., x_arg = "", y_arg = "") {
-  .Call(vctrs_df_ptype2, x, y, x_arg, y_arg, df_fallback = FALSE)
+  .Call(vctrs_df_ptype2_opts, x, y, opts = match_ptype2_opts(...), x_arg, y_arg)
 }
 #' @rdname df_ptype2
 #' @export
 df_cast <- function(x, to, ..., x_arg = "", to_arg = "") {
-  .Call(vctrs_df_cast_params, x, to, x_arg, to_arg, FALSE)
+  .Call(vctrs_df_cast_opts, x, to, opts = ptype2_opts(), x_arg, to_arg)
```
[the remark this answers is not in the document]
lionel-: I think it should, but I wanted to find a failing test case before matching the fallback parameters from the dots.

tests/testthat/test-bind.R
lionel-: The `foobaz()` is a `data.frame` subclass which contains a `foobar()` with a `c()` fallback. I've added a comment.

R/type-data-frame.R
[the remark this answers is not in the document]
lionel-: Found a failing case, now fixed: https://github.com/r-lib/vctrs/commit/cf9c5caadc989c6b8fff014f5645766745557712

PR #1112, opened 2020-05-21
Fix unspecified fallback

lionel-:
Branched from #1111.

This adds support for combining unspecified chunks to the `base::c()` fallback. We can't rely on `c()` for this because implementations might not handle unspecified inputs, and if unspecified is the first input this would prevent dispatch from reaching the expected method. Instead, we first combine non-unspecified chunks, then use `[` with `NA` integer locations to expand the result with missing values.

With this change, the fallback is now consistent when disjoint data frames are binded: 

```r
x <- sf::st_sfc(sf::st_point(1:2))
y <- sf::st_sfc(sf::st_multipoint(matrix(1:4, 2)), sf::st_multipoint(matrix(5:8, 2)))

vec_rbind(
  data_frame(x = x),
  data_frame(y = y)
)
#>             x                         y
#> 1 POINT (1 2)          MULTIPOINT EMPTY
#> 2 POINT EMPTY MULTIPOINT ((1 3), (2 4))
#> 3 POINT EMPTY MULTIPOINT ((5 7), (6 8))
```

### DIFF
--- a/src/type2.c
+++ b/src/type2.c
@@ -9,12 +9,6 @@ SEXP vec_ptype2_switch_native(const struct ptype2_opts* opts,
                               enum vctrs_type x_type,
                               enum vctrs_type y_type,
                               int* left);
-static
-SEXP vec_ptype2_from_unspecified(const struct ptype2_opts* opts,
-                                 enum vctrs_type other_type,
-                                 SEXP other,
-                                 struct vctrs_arg* other_arg);
-
 
 // [[ register() ]]
 SEXP vctrs_ptype2_opts(SEXP x,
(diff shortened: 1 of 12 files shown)


src/type2.c
```diff
@@ -115,6 +115,37 @@ SEXP vec_ptype2_switch_native(const struct ptype2_opts* opts,
   }
 }
 
+/**
+ * Return non-unspecified type.
+ *
+ * This is normally the `vec_ptype()` of the other input, but if the
+ * common class fallback is enabled we return the `vec_ptype2()` of
+ * this input. This way we may return a fallback sentinel which can be
```
DavisVaughan, line 123:
```suggestion
 * this input with itself. This way we may return a fallback sentinel which can be
```

lionel-:
> Like in #1111, I assume vec-unchop support will come later?

Yes, this might be a bit tricky with the indices.

PR #1124, opened 2020-05-27
Pass `ownership` through to `vec_restore()`

### LINKED ISSUE #1122: vec_rbind() performance regression on R 4.0 with df-cols
Performance regression when rbinding lots of data frames that have df-cols. I'm sure this has to do with making extra copies, but I'm not sure where yet. I'll take a look.

This is with dev vctrs.

R 3.6

```r
library(vctrs)

df_col <- new_data_frame(list(x = 1:2))
df <- new_data_frame(list(y = df_col))

x <- rep_len(list(df), 10000)
y <- rep_len(list(df_col), 10000)

lst_rbind <- function(x) {
  vec_rbind(!!!x)
}

bench::mark(lst_rbind(x))
#> # A tibble: 1 x 6
#> 1 lst_rbind(x)   37.5ms   37.9ms      26.2     277KB     52.4

bench::mark(lst_rbind(y))
#> # A tibble: 1 x 6
#> 1 lst_rbind(y)   16.5ms   18.2ms      55.1     274KB     16.5
```

R 4.0

```r
library(vctrs)

df_col <- new_data_frame(list(x = 1:2))
df <- new_data_frame(list(y = df_col))

x <- rep_len(list(df), 10000)
y <- rep_len(list(df_col), 10000)

lst_rbind <- function(x) {
  vec_rbind(!!!x)
}

bench::mark(lst_rbind(x))
#> # A tibble: 1 x 6
#> 1 lst_rbind(x)    316ms    352ms      2.84     764MB     48.2

bench::mark(lst_rbind(y))
#> # A tibble: 1 x 6
#> 1 lst_rbind(y)   15.8ms   17.6ms      55.6     274KB     23.4
```

DavisVaughan:
Closes #1122 

This PR fixes the `vec_rbind()` performance regression seen in #1122 by passing along the ownership parameter to `vec_restore()`. The sole purpose of this is to allow us to swap between `r_clone_shared()` and `r_clone_referenced()` here:

https://github.com/r-lib/vctrs/pull/1124/files#diff-90e5f87281649d48864c579abd8649c3R31-R35

The problem in #1122 is that df-cols created by `vec_init()` automatically get a refcount of 1 due to the `SET_VECTOR_ELT()` call that places them in the outer data frame. When `vec_restore()` is called from `df_assign()` on those inner df-cols it eventually hands off to `vec_bare_df_restore_impl()` which unconditionally called `r_clone_referenced()`. Because our df-col _was_ referenced, it makes a shallow duplicate of it even though it is the freshly created output object. The problem with this is that by shallow duplicating the df-col, the refcount of all the columns in the df-col gets bumped from 1 to 2. So on the next iteration of `vec_rbind()`, those columns get shallow duplicated (which is really a full duplication for non-lists). That was where the performance really suffered.

By passing the ownership parameter through, we can correctly switch to `r_clone_shared()` when restoring, which allows in place modification of the case where the refcount is 1. Since the df-col's refcount is 1, it doesn't get shallow duplicated and the inner columns of that df-col don't have their refcount bumped.

This is on R 4.0 with this PR. We now match R 3.6 performance 

```r
library(vctrs)

df_col <- new_data_frame(list(x = 1:2))
df <- new_data_frame(list(y = df_col))

x <- rep_len(list(df), 10000)
y <- rep_len(list(df_col), 10000)

lst_rbind <- function(x) {
    vec_rbind(!!!x)
}

bench::mark(lst_rbind(x))
#> # A tibble: 1 x 6
#> 1 lst_rbind(x)   37.8ms   39.1ms      25.6     278KB     44.7
```

---

In theory this needs a test, but I'm not really sure how to design one for this pathological case

### DIFF
--- a/src/ownership.h
+++ b/src/ownership.h
@@ -7,7 +7,7 @@ enum vctrs_ownership {
   vctrs_ownership_total
 };
 
-static inline const enum vctrs_ownership proxy_ownership(SEXP x) {
+static inline const enum vctrs_ownership vec_ownership(SEXP x) {
   return NO_REFERENCES(x) ? vctrs_ownership_total : vctrs_ownership_shared;
 }
 
--- a/src/ownership.h
+++ b/src/ownership.h
@@ -1,0 +1,14 @@
+#ifndef VCTRS_OWNERSHIP_H
+#define VCTRS_OWNERSHIP_H
+
+// Ownership is recursive
+enum vctrs_ownership {
+  vctrs_ownership_shared,
+  vctrs_ownership_total
+};
+
+static inline const enum vctrs_ownership proxy_ownership(SEXP x) {
+  return NO_REFERENCES(x) ? vctrs_ownership_total : vctrs_ownership_shared;
+}
+
+#endif
--- a/src/ownership.h
+++ b/src/ownership.h
@@ -3,12 +3,12 @@
 
 // Ownership is recursive
 enum vctrs_ownership {
-  vctrs_ownership_shared,
-  vctrs_ownership_total
+  VCTRS_OWNERSHIP_shared,
+  VCTRS_OWNERSHIP_total
 };
 
 static inline const enum vctrs_ownership vec_ownership(SEXP x) {
-  return NO_REFERENCES(x) ? vctrs_ownership_total : vctrs_ownership_shared;
+  return NO_REFERENCES(x) ? VCTRS_OWNERSHIP_total : VCTRS_OWNERSHIP_shared;
 }
 
 #endif
--- a/src/slice-chop.c
+++ b/src/slice-chop.c
@@ -203,7 +203,8 @@ static SEXP chop(SEXP x, SEXP indices, struct vctrs_chop_info info) {
       UNPROTECT(1);
     }
 
-    elt = vec_restore(elt, x, info.restore_size);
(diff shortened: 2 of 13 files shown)


src/ownership.h
```diff
@@ -0,0 +1,14 @@
+#ifndef VCTRS_OWNERSHIP_H
+#define VCTRS_OWNERSHIP_H
+
+// Ownership is recursive
+enum vctrs_ownership {
```
lionel-, line 5: Maybe take this opportunity to switch to the new naming scheme?

src/slice-chop.c
```diff
@@ -202,7 +203,8 @@ static SEXP chop(SEXP x, SEXP indices, struct vctrs_chop_info info) {
       UNPROTECT(1);
     }
 
-    elt = vec_restore(elt, x, info.restore_size);
+    const enum vctrs_ownership ownership = vec_ownership(elt);
```
lionel-, line 206: Probably no need to declare it const since it's passed by value anyway? Maybe inline this in the restore call?

DavisVaughan:
Alternative solution might be https://github.com/r-lib/vctrs/issues/1107. If `vec_restore()` was only done once at the very end in a recursive way then the duplication here wouldn't happen since it only occurs with repeated `vec_restore()` calls

PR #1137, opened 2020-06-04
Add coercion methods for sf data frames

### LINKED ISSUE #1136: Fix sf combination with multiple chunks
From https://github.com/r-spatial/sf/issues/1390#issuecomment-636536316

```r
sfc <- sf::st_sfc(sf::st_point())
sf <- sf::st_sf(geo = sfc)

vctrs::vec_rbind(sf, sf)
#> Simple feature collection with 2 features and 0 fields (with 2 geometries empty)
#> geometry type:  POINT
#> dimension:      XY
#> bbox:           xmin: NA ymin: NA xmax: NA ymax: NA
#> CRS:            NA
#>           geo
#> 1 POINT EMPTY
#> 2 POINT EMPTY

vctrs::vec_rbind(sf, sf, sf)
#> Error in st_geometry.sf(x) :
#>   attr(obj, "sf_column") does not point to a geometry column.
#> Did you rename it, without setting st_geometry(obj) <- "newname"?
```

This is a bad interaction between the `c()` fallback in `vec_rbind()`, which currently creates a sentinel column in the common type, and the same-type fallback for foreign data frames, which calls `[<-` to update the inputs with the common-type column structure.

lionel-:
Closes #1136.

Methods can't be included downstream yet because of workarounds to fix the bad interactions between ptype2 and the `c()` fallback in `vec_rbind()` on which we count to combine geometry columns. When the `c()` fallback is enabled, we return a bare data frame instead of an sf one. This is a stopgap to make dplyr work (via `dplyr_reconstruct()`).

Requires r-spatial/sf#1414.

### DIFF
--- a/R/type-sf.R
+++ b/R/type-sf.R
@@ -80,6 +80,15 @@ vec_ptype2.data.frame.sf = function(x, y, ...) {
 	sf_ptype2(x, y, ...)
 }
 
+# Maybe we should not have these methods, but they are currently
+# required to avoid the base-df fallback
+vec_ptype2.sf.tbl_df = function(x, y, ...) {
+	new_data_frame(sf_ptype2(x, y, ...))
+}
+vec_ptype2.tbl_df.sf = function(x, y, ...) {
+	new_data_frame(sf_ptype2(x, y, ...))
+}
+
 sf_cast = function(x, to, ...) {
 	data = vctrs::df_cast(x, to, ...)
 
--- a/R/type-sf.R
+++ b/R/type-sf.R
@@ -1,14 +1,20 @@
 
-# Imported at load-time
-st_crs = function(...) NULL
-st_precision = function(...) NULL
-st_as_sf = function(...) NULL
+# Imported at load-time in `sf_env`
+st_crs = function(...) stop_sf()
+st_precision = function(...) stop_sf()
+st_as_sf = function(...) stop_sf()
+stop_sf = function() abort("Internal error: Failed sf import.")
+
 sf_deps = c(
 	"st_crs",
 	"st_precision",
 	"st_as_sf"
 )
+sf_env = env()
 
+
+# sf namespace
+local(envir = sf_env, {
 
 # Registered at load-time (same for all other methods)
 vec_proxy.sf = function(x, ...) {
@@ -142,6 +148,10 @@ common_prec = function(x, y) {
 
 	lhs
 }
+
+}) # local(envir = sf_env)
+
+env_bind(ns_env("vctrs"), !!!as.list(sf_env))
 
 # Local Variables:
 # indent-tabs-mode: t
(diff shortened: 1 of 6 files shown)


lionel-:
Can you please take a cursory look @DavisVaughan?

R/type-sf.R
```diff
@@ -0,0 +1,138 @@
...
+
+vec_ptype2.sf.sf = function(x, y, ...) {
+	sf_ptype2(x, y, ...)
+}
+vec_ptype2.sf.data.frame = function(x, y, ...) {
+	sf_ptype2(x, y, ...)
+}
+vec_ptype2.data.frame.sf = function(x, y, ...) {
+	sf_ptype2(x, y, ...)
+}
```
DavisVaughan, line 81: Are `tbl_df` methods required?

R/type-sf.R
```diff
@@ -0,0 +1,138 @@
...
+		sf_column_name = sfc_name,
+		crs = crs,
+		precision = prec,
+		stringsAsFactors = FALSE
+	)
+}
+
+vec_cast.sf.sf = function(x, to, ...) {
+	sf_cast(x, to, ...)
+}
```
DavisVaughan, line 111: Are you relying on fallbacks to cast sf->data.frame? I can't seem to remember which fallbacks are in place and are working silently vs what we recommend people implement

R/type-sf.R
```diff
@@ -0,0 +1,138 @@
...
+
+
+# take conservative approach of requiring equal CRS and precision
+common_crs = function(x, y) {
+	lhs = st_crs(x)
+	rhs = st_crs(y)
+
+	if (lhs != rhs)
+		stop("coordinate reference systems not equal: use st_transform() first?")
```
DavisVaughan, line 120: Does this need to be a `stop_incompatible_type()` error with `details` set to this comment?

R/type-sf.R
```diff
@@ -0,0 +1,138 @@
...
+
+	lhs
+}
+common_prec = function(x, y) {
+	lhs = st_precision(x)
+	rhs = st_precision(y)
+
+	if (lhs != rhs)
+		stop("precisions not equal")
```
DavisVaughan, line 129: Same here re `stop_incompatible_type()`

R/type-sf.R
lionel-: hmm maybe, but this would need to be changed downstream too. Given all the work that is done in ptype2-default (e.g. `AsIs` handling), maybe this should call the latter. I'd prefer not making these decisions now, until we figure out higher order types.

R/type-sf.R
lionel-: Probably a good idea to add them. The df methods are definitely required to handle the case where we return a bare df when a sentinel is detected.

R/type-sf.R
lionel-: Good point, added the missing cast methods.

PR #1142, opened 2020-06-09
Fix default `vec_proxy_compare()` method

### LINKED ISSUE #1140: `vec_proxy_compare()` should default to `vec_proxy_equal()`, not `vec_proxy()`
Also it currently uses `vec_data()` instead of `vec_proxy()`, which currently changes the type of data frame proxies.

lionel-:
* Defaults to `vec_proxy_equal()` instead of `vec_data()`. Closes #1140.

* Update `xtfrm.vctrs_vctr()` to return an unclassed integer vector even when the comparison proxy is classed.

### DIFF
--- a/R/compare.R
+++ b/R/compare.R
@@ -49,7 +49,7 @@ vec_proxy_compare_default <- function(x, relax = FALSE) {
       stop_unsupported(x, "vec_proxy_compare")
     }
   } else {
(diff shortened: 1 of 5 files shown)


DavisVaughan:
This flattens df-cols in vec-proxy-compare now right? That seems reasonable

PR #1145, opened 2020-06-14
WIP: Remove arg_match() calls for testing

krlmlr:
because it's slow.


krlmlr:
We don't really need it too; handled in #1149.

PR #1148, opened 2020-06-14
Discussion: simpler layout for rlang_result

krlmlr:
We're using `result()` and `result_get()` to avoid the overhead of `tryCatch()` in code that is used in tight loops. The layout of the underlying object should be as simple as possible. Should we wrap in a simple unnamed list and let the class decide if it's an error or a valid value?

### DIFF
--- a/R/utils.R
+++ b/R/utils.R
@@ -77,17 +77,25 @@ paste_line <- function (...) {
 }
 
 # Experimental
-result <- function(ok = NULL, err = NULL) {
-  structure(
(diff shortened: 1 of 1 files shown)


lionel-:
This is for prototyping C code in R. If the code is slow it should be moved to C.

lionel-:
Closing this but please let me know if we should move this code to C to help with performance.

krlmlr:
I'll get back to you if I see a bottleneck in this part of the code.

PR #1151, opened 2020-06-15
Revert to a simpler ownership model - but always duplicate ALTREP

### LINKED ISSUE #1122: vec_rbind() performance regression on R 4.0 with df-cols

DavisVaughan:
Closes tidyverse/dplyr#5327
Closes #1124
Closes #1122 

Reverts back to a slightly simpler ownership model of `VCTRS_OWNED_true` and `VCTRS_OWNED_false`.

When we own the object, we only ever attempt to duplicate it if it is ALTREP. If `vec_init()` ever creates ALTREP objects in the future (https://github.com/r-lib/vctrs/pull/837), this will be required. When doing assignment, we have to duplicate ALTREP objects before dereferencing even if we own them, because we need access to the actual data that it is representing, not the ALTREP object's internals.

When we don't own the object, we use `r_clone_referenced()` to determine if we need to duplicate or not.

This fixed the repeated duplication that was occurring in `df_assign()`'s calls to `vec_proxy_assign_opts()`, but I then discovered that we still had repeated duplication in `vec_restore()`. To fix that, I had to borrow from the ideas in #1124 and pass through `owned` to `vec_restore()` as well. This allows us to avoid attempting duplication here as well.

With both of those in place, tidyverse/dplyr#5327 is fixed:

```r
library(dplyr)

date_frame <- tibble(date = rep(lubridate::date("2020-01-01"), 100))
date_frames_to_bind <- rep(list(date_frame), 10000)
bench::mark(bind_rows(date_frames_to_bind))
#> # A tibble: 1 x 6
#> 1 bind_rows(date_frames_to_bind)    133ms    142ms      6.87    8.39MB     25.8
```

---

It is worth noting that @lionel- and I both think that eventually we should probably have recursive proxy and restore functions (https://github.com/r-lib/vctrs/issues/1107). This would allow `df_assign()` to not have to proxy and restore the output columns repeatedly, which is where many of these duplication issues arise. That would also allow us to remove the `owned` argument from `vec_restore()` again, which I am currently viewing as a temporary fix.

---

This also closes #1124, because it fixes the original problem there where columns of df-cols were being duplicated.

### DIFF
--- a/src/owned.h
+++ b/src/owned.h
@@ -17,12 +17,16 @@ static inline enum vctrs_owned vec_owned(SEXP x) {
 }
 
 // Wrapper around `r_clone_referenced()` that only attempts to clone if
-// we indicate that we don't own `x`, or if we do own `x` but it is ALTREP.
-// If `x` is ALTREP, we must clone it before dereferencing, otherwise we get
-// a pointer into the ALTREP internals rather than into the object it
-// truly represents.
+// we indicate that we don't own `x`, or if `x` is ALTREP.
+// If `x` is ALTREP, we must unconditionally clone it before dereferencing,
+// otherwise we get a pointer into the ALTREP internals rather than into the
+// object it truly represents.
 static inline SEXP vec_clone_referenced(SEXP x, const enum vctrs_owned owned) {
-  if (owned == VCTRS_OWNED_false || ALTREP(x)) {
+  if (ALTREP(x)) {
+    return r_clone_referenced(x);
+  }
+
+  if (owned == VCTRS_OWNED_false) {
     return r_clone_referenced(x);
   } else {
     return x;
--- a/src/owned.h
+++ b/src/owned.h
@@ -1,0 +1,32 @@
+#ifndef VCTRS_OWNED_H
+#define VCTRS_OWNED_H
+
+#include "altrep.h"
+#include "utils.h"
+
+
+// Ownership is recursive
+enum vctrs_owned {
+  vctrs_owned_FALSE = 0,
+  vctrs_owned_TRUE
+};
+
+
+static inline enum vctrs_owned vec_owned(SEXP x) {
+  return NO_REFERENCES(x) ? vctrs_owned_TRUE : vctrs_owned_FALSE;
+}
+
+// Wrapper around `r_clone_referenced()` that only attempts to clone if
+// we indicate that we don't own `x`, or if we do own `x` but it is ALTREP.
+// If `x` is ALTREP, we must clone it before dereferencing, otherwise we get
+// a pointer into the ALTREP internals rather than into the object it
+// truly represents.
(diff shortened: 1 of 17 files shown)


src/owned.h
```diff
@@ -0,0 +1,32 @@
...
+}
+
+// Wrapper around `r_clone_referenced()` that only attempts to clone if
+// we indicate that we don't own `x`, or if we do own `x` but it is ALTREP.
+// If `x` is ALTREP, we must clone it before dereferencing, otherwise we get
+// a pointer into the ALTREP internals rather than into the object it
+// truly represents.
+static inline SEXP vec_clone_referenced(SEXP x, const enum vctrs_owned owned) {
+  if (owned == VCTRS_OWNED_false || ALTREP(x)) {
+    return r_clone_referenced(x);
```
lionel-, line 26: ALTREP objects should be copied unconditionally right? Even if they are maxed out in references, I think it would be clearer to copy unconditionally.

tests/testthat/helper-performance.R
```diff
@@ -0,0 +1,21 @@
...
+
+  skip("Not testing performance")
+}
+
+expect_time_lt <- function(expr, expect) {
+  time <- time_expr({{ expr }})
+  expect_lt(time, expect)
+}
+
+time_expr <- function(expr) {
```
lionel-, line 17: `time_of()`?

PR #1155, opened 2020-06-18
Let `vec_proxy_compare()` unwrap 1 col dfs and flatten df-cols

DavisVaughan:
Follow up to #1142 

- The `relax` argument has been removed from the generic of `vec_proxy_compare()`. Instead it is treated as an internal detail that gets passed to the default method from C. This should make it easier to remove it altogether later if we decide to do so, and _shouldn't_ break anything now.

- We now handle `vec_proxy_compare()` from C in a similar way to `vec_proxy_equal()`. We recurse over data frame columns, dispatching to the R level `vec_proxy_compare()` as we go. This also now uses the `df_flatten()` and `vec_proxy_unwrap()` infrastructure.

- I currently have a separate `vec_proxy_compare_recursive()` function, but it could be removed if we eventually remove `relax` entirely in favor of always seeing lists as orderable using `seq_along()`. Then we could just go through `vec_proxy_recursive()`


R/compare.R
```diff
@@ -1,19 +1,19 @@
...
+#' and `<=` (via [vec_compare()]); and [min()], [max()], [median()], and
+#' [quantile()].
 #'
 #' The default method assumes that all classes built on top of atomic
-#' vectors or records are orderable. If your class is not, you will need
+#' vectors or records are comparable. Internally the default calls
+#' [vec_proxy_equal()]. If your class is not comparable, you will need
 #' to provide a `vec_proxy_compare()` method that throws an error.
 #'
+#' Lists are not comparable, as it is unclear how to compare elements of
```
lionel-, line 13: "unclear" makes it sound like we don't know how to do it but it might be possible in the future. Perhaps:

> as comparing elements of different types is undefined.

R/compare.R
```diff
@@ -176,6 +157,44 @@ order_proxy <- function(proxy, direction = "asc", na_value = "largest") {
   }
 }
 
+# Order proxy ------------------------------------------------------------------
+
+#' Ordering proxy
+#'
+#' Returns a proxy object for ordering. For vctrs_vctr objects, this determines
```
lionel-, line 164:
```suggestion
#' Returns a proxy object for ordering. For [`vctrs_vctr`][vctr] objects, this determines
```

R/compare.R
lionel-: Though now I wonder if we should transform lists to a `list_of()` (at least conceptually). This would fail if no common type, and then we can compare the common prefix of the list elements, e.g. `list(c(1, 2, 4))` is lesser than `list(1:10)`. We'd have a rule in case of equal prefixes.

lionel-: I think the general concept is that comparison requires a common type but ordering doesn't. Maybe add this rule to the description?

src/proxy.c
```diff
@@ -30,108 +39,234 @@ SEXP vec_proxy(SEXP x) {
...
+  SEXP out = vec_proxy_equal_invoke(x, method);
+  UNPROTECT(1);
+  return out;
+}
 
-  if (is_data_frame(proxy)) {
-    // Flatten df-cols so we don't have to recurse to work with data
-    // frame proxies
-    proxy = PROTECT(df_flatten(proxy));
+static inline SEXP vec_proxy_equal_method(SEXP x) {
```
lionel-, line 69: Modifiers on own line?

src/proxy.c
```diff
@@ -30,108 +39,234 @@ SEXP vec_proxy(SEXP x) {
...
+
+// [[ register(); include("vctrs.h") ]]
+SEXP vec_proxy_compare(SEXP x) {
+  SEXP method = PROTECT(vec_proxy_compare_method(x));
+  SEXP out = vec_proxy_compare_invoke(x, method);
+  UNPROTECT(1);
+  return out;
+}
+
+static inline SEXP vec_proxy_compare_method(SEXP x) {
```
lionel-, line 114: If possible it would be nice to remove the code repetition, this way we see there is no difference.

src/proxy.c
```diff
@@ -30,108 +39,234 @@ SEXP vec_proxy(SEXP x) {
...
 
-SEXP vec_proxy_method(SEXP x) {
-  return s3_find_method("vec_proxy", x, vctrs_method_table);
+#undef DF_PROXY
+
+// [[ register() ]]
+SEXP vctrs_df_proxy(SEXP x, SEXP kind) {
+  enum vctrs_proxy_kind c_kind;
+
+  if (kind == Rf_install("default")) {
```
lionel-, line 225: This would be cleaner by passing enum values from the R side directly.

tests/testthat/test-compare.R
```diff
@@ -167,6 +164,37 @@ test_that("vec_proxy_compare.POSIXlt() correctly orders around DST", {
...
+  expect_identical(vec_proxy_compare(df), 1:2)
+
+  df_col <- new_data_frame(list(y = 1:2))
+  df <- new_data_frame(list(x = df_col))
+
+  expect_identical(vec_proxy_compare(df), 1:2)
+})
+
+test_that("vec_proxy_order() works on deeply nested lists", {
+  df_col <- new_data_frame(list(z = list("b", "a", "b")))
```
lionel-, line 188: Can you use `data_frame()` please? it's easier to understand the structure without the extra list.

R/compare.R
hadley: I wouldn't want to claim such a general principle; I think we should just state that lists have special behaviour and describe what it is.

R/compare.R
```diff
@@ -176,6 +157,44 @@ order_proxy <- function(proxy, direction = "asc", na_value = "largest") {
...
+#' `vec_order()` and `vec_sort()` for other vector types.
+#'
+#' The default is to call [vec_proxy_compare()].
+#'
+#' Lists are special cased, and return an integer vector that can be used to
+#' order the elements in the list by first appearance.
+#'
+#' @param x A vector x.
+#' @inheritParams ellipsis::dots_empty
+#' @return A 1d atomic vector or a data frame.
```
hadley, line 175: Should `vec_proxy_order()` always return an integer vector or data frame of integer vectors?

tests/testthat/test-compare.R
```diff
@@ -115,13 +106,19 @@ test_that("xtfrm.vctrs_vctr works for variety of base classes", {
   expect_equal(xtfrm.vctrs_vctr(letters[x]), x)
 })
 
-test_that("vec_proxy_compare() refuses to deal with lists", {
-  expect_error(vec_proxy_compare(list()), class = "vctrs_error_unsupported")
+test_that("vec_proxy_order() orders list using order of appearance", {
+  x <- 1:2
+  y <- 2:4
+  z <- 3
```
hadley, line 112: Might make sense to do `z <- "a"` just to emphasise that these aren't the same type.

R/compare.R
```diff
@@ -176,6 +157,44 @@ order_proxy <- function(proxy, direction = "asc", na_value = "largest") {
   }
 }
 
+# Order proxy ------------------------------------------------------------------
+
+#' Ordering proxy
+#'
+#' Returns a proxy object for ordering. For vctrs_vctr objects, this determines
+#' the behavior of `order()` and `sort()` (via `xtfrm()`). It also powers
+#' `vec_order()` and `vec_sort()` for other vector types.
```
hadley, line 166: I'd be tempted to document the comparison and order proxy together so that you can more clearly describe the difference (i.e. they're the same except for lists, which we special case in order to allow ordering of data frames containing list-columns).

R/compare.R
DavisVaughan: I don't think so, because fractional doubles and character vectors can still be ordered as is

Issue #1165, opened 2020-06-24
Crash in repeated `vec_unique()` calls with largeish data


DavisVaughan:
The following crashes RStudio for me with CRAN vctrs. It creates around 3 million rows. The crash seems to be a bit random

```r
library(vctrs)
library(nycflights13)

x <- flights[-19]
x <- vec_rbind(!!!rep_len(list(x), 10))

for (i in 1:20) {
  vec_unique(x)
}
```

This has 18 columns. I was also able to make it crash with 17, but couldn't seem to make it crash with 16 or less

lldb with `breakpoint set -E c` got me here:

```
Process 49553 stopped
* thread #1, queue = 'com.apple.main-thread', stop reason = EXC_BAD_ACCESS (code=EXC_I386_GPFLT)
    frame #0: 0x000000014050dc74 vctrs.so`dbl_equal_scalar_na_equal(x=0x000c00015a8b2030, y=0x000c00015ab43c70) at equal.h:44:21
   41     return isnan(x);
   42   }
   43   static inline int dbl_equal_scalar_na_equal(const double* x, const double* y) {
-> 44     const double xi = *x;
   45     const double yj = *y;
   46  
   47     switch (dbl_classify(xi)) {
Target 0: (R) stopped.
```

DavisVaughan:
@lionel- I have a guess about this, but I'm not sure.

Right here you assign this object that the dictionary should protect
https://github.com/r-lib/vctrs/blob/a518ead0b08be29beea287d11e17edc1017e16da/src/dictionary.c#L260

And right after that we drop into `init_dictionary_df()`
https://github.com/r-lib/vctrs/blob/a518ead0b08be29beea287d11e17edc1017e16da/src/dictionary.c#L271

This reassigns `d->protect` to a newly created handle
https://github.com/r-lib/vctrs/blob/a518ead0b08be29beea287d11e17edc1017e16da/src/dictionary.c#L221

But that `d->protect` spot is not reprotected on the way out at the calling site of `init_dictionary_df()`. Shouldn't it be? i.e. after this is called? https://github.com/r-lib/vctrs/blob/a518ead0b08be29beea287d11e17edc1017e16da/src/dictionary.c#L271

lionel-:
I think you're right, there should be a `PROTECT(d->protect);` after initialisation because there is an `R_alloc()` before we return.

PR #1167, opened 2020-06-26
Reprotect after calling `init_dictionary_df()`

DavisVaughan:
Possibly fixes https://github.com/r-lib/vctrs/issues/1165

It seems to not crash for me now? It was pretty random though, so it is hard to say for certain

### DIFF
--- a/src/dictionary.c
+++ b/src/dictionary.c
@@ -272,6 +272,10 @@ static struct dictionary* new_dictionary_opts(SEXP x, struct dictionary_opts* op
   default: Rf_error("Internal error: Unimplemented type in `new_dictionary()`.");
   }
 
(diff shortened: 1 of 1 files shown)


src/dictionary.c
```diff
@@ -272,6 +272,10 @@ static struct dictionary* new_dictionary_opts(SEXP x, struct dictionary_opts* op
   default: Rf_error("Internal error: Unimplemented type in `new_dictionary()`.");
   }
 
+  // Reprotect `d->protect` for the case of `init_dictionary_df()`,
```
lionel-, line 275: This comment is a bit too specific, it can just be "`init_` functions might allocate", if the protect is not self-explanatory.

PR #1174, opened 2020-07-06
Avoid concatenation of class vector if already contains data.frame

krlmlr:
I'd like to use `new_data_frame()` in tibble, it's much easier to create correct classes if callers are allowed to pass `class = c(..., "data.frame")` . Can provide timings if required.

### DIFF
--- a/src/type-data-frame.c
+++ b/src/type-data-frame.c
@@ -174,6 +174,9 @@ static SEXP c_data_frame_class(SEXP cls) {
   if (Rf_length(cls) == 0) {
     return classes_data_frame;
   }
+  if (STRING_ELT(cls, Rf_length(cls) - 1) == strings_data_frame) {
+    return cls;
+  }
 
   SEXP args = PROTECT(Rf_allocVector(VECSXP, 2));
   SET_VECTOR_ELT(args, 0, cls);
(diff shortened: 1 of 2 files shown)


lionel-:
As we have already discussed, this is not how we do constructors. `new_tibble()` should not do this either for consistency.

krlmlr:
The concatenation makes the constructor about two times as expensive, important e.g. in `nest()` with thousands of nests.

## This PR

``` r
library(vctrs)

sc <- c("tbl_df", "tbl")
c <- c("tbl_df", "tbl", "data.frame")

bench::mark(
  new_data_frame(class = sc),
  new_data_frame(class = c),
  iterations = 1e6
)
#> # A tibble: 2 x 6
#> 1 new_data_frame(class = sc)    1.9µs   2.26µs   342980.    6.39KB     43.6
#> 2 new_data_frame(class = c)     861ns   1.01µs   751211.        0B     24.0
```

<sup>Created on 2020-07-06 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

## 6bea6e3

``` r
library(vctrs)

sc <- c("tbl_df", "tbl")
c <- c("tbl_df", "tbl", "data.frame")

bench::mark(new_data_frame(class = sc), iterations = 1e6)
#> # A tibble: 1 x 6
#> 1 new_data_frame(class = sc)   1.72µs   2.09µs   368838.    6.39KB     46.8
```

<sup>Created on 2020-07-06 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

PR #1179, opened 2020-07-07
Add vctrs methods for `omit` and `exclude`

### LINKED ISSUE #1160: Special case objects of class omit
```R
x <- na.omit(c(1:5, NA))
y <- attr(x, "na.action")

x[y]
#> [1] NA
vctrs::vec_slice(x, y)
#> Error: Must subset elements with a valid subscript vector.
#> x Subscript has the wrong type `omit`.
```

From https://github.com/tidyverse/haven/issues/530

lionel-:
Closes #1160

These classes are coercible to numeric but they lose their class when combined with integer or double. I think the logic is similar to the data frame classes that @DavisVaughan implemented for tidymodels. The class identifies how the vector was generated and combining the vector with other data denatures it.


tslumley:
Great, thanks.

PR #1181, opened 2020-07-08
Explain double vs. numeric

krlmlr:
I tried `vec_ptype2.foo.numeric()` and wondered why it didn't work. Should we use `numeric()` in the text instead where appropriate, to illustrate the difference?

### DIFF
--- a/vignettes/s3-vector.Rmd
+++ b/vignettes/s3-vector.Rmd
@@ -19,10 +19,7 @@ This vignette shows you how to create your own S3 vector classes. It focuses on 
 
 I assume that you're already familiar with the basic machinery of S3, and the vocabulary I use in Advanced R: constructor, helper, and validator. If not, I recommend reading at least the first two sections of [the S3 chapter](https://adv-r.hadley.nz/s3.html) of _Advanced R_.
 
-This article refers to "vectors of numbers" as *double vectors*. Here, "double" stands for ["double precision floating point number"](https://en.wikipedia.org/wiki/Double-precision_floating-point_format). It is also the underlying *storage mode* for `numeric()` vectors in R, see the "Mode names" section in `?mode`. This is a deliberate choice, for two reasons:
-
-1. each vector class should be aware of its storage mode,
-2. using `numeric` instead of `double` will not work for method names for `vec_ptype2()` and `vec_cast()`.
+This article refers to "vectors of numbers" as *double vectors*. Here, "double" stands for ["double precision floating point number"](https://en.wikipedia.org/wiki/Double-precision_floating-point_format), see also `double()`.
 
 ```{r setup}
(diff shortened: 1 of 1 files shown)


vignettes/s3-vector.Rmd
```diff
@@ -19,6 +19,11 @@ This vignette shows you how to create your own S3 vector classes. It focuses on
 
 I assume that you're already familiar with the basic machinery of S3, and the vocabulary I use in Advanced R: constructor, helper, and validator. If not, I recommend reading at least the first two sections of [the S3 chapter](https://adv-r.hadley.nz/s3.html) of _Advanced R_.
 
+This article refers to "vectors of numbers" as *double vectors*. Here, "double" stands for ["double precision floating point number"](https://en.wikipedia.org/wiki/Double-precision_floating-point_format). It is also the underlying *storage mode* for `numeric()` vectors in R, see the "Mode names" section in `?mode`. This is a deliberate choice, for two reasons:
```
lionel-, line 22: I'm not sure we need to define "storage mode" as if it's a general concept to be aware of? Also isn't the storage mode "numeric" rather than "double"? Maybe make it clear that we generally prefer the `typeof()` than the `mode()` which is a legacy of S? The challenge is to stay clear and to the point so it doesn't feel like a digression right at the start of the vignette, and also not to introduce further confusion with these tricky concepts.

krlmlr:
I tweaked a bit, turns out `?double` isn't too bad at explaining this.

My other concern is the overuse of "double" in this text and the potential confusion with "double dispatch". Clearly `numeric` is used much more frequently in R code than `double` ([73k results](https://github.com/search?q=org%3Acran+numeric+extension%3A.R) vs. [17k results](https://github.com/search?q=org%3Acran+double+extension%3A.R) for a CRAN code search. Not sure we want to address this here and now, if we wanted to I have it almost ready: https://github.com/krlmlr/vctrs/tree/f-double.

vignettes/s3-vector.Rmd
```diff
@@ -203,6 +205,10 @@ Next we define methods that say that combining a `percent` and double should yie
 
 Because double dispatch is a bit of a hack, we need to provide two methods. It's your responsibility to ensure that each member of the pair returns the same result: if they don't you will get weird and unpredictable behaviour.
 
+The double dispatch mechanism requires us to refer to the underlying type, `double`, in the method name.
+If we implemented `vec_ptype2.vctrs_percent.numeric()`, it would never be called.
+See also `?mode`, especially the "Mode names" section.
```
lionel-, line 210: I'd rather not mention `?mode` since it's about the old storage modes from S.

krlmlr:
Done.

lionel-:
Thanks!

PR #1184, opened 2020-07-08
vec_ptype_abbr(check_named = , shape = )

### LINKED ISSUE #781: Add "named" for named vectors
in `vec_ptype_abbr()` (and perhaps `vec_ptype_full()`), because tibbles now allow named vectors.

Current:

``` r
vctrs::vec_ptype_abbr(c(a = 1))
#> [1] "dbl"
```

<sup>Created on 2020-01-28 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

Desired:

``` r
vctrs::vec_ptype_abbr(c(a = 1))
#> [1] "named dbl"
```

<sup>Created on 2020-01-28 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

krlmlr:
Closes #781.

Helps https://github.com/r-lib/pillar/issues/168.

Do we extend to `vec_ptype_full()`? Which parts?

### DIFF
--- a/R/ptype-abbr-full.R
+++ b/R/ptype-abbr-full.R
@@ -13,8 +13,12 @@
 #' characters where possible.
 #'
 #' @param x A vector.
-#' @param prefix_named If `TRUE`, to add a prefix for named vectors.
-#' @param suffix_shape If `TRUE` (the default), append the shape of the vector.
+#' @param prefix_named
+#'   If `TRUE`, to add a prefix for named vectors.
+#'   This argument is handled by the generic and will not be passed on to methods.
+#' @param suffix_shape
+#'   If `TRUE` (the default), append the shape of the vector.
+#'   This argument is handled by the generic and will not be passed on to methods.
 #' @inheritParams ellipsis::dots_empty
 #'
 #' @keywords internal
@@ -39,16 +43,16 @@ vec_ptype_abbr <- function(x, ..., prefix_named = FALSE, suffix_shape = TRUE) {
     ellipsis::check_dots_empty()
   }
 
-  abbr <- vec_ptype_abbr_(x, prefix_named = prefix_named, suffix_shape = suffix_shape)
+  abbr <- vec_ptype_abbr_dispatch(x)
 
   return(paste0(
-    if (prefix_named && !is.null(vec_names(x))) "named ",
+    if ((prefix_named || is_bare_list(x)) && !is.null(vec_names(x))) "named ",
     abbr,
     if (suffix_shape) vec_ptype_shape(x)
   ))
   UseMethod("vec_ptype_abbr")
 }
-vec_ptype_abbr_ <- function(x, ..., prefix_named = FALSE, suffix_shape = TRUE) {
+vec_ptype_abbr_dispatch <- function(x, ...) {
   UseMethod("vec_ptype_abbr")
 }
 
@@ -72,15 +76,12 @@ vec_ptype_full.default <- function(x, ...) {
 }
 
 #' @export
-vec_ptype_abbr.default <- function(x, ..., prefix_named = FALSE) {
+vec_ptype_abbr.default <- function(x, ...) {
   if (is.object(x)) {
     unname(abbreviate(vec_ptype_full(x), 8))
-  } else if (is_list(x)) {
-    named <- is_character(names(x))
-    # Always print "named" even if !prefix_named, for compatibility
-    paste0(if (named && !prefix_named) "named ", "list")
   } else if (is_vector(x)) {
     switch(typeof(x),
+      list = "list",
       logical = "lgl",
       integer = "int",
       double = "dbl",
--- a/R/ptype-abbr-full.R
+++ b/R/ptype-abbr-full.R
@@ -13,8 +13,8 @@
 #' characters where possible.
 #'
 #' @param x A vector.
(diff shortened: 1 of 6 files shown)


R/ptype-abbr-full.R
```diff
@@ -13,6 +13,8 @@
 #' characters where possible.
 #'
 #' @param x A vector.
+#' @param check_named Add a prefix for named vectors.
```
lionel-, line 16: `prefix_named` and `suffix_shape`?

R/ptype-abbr-full.R
```diff
@@ -56,11 +58,12 @@ vec_ptype_full.default <- function(x, ...) {
 }
 
 #' @export
-vec_ptype_abbr.default <- function(x, ...) {
+vec_ptype_abbr.default <- function(x, ..., check_named = FALSE, shape = TRUE) {
```
lionel-, line 61: Maybe these arguments should be handled in the generic rather than in methods? Otherwise all implementations will need to deal with them.

Alternatively, these could be arguments for the default method only, and we'd use `ellipsis::check_dots_used()` in the generic. This is more straightforward so might make more sense as a first step.

krlmlr: I don't see how this can be handled in the generic. The call to `UseMethod()` never returns. We could implement

```r
vec_ptype_label <- function(x, ... type = c("abbr", "full"), prefix_named = FALSE, suffix_shape = TRUE) {
  ...
}
```

and use that in pillar instead, would you support this?

(We now have quite a few packages implementing `vec_ptype_abbr()` methods, so changing `vec_ptype_abbr()` to a non-generic (and implementing a different generic) won't fly.)

lionel-: > The call to UseMethod() never returns.

You can call `UseMethod("vec_ptype_abbr")` from a helper function.

krlmlr: Wow, clever! Do you have an example or documentation for this... feature?

lionel-: oh just make sure there is also an unreachable `UseMethod()` call in the generic for static analysis (e.g. roxygen)

```r
out <- vec_ptype_abbr_dispatch(x, ...)

...

return(out)
UseMethod("vec_ptype_abbr")
```

krlmlr: Are there any requirements to the signature of the helper function?

lionel-: It should take the arguments dispatched to the methods. Here I'd make it `x, ...` to make it clearer that methods should take `...` (even though I think R CMD check will require these arguments based on the signature of the fake generic).

krlmlr: Thanks, I can do this.

R/ptype-abbr-full.R
krlmlr: I like `prefix_named` and `suffix_shape` too.

krlmlr:
I need to double-check what happens with pillar. Please let me know your preferred argument naming, it's easy to adapt now.

R/ptype-abbr-full.R
```diff
@@ -32,10 +34,21 @@ vec_ptype_full <- function(x, ...) {
...
+  abbr <- vec_ptype_abbr_(x, prefix_named = prefix_named, suffix_shape = suffix_shape)
+
+  return(paste0(
+    if (prefix_named && !is.null(vec_names(x))) "named ",
+    abbr,
+    if (suffix_shape) vec_ptype_shape(x)
+  ))
+  UseMethod("vec_ptype_abbr")
+}
+vec_ptype_abbr_ <- function(x, ..., prefix_named = FALSE, suffix_shape = TRUE) {
```
lionel-, line 51: We usually suffix these helpers with `_dispatch()`.

Since we deal with these prefixes and suffixes in the generic, `prefix_named` and `suffix_shape` shouldn't be passed on to methods.

R/ptype-abbr-full.R
```diff
@@ -59,14 +72,15 @@ vec_ptype_full.default <- function(x, ...) {
...
 #' @export
-vec_ptype_abbr.default <- function(x, ...) {
+vec_ptype_abbr.default <- function(x, ..., prefix_named = FALSE) {
   if (is.object(x)) {
     unname(abbreviate(vec_ptype_full(x), 8))
   } else if (is_list(x)) {
     named <- is_character(names(x))
-    paste0(if (named) "named ", "list", vec_ptype_shape(x))
+    # Always print "named" even if !prefix_named, for compatibility
+    paste0(if (named && !prefix_named) "named ", "list")
```
lionel-, line 81: Should move to the generic.

krlmlr:
Should be good to merge now. The covr build suffers from macOS problems.

R/ptype-abbr-full.R
```diff
@@ -13,6 +13,12 @@
 #' characters where possible.
 #'
 #' @param x A vector.
+#' @param prefix_named
+#'   If `TRUE`, to add a prefix for named vectors.
```
lionel-, line 17:
```suggestion
#'   If `TRUE`, add a prefix for named vectors.
```

lionel- (commented):
Looks good! Can you also add tests for data frames and matrices with row names please? Those should now get a `named` prefix.

PR #1188, opened 2020-07-15
Stop using `STRING_PTR()`

lionel-:
So we don't mess up refcounts of CHARSXP.

### DIFF
--- a/src/translate.c
+++ b/src/translate.c
@@ -213,7 +213,6 @@ static SEXP chr_translate_encoding(SEXP x, R_len_t size) {
   const SEXP* p_x = STRING_PTR_RO(x);
 
   SEXP out = PROTECT(r_clone_referenced(x));
-  SEXP* p_out = STRING_PTR(out);
 
   const void *vmax = vmaxget();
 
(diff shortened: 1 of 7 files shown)


src/utils.c
```diff
@@ -898,17 +898,40 @@ void* r_vec_deref(SEXP x) {
   case INTSXP: return INTEGER(x);
   case REALSXP: return REAL(x);
   case CPLXSXP: return COMPLEX(x);
-  case STRSXP: return STRING_PTR(x);
```
DavisVaughan, line 901: 👍

src/translate.c
```diff
@@ -213,19 +213,18 @@ static SEXP chr_translate_encoding(SEXP x, R_len_t size) {
...
-  SEXP* p_out = STRING_PTR(out);
 
   const void *vmax = vmaxget();
 
   for (int i = 0; i < size; ++i) {
     SEXP chr = p_x[i];
 
     if (Rf_getCharCE(chr) == CE_UTF8) {
-      p_out[i] = chr;
+      SET_STRING_ELT(out, i, chr);
```
DavisVaughan, line 223: This is the most common case when translating (do nothing), and `out` is just a clone of `x`, so I imagine that we could probably just skip this entirely (for performance). Not saying you have to make that change in this PR, just something I noticed that we might want to revisit.

PR #1195, opened 2020-07-19
Check against dev versions

krlmlr:
Currently failing with testthat.

### DIFF
--- a/.github/workflows/R-CMD-check-dev.yaml
+++ b/.github/workflows/R-CMD-check-dev.yaml
@@ -1,0 +1,96 @@
+on:
+  push:
+    branches:
+      - master
+  pull_request:
+    branches:
+      - master
+  schedule:
+    - cron:  '0 8 * * *'
+
+name: rcc dev
+
+jobs:
+  R-CMD-check:
+    runs-on: ${{ matrix.config.os }}
+
+    name: ${{ matrix.config.os }} (${{ matrix.config.r }})
+
+    strategy:
+      fail-fast: false
+      matrix:
+        config:
+          - {os: ubuntu-16.04,   r: '4.0', rspm: "https://packagemanager.rstudio.com/cran/__linux__/xenial/latest"}
+
+    env:
+      R_REMOTES_NO_ERRORS_FROM_WARNINGS: true
+      RSPM: ${{ matrix.config.rspm }}
+      GITHUB_PAT: ${{ secrets.GITHUB_TOKEN }}
+
+    steps:
+      - uses: actions/checkout@v2
+
+      - uses: r-lib/actions/setup-r@master
+        with:
+          r-version: ${{ matrix.config.r }}
+
+      - name: "Prepare timestamp for cache"
+        if: runner.os != 'Windows'
+        id: date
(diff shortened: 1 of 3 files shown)


.github/workflows/R-CMD-check-dev.yaml
```diff
@@ -0,0 +1,96 @@
+on:
```
lionel-, line 1: Why do we need a new workflow file? What does it contain? How do we maintain it as new R-CMD-check.yaml versions come in? Is it based on R-CMD-check.yaml? In that case do we need a commit with a clear diff?

tests/testthat/test-type-data-frame.R
```diff
@@ -110,8 +110,9 @@ test_that("combining data frames with foreign classes uses fallback", {
     foo <- structure(mtcars[1:3], class = c("foo", "data.frame"))
     bar <- structure(mtcars[4:6], class = c("bar", "data.frame"))
     baz <- structure(mtcars[7:9], class = c("baz", "data.frame"))
-    expect_warning(vec_ptype_common_df_fallback(foo, bar, baz))
-    expect_warning(vec_ptype_common_df_fallback(foo, baz, bar, baz, foo, bar))
+    # Nested expect_warning() require testthat > 2.3.2
+    suppressWarnings(expect_warning(vec_ptype_common_df_fallback(foo, bar, baz)))
```
lionel-, line 114: How about repeated expect-warning? I assume the number of warnings is deterministic?

hadley:
I don't think we should be routinely testing with dev testthat yet.

tests/testthat/test-type-data-frame.R
krlmlr: This works for the dev version of testthat, not for the CRAN version.

.github/workflows/R-CMD-check-dev.yaml
krlmlr: - Splitting workflows into multiple files (one per purpose) seems better than one big workflow with lots of branches.
- Dependencies are installed from GitHub via `remotes::install_dev()`, caching is enabled. Checks on push/PR and also once per day, it seems difficult to trigger a check on upstream changes
- This can be part of https://github.com/r-lib/actions, needs some exposure before that
- Yes
- We can have that commit in r-lib/actions if necessary

krlmlr:
testthat isn't special here, the workflow checks all dev versions of all imported and suggested packages. This helps catch problems early.

The changes here are necessary so that the checks pass with dev testthat. Should this be issues in the testthat repo instead?

lionel-:
If Hadley is uncomfortable with routinely checking packages against dev testthat we definitely shouldn't do that.

I'm a bit uncomfortable with introducing an entirely new workflow file to maintain, especially if there is no commit to show a proper diff.

Overall think we should better close this. You can push https://github.com/r-lib/vctrs/pull/1195/commits/5da77d4b47e9de34844a2120c16a1334ac60042e to master though.

krlmlr:
The workflow is used successfully in pillar and tibble, not entirely new. I will look into adding it to r-lib/actions.

krlmlr:
Pushed that commit.

PR #1199, opened 2020-07-21
Avoid calling `STRING_ELT()` on unnamed data frames

### LINKED ISSUE #1198: Allow for unnamed data frame input in dictionary functions
``` r
library(vctrs)
x <- new_data_frame(list(1, 2))
vec_unique_count(x)
#> Error in vec_unique_count(x): STRING_ELT() can only be applied to a 'character vector', not a 'NULL'
```

I don't think that an unnamed data frame is always a "corrupt" data frame.

I think that creating a data frame can be a useful signal that you want to use a row wise vctrs function, even if you don't know the column names. For example, consider this alternate `n_distinct()` implementation. The `set_names()` line is currently required and is actually pretty expensive when called a large number of times, but should not be necessary if `vec_unique_count()` didn't error here

```r
n_distinct <- function(..., na.rm = FALSE) {
  size <- vec_size_common(...)
  
  x <- vec_recycle_common(..., .size = size)
  #x <- set_names(x, seq_along(x))
  x <- new_data_frame(x, n = size)
  
  if (isTRUE(na.rm)){
    x <- vec_slice(x, !reduce(map(x, vec_equal_na), `|`))
  }
  
  vec_unique_count(x)
}
```

DavisVaughan:
Closes #1198 

Simple adjustment to `df_flatten_loop()` to avoid extracting names from an unnamed data frame.

Even with unnamed data frames, we return a data frame from `df_flatten()` that at least has `""` as column names. This makes it easy to ensure proper behavior in the edge case where an unnamed data frame has a df-col that has names (this is `df3` in the added tests)

### DIFF
--- a/src/type-data-frame.c
+++ b/src/type-data-frame.c
@@ -676,17 +676,23 @@ SEXP df_flatten(SEXP x) {
 static R_len_t df_flatten_loop(SEXP x, SEXP out, SEXP out_names, R_len_t counter) {
   R_len_t n = Rf_length(x);
   SEXP x_names = PROTECT(r_names(x));
+  bool has_names = x_names != R_NilValue;
 
   for (R_len_t i = 0; i < n; ++i) {
     SEXP col = VECTOR_ELT(x, i);
 
     if (is_data_frame(col)) {
       counter = df_flatten_loop(col, out, out_names, counter);
-    } else {
-      SET_VECTOR_ELT(out, counter, col);
-      SET_STRING_ELT(out_names, counter, STRING_ELT(x_names, i));
-      ++counter;
+      continue;
     }
+
+    SET_VECTOR_ELT(out, counter, col);
+
+    if (has_names) {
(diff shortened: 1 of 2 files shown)


lionel-:
I think all data frames should have a STRSXP for names.

DavisVaughan:
I feel like the pattern I proposed for `n_distinct()` might prove to be somewhat common. Is it possible to change `new_data_frame()` to always add the empty string for unnamed columns? Or, since we use `new_data_frame()` for testing these unnamed data frames, maybe we do need a slightly higher level `as_data_frame()` wrapper like you mentioned in https://github.com/tidyverse/dplyr/pull/5430#pullrequestreview-453085296, and that could enforce STRSXP column names.

lionel-:
>  Is it possible to change new_data_frame() to always add the empty string for unnamed columns?

The problem is that we then need another way to create unnamed data frames for unit tests and special cases. I think keeping `new_data_frame()` as a bare bones ctor (with optional input checking) and adding `as_data_frame()` could make sense.

PR #1201, opened 2020-07-22
Implement `data_frame()`

### LINKED ISSUE #1198: Allow for unnamed data frame input in dictionary functions

DavisVaughan:
Closes #1198 
Closes #1199 (an alternative to that PR)

This PR beefs up the existing internal `data_frame()` helper and exports it.

A few new features:

- `.size` argument
- `.name_repair` argument
- Tidy recycling of columns (which makes it safer to use too)
- Fully unnamed inputs are automatically given `""` for column names to prevent creation of corrupt data frames
- Auto-splicing of unnamed data frames
- Dynamic dots

Altogether this gives a pretty performant data frame constructor that follows tidyverse semantics.


src/type-data-frame.c
```diff
@@ -174,6 +174,153 @@ static SEXP c_data_frame_class(SEXP cls) {
...
+      SET_STRING_ELT(out_names, loc, p_names[i]);
+      ++loc;
+      continue;
+    }
+
+    SEXP col = VECTOR_ELT(x, i);
+    SEXP col_names = PROTECT(r_names(col));
+
+    if (TYPEOF(col_names) != STRSXP) {
+      Rf_errorcall(
```
lionel-, line 296: `stop_internal()`

tests/testthat/test-type-data-frame.R
```diff
@@ -410,6 +410,89 @@ test_that("new_data_frame() zaps existing attributes", {
...
+})
+
+test_that("named data frames are not spliced", {
+  df_col <- data_frame(x = 2, y = 3)
+  df <- data_frame(w = 1, col = data_frame(x = 2, y = 3), z = 4)
+
+  expect_identical(df$col, df_col)
+})
+
+test_that("spliced data frames without names are caught", {
```
lionel-, line 461: Should we repair recursively instead? That would seem more consistent.

src/type-data-frame.c
```diff
@@ -174,6 +174,153 @@ static SEXP c_data_frame_class(SEXP cls) {
...
+      );
+    }
+
+    const SEXP* p_col_names = STRING_PTR_RO(col_names);
+    r_ssize col_i = 0;
+
+    r_ssize stop = loc + r_length(col);
+
+    for (; loc < stop; ++loc, ++col_i) {
+      SET_VECTOR_ELT(out, loc, VECTOR_ELT(col, col_i));
```
lionel-, line 309: Do we need to check the size of columns? I guess it's ok to assume a correctly constructed data frame when we don't read or write to the column (no risk of crash).

tests/testthat/test-type-data-frame.R
lionel-: oh in that case that would just be repair after splicing, which seems fine.

R/type-data-frame.R
```diff
@@ -28,13 +28,12 @@ new_data_frame <- function(x = list(), n = NULL, ..., class = NULL) {
...
 
-# Light weight constructor used for tests - avoids having to repeatedly do
-# stringsAsFactors = FALSE etc. Should not be used in internal code as is
-# not a real helper as it lacks value checks.
-data_frame <- function(...) {
-  cols <- list(...)
-  new_data_frame(cols)
+data_frame <- function(...,
+                       .size = NULL,
+                       .name_repair = c("minimal", "unique", "universal", "check_unique")) {
```
lionel-, line 33: The default should be `"unique"`.

lionel-:
@krlmlr Maybe `df_splice()` could be useful for tibble?

src/type-data-frame.c
```diff
@@ -174,6 +174,153 @@ static SEXP c_data_frame_class(SEXP cls) {
...
+      continue;
+    }
+
+    SEXP col = VECTOR_ELT(x, i);
+
+    if (is_data_frame(col)) {
+      any_needs_splice = true;
+      break;
+    }
+  }
```
krlmlr, line 245: Can we get rid of the first loop and compute the width right away, with shortcut if the width is unchanged?

src/type-data-frame.c
DavisVaughan: Doesn't `vec_recycle_common()` handle this?

src/type-data-frame.c
DavisVaughan: This does short circuit if the width is unchanged. If we get through the first loop and don't find any unnamed data frames then we are done (i.e. the width is unchanged)

src/type-data-frame.c
lionel-: In the nested data frames? I wouldn't think so.

lionel-: (as mentioned above it's probably ok not to check consistency here)

R/type-data-frame.R
DavisVaughan: The default in tibble is `"check_unique"`. The default in `vec_as_names()` is `"minimal"`. Do we need to worry about having so many different defaults? (Especially if we add `tibble0()` on top of this?)

src/type-data-frame.c
krlmlr: I'm proposing to get rid of the first loop and short-cirtuiting after the second loop if width is unchanged. Reduces code complexity?

R/type-data-frame.R
lionel-: oh yeah I meant to be consistent with tibble, my mistake.

src/type-data-frame.c
DavisVaughan: Oh, the point of the first loop is to never allocate the `splice` vector in the common case of having no data frame inputs to splice. It does add a little complexity, but I think it is worth it here to avoid the allocation

DavisVaughan:
Rather than export `df_splice()`, I was hoping that, at the end of `tibble_quos()`, `splice_dfs()` and `new_tibble()` could just be replaced with `tibble0()` (i.e. `data_frame()` + a small wrapper that adds the tibble class). Since tibble subclasses name repair errors in `set_repaired_names()`, that will still have to be done separately, but I imagine that a call to `data_frame()` with `.name_repair = "minimal"` might work nicely?

lionel-:
We also need to think about scalar type errors:

```r
fn <- function() NULL
fit <- lm(1:3 ~ 1)

df_list(x = fn, y = fit)
#> Error: All columns in a data frame must be vectors.
#> * Column `x` is a function.
#> * Column `y` is a `lm` object.
```

NEWS.md
```diff
@@ -7,6 +7,11 @@
   list-columns are easier to create, and unnamed data frame input is
   automatically spliced.
 
+* New `df_list()` for safely and consistently constructing the data structure
+  underlying a data frame, a named list of equal-length vectors. It is useful
+  in combination with `new_data_frame()` for creating user-friendly
+  constructors for data frame subclasses.
```
lionel-, line 13: that use the tidyverse rules for recycling and determining types.

R/type-data-frame.R
```diff
@@ -48,6 +60,47 @@ new_data_frame <- fn_inline_formals(new_data_frame, "x")
 #' - The dots are dynamic, allowing for splicing of lists with `!!!` and
 #'   unquoting.
 #'
+#' @seealso
+#' [new_data_frame()] for constructing data frame subclasses from validated
```
lionel-, line 64:
```suggestion
#' [new_data_frame()] for constructing data frame subclasses from a validated
```

PR #1206, opened 2020-08-04
Add a user FAQ about type compatibility

lionel-:


lionel-:
@DavisVaughan If you have time can you take a quick look please? The goal of this FAQ is to have somewhere to link to in the documentation of functions that take a common type. For example in `purrr::simplify()` or in `dplyr::bind_rows()`.

man/faq/user/faq-compatibility-types.Rmd
```diff
@@ -0,0 +1,62 @@
...
+```
+
+```{r, include = FALSE}
+stopifnot(rlang::is_installed("dplyr"))
+```
+
+Two vectors are __compatible__ when you can safely:
+
+- Combine them into one larger vector.
+- Assign values from one of the vector in the other vector.
```
DavisVaughan, line 12:
```suggestion
- Assign values from one of the vectors into the other vector.
```

man/faq/user/faq-compatibility-types.Rmd
```diff
@@ -0,0 +1,62 @@
...
+
+Examples of compatible types are integer and double vectors. On the other hand, integer and character vectors are not compatible.
```
DavisVaughan, line 14: Integer and doubles are interesting because integer->double is always compatible, but you can't always assign a double into an integer vector. Not sure if this is worth mentioning, but safe assignment was one of your two defining principles for compatibility so I figured I'd bring it up.

DavisVaughan: Update: I see now that you mention the integer/double issue below in the conversion section. Is the compatibility principle above regarding assignment sufficiently met when you can always assign in at least one of the two directions? (i.e. we can always assign integer->double, but not always double->integer, but that is enough for them to be compatible)

man/faq/user/faq-compatibility-types.Rmd
```diff
@@ -0,0 +1,62 @@
...
+   ```
+
+In general, the common type is the _richer_ type, in other words the type that can represent the most values. Logical vectors are at the bottom of the hierarchy of numeric types because they can only represent two values (not counting missing values). Then come integer vectors, and then doubles. Here is the vctrs type hierarchy for the fundamental vectors:
+
+\\figure{coerce.png}
+
+
+# Type conversion and lossy cast errors
+
+Type compatibility does not necessarily mean that you can __convert__ one type to the other type. That's because one of the type might support a larger set of possible values. For instance, integer and double vectors are compatible but double vectors can't be converted to integer if they contain fractional values.
```
DavisVaughan, line 44:
```suggestion
Type compatibility does not necessarily mean that you can __convert__ one type to the other type. That's because one of the types might support a larger set of possible values. For instance, integer and double vectors are compatible, but double vectors can't be converted to integer if they contain fractional values.
```

man/faq/user/faq-compatibility-types.Rmd
lionel-: Yes I framed "compatible" to mean "have a common type". Do you think that makes sense? The lossy cast errors are not compatible type errors:

```r
vec_assign(1:3, 2, 1.5)
#> Error: Can't convert from <double> to <integer> due to loss of precision.
#> * Locations: 1

class(last_error())
#> [1] "vctrs_error_cast_lossy" "vctrs_error"            "rlang_error"
#> [4] "error"                  "condition"
```

DavisVaughan: Yea, after reading the whole article I think it makes sense as is. If I was a new user, I think this would help a lot.

PR #1207, opened 2020-08-04
Add a user FAQ about scalar type errors

lionel-:


lionel-:
- Adds user FAQ about "Input must be a vector" error.

- Adds developer FAQ about fixing that error for list and (improperly constructed) data frame subclasses.

Issue #1135, opened 2020-06-04
Generalise base type fallback


lionel-:
Automatically fall back to base type if it's explicitly inherited from. This generalises the data frame fallback.

Would help with cases like tidyverse/dplyr#5296.

deann88:
Probably this one is also connected:
```
> # doesn't bind on common_class_fallback 
> bind_rows(
+   tibble(a = jsonlite:::as.scalar("chr"), b = 2), 
+   tibble(b = 3, c = TRUE, d = FALSE)
+ )
Error: Can't combine `..1` <scalar> and `..2` <vctrs:::common_class_fallback>.
Run `rlang::last_error()` to see where the error occurred.
> 
> bind_rows(
+   tibble(a = "chr", b = 2), 
+   tibble(b = 3, c = TRUE, d = jsonlite:::as.scalar(2))
+ )
Error: Can't combine `..1` <vctrs:::common_class_fallback> and `..2` <scalar>.
Run `rlang::last_error()` to see where the error occurred.
```

PR #1213, opened 2020-08-07
Inherit from the base type by default in `new_vctr()`

lionel-:
No impact on revdeps.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,6 +1,9 @@
 
 # vctrs (development version)
 
+* `new_vctr()` now inherits from the base type of its input by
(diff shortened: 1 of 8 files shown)


DavisVaughan:
Original issue where we flipped it to `FALSE` https://github.com/r-lib/vctrs/pull/778

lionel-:
Right I think base type inheritance caused issues when we still had inheritance of coercion methods because it was causing all sorts of unrelated methods to be inherited by default. Once #1135 is resolved this will only be the case in a limited way, as part of a fallback coercion.

Issue #1217, opened 2020-08-13
Concatenating df-cols with row names causes excessive duplication


lionel-:
Data frame with rownames (non-repaired, non-recursive case)

```r
df <- data_frame(x = 1:2)
dfs <- rep(list(df), 1e2)
dfs <- map2(dfs, seq_along(dfs), set_rownames_recursively)
with_memory_prof(vec_rbind(!!!dfs))
#> [1] 10.3KB
```

Data frame with rownames (non-repaired, recursive case)

```r
df <- data_frame(
  x = 1:2,
  y = data_frame(x = 1:2)
)
dfs <- rep(list(df), 1e2)
dfs <- map2(dfs, seq_along(dfs), set_rownames_recursively)
with_memory_prof(vec_rbind(!!!dfs))
#> [1] 1.01MB
```

This is because row names are repaired on each assignment. I think the general way to fix this (and a bunch of other issues) is to proxy recursively (#1107).

PR #1218, opened 2020-08-13
Fix excessive duplication during concatenation of named vectors

lionel-:
* Add `vec_proxy_set_names()` which takes a `vctrs_owned` argument. This variant of `vec_set_names()` never dispatches and uses `vec_clone_referenced(x, owned)` instead of `r_clone_referenced()`.

```r
ints <- rep(list(set_names(1:3, letters[1:3])), 1e4)
bench::mark(
  base = unlist(ints),
  vctrs = vec_unchop(ints),
  check = FALSE
)[1:8]
#> # Before
#> 1 base        669.7µs  843.4µs  1101.     351.66KB     7.82   563     4
#> 2 vctrs          1.5s     1.5s     0.667    2.24GB    29.3      1    44

#> # After
#> 1 base       663.31µs 833.14µs     1203.     352KB    10.9    554     5
#> 2 vctrs        4.52ms   4.93ms      201.     625KB     4.18    96     2
```

* Don't repair row names on proxies. Instead, row names are now repaired as part of restoration of data frames.

```r
df <- new_data_frame(
  df_list(
    x = as.list(1:2),
    y = 1:2
  ),
  row.names = c("foo", "bar")
)
dfs <- rep(list(df), 3)
vec_unchop(dfs)

dfs <- rep(list(df), 1e2)
bench::mark(
  vctrs = vec_unchop(dfs),
  check = FALSE
)[1:8]
#> # Before
#> 1 vctrs         449ms    635ms      1.57    85.4MB     1.57     2     2

#> # After
#> 1 vctrs         905µs   1.04ms      948.    14.8KB     18.0   421     8
```

This issue isn't fixed in the case of df-cols because we proxy and restore these at each assignment. This particular bug is tracked in #1217 and will be solved once #1107 is resolved.

We are now tracking memory allocations with human readable sizes in golden tests. This should allow us to detect memory issues early.

This PR also imports the pairlist and attributes utils from the rlang C library. This is because I initially tried to solve the names duplication issue with `r_clone_shared()`. The attribute getters from the GNU R lib are problematic in that regard because they max-bump the namedness/refcount of attributes. I don't think this is appropriate behaviour because in most cases we are only inspecting the attribute. The rlang utils never bump refcounts, and they also never allocate (no need for protection). Even though these rlang utils are no longer necessary to fix the duplication issues since I switched to the `vctrs_owned` approach that Davis introduced, I kept them in the PR because they have better properties.


src/names.c
```diff
@@ -742,19 +755,18 @@ SEXP vec_set_df_rownames(SEXP x, SEXP names) {
...
+SEXP vec_set_names_impl(SEXP x, SEXP names, bool proxy, const enum vctrs_owned owned) {
   check_names(x, names);
 
   if (is_data_frame(x)) {
-    return vec_set_df_rownames(x, names);
+    return vec_set_df_rownames(x, names, proxy, owned);
   }
 
   if (has_dim(x)) {
     return vec_set_rownames(x, names);
```
DavisVaughan, line 766: Does this need to know about `proxy`?

src/slice-assign.c
```diff
@@ -378,25 +379,27 @@ static SEXP vec_assign_fallback(SEXP x, SEXP index, SEXP value) {
...
+  SEXP proxy_nms = PROTECT(vec_proxy_names(proxy));
   if (proxy_nms == R_NilValue) {
     proxy_nms = PROTECT(Rf_allocVector(STRSXP, vec_size(proxy)));
   } else {
-    proxy_nms = PROTECT(r_clone_referenced(proxy_nms));
+    proxy_nms = PROTECT(vec_clone_referenced(proxy_nms, owned));
   }
+  proxy_nms = PROTECT(chr_assign(proxy_nms, index, value_nms, owned));
 
-  proxy_nms = PROTECT(chr_assign(proxy_nms, index, value_nms, VCTRS_OWNED_true));
```
DavisVaughan, line 396: I always found it a bit odd that `VCTRS_OWNED_true` was constant here, so this is a nice change 👍

tests/testthat/test-c.R
```diff
@@ -438,5 +438,47 @@ test_that("concatenation performs expected allocations", {
 
     "Doubles to integer"
     with_memory_prof(vec_unchop(dbls, ptype = int()))
+
+
+    "# Concatenation with with names"
```
DavisVaughan, line 443:
```suggestion
    "# Concatenation with names"
```

src/names.c
lionel-: Probably, now done!

PR #1219, opened 2020-08-13
Allow returning `NULL` names from a `name_spec`

### LINKED ISSUE #1215: Allow returning NULL names from a name-spec
So it is possible to write a name spefication that ignores outer names:

```r
zap_outer_spec <- function(outer, inner) if (is_character(inner)) inner

names(vec_c(a = 1:2, .name_spec = zap_outer_spec))
#> NULL

names(vec_c(a = 1:2, c(b = 3L), .name_spec = zap_outer_spec))
#> [1] ""  ""  "b"
```

Currently it's not possible to use `NULL` names when a name-spec is invoked.

lionel-:
Branched from #1218.
Closes #1215.

This makes it possible to return `NULL` from a `name_spec`. The names vector of the concatenated vector is never allocated or assigned if the name-spec only returns `NULL`. To achieve this, the names vectors are lazily allocated.

This will be useful for implementing `purrr::flatten()` in a backward-compatible way (tidyverse/purrr#785).

The internals of `vec_c()`, `vec_unchop()` with supplied indices, and `vec_rbind()` are now a little more symmetric in style, naming, and structure.

### DIFF
--- a/src/bind.c
+++ b/src/bind.c
@@ -129,8 +129,8 @@ static SEXP vec_rbind(SEXP xs,
   SEXP out = vec_init(proxy, n_rows);
   PROTECT_WITH_INDEX(out, &out_pi);
 
-  SEXP idx = PROTECT_N(compact_seq(0, 0, true), &n_prot);
-  int* idx_ptr = INTEGER(idx);
+  SEXP loc = PROTECT_N(compact_seq(0, 0, true), &n_prot);
+  int* p_loc = INTEGER(loc);
 
   PROTECT_INDEX rownames_pi;
   SEXP rownames = R_NilValue;
@@ -178,10 +178,10 @@ static SEXP vec_rbind(SEXP xs,
     }
     SEXP x = VECTOR_ELT(xs, i);
 
-    init_compact_seq(idx_ptr, counter, size, true);
+    init_compact_seq(p_loc, counter, size, true);
 
     // Total ownership of `out` because it was freshly created with `vec_init()`
-    out = df_assign(out, idx, x, VCTRS_OWNED_true, &bind_assign_opts);
+    out = df_assign(out, loc, x, VCTRS_OWNED_true, &bind_assign_opts);
     REPROTECT(out, out_pi);
 
     // FIXME: This work happens in parallel to the names assignment in
@@ -203,7 +203,7 @@ static SEXP vec_rbind(SEXP xs,
       PROTECT(rn);
 
       if (rownames_type(rn) == ROWNAMES_IDENTIFIERS) {
-        rownames = chr_assign(rownames, idx, rn, VCTRS_OWNED_true);
+        rownames = chr_assign(rownames, loc, rn, VCTRS_OWNED_true);
         REPROTECT(rownames, rownames_pi);
       }
 
--- a/src/bind.c
+++ b/src/bind.c
@@ -77,18 +77,22 @@ static SEXP vec_rbind(SEXP xs,
     Rf_errorcall(R_NilValue, "Can't bind objects that are not coercible to a data frame.");
   }
 
-  SEXP nms = PROTECT_N(r_names(xs), &n_prot);
-  bool has_names = nms != R_NilValue;
+  bool assign_names = !Rf_inherits(name_spec, "rlang_zap");
+
   bool has_names_to = names_to != R_NilValue;
   R_len_t names_to_loc = 0;
 
   if (has_names_to) {
+    if (!assign_names) {
+      r_abort("Can't zap outer names when `.names_to` is supplied.");
+    }
+
     SEXP ptype_nms = PROTECT(r_names(ptype));
     names_to_loc = r_chr_find(ptype_nms, names_to);
     UNPROTECT(1);
 
     if (names_to_loc < 0) {
-      ptype = PROTECT_N(cbind_names_to(has_names, names_to, ptype), &n_prot);
+      ptype = PROTECT_N(cbind_names_to(r_names(xs) != R_NilValue, names_to, ptype), &n_prot);
       names_to_loc = 0;
     }
   }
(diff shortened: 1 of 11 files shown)


src/bind.c
```diff
@@ -77,18 +77,22 @@ static SEXP vec_rbind(SEXP xs,
...
-  SEXP nms = PROTECT_N(r_names(xs), &n_prot);
-  bool has_names = nms != R_NilValue;
+  bool assign_names = !Rf_inherits(name_spec, "rlang_zap");
+
   bool has_names_to = names_to != R_NilValue;
   R_len_t names_to_loc = 0;
 
   if (has_names_to) {
+    if (!assign_names) {
+      r_abort("Can't zap outer names when `.names_to` is supplied.");
```
DavisVaughan, line 87: Is something backwards here? `?vec_rbind` shows that `.names_to` is the argument that accepts `zap()`, not `.name_spec`.

src/c-unchop.c
```diff
@@ -38,20 +38,20 @@ static SEXP vec_unchop_fallback(SEXP ptype,
...
     Rf_errorcall(R_NilValue, "`x` must be a list");
   }
 
   if (indices == R_NilValue) {
-    return vec_c(x, ptype, name_spec, name_repair);
+    return vec_c(xs, ptype, name_spec, name_repair);
   }
 
-  R_len_t x_size = vec_size(x);
+  R_len_t x_size = vec_size(xs);
```
DavisVaughan, line 54: maybe `xs_size`

src/c-unchop.c
```diff
@@ -82,88 +82,74 @@ static SEXP vec_unchop(SEXP x,
...
       continue;
     }
 
-    R_len_t index_size = vec_size(VECTOR_ELT(indices, i));
+    R_len_t index_size = Rf_length(VECTOR_ELT(indices, i));
     out_size += index_size;
 
-    if (has_names) {
-      p_sizes[i] = index_size;
-    }
```
DavisVaughan, line 119: I guess `Rf_length()` is fast enough that there is no need to store them in `sizes`. It certainly makes things less complicated

src/c-unchop.c
```diff
@@ -82,88 +82,74 @@ static SEXP vec_unchop(SEXP x,
...
-    // Each element of `x` is recycled to its corresponding index's size
-    elt = vec_recycle(elt, index_size, args_empty);
-    SET_VECTOR_ELT(x, i, elt);
+    // Each element of `xs` is recycled to its corresponding index's size
+    x = vec_recycle(x, index_size, args_empty);
+    SET_VECTOR_ELT(xs, i, x);
   }
 
-  indices = PROTECT(vec_as_indices(indices, out_size, R_NilValue));
+  SEXP locs = PROTECT(vec_as_indices(indices, out_size, R_NilValue));
```
DavisVaughan, line 109: I have a tough time naming this. We often use `locs` as an integer vector of locations, but this is a list-of-locs and we don't really have a good name for that. That was why I used `indices`

src/c-unchop.c
```diff
@@ -82,88 +82,74 @@ static SEXP vec_unchop(SEXP x,
...
+      SEXP outer = xs_is_named ? STRING_ELT(xs_names, i) : R_NilValue;
+      SEXP inner = PROTECT(vec_names(x));
+      SEXP x_nms = PROTECT(apply_name_spec(name_spec, outer, inner, size));
+
+      if (x_nms != R_NilValue) {
+        R_LAZY_ALLOC(out_names, out_names_pi, STRSXP, out_size);
+        if (x_nms != chrs_empty) {
+          out_names = chr_assign(out_names, loc, x_nms, VCTRS_OWNED_true);
+          REPROTECT(out_names, out_names_pi);
+        }
```
DavisVaughan, line 150: What is this if check for? Can `chr_assign()` not handle this or is it special behavior?

src/names.c
```diff
@@ -537,7 +537,11 @@ SEXP apply_name_spec(SEXP name_spec, SEXP outer, SEXP inner, R_len_t n) {
   }
 
   if (outer == strings_empty || outer == NA_STRING) {
-    return inner;
+    if (inner == R_NilValue) {
+      return chrs_empty;
```
DavisVaughan, line 541: Oh maybe this answers my question from before about special behavior of `chrs_empty`

src/bind.c
DavisVaughan: Oh now I see what `.name_spec` can also be `zap()`. I had to go hunting in the `?vec_c` docs to see that (but `vec_rbind` does link to `vec_c` so i think that is ok)

tests/testthat/test-bind.R
```diff
@@ -930,6 +931,43 @@ test_that("vec_rbind() fallback works with tibbles", {
...
+})
+
+test_that("can't zap names when `.names_to` is supplied", {
+  expect_identical(
+    vec_rbind(foo = c(x = 1), .names_to = zap(), .name_spec = zap()),
+    data.frame(x = 1)
+  )
+  expect_error(
+    vec_rbind(foo = c(x = 1), .names_to = "id", .name_spec = zap()),
+    "Can't zap outer names when `.names_to` is supplied.",
```
DavisVaughan, line 948: I wonder if this is too strict. I almost thing that this should result in an integer `id` column (i.e. all names were zapped, which is equivalent to no names being present to start with).

src/c-unchop.c
lionel-: hmm I think we use `loc` for an integer vector of location. In your vec-prefixes proposals, `locs` refers to a list of loc vectors.

tests/testthat/test-bind.R
lionel-: I'd prefer to restrict things for now. This zapping scheme is complicated. Currently `.name_spec = zap()` ignores all names, including inner names in the recursive case (data frames, df-cols and their columns). This doesn't seem completely right to me. Though this can now be worked around by supplying `~ NULL` instead, which is only applied to outer names.

src/c-unchop.c
lionel-: I'll add a comment that this is to prevent recycling. Probably can be removed once we support ALTREP repetitions in recycling and assignment.

lionel-: Even with altrep we'll be better off with this check to avoid the assignment loop.

PR #1223, opened 2020-08-18
Return data frames from `vec_data()`

lionel-:
Shouldn't affect dplyr as it uses `dplyr_vec_data()` which already preserves data frames.


R/proxy.R
```diff
@@ -207,6 +207,10 @@ vec_data <- function(x) {
   vec_assert(x)
   x <- vec_proxy(x)
 
+  if (is.data.frame(x)) {
+    return(new_data_frame(x, row.names = attr(x, "row.names")))
```
DavisVaughan, line 211: Does this have performance implications? Doesn't this materialize compact row names?

lionel-: oops you're right, `attr()` doesn't return a character vector unlike `row.names()`, but it does materialise. We should probably use `.row_names_info(x, 0)` here.

DavisVaughan: we have used `.row_names_info(data, type = 0L)` many times in dplyr. is it worth making an rlang helper for that?

lionel-: Now done, thanks!

PR #1224, opened 2020-08-18
Use `df_list()` to validate fields of record vectors

lionel-:
Branched from #1223.

- The fields of record vectors are validated with `df_list()`.
- We now use vctrs primitives in the base vector methods.

This makes record vectors more flexible (recycling of fields, allows df-cols and any vctrs types as field), simplifies internals, and will make it easier to support record lists and record data frames.


lionel-:
I reintroduced an equal-size check because the recycling of fields caused a failure in tidymodels/textrecipes.

https://github.com/r-lib/vctrs/commit/c5b8aa9091a8fd5a3726b724c79d0cdc0eeae6a6

Issue #1164, opened 2020-06-23
S4 list columns do not work in tidyr::crossing


kendonB:
``` r
library(terra)
#> This is terra version 0.7.1 (alpha-release)
library(tidyr)
#> 
#> Attaching package: 'tidyr'
#> The following objects are masked from 'package:terra':
#> 
#>     extract, pack
f <- system.file("ex/lux.shp", package="terra")
v <- vect(f)
typeof(v)
#> [1] "S4"
crossing(tibble(v= list(v)),
         tibble(v= list(v)))
#> Error: Unsupported type S4
```

<sup>Created on 2020-06-24 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

kendonB:
related https://github.com/tidyverse/tidyr/issues/735

hadley:
Duplicate of https://github.com/r-lib/vctrs/issues/1164

hadley:
Oops, I just accidentally closed this as duplicate of itself 🤣 . Here a vctrs reprex:

``` r
library(terra)
#> Warning: multiple methods tables found for 'rectify'
#> This is terra version 0.7.4 (alpha-release)
#> 
#> NOTE: You are using GDAL version 2.4.2 
#> For full functionality you need at least version 3.0.4

f <- system.file("ex/lux.shp", package="terra")
v <- vect(f)
vctrs::vec_unique(list(v))
#> Error: Unsupported type S4
```

<sup>Created on 2020-06-29 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

PR #1226, opened 2020-08-19
Implement `vec_chop2()`

lionel-:
Add `vec_chop2()` which is the `[[` equivalent of `vec_chop()`. It is still unexported for now  because we probably don't want to rush the "2" API but it will be useful internally for implementing `vec_map()`.

With atomic vectors it always transforms inner names into outer names:

```r
vec_chop2(c(foo = 1, bar = 2))
#> $foo
#> [1] 1
#>
#> $bar
#> [1] 2

df <- data.frame(x = 1:2, row.names = c("foo", "bar"))
vec_chop2(df)
#> $foo
#>   x
#> 1 1
#>
#> $bar
#>   x
#> 1 2

mat <- matrix(1:4, 2, dimnames = list(c("foo", "bar"), c("x", "y")))
vec_chop2(mat)
#> $foo
#>      x y
#> [1,] 1 3
#>
#> $bar
#>      x y
#> [1,] 2 4
```

Since we now require lists to have list storage, it just unstructures them:

```r
x <- list(x = 1, y = 2)
vec_chop2(new_vctr(x))
#> $x
#> [1] 1
#>
#> $y
#> [1] 2
```

Having this storage requirement really simplifies working with lists. We won't be able to support S4 lists like `S4Vectors::List()` without introducing something like `list_proxy()` and `list_unproxy()` though.

To make the implementation generic, `vctrs_rcrd` now supports assigning `NULL` names. This also fixes a bug as `names<-` would previously be applied to the fields. I still plan to add names support to rcrd vectors in the future but setting names to a character vector is an error for now.

There is a new helper class for unit tests which is a record list that caches the sizes of its elements in attributes.


src/slice-chop.c
```diff
@@ -414,3 +414,41 @@ SEXP vec_as_indices(SEXP indices, R_len_t n, SEXP names) {
...
+    // Lists necessarily have list storage and so don't require any
+    // genericity for extracting elements
+    out = PROTECT(r_clone_referenced(x));
+    SET_ATTRIB(out, R_NilValue);
+    SET_OBJECT(out, 0);
+
+    UNPROTECT(1);
+  } else {
+    // Zap inner names of atomic vectors. They become outer names.
+    out = PROTECT(vec_set_names(x, R_NilValue));
```
DavisVaughan, line 443: I think this can modify x in place if it isn’t referenced? Maybe capture the vec_names earlier to prevent issues if that is the case?

lionel-: oh yeah

lionel-:
Right we can probably wait until there is a need for `indices`.

lionel-:
> Having this storage requirement really simplifies working with lists. We won't be able to support S4 lists like S4Vectors::List() without introducing something like list_proxy() and list_unproxy() though.

Ah there's a feature request about this: #1164.

PR #1228, opened 2020-08-19
Add `vec_slice2()` and `vec_assign2()`

lionel-:
Branched from #1226.

Following `vec_chop2()` and `vec_map()`, this further explores the idea of "extracting" variants of vctrs primitives.

- With atomic vectors, `vec_slice2()` and `vec_assign2()` are equivalent to `vec_slice()` and `vec_assign()`. The only difference is that `vec_slice2()` zaps the names of the result.

- `vec_slice2()` with lists makes use of the list storage requirement and calls `.subset2()`.

- `vec_assign2()` with lists wraps the RHS into a list. Genericity is ensured by `vec_assign()` which coerces the RHS (now a list) to the type of the LHS (a list type). This allows the class of the LHS to initialise attributes, possibly based on the contents of the RHS.

One big difference with `[[<-` is that `NULL` values are not treated as a deletion sentinel, and `list(NULL)` is not unwrapped to escape the sentinel usage.

```r
x <- list(foo = 1, bar = 2)

vec_assign2(x, 2, NULL)
#> $foo
#> [1] 1
#>
#> $bar
#> NULL

vec_assign2(x, 2, list(NULL))
#> $foo
#> [1] 1
#>
#> $bar
#> $bar[[1]]
#> NULL
```

There is currently no way to remove elements:

```r
vec_assign2(x, 2, zap())
#> Error: Can't zap elements.
```

These operations are unexported for now.

### DIFF
--- a/R/slice.R
+++ b/R/slice.R
@@ -263,3 +263,24 @@ vec_slice2 <- function(x, i) {
     }
   )
 }
+
+vec_assign2 <- function(x, i, value, ..., x_arg = "", value_arg = "") {
+  if (!missing(...)) {
+    ellipsis::check_dots_empty()
+  }
+
+  # We may relax this in the future, e.g. for character `i`
+  if (is_zap(value)) {
+    abort("Can't zap elements.")
+  }
+
+  # If `x` is recursive, wrap RHS in a list before calling
+  # `vec_assign()`. The class of `x` must be coercible with lists. We
+  # intentionally wrap `NULL` values instead of treating them as a
+  # sentinel to zap elements.
+  if (vec_is_list(x)) {
+    value <- list(value)
+  }
+
+  vec_assign(x, i, value, x_arg = x_arg, value_arg = value_arg)
+}
--- a/R/slice.R
+++ b/R/slice.R
@@ -249,3 +249,17 @@ vec_slice_seq <- function(x, start, size, increasing = TRUE) {
 vec_slice_rep <- function(x, i, n) {
   .Call(vctrs_slice_rep, x, i, n)
 }
+
+vec_slice2 <- function(x, i) {
+  with_extract(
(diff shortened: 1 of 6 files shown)


R/slice.R
```diff
@@ -249,3 +249,38 @@ vec_slice_seq <- function(x, start, size, increasing = TRUE) {
...
+
+vec_slice2 <- function(x, i) {
+  with_extract(
+    if (vec_is_list(x)) {
+      # Lists are currently guaranteed to have list storage so we can
+      # just subset them directly
+      i <- vec_as_location2(i, vec_size(x))
+      .subset2(x, i)
+    } else {
+      out <- vec_slice(x, i)
```
DavisVaughan, line 261: Use `vec_as_location2()` on `i`?

R/slice.R
```diff
@@ -249,3 +249,38 @@ vec_slice_seq <- function(x, start, size, increasing = TRUE) {
...
+
+  # If `x` is recursive, wrap RHS in a list before calling
+  # `vec_assign()`. The class of `x` must be coercible with lists. We
+  # intentionally wrap `NULL` values instead of treating them as a
+  # sentinel to zap elements.
+  if (vec_is_list(x)) {
+    value <- list(value)
+  }
+
+  vec_assign(x, i, value, x_arg = x_arg, value_arg = value_arg)
```
DavisVaughan, line 285: Use `vec_as_location2()` on `i`?

DavisVaughan:
Closes https://github.com/r-lib/vctrs/issues/141
Closes https://github.com/r-lib/vctrs/pull/626

DavisVaughan:
Should `[[<-.vctrs_vctr` use `vec_assign2()`? I guess it would change the behavior of assigning `NULL` into list-based vctrs

PR #1236, opened 2020-09-01
Fix `obj_maybe_translate_encoding()` crash

### LINKED ISSUE #1233: obj_maybe_translate_encoding causes a crash
Running the following code causes R to crash

```
# no error
x <-  vctrs::new_rcrd(list(foo = 1))
vctrs:::obj_maybe_translate_encoding(x)

# crash
x <-  vctrs::new_rcrd(list(foo = 1:2))
vctrs:::obj_maybe_translate_encoding(x)
```

I was lead to this trying to debug the following error:  https://github.com/tidyverse/tidyr/issues/1004

DavisVaughan:
Closes #1233 

The crash occurred when we had the following scenario:

- A list
- A data frame element in that list
- A rcrd column of that data frame, which has a `vec_size()` that is different from the low level `Rf_length()`

A proxy of the outer list won't proxy the elements of the list, so the rcrd column of the data frame didn't get proxied. When we eventually processed that data frame element, we assumed it had been proxied and passed the `size` of the data frame on to each column as the number of elements to iterate over. Since the rcrd wasn't proxied, the `size` didn't match up with its actual `Rf_length()` and we indexed OOB.

The fix is to treat data frames that are elements of a list as lists themselves. This forces us to compute the length of each list/character column individually with `Rf_length()`, so we correctly compute the unproxied length of the rcrd column. I think this is a more appropriate fix vs a potential alternative of proxying each list element.

I don't really expect this to have any performance impact.

I somehow forgot to add the meaningful part of the fix as its own commit, so it got tied in with the commit "Revert `size` to `n` change" 😞 , sorry. I only just noticed that.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,6 +1,9 @@
 
 # vctrs (development version)
 
+* Fixed an encoding translation bug with lists containing data frames which
+  have columns where `vec_size()` is different from the low level
+  `Rf_length()` (#1233).
 
 # vctrs 0.3.4
 
--- a/src/translate.c
+++ b/src/translate.c
@@ -13,8 +13,8 @@
 // UTF-8 translation is not attempted in these cases:
 // - (utf8 + utf8), (latin1 + latin1), (unknown + unknown), (bytes + bytes)
 
-static bool chr_translation_required_impl(const SEXP* x, R_len_t size, cetype_t reference) {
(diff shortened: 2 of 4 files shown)


DavisVaughan:
Superseded by #1238, which has a better commit structure

Issue #1063, opened 2020-05-01
Use custom double dispatch in `vec_arith()`


lionel-:

DavisVaughan:
First steps in https://github.com/r-lib/vctrs/pull/1241

PR #1241, opened 2020-09-03
Remove `vec_arith()` numeric `y` methods

DavisVaughan:
I'm thinking about how to add custom double dispatch for `vec_arith()`, and I think the first step is to try and clean up existing numeric methods by separating them into integer/double.

We can't touch any methods that are of the form `vec_arith.numeric.*()` yet, because we don't export `vec_arith.integer` or `vec_arith.double` (and I don't think we want to).

However, we can replace `vec_arith.*.numeric()` methods with their explicit integer/double counterparts. That is what this PR does.

It also leaves in some FIXME commented out sections that will be removed in a subsequent PR when custom double dispatch is added.

Tests have been updated to be more explicit about testing both integer and double cases.


lionel-:
Probably best to leave this for 0.4.0.

DavisVaughan:
Yea definitely

DavisVaughan:
This is probably the right first step, but we will rebuild it from scratch when we tackle https://github.com/r-lib/vctrs/issues/1063 for real

PR #1242, opened 2020-09-04
Extract equality comparator from dictionary

DavisVaughan:
I think that the performance upgrades to `dictionary` regarding storing pointers for faster equality comparisons could be generally useful.

This PR extracts the pieces of `dictionary` specific to comparisons into new `comparator` and `comparator_vec` objects. The `comparator` holds the comparison functions, and the `comparator_vec` holds the object and pointer used for the comparison.

I think this generally makes `dictionary.c` a bit more focused as well, and there are no performance implications.

One use case of the comparator on its own is to implement POCs for functions like `vec_identify_runs()`. For example, below is all the code required to implement this in a generic way. That said, it can be done more efficiently using macros, especially for the atomic vector cases - so for this specific function the final implementation probably wouldn't use it.

```c
static SEXP vec_identify_runs(SEXP x) {
  int nprot = 0;

  R_len_t size = vec_size(x);

  SEXP out = PROTECT_N(Rf_allocVector(INTSXP, size), &nprot);
  int* p_out = INTEGER(out);

  x = PROTECT_N(vec_proxy_equal(x), &nprot);
  x = PROTECT_N(obj_maybe_translate_encoding(x, size), &nprot);

  R_len_t id = 0;

  struct comparator* p_comparator = new_comparator(x);
  PROTECT_COMPARATOR(p_comparator, &nprot);

  struct comparator_vec* p_comparator_vec = new_comparator_vec(x);
  PROTECT_COMPARATOR_VEC(p_comparator_vec, &nprot);
  const void* vec_p = p_comparator_vec->vec_p;

  for (R_len_t i = 0; i < size; ++i) {
    if (!p_comparator->equal(vec_p, i, vec_p, id)) {
      id = i;
    }

    p_out[i] = id + 1;
  }

  UNPROTECT(nprot);
  return out;
}
```


src/comparator.h
```diff
@@ -0,0 +1,32 @@
...
+
+struct comparator {
+  int (*equal)(const void* x, R_len_t i, const void* y, R_len_t j);
+  int (*equal_missing)(const void* x, R_len_t i);
+  SEXP self;
+};
+
+struct comparator_vec {
+  SEXP vec;
+  const void* vec_p;
```
lionel-, line 14: `p_vec`?

lionel- (commented):
`struct comparator` doesn't seem exactly the right term given that "compare" has a different meaning in vctrs.

In principle this could take any binary predicate taking scalar elements of vector arrays. Maybe `struct oo_binary_op`?

And then the ctors could be `new_oo_equal()`, `init_oo_equal_dbl()`, etc?

DavisVaughan:
@lionel- I've refactored this on top of the "poly" ideas we discussed over slack.

There are now two poly function types:

- A binary `poly_binary_op_fn_t` that returns an `int`
- A unary `poly_unary_op_fn_t` that returns an `int`

These should be fairly generic, but we might need to change the return type one day.

A proxy-specific binary equal function can be retrieved with `new_poly_op_equal_scalar_na_equal_p(proxy)`. A unary equal missing function is retrieved with `new_poly_op_equal_missing_p(proxy)`. We can add more of these as needed much easier now.

These op functions take void pointers as before, which are the `->p_vec` members of a poly vec, created by `new_poly_vec(proxy)`.

The dictionary struct now holds a `struct poly_vec*` corresponding to `x`, and function pointers to the proxy-specific `equal()` and `equal_missing()` functions.

DavisVaughan:
In light of https://github.com/r-lib/vctrs/issues/1276, `new_poly_p_is_missing()` should really be reworked `new_poly_p_is_complete()` and `p_df_is_missing()` might should work off completeness instead. For now I've kept the names and implementations consistent with what is currently there

src/poly-op.h
```diff
@@ -0,0 +1,27 @@
+#ifndef VCTRS_POLY_OP
+#define VCTRS_POLY_OP
+
+#include "vctrs.h"
+
+typedef int (*poly_binary_int_fn_t)(const void* x, r_ssize i, const void* y, r_ssize j);
```
lionel-, line 6: Should we use the `_fn_ptr` suffix instead of `_fn_t`? This is what I use in rlang. I find it makes it slightly clearer what the type is about.

lionel- (approved):

Issue #1210, opened 2020-08-06
`vec_locate_runs()` and `vec_partition()`


lionel-:
`vec_locate_runs()` finds the locations of the last values of consecutive runs in a vector:

```r
vec_detect_runs <- function(x) {
  n <- vec_size(x)
  if (!n) {
    return(int())
  }

  diffs <- vec_slice(x, -1L) != vec_slice(x, -n)
  c(diffs, TRUE)
}
vec_locate_runs <- function(x) {
  which(vec_detect_runs(x))
}

x <- sample(1:2, 10, TRUE)
x
#>  [1] 2 2 1 1 1 1 2 1 1 1

vec_locate_runs(x)
#> [1]  2  6  7 10

vec_locate_runs(rep(1, 10))
#> [1] 10
vec_locate_runs(c(0, rep(1, 9)))
#> [1]  1 10
vec_locate_runs(c(rep(1, 9), 0))
#> [1]  9 10
vec_locate_runs(1:2)
#> [1] 1 2
```

`vec_partition()` partitions a vector into a list based on pivotal values:

```r
seq0 <- function(from, to) {
  if (length(from) != 1) {
    abort("`from` must be length one")
  }
  if (length(to) != 1) {
    abort("`to` must be length one")
  }

  if (from > to) {
    abort("Can't create decreasing sequence")
  }

  seq.int(from, to)
}

vec_partition <- function(x, runs) {
  runs <- vec_cast(runs, integer())

  runs_size <- length(runs)
  if (!runs_size) {
    return(list(x))
  }

  x_size <- vec_size(x)
  runs <- sort(runs)

  last_run <- runs[[runs_size]]
  if (last_run > x_size) {
    abort("`runs` can't exceed the vector size.")
  }
  if (last_run != x_size) {
    # Ensure a full partition of `x`
    runs <- c(runs, x_size)
  }

  out <- vec_init(list(), runs_size)
  first <- 1L

  for (i in seq_along(runs)) {
    last <- runs[[i]]
    out[[i]] <- vec_slice(x, seq0(first, last))
    first <- last + 1L
  }

  out
}

vec_partition(1:10, c(1, 10))
#> [[1]]
#> [1] 1
#>
#> [[2]]
#> [1]  2  3  4  5  6  7  8  9 10

vec_partition(1:10, c(1, 9))
#> [[1]]
#> [1] 1
#>
#> [[2]]
#> [1] 2 3 4 5 6 7 8 9
#>
#> [[3]]
[...]

PR #1248, opened 2020-09-08
Implement `vec_locate_runs()` and `vec_detect_runs()`

DavisVaughan:
Follow up to #1243 
Part of https://github.com/r-lib/vctrs/issues/1210

I've given these functions a `start = TRUE` option to detect either the start / end of each run (it isn't completely straightforward to get from starts to ends, so the option is useful). I've defaulted to `TRUE` since I think that would play nicely with `vec_partition(x, at = vec_locate_runs(x))`, which I think could take starting values, but that is up for debate still and might affect the default here.


lionel-:
It's probably best to unexport these until we have a clearer view of how these functions will be used? Now that we have `vec_identify_runs()` these might not be as useful. Having them unexported will still allow us to experiment with them as we are solving concrete problems.

Issue #1245, opened 2020-09-08
Consider always translating in `vec_proxy_equal()`


DavisVaughan:
As noted here https://github.com/r-lib/vctrs/pull/1243#discussion_r484732568, we seem to always translate before calling `vec_proxy_equal()`.

Like in the dictionary functions:
https://github.com/r-lib/vctrs/blob/7f5134f3b0ae747407321c57451f2eee60722ce3/src/dictionary.c#L374-L375

One reason that we might _not_ do this is that `vec_proxy_equal()` can always be called with a single vector, but sometimes you need to make sure that encodings between two vectors are consistent. That requires using `obj_maybe_translate_encoding2()` as done in `vec_match()`
https://github.com/r-lib/vctrs/blob/7f5134f3b0ae747407321c57451f2eee60722ce3/src/dictionary.c#L537-L545

An example of this is when `x` is all Latin1, but `y` is UTF-8. In `obj_maybe_translate_encoding()`, we allow both `x` and `y` to pass through unchanged, but `obj_maybe_translate_encoding2()` would translate both to UTF-8.

We currently let an all Latin1 vector through a) for performance reasons b) this is what base R does. However, the new `vec_unique()` will always convert to UTF-8, and it would certainly simplify things if we always converted to UTF-8 everywhere. I think this would remove the need for `obj_maybe_translate_encoding2()`, and would generally simplify the translation code greatly. It is definitely worth considering it.

DavisVaughan:
Counterpoint - there are functions where we want the equality proxy, but the underlying encoding of the strings doesn't matter. For example, `vec_equal_na()`

DavisVaughan:
I have determined that keeping them separate is probably best. I'd rather see an explicit call to `vec_normalize_encoding()` than be surprised about the fact that it translated when I didn't want it to

PR #1249, opened 2020-09-08
Simplify translation helpers to always convert to ASCII or UTF-8

### LINKED ISSUE #1246: Consider making `obj_maybe_translate_encoding()` always translate to UTF-8
Would likely make #1245 more straightforward

We currently let a vector that is consistently in a non-UTF-8 encoding (like an all Latin1 vector) through `obj_maybe_translate_encoding()` without translation. We do this:

a) For performance reasons with those vectors
b) This is what base R does

However, the new `vec_order()` will always convert to UTF-8, and it would certainly simplify things if we always converted to UTF-8 everywhere.

I don't think it would change the actual behavior of any functions that use this translation helper, since you'd still end up with a consistent encoding that you can compare in.

I _think_ this would remove the need for obj_maybe_translate_encoding2(), because you could always expect that the result of `obj_maybe_translate_encoding()` would be in UTF-8, making it comparable to any other UTF-8 vector.

It would also greatly simplify the translation code, especially for lists, which already have weird translation rules (it does a first pass through the list to check if any element needs translation, then does a second pass to translate all elements if any elements required translation).

---

Also consider renaming it to `proxy_translate_utf8()`

DavisVaughan:
Closes #1246 

New `proxy_normalize_encoding()` to replace `obj_maybe_translate_encoding()` and `obj_maybe_translate_encoding2()`. It has been rewritten from scratch, so the diff for `translate.c` isn't really meaningful.

We now translate to UTF-8 unconditionally (the exact rule is to translate to UTF-8 if the CHARSXP is both not marked as ASCII and not marked as UTF-8). There are a few reasons for this:

- This greatly simplifies the translation code, and is much easier to explain and maintain.

- The result of `proxy_normalize_encoding()` will work _within_ a single vector (`vec_unique()`) and _between_ multiple vectors (`vec_match()`). We no longer need a version that takes 1 vector and a version that takes 2.

- This would allow us to always translate in `vec_proxy_equal()` and friends, i.e. #1245 

- I am hopeful that this would mean that we can simplify the `compare.c` and `equal.h` code related to strings, since we could assume that all string input was UTF-8. This would make string comparisons a simple SEXP pointer comparison, which would be very fast.

- The new `vec_order()` always compares in a UTF-8 encoding, so this makes other vctrs functions consistent with that. In fact, calling `proxy_normalize_encoding()` up front might allow us to simplify the internal of `vec_order()` as well, since we can assume UTF-8 already.

- Packages like readr and readxl always return results with UTF-8 encoding, which should make translation from other encodings less likely.

---

Generally, the behavior of functions that use this translation helper hasn't changed. Previously they always tried to convert to some common encoding, and they still do. The way we figure out that common encoding is just different now.

This does have the following side effects:

- Character vectors that only have "bytes" encodings are no longer supported in any function using these translation features. This should be incredibly rare though. They are no longer supported because `Rf_translateCharUTF8()` can't translate bytes->UTF8.

- Character vectors that are marked completely with Latin-1 encodings will be translated to UTF-8 now.  It is worth noting that base R will also translate these to UTF-8. Previously we tried to avoid translating these. This was nice for performance, but was complicated for two reasons:

   1) When comparing within a single list, if any element of the list required translation, we'd have to go back and translate the whole list to a common encoding. 

   2) When comparing between two vectors, we had to reconcile their two individual encodings with each other. For example, if `x` was all UTF8 and `y` was all Latin1, we'd have to make sure that we translate these to the same encoding, which is why we needed `obj_maybe_translate_encoding2()`.

  Always translating to UTF8 avoids these issues.

### DIFF
--- a/R/translate.R
+++ b/R/translate.R
@@ -1,3 +1,3 @@
-proxy_normalize_encoding <- function(x) {
-  .Call(vctrs_proxy_normalize_encoding, x)
+vec_normalize_encoding <- function(x) {
+  .Call(vctrs_normalize_encoding, x)
 }
--- a/R/translate.R
+++ b/R/translate.R
@@ -1,3 +1,3 @@
-vec_normalize_encoding <- function(x) {
-  .Call(vctrs_normalize_encoding, x)
+proxy_normalize_encoding <- function(x) {
+  .Call(vctrs_proxy_normalize_encoding, x)
 }
--- a/R/translate.R
+++ b/R/translate.R
@@ -1,7 +1,3 @@
-obj_maybe_translate_encoding <- function(x) {
-  .Call(vctrs_maybe_translate_encoding, x)
-}
-
-obj_maybe_translate_encoding2 <- function(x, y) {
-  .Call(vctrs_maybe_translate_encoding2, x, y)
+vec_normalize_encoding <- function(x) {
+  .Call(vctrs_normalize_encoding, x)
 }
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,6 +1,10 @@
 
 # vctrs (development version)
 
+* Functions that make comparisons within a single vector, such as
+  `vec_unique()`, or between two vectors, such as `vec_match()`, now
+  convert all character input to UTF-8 before making comparisons (#1246).
+
 * New `df_detect_complete()` which returns a logical vector that detects
   if rows of `x` are complete, i.e. they have no missing values.
 
--- a/src/dictionary.c
+++ b/src/dictionary.c
@@ -372,7 +372,7 @@ SEXP vctrs_unique_loc(SEXP x) {
   R_len_t n = vec_size(x);
 
   x = PROTECT_N(vec_proxy_equal(x), &nprot);
-  x = PROTECT_N(obj_maybe_translate_encoding(x, n), &nprot);
+  x = PROTECT_N(proxy_normalize_encoding(x), &nprot);
 
   struct dictionary* d = new_dictionary(x);
   PROTECT_DICT(d, &nprot);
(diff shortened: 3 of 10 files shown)


R/translate.R
```diff
@@ -1,7 +1,3 @@
-obj_maybe_translate_encoding <- function(x) {
-  .Call(vctrs_maybe_translate_encoding, x)
-}
-
-obj_maybe_translate_encoding2 <- function(x, y) {
-  .Call(vctrs_maybe_translate_encoding2, x, y)
+proxy_normalize_encoding <- function(x) {
```
lionel-, line 1: The `vec_proxy_` prefix would be more consistent?

lionel- (commented):
I think this generally makes sense.

> Packages like readr and readxl always return results with UTF-8 encoding, which should make translation from other encodings less likely.

Another common way in which vectors in native or latin encodings arise is through user code like `rep("café", n)`. Do we want to always translate in `vec_slice()` too? I guess the impact on performance would be too important. I wish we had a way to mark encoding on character vectors.

R/translate.R
DavisVaughan: Hmm, would it? This function has to be called on a proxy. `vec_proxy_equal()`, for example, is called on a vector but returns a proxy.

DavisVaughan: I should make the input argument here at the R level `proxy` rather than `x`, like I did at the C level

lionel-: > Hmm, would it?

I think so. A proxy is still a vector. `vec_proxy_assign()` takes a proxy. `vec_proxy_typeof()` returns the type of a proxy.

DavisVaughan: Hmm, on second thought I'm no longer convinced that this _has_ to be called on a proxy anymore. It previously had to be because of how it handled data frames, but now we treat those like lists. It should really be able to be called on any vector without issue.

I think that I can call this `vec_normalize_encoding(x)`, with a note that it is generally called on a proxy.

That said, even though this function would "work" on any vector, sometimes we have to call it on the proxy to get "correct" behavior. Consider a list-rcrd with a vectorized character attribute. `vec_normalize_encoding()` doesn't translate attributes, so you'd have to call `vec_proxy_equal()` before `vec_normalize_encoding()` to promote the attribute to a data frame column first, so it gets translated. I think I can just add a note about this.

Issue #1247, opened 2020-09-08
Use `VECTOR_PTR_RO()` in `list_equal_scalar_na_equal()`


DavisVaughan:
For performance, as discussed in https://github.com/r-lib/vctrs/pull/1243#discussion_r484744067

PR #1250, opened 2020-09-09
Implement "complete" function family

DavisVaughan:
Closes #899 (supersedes it)

Implements three new functions related to complete values:

- `vec_detect_complete()`
- `vec_locate_complete()`
- `vec_slice_complete()`

It might seem like `vec_slice_complete()` is overkill, but I think that it is one of the most common things that you'll do with this. For example, `tidyr::drop_na()` would use it. It is similar to how `vec_unique()` is a thin wrapper around `vec_slice()` and `vec_unique_loc()`, and I think `vec_unique()` could become `vec_slice_unique()` (I've added it to the google sheet).

List handling will be improved when #1247 is tackled.

---

Regarding performance, much faster than `complete.cases()`!

`complete.cases()` also doesn't handle data frames with POSIXlt columns or df-cols very well, as it expects each data frame column to have the same `Rf_length()`. We don't have this problem here.

``` r
library(vctrs)

set.seed(123)

make_df <- function(expr) {
  expr <- rlang::enexprs(expr = expr)
  call <- rlang::call2("replicate", n = 100, simplify = FALSE, !!!expr)
  cols <- rlang::eval_bare(call)
  names(cols) <- paste0("a", 1:length(cols))
  new_data_frame(cols)
}

x <- rep(1L, 1e6)
y <- sample(c(1, NA, 3), size = 1e6, replace = TRUE)

bench::mark(vec_detect_complete(x), complete.cases(x), iterations = 200)
#> # A tibble: 2 x 6
#>   expression                  min   median `itr/sec` mem_alloc `gc/sec`
#>   <bch:expr>             <bch:tm> <bch:tm>     <dbl> <bch:byt>    <dbl>
#> 1 vec_detect_complete(x) 358.69µs 594.88µs     1159.    3.82MB    143. 
#> 2 complete.cases(x)        3.79ms   4.26ms      226.    3.82MB     28.0

bench::mark(vec_detect_complete(y), complete.cases(y), iterations = 200)
#> # A tibble: 2 x 6
#>   expression                  min   median `itr/sec` mem_alloc `gc/sec`
#>   <bch:expr>             <bch:tm> <bch:tm>     <dbl> <bch:byt>    <dbl>
#> 1 vec_detect_complete(y)    707µs 848.52µs     1132.    3.81MB    147. 
#> 2 complete.cases(y)         7.2ms   7.88ms      122.    3.81MB     15.1

# Data frame with no missing values
df1 <- make_df(rep(1, 1e6))

bench::mark(vec_detect_complete(df1), complete.cases(df1))
#> # A tibble: 2 x 6
#> 1 vec_detect_complete(df1)   81.8ms     83ms     11.9     3.83MB        0
#> 2 complete.cases(df1)       266.4ms    279ms      3.58    3.81MB        0

# Data frame with missing values near the front
df2 <- vec_cbind(data_frame(y = y), df1)

bench::mark(vec_detect_complete(df2), complete.cases(df2))
#> # A tibble: 2 x 6
#> 1 vec_detect_complete(df2)   79.9ms   82.7ms     12.1     3.82MB     0   
#> 2 complete.cases(df2)       267.3ms  267.3ms      3.74    3.81MB     3.74

# Data frame with missing values all throughout
df3 <- make_df(sample(c(1, NA, NA, NA), size = 1e6, replace = TRUE))

bench::mark(vec_detect_complete(df3), complete.cases(df3))
#> # A tibble: 2 x 6
#> 1 vec_detect_complete(df3)    261ms    262ms      3.82    3.82MB        0
#> 2 complete.cases(df3)         594ms    594ms      1.68    3.81MB        0
```

<sup>Created on 2020-09-09 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0.9001)</sup>


R/complete.R
```diff
@@ -0,0 +1,67 @@
...
+#' vec_detect_complete(x)
+#' !vec_equal_na(x)
+#'
+#' y <- c("a", "b", NA, "d", "e")
+#'
+#' df <- data_frame(x = x, y = y)
+#'
+#' # For data frames, this returns `TRUE` where all elements
+#' # of the row are non-missing. Compare that with `!vec_equal_na()`,
+#' # which detect where any elements of the row are non-missing.
```
lionel-, line 46: > detects rows that have at least one non-missing value.

src/complete.c
```diff
@@ -0,0 +1,258 @@
...
+  case vctrs_type_logical: lgl_detect_complete(proxy, size, p_out); break;
+  case vctrs_type_integer: int_detect_complete(proxy, size, p_out); break;
+  case vctrs_type_double: dbl_detect_complete(proxy, size, p_out); break;
+  case vctrs_type_complex: cpl_detect_complete(proxy, size, p_out); break;
+  case vctrs_type_character: chr_detect_complete(proxy, size, p_out); break;
+  case vctrs_type_raw: raw_detect_complete(proxy, size, p_out); break;
+  case vctrs_type_list: list_detect_complete(proxy, size, p_out); break;
+  case vctrs_type_dataframe: df_detect_complete(proxy, size, p_out); break;
+  case vctrs_type_scalar: Rf_errorcall(R_NilValue, "Can't detect missing values in scalars with `vec_detect_complete()`.");
+  default: Rf_errorcall(R_NilValue, "Unimplemented type in `vec_detect_complete()`.");
```
lionel-, line 77: `stop_unimplemented_vctrs_type()`

src/complete.c
```diff
@@ -0,0 +1,258 @@
...
+  switch (vec_proxy_typeof(proxy)) {
+  case vctrs_type_logical: lgl_detect_complete(proxy, size, p_out); break;
+  case vctrs_type_integer: int_detect_complete(proxy, size, p_out); break;
+  case vctrs_type_double: dbl_detect_complete(proxy, size, p_out); break;
+  case vctrs_type_complex: cpl_detect_complete(proxy, size, p_out); break;
+  case vctrs_type_character: chr_detect_complete(proxy, size, p_out); break;
+  case vctrs_type_raw: raw_detect_complete(proxy, size, p_out); break;
+  case vctrs_type_list: list_detect_complete(proxy, size, p_out); break;
+  case vctrs_type_dataframe: df_detect_complete(proxy, size, p_out); break;
+  case vctrs_type_scalar: Rf_errorcall(R_NilValue, "Can't detect missing values in scalars with `vec_detect_complete()`.");
```
lionel-, line 76: `stop_internal()`

src/complete.c
```diff
@@ -0,0 +1,258 @@
...
+  for (R_len_t i = 0; i < size; ++i) {
+    p_out[i] = 1;
+  }
+
+  int n_col = Rf_length(x);
+  const SEXP* p_x = VECTOR_PTR_RO(x);
+
+  for (R_len_t i = 0; i < n_col; ++i) {
+    const SEXP col = p_x[i];
+    vec_detect_complete_col(col, size, p_out);
```
lionel-, line 149: A bit strange to declare a const SEXP here.

```suggestion
    vec_detect_complete_col(p_x[i], size, p_out);
```

src/equal.h
```diff
@@ -4,6 +4,9 @@
 #include "vctrs.h"
 
 
+static inline bool lgl_scalar_equal_missing(int x) {
```
lionel-, line 7: Follow the surrounding `_equal_scalar_` naming scheme?

lionel-: Did you add these variants for performance or for readability? I would assume not performance since `NA_LOGICAL` etc are not compile time constants anyway.

src/utils.c
```diff
@@ -981,17 +981,27 @@ R_len_t r_lgl_sum(SEXP x, bool na_true) {
   }
 
   R_len_t n = Rf_length(x);
```
lionel-, line 983: We could take this opportunity to switch to long vector sizes.

src/complete.c
```diff
@@ -0,0 +1,258 @@
...
+static void df_detect_complete(SEXP x, R_len_t size, int* p_out);
+
+// [[ register() ]]
+SEXP vctrs_detect_complete(SEXP x) {
+  return vec_detect_complete(x);
+}
+
+static
+SEXP vec_detect_complete(SEXP x) {
+  SEXP proxy = PROTECT(vec_proxy_equal(x));
```
lionel-, line 61: We currently define missing values in record vectors to be the proxy rows where all values are missing. In the complete cases though, I suspect we should treat record vectors the same way as atomic vectors, i.e. an element is complete unless it's missing. I.e. if there are some missing values in a row, it's still a complete element.

src/complete.c
```diff
@@ -0,0 +1,258 @@
...
+  case vctrs_type_raw: raw_detect_complete_col(x, size, p_out); break;
+  case vctrs_type_list: list_detect_complete_col(x, size, p_out); break;
+  case vctrs_type_dataframe: stop_internal("vec_detect_complete_col", "Data frame columns should have been handled already.");
+  case vctrs_type_scalar: Rf_errorcall(R_NilValue, "Can't detect missing values in scalars with `vec_detect_complete()`.");
+  default: Rf_errorcall(R_NilValue, "Unimplemented type in `vec_detect_complete()`.");
+  }
+}
+
+
+/*
```
lionel-, line 178: 👍

lionel- (approved):

DavisVaughan:
@lionel- to be conservative we could just export `vec_detect_complete()` for now and leave `vec_slice_complete()` and `vec_locate_complete()` as internal functions.

I take back what I said about `drop_na()` using `vec_slice_complete()`, it would really use `vec_detect_complete()` on a subset of columns and then call `vec_slice()` on the whole data frame. Which means that I feel less strongly that it should be exported.

src/equal.h
DavisVaughan: Mainly for readability, but also:
- I think the double one is more correct this way actually, something like this doesn't cover the `NaN` case? It should also be faster https://github.com/r-lib/vctrs/blob/05968ce8e669f73213e3e894b5f4424af4f46316/src/dictionary.c#L54-L56
- the character variant is faster because it doesn't try to translate
- the list variant is faster because we don't have to use `equal_object()`, and we don't have to pass in and subset `vctrs_shared_na_list`
- it is nice to have a raw variant that just returns false

src/complete.c
DavisVaughan: This is tricky because we've worked hard to not special case vctrs-rcrd in the C code. To be completely general we might need `vec_proxy_complete()`, which would default to `vec_proxy_equal()` but we could define:

```r
vec_proxy_complete.vctrs_rcrd <- function(x) {
  proxy <- vector(length = vec_size(x))
  incomplete <- vec_equal_na(x)
  proxy[incomplete] <- NA
  proxy
}
```

lionel-: oh definitely not a new proxy. I was thinking `vec_detect_complete()` to special case vectors that inherit from `data.frame`. For other kinds of vectors it'd just use the equal-na code path and flip the boolean vector.

It's also possible that this function should be `df_detect_complete()`.

DavisVaughan: > It's also possible that this function should be df_detect_complete().

That seems....much simpler. I like it!

DavisVaughan: Note: Consider what happens with a rcrd column in this `df_detect_complete()` function.

DavisVaughan:
Ok @lionel- I've changed this to `df_detect_complete()` and friends. I think that makes more sense, as for other types it is the same as `!vec_equal_na()`.

A few notes:

- The data frame itself is not proxied, as this would recursively proxy each column in a way that would make it hard to track which columns came from a rcrd column's df proxy. Instead, the data frame is flattened to remove df-cols, and then each column is proxied individually. At this point it mostly works the same as before, except data frame proxies get their own path here.

- The data frame proxy column path ensures that only data frame rows that are completely missing are marked as "incomplete". It uses the `df_short_circuit_info` struct, but uses the same trick that you gave a 👍  to earlier to avoid too many branches. Really the `df_short_circuit_info` struct is used here just for access to a boolean array.

- I've given the `df_short_circuit_info` struct the ability to be initialized lazily. This allows us to reuse the memory across multiple columns that have data frame proxies, but only create it when it is needed.

lionel- (approved):

PR #1256, opened 2020-09-14
Early exit from `df_flatten()` when no flattening is required

### LINKED ISSUE #1253: `df_flatten()` early exit
If `width == Rf_length(x)`, isn't there nothing to do? No need to allocate and do all those `VECTOR_ELT()` and `SET_VECTOR_ELT()` calls?

https://github.com/r-lib/vctrs/blob/ebcf82240d65c1b915859b8d72b1dad1dcd2cd99/src/type-data-frame.c#L858-L869

DavisVaughan:
Closes #1253 

Small PR to early exit from `df_flatten()` when there is nothing to do.

Test needed adjustment because it used to always allocate a new vector and row names were not being copied over. Now it can just return its input unmodified so the row names are maintained

### DIFF
--- a/src/type-data-frame.c
+++ b/src/type-data-frame.c
@@ -856,7 +856,13 @@ static R_len_t df_flatten_loop(SEXP x, SEXP out, SEXP out_names, R_len_t counter
 //
 // [[ register(); include("type-data-frame.h") ]]
 SEXP df_flatten(SEXP x) {
+  R_len_t n_cols = Rf_length(x);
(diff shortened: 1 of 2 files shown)


lionel- (approved):
Did you check that callers of `df_flatten()` do not assume a fresh list?

DavisVaughan:
- Called by `df_proxy()`, which calls `r_clone_referenced()` earlier on
- Called by `df_detect_complete()`, which doesn't modify the result after calling this

PR #1260, opened 2020-09-16
Translate attributes in `vec_normalize_encoding()`

### LINKED ISSUE #1258: Spurious test failure with dictionary functions, lists, and translatable attributes
This translation test fails on a very rare basis

https://github.com/r-lib/vctrs/blob/b338e79f80428e82312a695e72c63e42a93fa11b/tests/testthat/test-translate.R#L117

First seen on a github actions CI build

I confirmed that this fails even on the current master branch, which means that it is completely unrelated to changes in vec-order (which was my first thought, since it messes with the global string pool)

The only way I can force it to fail so far is to run the following and wait

```r
for (i in 1:100) {
  devtools::test()
}
```

I've inserted the following browser statement into the problematic test, but even when I'm in there I can't figure out what is happening

```r
  if (!identical(vec_unique(x), x)) {
    browser()
  }
```

My guess is that it has to do with the fact that attributes aren't translated by `vec_normalize_encoding()`. In `vec_unique()` we compute hash values using the attributes as part of the hash, but we don't translate, so we probably get different hashes. But then in `dict_hash_with()` we use `d->equal()` which would compare the two values with translation, so they'd look equal there. It probably has something to do with that?

DavisVaughan:
Closes #1258 

After much head-banging, I think I've finally gotten the recursive pattern correct in such a way that we can elegantly translate attributes recursively. It required rewriting it again, but I think it is overall even more elegant than before.

This is a bit slower than what we used to do (not translate attributes), but it really is only noticeable if you have deeply nested lists. The most common place this could arise is with multiple list columns, as is often done in tidymodels.

That said, translating the attributes here means that we can rip out the character translation parts of `equal.h` and `compare.c`, which I'm hoping will recover a bit of the lost performance.

Overall, I think it is worth it to put the translation code in one place, so we don't have to worry about it in other vctrs functions.

Here are a few examples, side by side with `vec_unique()`, which internally calls `vec_normalize_encoding()`, so you can get a feel for its impact.

Seems to have no impact on a sampled version of `flights`, which I consider to be fairly realistic.

``` r
library(nycflights13)
library(tidymodels)
library(vctrs)

vec_normalize_encoding <- vctrs:::vec_normalize_encoding

# sample the first 5 rows ~400k times
# 19 cols, 1 has attributes
flights <- vec_slice(flights, sample(1:5, nrow(flights), replace = TRUE))

# A fairly realistic dataset
bench::mark(
  vec_unique(flights),
  vec_normalize_encoding(flights), 
  iterations = 100,
  check = FALSE
)

# before
#> # A tibble: 2 x 6
#>   expression                           min   median `itr/sec` mem_alloc `gc/sec`
#>   <bch:expr>                      <bch:tm> <bch:tm>     <dbl> <bch:byt>    <dbl>
#> 1 vec_unique(flights)               38.7ms  41.33ms      24.0    3.31MB     1.53
#> 2 vec_normalize_encoding(flights)    2.2ms   2.39ms     416.         0B     0

# after
#> # A tibble: 2 x 6
#>   expression                           min   median `itr/sec` mem_alloc `gc/sec`
#>   <bch:expr>                      <bch:tm> <bch:tm>     <dbl> <bch:byt>    <dbl>
#> 1 vec_unique(flights)              36.03ms  38.84ms      25.8    3.31MB     1.65
#> 2 vec_normalize_encoding(flights)   1.88ms   2.17ms     472.         0B     0
```

Does have some impact on a deeply nested tibble that occurs naturally in tidymodels, but it isn't too bad.

```r
resamples <- bootstraps(mtcars, times = 100)

spline_rec <- recipe(mpg ~ ., data = mtcars) %>%
  step_ns(disp) %>%
  step_ns(wt)

lin_mod <- linear_reg() %>%
  set_engine("lm")

spline_res <- fit_resamples(lin_mod, spline_rec, resamples)

# 100 rows, some list columns that have tibbles in them
spline_res
#> # Resampling results
#> # Bootstrap sampling 
#> # A tibble: 100 x 4
#>    splits          id           .metrics         .notes          
#>    <list>          <chr>        <list>           <list>          
#>  1 <split [32/13]> Bootstrap001 <tibble [2 × 3]> <tibble [0 × 1]>
#>  2 <split [32/9]>  Bootstrap002 <tibble [2 × 3]> <tibble [0 × 1]>
#>  3 <split [32/13]> Bootstrap003 <tibble [2 × 3]> <tibble [0 × 1]>
#>  4 <split [32/13]> Bootstrap004 <tibble [2 × 3]> <tibble [0 × 1]>
#>  5 <split [32/10]> Bootstrap005 <tibble [2 × 3]> <tibble [0 × 1]>
#>  6 <split [32/10]> Bootstrap006 <tibble [2 × 3]> <tibble [0 × 1]>
#>  7 <split [32/11]> Bootstrap007 <tibble [2 × 3]> <tibble [0 × 1]>
#>  8 <split [32/11]> Bootstrap008 <tibble [2 × 3]> <tibble [0 × 1]>
#>  9 <split [32/12]> Bootstrap009 <tibble [2 × 3]> <tibble [0 × 1]>
#> 10 <split [32/9]>  Bootstrap010 <tibble [2 × 3]> <tibble [0 × 1]>
#> # … with 90 more rows

bench::mark(
  vec_unique(spline_res),
  vec_normalize_encoding(spline_res), 
  iterations = 100,
  check = FALSE
)

# before
#> # A tibble: 2 x 6
#>   expression                             min  median `itr/sec` mem_alloc
#>   <bch:expr>                         <bch:t> <bch:t>     <dbl> <bch:byt>
#> 1 vec_unique(spline_res)             777.3µs 852.6µs     1133.     103KB
#> 2 vec_normalize_encoding(spline_res)  11.7µs  11.8µs    76472.        0B

# after
#> # A tibble: 2 x 6
#>   expression                             min  median `itr/sec` mem_alloc
#>   <bch:expr>                         <bch:t> <bch:t>     <dbl> <bch:byt>
#> 1 vec_unique(spline_res)             823.8µs 886.9µs     1116.     103KB
#> 2 vec_normalize_encoding(spline_res)  63.1µs  66.2µs    14950.        0B
```

### DIFF
--- a/src/translate.c
+++ b/src/translate.c
@@ -129,8 +129,9 @@ SEXP list_normalize_encoding(SEXP x) {
     }
     PROTECT(elt_new);
 
+    // Only cloned once, at which point `x` is free of references
     if (MAYBE_REFERENCED(x)) {
-      x = PROTECT(r_clone_referenced(x));
+      x = PROTECT(r_clone(x));
       ++nprot;
       p_x = VECTOR_PTR_RO(x);
     }
@@ -182,8 +183,9 @@ SEXP attrib_normalize_encoding(SEXP x) {
     }
     PROTECT(elt_new);
 
+    // Only cloned once, at which point `x` is free of references
     if (MAYBE_REFERENCED(x)) {
-      x = PROTECT(r_clone_referenced(x));
+      x = PROTECT(r_clone(x));
       ++nprot;
       node = x;
 
--- a/src/translate.c
+++ b/src/translate.c
@@ -172,6 +172,7 @@ static
 SEXP attrib_normalize_encoding(SEXP x) {
   int nprot = 0;
   r_ssize loc = 0;
+  bool owned = false;
 
   for (SEXP node = x; node != r_null; node = r_node_cdr(node), ++loc) {
     SEXP elt_old = r_node_car(node);
@@ -182,9 +183,12 @@ SEXP attrib_normalize_encoding(SEXP x) {
     }
     PROTECT(elt_new);
 
-    // Cloned once, at which point `x` and all `node`s are free of references
-    if (MAYBE_REFERENCED(x)) {
+    if (!owned) {
+      // Shallow clone entire pairlist if not owned.
+      // Should be fast because these are generally short.
       x = PROTECT_N(r_clone(x), &nprot);
+      owned = true;
+
       node = x;
 
       // Restore original positioning post-clone
--- a/src/translate.c
+++ b/src/translate.c
@@ -183,7 +183,7 @@ SEXP attrib_normalize_encoding(SEXP x) {
     PROTECT(elt_new);
 
     // Cloned once, at which point `x` and all `node`s are free of references
-    if (MAYBE_REFERENCED(x)) {
+    if (MAYBE_REFERENCED(node)) {
       x = PROTECT_N(r_clone(x), &nprot);
       node = x;
 
--- a/src/translate.c
+++ b/src/translate.c
@@ -183,7 +183,7 @@ SEXP attrib_normalize_encoding(SEXP x) {
     PROTECT(elt_new);
 
     // Cloned once, at which point `x` and all `node`s are free of references
-    if (MAYBE_REFERENCED(node)) {
+    if (MAYBE_REFERENCED(x)) {
       x = PROTECT_N(r_clone(x), &nprot);
       node = x;
 
--- a/src/translate.c
+++ b/src/translate.c
@@ -131,8 +131,7 @@ SEXP list_normalize_encoding(SEXP x) {
 
     // Cloned once, at which point `x` is free of references
     if (MAYBE_REFERENCED(x)) {
-      x = PROTECT(r_clone(x));
-      ++nprot;
+      x = PROTECT_N(r_clone(x), &nprot);
       p_x = VECTOR_PTR_RO(x);
     }
 
@@ -185,8 +184,7 @@ SEXP attrib_normalize_encoding(SEXP x) {
 
     // Cloned once, at which point `x` and all `node`s are free of references
     if (MAYBE_REFERENCED(x)) {
-      x = PROTECT(r_clone(x));
-      ++nprot;
+      x = PROTECT_N(r_clone(x), &nprot);
       node = x;
 
       // Restore original positioning post-clone
--- a/src/translate.c
+++ b/src/translate.c
@@ -129,7 +129,7 @@ SEXP list_normalize_encoding(SEXP x) {
     }
     PROTECT(elt_new);
 
-    // Only cloned once, at which point `x` is free of references
+    // Cloned once, at which point `x` is free of references
     if (MAYBE_REFERENCED(x)) {
       x = PROTECT(r_clone(x));
       ++nprot;
@@ -183,7 +183,7 @@ SEXP attrib_normalize_encoding(SEXP x) {
     }
     PROTECT(elt_new);
 
-    // Only cloned once, at which point `x` is free of references
+    // Cloned once, at which point `x` and all `node`s are free of references
     if (MAYBE_REFERENCED(x)) {
       x = PROTECT(r_clone(x));
       ++nprot;
--- a/src/translate.c
+++ b/src/translate.c
@@ -1,28 +1,24 @@
 #include "translate.h"
+#include "vctrs.h"
+#include "utils.h"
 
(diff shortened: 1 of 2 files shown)


DavisVaughan:
One more example that is a bit simpler, and really shows that there is a performance impact here with lists that we might have to just accept.

```r
library(vctrs)

vec_normalize_encoding <- vctrs:::vec_normalize_encoding

x <- rep_len(list(1), 1e6)

bench::mark(
  vec_unique(x),
  vec_normalize_encoding(x),
  check = FALSE,
  iterations = 50
)

#before 
#> # A tibble: 2 x 6
#> 1 vec_unique(x)                 23ms  24.49ms      39.9    11.8MB     18.8
#> 2 vec_normalize_encoding(x)    1.6ms   1.73ms     548.         0B      0

# after
#> # A tibble: 2 x 6
#> 1 vec_unique(x)              29.38ms  31.13ms      30.9    11.8MB     14.6
#> 2 vec_normalize_encoding(x)   7.44ms   8.23ms     121.         0B      0
```

src/translate.c
```diff
@@ -110,87 +113,90 @@ static r_ssize chr_find_normalize_start(SEXP x, r_ssize size) {
...
+    SEXP elt_old = r_node_car(node);
+
+    SEXP elt_new = obj_normalize_encoding(elt_old);
+    if (elt_old == elt_new) {
+      continue;
+    }
+    PROTECT(elt_new);
+
+    if (MAYBE_REFERENCED(x)) {
+      x = PROTECT(r_clone_referenced(x));
```
lionel-, line 186: The referenced check is not correct here since tails of pairlists can be reused by fresh nodes.

We only need to duplicate the pairlist nodes up to `loc` and we could reuse the tail of the pairlist. Then recurse back into `attrib_normalise_encoding()` from the `loc`th node (or equivalently update the state and `continue;`, that's probably better). That would avoid cloning and traversing the attributes multiple times when multiple attributes need normalisation.

src/translate.c
```diff
@@ -110,87 +113,90 @@ static r_ssize chr_find_normalize_start(SEXP x, r_ssize size) {
...
+    PROTECT(elt_new);
 
-      elt = list_normalize_encoding(elt, elt_size, elt_start);
-      SET_VECTOR_ELT(x, i, elt);
-      break;
-    }
-    default:
-      continue;
+    if (MAYBE_REFERENCED(x)) {
+      x = PROTECT(r_clone_referenced(x));
```
lionel-, line 133:
```suggestion
      x = PROTECT(r_clone(x));
```

src/translate.c
```diff
@@ -110,87 +113,90 @@ static r_ssize chr_find_normalize_start(SEXP x, r_ssize size) {
...
-      }
+    PROTECT(elt_new);
 
-      elt = list_normalize_encoding(elt, elt_size, elt_start);
-      SET_VECTOR_ELT(x, i, elt);
-      break;
-    }
-    default:
-      continue;
+    if (MAYBE_REFERENCED(x)) {
```
lionel-, line 132: Mention that this can only be cloned once, at which point `x` is free of references.

lionel- (commented):
> Overall, I think it is worth it to put the translation code in one place, so we don't have to worry about it in other vctrs functions.

I think that makes sense!

DavisVaughan:
@lionel- I _think_ I have this correct, but could use a second set of eyes.

This tries to be as efficient as possible by copying the pairlist up to the element that needs translation, and then it uses the tail of the rest of the pairlist.

If a second element need translation, then it only copies nodes after the node of the first element that needed translation. So it reuses a "head" and "tail" and only makes a copy of the "middle".

DavisVaughan:
I'm not sure if this protect is correct?
https://github.com/r-lib/vctrs/blob/f71c8e0b412bf344ada754b4bfa5fdf03a86cab8/src/translate.c#L221

I made an educated guess based on the fact that `val` gets protected here
https://github.com/wch/r-source/blob/bac4cd3013ead1379e20127d056ee036278b47ff/src/main/duplicate.c#L266

lionel-:
I think this implementation is a bit too complicated. Ideally we'd have some sort of `node_clone_until()` function that clones the top of a pairlist. It would return the new head and the new tail in an output parameter.

In this case though, since attribute pairlists tend to be short, we could just use shallow duplication of the whole pairlist with a boolean flag to keep track of whether we own the pairlist or not.

PR #1262, opened 2020-09-16
Implement `vec_rle()`

### LINKED ISSUE #1081: Implement `vec_identify_runs()` and `vec_rle()`
Inspired by the adjacency grouping idea in https://github.com/tidyverse/dplyr/issues/5184

We could do much better than this in terms of performance with a C implementation. `vec_group_rle()` does a _lot_ more work than is necessary here because it uses a dictionary to keep track of what it has already seen. 

I've added this to the vec-prefixes google sheet

``` r
library(vctrs)

vec_runs <- function(x) {
  rle <- vec_group_rle(x)
  lengths <- field(rle, "length")
  rep(seq_along(lengths), times = lengths)
}

mtcars <- as_tibble(mtcars)

mtcars %>%
  select(vs, am) %>%
  mutate(runs = vec_runs(across()))
#> # A tibble: 32 x 3
#>       vs    am  runs
#>    <dbl> <dbl> <int>
#>  1     0     1     1
#>  2     0     1     1
#>  3     1     1     2
#>  4     1     0     3
#>  5     0     0     4
#>  6     1     0     5
#>  7     0     0     6
#>  8     1     0     7
#>  9     1     0     7
#> 10     1     0     7
#> # … with 22 more rows
```

<sup>Created on 2020-05-05 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

DavisVaughan:
Closes #1081 

This is internal for now, but feels pretty useful.

Unlike `vec_group_rle()`, which returns a custom rle object, this just returns a two column data frame. I'm not sure why the other one doesn't return a data frame, it feels like the natural thing here.

```r
library(vctrs)

vec_rle <- vctrs:::vec_rle

x <- vec_rle(c("b", "b", "a", "a", "d"))
x
#>   key size
#> 1   b    2
#> 2   a    2
#> 3   d    1
```

Inverse rle doesn't need its own function, its just `vec_rep_each()`

```r
# Inverse rle
vec_rep_each(x$key, x$size)
#> [1] "b" "b" "a" "a" "d"
```

I also realized that this can be used to implement `vec_group_rle()` in a much simpler way (really we could deprecate that function, I think). In an ideal world I'd rename `vec_group_id() -> vec_identify_duplicates()`, and then we'd have:

```r
library(nycflights13)
vec_identify_duplicates <- vec_group_id

# 10 rows of flights, sampled
slice <- vec_slice(flights[4:6], sample(1:10, nrow(flights), replace = TRUE))
head(slice, 2)
#> # A tibble: 2 x 3
#>   dep_time sched_dep_time dep_delay
#>      <int>          <int>     <dbl>
#> 1      558            600        -2
#> 2      544            545        -1

# Restructure return value of `vec_group_rle()`
vec_group_rle2 <- function(x) {
  out <- vec_group_rle(x)
  
  key <- structure(field(out, "group"), n = attr(out, "n"))
  size <- field(out, "length")
  
  data_frame(key = key, size = size)
}

bench::mark(
  vec_group_rle2(slice),
  vec_rle(vec_identify_duplicates(slice))
)
#> # A tibble: 2 x 6
#>   expression                                  min median `itr/sec` mem_alloc
#>   <bch:expr>                              <bch:t> <bch:>     <dbl> <bch:byt>
#> 1 vec_group_rle2(slice)                   10.13ms 12.5ms      77.1   10.16MB
#> 2 vec_rle(vec_identify_duplicates(slice))  8.61ms 10.4ms      97.6    9.25MB
```


lionel- (approved):
Maybe should be called `vec_unrep()`? That would be more telling for those who don't know what RLE is. I also find it makes it easier to reason about composed operations like`vec_unrep(vec_identify_duplicates(x))`

PR #1266, opened 2020-09-18
Refactor vec-order structs to allow "self protection"

DavisVaughan:
This PR is an intermediate step that allows the internal parts of `vec_order()` to be called from other files. This allows additional information like the group sizes to be returned as well. It comes back in a new "meta struct" called `order_info`. 

Other vctrs functions can call `new_order_info()` to create an `order_info` struct, and then call `vec_order_info()` to compute the order and populate the `order_info` struct.

`order_info` also holds everything required to call `vec_order_info()` a second time very efficiently, which is extremely useful for reordering the results of the first call by first appearance (i.e. something that is required for `vec_unique()` and friends).

The main change to the existing structs is that they can now "self protect" themselves so they can be returned from functions. This follows the same model as `struct dictionary`. The trickiest one was `group_infos`, because now the array of 2 `group_info` pointers can't be allocated on the stack anymore.

I've also tweaked where `vec_order_expand_args()` is called from, and exposed it, `parse_direction()`, and `parse_na_value()` in `order-radix.h` for eventual use by `vec_rank()`

As a POC, I have a separate branch from this that has reimplementations of a number of vctrs functions that were dictionary based. Most of them are more efficient, with the exception of `vec_duplicate_any()`, which was more efficient with the dictionary https://github.com/DavisVaughan/vctrs/commit/d73f35f3e00917683e552f2004956cdaf18aaa23

These POCs also don't suffer from the issue where the dictionary gets full, as was seen in https://github.com/tidyverse/dplyr/issues/5291 from `vec_group_loc()` erroring (`vec_group_loc2()` works on the 1e9 strings in about 19 seconds)

### DIFF
--- a/src/lazy.h
+++ b/src/lazy.h
@@ -6,6 +6,7 @@
 // -----------------------------------------------------------------------------
 
 /*
+ * @member self A RAWSXP for the struct memory.
  * @member data The RAWSXP that gets allocated lazily.
  * @member p_data A void pointer to the RAWSXP.
  * @member data_pi A protection index to `data` so it can reprotect itself
@@ -15,6 +16,7 @@
  *   is from `sizeof(<type>)`.
  */
 struct lazy_raw {
+  SEXP self;
   SEXP data;
   void* p_data;
   PROTECT_INDEX data_pi;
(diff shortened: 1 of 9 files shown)


DavisVaughan:
Another thing of note in the POCs. I've added `sort = FALSE/TRUE` options to functions like `vec_unique()`. The results come back sorted from calling `vec_unique()`, so its often easier to return the sorted unique values if that is what the user wants (this probably won't be the case with character vectors after I add `chr_order_appearance()`).

I think Hadley had mentioned in the past that `vec_order_locs()` could really be `vec_group_locs(x, sort = TRUE)`, so it seems like that would be a natural argument to add to a few of these functions (This would require solving the C locale sort order issue though)

lionel- (commented):
We should probably move to the approach now used in rlang, which is to include a `sexp* shelter` as first element of heap-allocated structs.

I wonder if rlang's dynamic vectors could be used instead of these lazy structs?

DavisVaughan:
I think we could just rename `self` to `shelter`? This PR will be broken into smaller pieces when I get back around to it (likely for 0.4.0).

> rlang's dynamic vectors

I _think_ they could be the core data structure here. We might still need a little struct wrapper around them that holds the dynamic vector alongside the eventual `size` to resize to if the vector is required.

PR #1267, opened 2020-09-22
Assign names before skipping empty input in `vec_c()`

DavisVaughan:
Closes https://github.com/r-lib/vctrs/issues/1263

This also has the side effect of avoiding the `vec_c()` issue noted in https://github.com/r-lib/vctrs/issues/623#issuecomment-696946634, since names are now extracted before the cast. `vec_unchop()` still exhibits this issue, since it uses `vec_cast_common()` before the loop. The real solution is to fix `vec_cast()`, I think.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,6 +1,9 @@
 
 # vctrs (development version)
 
+* `vec_c()` now correctly returns a named result with named empty inputs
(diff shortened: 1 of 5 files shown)


lionel- (approved):
Nice!

PR #1269, opened 2020-09-23
Unify equality helper names

### LINKED ISSUE #1252: `vec_match(na_equal = FALSE)` and `NaN`
I'm fairly certain this is wrong

```r
> vec_match(c(NaN, NA_real_), c(NaN, NA_real_), na_equal = FALSE)
[1]  1 NA
```

Because of:
https://github.com/r-lib/vctrs/blob/05968ce8e669f73213e3e894b5f4424af4f46316/src/dictionary.c#L54-L56

Noted in:
https://github.com/r-lib/vctrs/pull/1250#discussion_r487093321

If the equal-missing functions switch to use new ones added in https://github.com/r-lib/vctrs/pull/1250 then it should fix itself

DavisVaughan:
Closes #1252 

This PR refactors everything in `equal.h` to have a unified naming scheme. All of the utilities now live here, which are then used by `dictionary.c` and `equal.c`.

This is the last step required before continuing work on the poly-op PR started in https://github.com/r-lib/vctrs/pull/1242

We now have the following set of equality utilities (replace lgl with any atomic type):

- Missing values:

  - `lgl_equal_missing_scalar(x)`
  - `p_lgl_equal_missing_scalar(p_x, i)`
  - `p_equal_missing_scalar(p_x, i, type)`

- Equality treating NAs as equal

  - `lgl_equal_scalar_na_equal(x, y)`
  - `p_lgl_equal_scalar_na_equal(p_x, i, p_y, j)`
  - `p_equal_scalar_na_equal(p_x, i, p_y, j, type)`

- Equality propagating NAs

  - `lgl_equal_scalar_na_propagate(x, y)`
  - `p_lgl_equal_scalar_na_propagate(p_x, i, p_y, j)`
  - `p_equal_scalar_na_propagate(p_x, i, p_y, j, type)`

The pointers `p_x` and `p_y` are `const void*` so the function can be stored in the dictionary struct

We also have `lgl_equal_scalar(x, y, na_equal)` and its variants. These are only used by `vec_equal()` in an inefficient way, and will be removed in a future PR in favor of pushing the `na_equal` handling outside the loop.

I've removed the `equal_scalar()` function. It wasn't used anywhere, and I'm fairly certain we have better ways of handling generic equality. We can use the poly-op ideas instead, and in the case of data frames we can also iterate column-wise rather than row-wise, so we don't need this helper.

It is probably easiest to look at `equal.h` in its entirety, rather than the diff.


lionel-:
Is the `_scalar_` particle useful? It looks like we could remove it without ambiguity.

DavisVaughan:
I think it is useful. We have `lgl_hash()` for a logical vector and `lgl_hash_scalar()` for a scalar. I think that is a useful distinction.

I was planning on cleaning up the `vec_equal_na()` internals by replacing the macro with type specific functions that work on vectors, like: `lgl_equal_missing()`. So we need the scalar version of `lgl_equal_missing_scalar()` to prevent name clashes. https://github.com/r-lib/vctrs/blob/f2f02a4dccb8233f80938104eb3e52597b0f0440/src/equal.c#L447

DavisVaughan:
After speaking with @lionel-, we have further revised to:

- Missing values:

  - `lgl_is_missing(x)`
  - `p_lgl_is_missing(p_x, i)`
  - `p_is_missing(p_x, i, type)`

- Equality treating NAs as equal

  - `lgl_equal_na_equal(x, y)`
  - `p_lgl_equal_na_equal(p_x, i, p_y, j)`
  - `p_equal_na_equal(p_x, i, p_y, j, type)`

- Equality propagating NAs

  - `lgl_equal_na_propagate(x, y)`
  - `p_lgl_equal_na_propagate(p_x, i, p_y, j)`
  - `p_equal_na_propagate(p_x, i, p_y, j, type)`

PR #1274, opened 2020-10-06
Add missing references contents

DavisVaughan:
@lionel- do these categories make sense to you? The broken pkgdown build was annoying me 😛


lionel-:
It seems strange to document the user-facing detect-complete with low level generic operations. Maybe we need a new "Data frame" section? Then the new constructors could be documented there.

PR #1275, opened 2020-10-06
Rewrite `vec_equal()`

DavisVaughan:
This PR rewrites `vec_equal()`. This is done for a few reasons:

- It is now a little faster, maybe 10-15% or so, by branching off `na_equal` outside the inner equality loop

- By doing the above bullet point, we can now remove the scalar `*_equal()` helpers that were left in `equal.h`. They weren't used anywhere else, and it is generally better to use the `*_equal_na_equal()` and `*_equal_na_propagate()` instead

- I used this chance to give vec-equal a major refresh to update to the current coding styles we are using

``` r
library(vctrs)

set.seed(123)

x <- sample(1:10, 1e8, replace = TRUE)
y <- sample(1:10, 1e8, replace = TRUE)

bench::mark(vec_equal(x, y), iterations = 50)

# before
#> # A tibble: 1 x 6
#> 1 vec_equal(x, y)   90.2ms    101ms      8.88     382MB     9.62

# after
#> # A tibble: 1 x 6
#> 1 vec_equal(x, y)   81.6ms   85.9ms      10.4     382MB     11.2
```

```r
df <- data.frame(x = x, y = y)
df2 <- data.frame(x = y, y = x)

bench::mark(vec_equal(df, df2), iterations = 20)

# before
#> # A tibble: 1 x 6
#> 1 vec_equal(df, df2)    458ms    500ms      1.92     477MB     2.35

# after
#> # A tibble: 1 x 6
#> 1 vec_equal(df, df2)    419ms    441ms      2.25     477MB     2.76
```

<sup>Created on 2020-10-06 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0.9001)</sup>

### DIFF
--- a/src/equal.c
+++ b/src/equal.c
@@ -27,27 +27,27 @@ static SEXP df_equal(SEXP x, SEXP y, R_len_t size, bool na_equal);
  */
 static
 SEXP vec_equal(SEXP x, SEXP y, bool na_equal) {
-  x = PROTECT(vec_proxy_equal(x));
-  y = PROTECT(vec_proxy_equal(y));
+  SEXP x_proxy = PROTECT(vec_proxy_equal(x));
+  SEXP y_proxy = PROTECT(vec_proxy_equal(y));
 
-  R_len_t size = vec_size(x);
+  R_len_t size = vec_size(x_proxy);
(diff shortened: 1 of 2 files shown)


lionel- (approved):
Nice!

PR #1277, opened 2020-10-07
Implement internal `vec_proxy_complete()`

### LINKED ISSUE #1276: Reconsider a `vec_proxy_complete()` proxy
In #1250 I mentioned that a `vec_proxy_complete()` proxy would ease implementation of `df_detect_complete()`. The current implementation of `df_detect_complete()` is rather complicated because we can't use `vec_proxy_equal()`. We can't use it because we want to special case classes like rcrds that have a df equality proxy to ensure that they are only considered "incomplete" if ALL values in a row are missing. This is at odds with what `df_detect_complete()` normally does with data frames, which is to consider them incomplete if ANY values in a row are missing. This keeps us from using `vec_proxy_equal()` because that flattens df-cols, so we lose the distinction between columns that came from a rcrd, and columns that came from a normal data frame.

https://github.com/r-lib/vctrs/pull/1250#discussion_r487110134

As it turns out, `df_detect_complete()` isn't the only place where we use the idea of "competeness". It is also used in this `p_df_is_missing()` function that is used by `vec_match(na_equal = FALSE)` and `vec_in(na_equal = FALSE)`.

https://github.com/r-lib/vctrs/blob/0263587907dcf2236d715cf6746a7e816caea244/src/dictionary.c#L102-L116

This function would be more properly called `p_df_is_incomplete()`, as it early exits at the first missing value. This gives us this `vec_match()` behavior, where we return `NA` if any `needle` row value is missing, which I _think_ we want?

``` r
library(vctrs)

df <- data_frame(x = 1, y = c(2, NA))
df
#>   x  y
#> 1 1  2
#> 2 1 NA

vec_match(df, df, na_equal = FALSE)
#> [1]  1 NA
```

However, it suffers from the same problem that `df_detect_complete()` had to work around with rcrd types. If a rcrd col has any `NA` value in a particular row, that entire row is considered incomplete, which we don't want:

```r
library(vctrs)

# row 1 is missing, row 2 is not
rcrd <- new_rcrd(list(a = c(NA, 1), b = c(NA, NA)))

df <- data_frame(x = 1:2, y = rcrd)

# flattens df-cols, but now we can't tell that cols `a` and `b` must be jointly
# missing to be considered incomplete
vec_proxy_equal(df)
#>   x  a  b
#> 1 1 NA NA
#> 2 2  1 NA

# says that both rows are missing, which isn't right
vec_match(df, df, na_equal = FALSE)
#> [1] NA NA

# alternative proxy that defaults to `vec_proxy_equal()`,
# but allows customization for things like rcrds
vec_proxy_complete.vctrs_rcrd <- function(x) {
  proxy <- vector(length = vec_size(x))
  incomplete <- vec_equal_na(x)
  proxy[incomplete] <- NA
  proxy
}

df2 <- df
df2$y <- vec_proxy_complete.vctrs_rcrd(df2$y)

vec_match(df2, df2, na_equal = FALSE)
#> [1] NA  2
```

One of two things could solve this:

1. It is possible `vec_match()` should work off missingness rather than completeness (as it currently attempts to do). In this case, we just need to change `p_df_is_missing()` to only return `true` if the entire row is missing

2. A `vec_proxy_complete()` proxy could be used, which would let us flatten data frames and still have correct results since we can write a custom method for vctrs_rcrd. This would also simplify `df_detect_complete()` greatly

DavisVaughan:
Closes #1276 

This PR adds an internal `vec_proxy_complete()`. It is identical to `vec_proxy_equal()`, except in the case where the following two conditions are met:

1. `x` is not a data frame

2. `vec_proxy_equal(x)` is a data frame

In these cases, we follow a heuristic where each element of `x` is only considered incomplete if the entire row of the proxy is missing. This seems more appropriate for rcrd classes.

Arrays are special cased as an exception here. It technically meets this criteria, but the data frame proxy is really a stop gap and we treat it as if `x` itself was a data frame.

---

I went ahead and used it in `df_detect_complete()`. It is a little tricky because we have to check for proxy-unwrapping, but it simplified the code a lot.

We might consider just making this `vec_detect_complete()`. All the infrastructure is already there now.

---

I haven't added support for this into `vec_match()` yet. That can be a separate PR


lionel- (approved):
> We might consider just making this vec_detect_complete(). All the infrastructure is already there now.

One advantage is that it would also work with matrices right?

DavisVaughan:
Oh yea, that's true. We'd end up with this, which is nice

```r
> x <- matrix(c(NA, NA, NA, 1, 2, 3), nrow = 3, byrow = T)
> x
     [,1] [,2]
[1,]   NA   NA
[2,]   NA    1
[3,]    2    3
> df_detect_complete(vec_proxy_complete(x))
[1] FALSE FALSE  TRUE
```

PR #1280, opened 2020-10-08
Avoid `PROTECT()` / `UNPROTECT()` calls when no attributes are present

DavisVaughan:
Slight but meaningful improvement to `vec_normalize_encoding()` by avoiding `PROTECT()` / `UNPROTECT()` calls when we don't have any attributes. This mainly helps with lists, where we have to iterate over each element. Normally I wouldn't try to avoid protect/unprotect calls like this, but I'd like `vec_normalize_encoding()` to be as lightweight as possible when there aren't any character vectors to worry about, and this actually helped a good bit

``` r
x <- rep_len(list(1), 1e6)

vec_normalize_encoding <- vctrs:::vec_normalize_encoding

bench::mark(vec_normalize_encoding(x), iterations = 2000)

# Before
#> # A tibble: 1 x 6
#> 1 vec_normalize_encoding(x)   7.31ms    8.1ms      123.        0B        0

# After
#> # A tibble: 1 x 6
#> 1 vec_normalize_encoding(x)   4.83ms   5.28ms      189.        0B        0
```

<sup>Created on 2020-10-08 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0.9001)</sup>

### DIFF
--- a/src/translate.c
+++ b/src/translate.c
@@ -45,19 +45,24 @@ SEXP vec_normalize_encoding(SEXP x) {
 
 static SEXP chr_normalize_encoding(SEXP x);
 static SEXP list_normalize_encoding(SEXP x);
-static SEXP obj_attrib_normalize_encoding(SEXP x);
+static SEXP obj_attrib_normalize_encoding(SEXP x, SEXP attrib);
 
 static inline
 SEXP obj_normalize_encoding(SEXP x) {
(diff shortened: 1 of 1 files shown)


lionel- (approved):
I was surprised at first, but if we're optimising at the milliseconds scale for millions of operations, it makes sense that every nanosecond counts. As long as they don't make the code too complicated or take us too much time I think these optimisations are good.

PR #1282, opened 2020-10-16
Ensure that size is preserved when cbind-ing zero column data frames

### LINKED ISSUE #1281: vec_cbind() drops rows with zero column data frames
I expect 1 row, 0 cols:

``` r
vctrs::vec_cbind(vctrs::new_data_frame(n = 1L), vctrs::new_data_frame(n = 1L))
```

<sup>Created on 2020-10-16 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0.9001)</sup>

Comes up in tune where i make pretty heavy usage of zero col data frames

DavisVaughan:
Closes #1281 

Fairly straightforward fix. Just need to pass the size through to vec-restore

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,6 +1,9 @@
 
 # vctrs (development version)
 
+* `vec_cbind()` with only empty data frames now preserves the common size of
+  the inputs in the result (#1281).
+
 * `vec_c()` now correctly returns a named result with named empty inputs
   (#1263).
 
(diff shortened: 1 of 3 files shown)


lionel-:
Can we pass a data frame instead? Now that data frames are a primary type in vctrs, I would like to remove the `n` argument of `vec_restore()`. We introduced it only to deal with restoration from list to data frame. This might be a good place to start making this change.

PR #1284, opened 2020-10-20
Implement `vec_fill()`

DavisVaughan:
This PR implements a generic version of `tidyr::fill()`. It uses `vec_equal_na()` to detect missingness, then uses that to construct a vector of locations to slice with to mimic "filling" in the missing values

I haven't exported it yet, but could if that seems useful

Sidenote: I think it would be fairly useful for the C level version of `vec_equal_na()` to also return the number of missing values, possibly as an attribute, or maybe as an output parameter? If we detect 0 missing values here from the call to `vec_equal_na()`, then we can just return the input.


DavisVaughan:
Performs about as well as you might expect when compared with the low level helper in tidyr. Remember that tidyr's version isn't generic, so rather than building up a vector of locations and slicing it just inserts elements of `x` into the output object directly as it goes along. All things considered, I think this is pretty good

The downup/updown strategies used here are more efficient than the ones in tidyr

``` r
vec_fill <- vctrs:::vec_fill

tidyr_fill <- function(x, direction = c("down", "up", "downup", "updown")) {
  tidyr:::vec_fill_fun(direction)(x)
}

# int
x <- sample(c(1L, NA, 3L), 1e6, TRUE) + 0L

bench::mark(
  vctrs = vec_fill(x, direction = "down"),
  tidyr = tidyr_fill(x, direction = "down")
)
#> # A tibble: 2 x 6
#> 1 vctrs        2.65ms   3.16ms      259.   11.44MB     229.
#> 2 tidyr      555.82µs 808.15µs     1059.    8.63MB     129.

bench::mark(
  vctrs = vec_fill(x, direction = "downup"),
  tidyr = tidyr_fill(x, direction = "downup")
)
#> # A tibble: 2 x 6
#> 1 vctrs        2.18ms   2.89ms      354.   11.44MB     354.
#> 2 tidyr         1.3ms   1.79ms      579.    7.63MB     219.

# dbl
x <- sample(c(1L, NA, 3L), 1e6, TRUE) + 0

bench::mark(
  vctrs = vec_fill(x, direction = "down"),
  tidyr = tidyr_fill(x, direction = "down")
)
#> # A tibble: 2 x 6
#> 1 vctrs        2.69ms    3.6ms      250.   15.26MB    154. 
#> 2 tidyr        4.37ms   5.37ms      187.    7.63MB     26.3

bench::mark(
  vctrs = vec_fill(x, direction = "downup"),
  tidyr = tidyr_fill(x, direction = "downup")
)
#> # A tibble: 2 x 6
#> 1 vctrs        2.97ms    3.4ms      286.    15.3MB    157. 
#> 2 tidyr        8.41ms   8.93ms      110.    15.3MB     56.8
```

<sup>Created on 2020-10-20 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0.9001)</sup>

src/equal.c
```diff
@@ -432,8 +437,8 @@ bool equal_names(SEXP x, SEXP y) {
 
 static SEXP df_equal_na(SEXP x, R_len_t size);
 
-// [[ register() ]]
-SEXP vctrs_equal_na(SEXP x) {
+// [[ include("vctrs.h") ]]
```
lionel-, line 440: Are we now avoiding the main header?

```suggestion
// [[ include("equal.h") ]]
```

lionel- (approved):
Cool!

My only concern is with the name of the function which doesn't seem specific enough. Also, at the C level, I would expect a fill function to assign a specific value as in `std::fill()` from the STL.

But it does have the advantage of matching `tidyr::fill()` which is here to stay.

lionel-:
Would the UI be clearer with `direction = c("up", "down"), backtrack = FALSE`? Or some other better named boolean? I find `updown` and `downup` not as clear.

DavisVaughan:
`leading` - Should leading `NA`s be backfilled with the first encountered non-missing value in the chosen `direction`?

```r
vec_fill(x, direction = c("down", "up"), leading = FALSE)
```

I was also thinking about adding `limit = NULL` which could be set to a positive integer value. This would limit the number of missing values that could be filled in a row. The default would assume no limit. This is offered in other places like by `zoo::na.locf()` and in Pandas and was recently requested in tidyr https://github.com/tidyverse/tidyr/issues/1053

lionel-:
`leading` seems to make sense. Regarding `limit`, I think `max_gap` as in the zoo package would be clearer. cc @hadley

DavisVaughan:
Regarding the name, we could use something like `vec_patch_sliding()`. In dplyr there is now `rows_patch()` for overwriting missing values. I could see us having some `vec_patch(x, with)` function in the future that replaces `NA` values with `with`

R/fill.R
```diff
@@ -0,0 +1,22 @@
+#' Fill in missing values with the previous or following value
+#'
+#' `vec_fill()` fills gaps of missing values with the previous or following
+#' non-missing value.
+#'
+#' @param x A vector
+#' @param direction Direction in which to fill missing values. Must be either
+#'   `"down"` or `"up"`.
+#' @param leading A logical specifying if leading missing values should be
```
hadley, line 9: I find `downup`/`updown` easier to understand, and it has the advantage of matching the existing `tidyr::fill()` syntax which some people will be familiar with.

R/fill.R
```diff
@@ -0,0 +1,22 @@
...
+#'   back-filled with the first non-missing value found in the chosen
+#'   `direction`.
+#' @param max_gap A single positive integer specifying the maximum size of the
```
hadley, line 12: I don't love "gap" here because it somehow feels the wrong way around — like that's the maximum size gap that would be created. Maybe `max_fill` or just `max_n`?

R/fill.R
```diff
@@ -0,0 +1,22 @@
...
+#'   gap of sequential missing values that will be filled. If `NULL`, there is
+#'   no limit to the gap size.
+#'
+#' @keywords internal
```
hadley, line 16: Needs some examples, including a data frame.

hadley: And maybe doesn't need to be internal but should be marked as experimental?

hadley (commented):
I'm fine with `vec_fill()` as a name. I think if you wanted something more specific, `vec_fill_na()` would be reasonable.

PR #1285, opened 2020-10-22
Export `vec_unrep()`

DavisVaughan:
@lionel- do you like the name `vec_compress()` any more than `vec_unrep()`? It occurred to me that "compress" is pretty expressive of what this function does, but "unrep" fits the function family a bit more.


R/rep.R
```diff
@@ -5,10 +5,22 @@
...
+#' These functions work along the size of `x`, rather than its length. For data
+#' frames, this means that rows are repeated rather than columns.
+#'
+#' Using `vec_unrep()` and `vec_rep_each()` together is very similar to using
+#' [base::rle()] and [base::inverse.rle()]. The following invariant shows
+#' the relationship between the two functions:
+#'
+#' ```
+#' compressed <- vec_unrep(x)
+#' x == vec_rep_each(compressed$key, compressed$times)
```
lionel-, line 22: The `==` is a bit confusing because of the vectorised R operation. Perhaps `identical()` or `===`?

R/rep.R
```diff
@@ -5,10 +5,22 @@
...
 #' @details
-#' `vec_rep()` and `vec_rep_each()` work along the size of `x`, rather than
-#' its length. For data frames, this means that rows are repeated rather
-#' than columns.
+#' These functions work along the size of `x`, rather than its length. For data
+#' frames, this means that rows are repeated rather than columns.
+#'
+#' Using `vec_unrep()` and `vec_rep_each()` together is very similar to using
+#' [base::rle()] and [base::inverse.rle()]. The following invariant shows
+#' the relationship between the two functions:
```
lionel-, line 18: Add a boundary unit test for the empty case?

```r
compressed <- vec_unrep(integer())
expect_equal(
  vec_rep_each(compressed$key, compressed$times),
  integer()
)
```

lionel- (approved):
To me "unrep" is clearer than "compress" because it's more explicit about how the vector is compressed. What do you think @hadley?

R/rep.R
```diff
@@ -5,10 +5,22 @@
...
+#'   repeated.
+#'
 #' @details
-#' `vec_rep()` and `vec_rep_each()` work along the size of `x`, rather than
-#' its length. For data frames, this means that rows are repeated rather
-#' than columns.
+#' These functions work along the size of `x`, rather than its length. For data
+#' frames, this means that rows are repeated rather than columns.
+#'
+#' Using `vec_unrep()` and `vec_rep_each()` together is very similar to using
```
hadley, line 16: I'd add something here just being explicit about how it's different to rle. Also might worth drawing connection to `count()`/`uncount()`?

hadley (approved):
`vec_unrep()` seems like a decent name to me.

PR #1306, opened 2020-12-18
Use `rlang::hash()` to remove dependency on digest

### LINKED ISSUE #107: Optional hashing of string values

### LINKED ISSUE #664: Expose vec_hash()

DavisVaughan:
Closes #107 
Closes #664 
Closes #667 

I imagine that we now will _not_ want to expose `obj_hash()` as was done in #667.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,7 @@
 # vctrs (development version)
 
+* Removed dependency on digest.
+
 * Fixed an issue where `vctrs_rcrd` objects were not being proxied correctly
   when used as a data frame column (#1318).
 
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,6 +1,6 @@
 # vctrs (development version)
 
-* Removed dependency on digest.
+* Removed dependency on digest in favor of `rlang::hash()`.
 
 * Fixed an issue where `vctrs_rcrd` objects were not being proxied correctly
   when used as a data frame column (#1318).
--- a/R/type-factor.R
+++ b/R/type-factor.R
@@ -225,7 +225,7 @@ hash_label <- function(x, length = 5) {
   } else {
     # Can't use hash() currently because it hashes the string pointers
     # for performance, so the values in the test change each time
-    substr(digest::digest(x), 1, length)
+    substr(rlang::hash(x), 1, length)
   }
 }
 
--- a/R/type-factor.R
+++ b/R/type-factor.R
@@ -223,7 +223,7 @@ hash_label <- function(x, length = 5) {
   if (length(x) == 0) {
     ""
   } else {
-    # Can't use hash() currently because it hashes the string pointers
+    # Can't use obj_hash() because it hashes the string pointers
     # for performance, so the values in the test change each time
     substr(rlang::hash(x), 1, length)
   }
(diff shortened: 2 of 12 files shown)


NEWS.md
```diff
@@ -1,5 +1,7 @@
 # vctrs (development version)
 
+* Removed dependency on digest.
```
hadley, line 3: in favour of ...

R/type-factor.R
```diff
@@ -223,9 +223,9 @@ hash_label <- function(x, length = 5) {
   if (length(x) == 0) {
     ""
   } else {
-    # Can't use hash() currently because it hashes the string pointers
+    # Can't use obj_hash() because it hashes the string pointers
     # for performance, so the values in the test change each time
-    substr(digest::digest(x), 1, length)
+    substr(rlang::hash(x), 1, length)
```
hadley, line 228: Probably worth doing a revdep run just to make sure this doesn't break any existing tests.

tests/testthat/error/test-conditions.txt
```diff
@@ -101,15 +101,15 @@ lossy cast from character to factor mentions loss of generality
 ===============================================================
 
 > vec_cast("a", factor("b"))
-Error: Can't convert from <character> to <factor<ddf10>> due to loss of generality.
+Error: Can't convert from <character> to <factor<9b7e3>> due to loss of generality.
```
hadley, line 104: Is this now being OS/R version dependent going to cause problems?

DavisVaughan: It should not be OS dependent, so that should be okay.

For somewhat complex types it may be R version dependent (I noticed it was with data frames), but simple R objects seem to give the same hash across R versions. The tests pass here, so it seems like factors are okay?

I'm not sure what is different about data frames, but R core must have changed something about the way they are structured somewhere along the way, which changed the hash - possibly they changed something about the way internal row names work?

DavisVaughan: Oh I figured out what changed with data frames. Some time between R 3.3 and now, the order of the attributes changed. `class` now comes before `row.names`, and that changed the internal hash. Otherwise, barring ALTREP differences, the hashes are _generally_ also stable across R versions, but we can't promise that because of things like this.

R 3.3.3:

```r
> digest::digest(data.frame(x=1))
[1] "e74d491629d919e23692a94cd9e5b935"

> digest::digest(structure(list(x = 1), row.names = c(NA, -1L), class = "data.frame"))
[1] "e74d491629d919e23692a94cd9e5b935"

> digest::digest(structure(list(x = 1), class = "data.frame", row.names = c(NA, -1L)))
[1] "65a37f3d72f58de9eb80cd1674896a60"
```

R 4.0.3

```r
> digest::digest(data.frame(x=1))
[1] "65a37f3d72f58de9eb80cd1674896a60"

> digest::digest(structure(list(x = 1), row.names = c(NA, -1L), class = "data.frame"))
[1] "e74d491629d919e23692a94cd9e5b935"

> digest::digest(structure(list(x = 1), class = "data.frame", row.names = c(NA, -1L)))
[1] "65a37f3d72f58de9eb80cd1674896a60"
```

R/type-factor.R
DavisVaughan: I revdep run shows no breakage!

lionel- (approved):

PR #1314, opened 2021-02-01
add unlicense disclaimer

### LINKED ISSUE #1254: License s3_register with unlicense
https://choosealicense.com/licenses/unlicense/ — to make it very clear that it's fine to copy and paste the source code into your package.  I think this just needs a comment in the code and new section in the documentation that emphasises that we expect you to copy and paste into your package, and the function is licensed to make that as easy as possible.

maxheld83:
this adds the unlicense to the `s3_register()` source, closes #1254.


LICENSE.unlicense
```diff
@@ -0,0 +1,24 @@
+This is free and unencumbered software released into the public domain.
```
hadley, line 1: There's no need to include the source of this license here

R/register-s3.R
```diff
@@ -1,3 +1,5 @@
+# This source code file is licensed under the unlicense license found in the 
+# `LICENSE.unlicense` file in the root directory of this source tree.
```
hadley, line 2: Just use a link here

LICENSE.unlicense
maxheld83: done

PR #1315, opened 2021-02-01
mention alternative licenses in one place; superset of #1314

### LINKED ISSUE #1254: License s3_register with unlicense

maxheld83:
#1314 applies the unlicense to `s3_register()` and closes #1254.

It turns out that there are already some alternatively licensed source files, licensed under the MPL-2.0.
This was already documented in `LICENSE.note`.

So I've streamlined this and now all alternative licenses are mentioned in `LICENSE.note`, and included in separate files.

So this is a superset of #1314.


maxheld83:
as per the review of #1314, the unlicense license text is *not* added to the package source.
There is, therefore, no need to streamline the documentation of multiple licenses.

PR #1328, opened 2021-02-26
Delegate encoding normalization to `vec_normalize_encoding()`

DavisVaughan:
This PR strips out all of the encoding logic from `vec_order_radix()`, replacing it with a single call to `vec_normalize_encoding()` at the top.

This greatly simplifies the character encoding logic, at the cost of some performance in two ways:

- Previously, we only checked the encodings of the _unique_ strings to determine if any re-encoding was required. Now we are forced to loop through the whole vector up front, checking the encoding of each string. This generally isn't that bad, but...

- ...When there are _lots_ of unique values, `vec_normalize_encoding()` is slower than when there are just a few. I assume this has something to do with the CPU caching and reusing the result of `LEVELS(x)` (which we use in `string_is_normalized()`) when there are just a few unique values.

I still think this tradeoff is worth it, as it further standardizes how we handle encodings throughout vctrs and is much easier to reason about. The performance hit from calling `vec_normalize_encoding()` will really only be noticeable if there are a massive amount of unique values.

``` r
library(vctrs)

set.seed(123)

vec_order_radix <- vctrs:::vec_order_radix
vec_normalize_encoding <- vctrs:::vec_normalize_encoding
```

```r
# few uniques
dict <- c("apples", "bananas")
x1 <- sample(dict, size = 1e6, replace = TRUE)
x2 <- sample(dict, size = 1e7, replace = TRUE)

## few uniques / short
# before
bench::mark(vec_order_radix(x1), iterations = 50)
#> # A tibble: 1 x 6
#>   expression               min   median `itr/sec` mem_alloc `gc/sec`
#>   <bch:expr>          <bch:tm> <bch:tm>     <dbl> <bch:byt>    <dbl>
#> 1 vec_order_radix(x1)   8.72ms   11.3ms      88.6    12.7MB     19.4

# after
bench::mark(vec_order_radix(x1), iterations = 50)
#> # A tibble: 1 x 6
#>   expression               min   median `itr/sec` mem_alloc `gc/sec`
#>   <bch:expr>          <bch:tm> <bch:tm>     <dbl> <bch:byt>    <dbl>
#> 1 vec_order_radix(x1)   10.6ms   13.3ms      75.4    12.7MB     16.5

# normalization cost
bench::mark(vec_normalize_encoding(x1))
#> # A tibble: 1 x 6
#>   expression                      min   median `itr/sec` mem_alloc `gc/sec`
[...]

PR #1334, opened 2021-03-04
Add `string_key` argument to `vec_order_radix()`

DavisVaughan:
Will need to rebase on #1328, which will ensure that `x` is UTF-8 _before_ we apply `string_key`

This PR adds the long awaited `string_key` argument to `vec_order_radix()`. This adds a lot of flexibility, giving the user the power to do things like:

``` r
library(vctrs)

vec_sort_radix <- function(x, string_key = NULL) {
  o <- vctrs:::vec_order_radix(x, string_key = string_key)
  vec_slice(x, o)
}

x <- c("B", "A", "a", "A")

# capital before lower case
vec_sort_radix(x)
#> [1] "A" "A" "B" "a"

# case insensitive
vec_sort_radix(x, tolower)
#> [1] "A" "a" "A" "B"

# en_US
vec_sort_radix(x, ~stringi::stri_sort_key(.x, locale = "en_US"))
#> [1] "a" "A" "A" "B"
```

The only thing that is mildly surprising is that, with `vec_order_locs()` (which I doubt we ever expose as is), if the `string_key` makes strings that were previously different look identical (as with `tolower()`), then the keys and locations returned reflect that. I think this is expected behavior. Also note that it doesn't affect typical usage of `stri_sort_key()`, since this orders `"a"` before `"A"` rather than making them look identical.

```r
library(vctrs)

x <- c("B", "A", "a", "A")

vctrs:::vec_order_locs(x, string_key = tolower)
#>   key     loc
#> 1   A 2, 3, 4
#> 2   B       1

vctrs:::vec_order_locs(x, string_key = ~stringi::stri_sort_key(.x, locale = "en_US"))
#>   key  loc
#> 1   a    3
#> 2   A 2, 4
#> 3   B    1
```

### DIFF
--- a/R/order-radix.R
+++ b/R/order-radix.R
@@ -52,6 +52,10 @@
 #'     length as `x`, also encoded as UTF-8.
 #'   - For data frames, `chr_transform` will be applied to all character
 #'     columns.
+#'
+#'   Common transformation functions include: `tolower()` for case-insensitive
+#'   ordering and `stringi::str_sort_key()` for locale-aware ordering. See the
+#'   Details section for more information.
 #' @return
 #' * `vec_order()` an integer vector the same size as `x`.
 #' * `vec_sort()` a vector with the same size and type as `x`.
--- a/R/order-radix.R
+++ b/R/order-radix.R
@@ -20,7 +20,7 @@
 #' the C-locale. While sorting with the C-locale can be useful for
 #' algorithmic efficiency, in many real world uses it can be the cause of
 #' data analysis mistakes. To balance these trade-offs, you can supply a
-#' `string_key` to transform character vectors into an alternative
+#' `chr_transform` to transform character vectors into an alternative
 #' representation that orders in the C-locale in a less surprising way. For
 #' example, providing [base::tolower()] as a string key will order the original
 #' string in a case-insensitive manner. Locale-aware ordering can be achieved
@@ -42,14 +42,15 @@
 #'   - For data frames, a length `1` or `ncol(x)` character vector containing
 #'     only `"largest"` or `"smallest"`, specifying how `NA`s should be treated
 #'     in each column.
-#' @param string_key Transformation of character vectors for sorting in
+#' @param chr_transform Transformation of character vectors for sorting in
 #'   alternate locales.
 #'   - If `NULL`, no transformation is done.
 #'   - Otherwise, this must be a function of one argument. The function will be
 #'     invoked with `x`, if it is a character vector, after it has been
 #'     translated to UTF-8, and should return a character vector with the same
 #'     length as `x`, also encoded as UTF-8.
-#'   - For data frames, `string_key` will be applied to all character columns.
+#'   - For data frames, `chr_transform` will be applied to all character
+#'     columns.
 #' @return
 #' * `vec_order()` an integer vector the same size as `x`.
 #' * `vec_sort()` a vector with the same size and type as `x`.
@@ -87,15 +88,15 @@
 #' y <- c("B", "A", "a")
 #' vec_sort(y)
 #'
-#' # To order in a case-insensitive manner, provide a `string_key` that
+#' # To order in a case-insensitive manner, provide a `chr_transform` that
 #' # transforms the strings to all lowercase
-#' vec_sort(y, string_key = tolower)
+#' vec_sort(y, chr_transform = tolower)
 #' @noRd
 vec_order_radix <- function(x,
                             direction = "asc",
                             na_value = "largest",
-                            string_key = NULL) {
-  .Call(vctrs_order, x, direction, na_value, string_key)
+                            chr_transform = NULL) {
+  .Call(vctrs_order, x, direction, na_value, chr_transform)
 }
 
 #' Identify ordered groups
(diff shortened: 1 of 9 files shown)


DavisVaughan:
A few benchmarks suggests that it is extremely compelling to use `stri_sort_key()` + radix ordering of character vectors over the locale aware shell sort that R does

```r
library(vctrs)
library(stringi)

set.seed(123)
vec_order2 <- vctrs:::vec_order_radix

dict <- stri_rand_strings(100, 5)

x <- sample(dict, 1e5, TRUE)
y <- sample(dict, 1e6, TRUE)
z <- sample(dict, 1e7, TRUE)

bench::mark(
  base = order(x),
  radix = vec_order2(x, string_key = ~stri_sort_key(.x, locale = "en_US"))
)
#> # A tibble: 2 x 6
#> 1 base        267.9ms  268.3ms      3.73  390.67KB        0
#> 2 radix        27.7ms   30.9ms     30.9     2.34MB        0

bench::mark(
  base = order(y),
  radix = vec_order2(y, string_key = ~stri_sort_key(.x, locale = "en_US"))
)
#> # A tibble: 2 x 6
#> 1 base          4.64s    4.64s     0.216    3.81MB     0   
#> 2 radix      324.32ms 342.73ms     2.92    20.33MB     2.92

bench::mark(
  base = order(z),
  radix = vec_order2(z, string_key = ~stri_sort_key(.x, locale = "en_US"))
)
#> # A tibble: 2 x 6
#> 1 base          1.16m    1.16m    0.0144    38.1MB    0    
#> 2 radix         3.13s    3.13s    0.320    200.6MB    0.320
```

DavisVaughan:
A bit more information to consider here. This argument will have to bubble up in a few places. I can think of:

```r
vec_order()
vec_order_*() # lower level helpers that return group size info
vec_sort()
vec_rank()
```

It should not have to appear in `vec_match()`, if that ever used `vec_order()` internally, as it would order by first appearance and could use the C locale with no issues.

I don't think this should stop us from adding it. Pros of having it as an argument that gets passed down into the guts of vec-order are:

- It is guaranteed to get applied after `x` has been translated to UTF8 with `vec_normalize_encoding()`
- It is applied after `vec_proxy_order()`, which could change the structure of `x`, potentially adding or removing character vectors
- It automatically handles traversal over data frame `x` to ensure we apply it on all character columns

I had thought about supplying some helper like `proxy_string_transform(x, fn)`, but I think that would be difficult to use robustly, as you'd have to manually proxy, call this new helper, then pass off to `vec_order()` (and even with all that the first bullet point about translation to UTF8 isn't solved).

lionel- (commented):
Performance with sort keys looks good!

- Would it be possible to apply the key transformation early instead of passing the key function through all these layers?

- Similarly, rather than adding a key argument in many places, it might be worth exploring how to attach a key to a character vector. This would have the advantage of avoiding repeated transformations of the character vector. It seems like a record vector would work but we'd need to keep the key in the attributes to avoid changing the underlying type for compatibility (so we couldn't use vctrs-rcrd).

- I feel that we could do better than `string_key` for the argument name. "string" does not seem appropriate because it's about character vectors. "key" does not indicate a transformation function. Maybe `chr_order_trans` or `chr_order_proxy`?

DavisVaughan:
> Would it be possible to apply the key transformation early instead of passing the key function through all these layers?

Good idea. We could introduce `proxy_chr_transform(proxy, fn)` as I mentioned above. It would be documented to expect that `proxy` has already been proxied by `vec_proxy_order()`, and that it has been normalized to UTF-8 with `vec_normalize_encoding()`.

Because it would be called on the order proxy, any nested data frames would already be flattened, so traversal should be straightforward.

---

Maybe just `chr_transform` instead of `string_key`?

DavisVaughan:
@lionel- could you please take one more look at this regarding the usage of `proxy_chr_transform()`? I think it does simplify things nicely, thanks for the suggestion there.

R/order-radix.R
```diff
@@ -36,6 +43,15 @@
 #'   - For data frames, a length `1` or `ncol(x)` character vector containing
 #'     only `"largest"` or `"smallest"`, specifying how `NA`s should be treated
 #'     in each column.
+#' @param chr_transform Transformation of character vectors for sorting in
```
lionel-, line 46: Mention the relevant stringi function to sort in alternate locale?

lionel- (approved):

lionel-:
If we do end up with superclasses of character that embed a locale or sort key, will `chr_transform` have precedence or will `chr_transform` only apply as a default?

DavisVaughan:
> will chr_transform have precedence or will chr_transform only apply as a default?

_Probably_ the sort key in the superclass would have precedence?

I imagine that if you specifically set a character vector's locale to `"de_DE"` with this superclass, then if I called `vec_order_radix(x, chr_transform = ~str_sort_key(.x, "en_US"))` and `x` was a data frame with that localized character vector as a column, then you'd probably want it to keep its special locale.

But you could also specify a transform of `chr_transform = tolower`, which isn't really tied to locale as much, and I'm not sure which should win there.

PR #1342, opened 2021-03-29
Correctly flip parsed `na_last` value when `direction = "desc"`

DavisVaughan:
I'm not entirely sure how I did this, but this bit of `order_proxy()` wasn't faithfully translated over to `vec_order_radix()`

https://github.com/r-lib/vctrs/blob/f255bca3cf118e85701966ffca0c4692c29ca5f3/R/order.R#L53-L55

That resulted in this inconsistency when `direction = "desc"`

``` r
x <- c(1, 2, NA)
x[vctrs::vec_order(x, direction = "desc", na_value = "smallest")]
#> [1]  2  1 NA
x[vctrs:::vec_order_radix(x, direction = "desc", na_value = "smallest")]
#> [1] NA  2  1
```

We now flip the `na_last` argument (parsed from `na_value`) depending on the `decreasing` value (parsed from `direction`), being careful to recycle them in case the user does `direction = c("desc", "asc"), na_value = "smallest"`.

Luckily there were a large number of tests for this, they were just coded incorrectly. Fixing those tests makes me confident that this is working correctly now.

### DIFF
--- a/src/order-radix.c
+++ b/src/order-radix.c
@@ -184,50 +184,38 @@
 
 // -----------------------------------------------------------------------------
 
-static SEXP vec_order(SEXP x, SEXP decreasing, SEXP na_last, SEXP chr_transform);
+static SEXP vec_order(SEXP x, SEXP direction, SEXP na_value, SEXP chr_transform);
 
 // [[ register() ]]
 SEXP vctrs_order(SEXP x, SEXP direction, SEXP na_value, SEXP chr_transform) {
-  SEXP decreasing = PROTECT(parse_direction(direction));
-  SEXP na_last = PROTECT(parse_na_value(na_value));
-
-  SEXP out = vec_order(x, decreasing, na_last, chr_transform);
-
-  UNPROTECT(2);
-  return out;
+  return vec_order(x, direction, na_value, chr_transform);
 }
 
 
 static SEXP vec_order_impl(SEXP x,
-                           SEXP decreasing,
-                           SEXP na_last,
+                           SEXP direction,
+                           SEXP na_value,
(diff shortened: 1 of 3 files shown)


DavisVaughan:
@lionel- I'm going to go ahead and merge this to work on a follow up PR, but any feedback is welcome

tests/testthat/test-order-radix.R
```diff
@@ -56,11 +56,11 @@ test_that("all combinations of `direction` and `na_value` work", {
   )
   expect_identical(
     x[vec_order_radix(x, na_value = "largest", direction = "desc")],
-    x[order(x, na.last = TRUE, decreasing = TRUE)]
+    x[order(x, na.last = FALSE, decreasing = TRUE)]
```
lionel-, line 59: These tests confused me for a moment. They are a bit misleading because they imply a one-to-one correspondence between the two ordering behaviours. Should we test against `vec_order()` instead?

Also we should document in `?vec_order::na_value`  that it behaves differently from `order::na.last` (`na.last` doesn't interact with `decreasing` whereas `na_value` interacts with `direction`).

src/order-radix.c
```diff
@@ -4065,9 +4065,42 @@ int df_decreasing_expansion(SEXP x) {
 
 // -----------------------------------------------------------------------------
 
+/*
+ * `na_value` -> `na_last` is originally parsed as:
```
lionel-, line 4069: May be confusing to call this argument `na_last` since it doesn't behave like `order::na.last`. `na_largest` might be a better name for this boolean. No need for action now though.

lionel- (approved):

PR #1343, opened 2021-03-29
Add the capability to differentiate `NaN` from `NA_real_` in `vec_order_radix()`

DavisVaughan:
~Requires #1342 to be merged, and some changes will have to be done here after that~ Done

Dictionary functions generally treat `NA_real_` different from `NaN` for double and complex vectors:

``` r
x <- c(NA_real_, NaN)
vctrs::vec_unique(x)
#> [1]  NA NaN
vctrs::vec_match(x, x, na_equal = TRUE)
#> [1] 1 2

x <- complex(real = c(NA, NA, NaN, NaN), imaginary = c(NA, NaN, NaN, NA))
vctrs::vec_unique(x)
#> [1]       NA       NA NaN+NaNi       NA
```

If we are thinking about using `vec_order_radix()` for these internally, we will need to allow `NaN` to look different from `NA_real_` in the ordering code.

I've added a boolean `nan_distinct` argument to `vec_order_radix()` to control this. If `TRUE`, `NaN` values are always placed between `NA_real_` and non-missing numbers. Because there is no obvious ordering between `NA_real_` and `NaN`, this seems like a reasonable heuristic, and is also what data.table does (unconditionally, I might add).

Unfortunately, we do have to thread this `nan_distinct` argument all the way through to `dbl_adjust()`. I don't see any way around that, but it isn't too bad. It also gets threaded through the complex (`cpl_*()`) calls.

``` r

x <- c(NaN, 1, 2, NA_real_, NaN, NA_real_)

x[vec_order_radix(x, direction = "asc", na_value = "largest", nan_distinct = TRUE)]
#> [1]   1   2 NaN NaN  NA  NA
x[vec_order_radix(x, direction = "asc", na_value = "smallest", nan_distinct = TRUE)]
#> [1]  NA  NA NaN NaN   1   2
x[vec_order_radix(x, direction = "desc", na_value = "largest", nan_distinct = TRUE)]
#> [1]  NA  NA NaN NaN   2   1
x[vec_order_radix(x, direction = "desc", na_value = "smallest", nan_distinct = TRUE)]
#> [1]   2   1 NaN NaN  NA  NA
```

### DIFF
--- a/src/order-sortedness.c
+++ b/src/order-sortedness.c
@@ -15,10 +15,15 @@
 
 // -----------------------------------------------------------------------------
 
-static inline int dbl_cmp(double x, double y, const int direction, const int na_order);
+static inline int dbl_cmp(const double x,
+                          const double y,
+                          const int direction,
+                          const int na_order,
+                          const int na_nan_cmp);
 
 /*
- * Check if a double vector is ordered, handling `decreasing` and `na_last`
+ * Check if a double vector is ordered, handling `decreasing`, `na_last`, and
+ * `nan_distinct`.
  *
  * If the double vector is in the expected ordering, no sorting needs to
  * occur. In these cases, if `p_x` is in exactly the expected ordering.
@@ -31,6 +36,7 @@ enum vctrs_sortedness dbl_sortedness(const double* p_x,
                                      r_ssize size,
                                      bool decreasing,
                                      bool na_last,
+                                     bool nan_distinct,
                                      struct group_infos* p_group_infos) {
   if (size == 0) {
     return VCTRS_SORTEDNESS_sorted;
@@ -43,6 +49,7 @@ enum vctrs_sortedness dbl_sortedness(const double* p_x,
 
   const int direction = decreasing ? -1 : 1;
   const int na_order = na_last ? 1 : -1;
+  const int na_nan_cmp = nan_distinct ? na_order : 0;
 
   double previous = p_x[0];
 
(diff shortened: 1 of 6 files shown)


src/order-sortedness.c
```diff
@@ -127,25 +147,45 @@ enum vctrs_sortedness dbl_sortedness(const double* p_x,
...
+ * Compare two doubles, handling `na_order`, `direction`, and `na_nan_cmp`
  */
 static inline
-int dbl_cmp(double x, double y, const int direction, const int na_order) {
-  if (isnan(x)) {
-    if (isnan(y)) {
-      return 0;
-    } else {
-      return na_order;
+int dbl_cmp(const double x,
```
lionel-, line 156: Since we are not passing addresses, I would have made the signature consistent by removing all `const` qualifiers rather than add some.

src/order-sortedness.c
```diff
@@ -127,25 +147,45 @@ enum vctrs_sortedness dbl_sortedness(const double* p_x,
...
+            const double y,
+            const enum vctrs_dbl_class x_type,
+            const enum vctrs_dbl_class y_type,
+            const int direction,
+            const int na_order,
+            const int na_nan_cmp) {
```
lionel-, line 162: Name this argument `na_nan_order` since it has the same sort of meaning as `na_order`? Also to me a `_cmp` suffix evokes a function pointer.

lionel- (approved):
👍

PR #1347, opened 2021-03-31
Document interaction between `na_value` and `direction`

lionel-:


DavisVaughan (approved):

DavisVaughan:
Oh wait, you should put these docs in `order-radix.R` not `order.R`!

R/order.R
```diff
@@ -7,6 +7,14 @@
 #' * `vec_order()` an integer vector the same size as `x`.
 #' * `vec_sort()` a vector with the same size and type as `x`.
 #'
+#' @section Differences with `base::order()`:
```
DavisVaughan, line 10: Should be in `order-radix.R` for now

DavisVaughan (changes requested):

R/order.R
DavisVaughan: Oh never mind, `vec_order()` does this too, duh

PR #1350, opened 2021-04-02
Modularize TRUELENGTH saving/reallocating helpers

DavisVaughan:
For `vec_order_radix()`'s `struct truelength_info`, we only have to store the original TRUELENGTH value in `strings` / `truelengths` if R is actually using it, i.e. the value is >0, otherwise we can always reset it to the default value of 0. It is rare for R to use the TRUELENGTH, so we should rarely have to reallocate the `strings` and `truelengths` vectors. This saves us some time with large vectors when most of the strings are unique, because we previously saved every unique value and its TRUELENGTH.

Base R actually does this optimization too, I just didn't see the benefit of it until now.

I've also separated out the saving/reallocating of individual CHARSXP `sizes` into its own helper as well. I've got an upcoming PR that allows for ordering character vectors by appearance, which doesn't require the `sizes` vector, so this is preparation for that. That PR will be used for reimplementing the dictionary functions.

In the numbers below, I'm mainly interested in the big drop in memory usage required to sort a character vector with lots of unique values

``` r
set.seed(123)

random_strings <- function(size, min_length, max_length) {
  lengths <- rlang::seq2(min_length, max_length)
  stringi::stri_rand_strings(
    size,
    sample(lengths, size = size, replace = TRUE)
  )
}

# all unique strings!
x <- random_strings(size = 1e7, min_length = 5, max_length = 20)

# before
bench::mark(vec_order_radix(x), iterations = 20)
#> # A tibble: 1 x 6
#> 1 vec_order_radix(x)    5.86s    5.97s     0.162     856MB    0.341

# after
bench::mark(vec_order_radix(x), iterations = 20)
#> # A tibble: 1 x 6
#> 1 vec_order_radix(x)    5.12s    5.23s     0.185     547MB    0.222
```

### DIFF
--- a/src/order-radix.c
+++ b/src/order-radix.c
@@ -3036,7 +3036,7 @@ void chr_mark_sorted_uniques(const SEXP* p_x,
                              struct truelength_info* p_truelength_info) {
   chr_mark_uniques(p_x, size, p_truelength_info);
 
-  r_ssize n_uniques = p_truelength_info->size_used;
+  r_ssize n_uniques = p_truelength_info->n_uniques_used;
 
   SEXP* p_x_aux = (SEXP*) init_lazy_raw(p_lazy_x_aux);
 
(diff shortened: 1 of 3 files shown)


DavisVaughan:
@lionel- I feel confident about this, and it'll be simpler for me to add further PRs if I go ahead and merge, so I'm going to merge it in. Feel free to review if you have time though!

Issue #1004, opened 2020-04-10
Behavior of `max/min.vctrs_vctr` with empty vectors


DavisVaughan:
Currently, `max.vctrs_vctr` tries to cast `-Inf` to the type of `x` if `x` is empty. I'm not sure that that makes sense in all cases, and it assumes that there is a cast method from double to the type of `x` (and that that cast method works with infinity).

For example, we can generally take the max/min of a character vctr, but if it is empty we get an error. It doesn't error for `pmax()` because its size invariant is different. `max()` is required to return an object of size 1, `pmax()` returns an object of the common size of all its input.

``` r
library(vctrs)

y <- new_vctr("x")
max(y)
#> <vctrs_vctr[1]>
#> [1] x
pmax(y)
#> <vctrs_vctr[1]>
#> [1] x

x <- new_vctr(character())
max(x)
#> Error: Can't convert <double> to <vctrs_vctr>.
pmax(x)
#> <vctrs_vctr[0]>
```

I think I would argue that for `vctrs_vctr` objects, we should return a typed `NA` for `max()` and `min()` if the input is empty. This would keep it from ever erroring, wouldn't require cast methods to work, and wouldn't be a completely new idea because it's what `max()` does with empty character input.

```r
max(integer())
#> Warning in max(integer()): no non-missing arguments to max; returning -Inf
#> [1] -Inf
max(numeric())
#> Warning in max(numeric()): no non-missing arguments to max; returning -Inf
#> [1] -Inf
max(character())
#> Warning in max(character()): no non-missing arguments, returning NA
#> [1] NA
```

lionel-:
The problem of returning `NA` is that we are losing the information that an empty vector doesn't contain any value that might be lower or larger than a future input.

DavisVaughan:
It really seems like the minimum and maximum allowed values of a specific ptype are specific to that ptype, and can often be impossible to guess using general heuristics. Maybe we need something like `vec_maximum(ptype)` and `vec_minimum(ptype)` which could be generic and return the min/max values of the given ptype?

For example, in clock the minimum allowed calendar year is around year `-32768` but there is no R level heuristic that can tell you that (it's the minimum value of a C++ `short`)

lionel-:
The base methods emit a warning. Maybe we should just fail?

PR #1358, opened 2021-04-10
Fix `min()`, `max()` and `range()` where `na.rm = TRUE` and all values are `NA`

### LINKED ISSUE #1357: min and max throw an error if `na.rm = TRUE` and all values are `NA`
From tidyverse/haven#528 - `min()` and `max()` throw an error if all values are `NA` and `na.rm = TRUE`.
The base R versions return `Inf` or `-Inf` respectively.

They should possibly return `NA` instead of `Inf`/`-Inf`, see #1004.

``` r
min(NA_real_, na.rm = TRUE)
#> Warning in min(NA_real_, na.rm = TRUE): no non-missing arguments to min;
#> returning Inf
#> [1] Inf
vctrs:::min.vctrs_vctr(NA_real_, na.rm = TRUE)
#> Error in idx[[1]]: subscript out of bounds

max(NA_real_, na.rm = TRUE)
#> Warning in max(NA_real_, na.rm = TRUE): no non-missing arguments to max;
#> returning -Inf
#> [1] -Inf
vctrs:::max.vctrs_vctr(NA_real_, na.rm = TRUE)
#> Error in idx[[1]]: subscript out of bounds
```

<sup>Created on 2021-04-09 by the [reprex package](https://reprex.tidyverse.org) (v1.0.0)</sup>

gorcha:
This PR fixes #1357. Where `na.rm = TRUE` and all values are `NA`, these functions now return `Inf`/`-Inf` or `NA` depending on the input type.

As noted in #1004, base R returns `Inf`/`-Inf` for numeric types, and `NA_character_` for character vectors, so it should do so in vctrs as well. I've followed almost the same rules here.

If `Inf`/`-Inf` can be successfully cast to the input type that will be the return value. Otherwise it will return `NA` cast to the appropriate type. This is kind of the same as base R in theory, but in vctrs doubles can't be cast to integer or logical, so these return `Inf` in base R but `NA` here.

This behaviour has been extended to empty vectors as well, so they no longer throw an error if `Inf` can't be cast to the input type.

I'm thinking that #1004 is right, and it should just be `NA` in all cases, but I've left `Inf` as the default where possible to match existing behaviour.

``` r
library(vctrs)

# Integer
min(integer())
#> Warning in min(integer()): no non-missing arguments to min; returning Inf
#> [1] Inf
min(new_vctr(integer()))
#> <vctrs_vctr[1]>
#> [1] NA

min(NA_integer_, na.rm = TRUE)
#> Warning in min(NA_integer_, na.rm = TRUE): no non-missing arguments to min;
#> returning Inf
#> [1] Inf
min(new_vctr(NA_integer_), na.rm = TRUE)
#> <vctrs_vctr[1]>
#> [1] NA

# Double
min(numeric())
#> Warning in min(numeric()): no non-missing arguments to min; returning Inf
#> [1] Inf
min(new_vctr(numeric()))
#> <vctrs_vctr[1]>
#> [1] Inf

min(NA_real_, na.rm = TRUE)
#> Warning in min(NA_real_, na.rm = TRUE): no non-missing arguments to min;
#> returning Inf
#> [1] Inf
min(new_vctr(NA_real_), na.rm = TRUE)
#> <vctrs_vctr[1]>
#> [1] Inf

# Character
min(character())
#> Warning in min(character()): no non-missing arguments, returning NA
#> [1] NA
min(new_vctr(character()))
#> <vctrs_vctr[1]>
#> [1] NA

min(NA_character_, na.rm = TRUE)
#> Warning in min(NA_character_, na.rm = TRUE): no non-missing arguments, returning
#> NA
#> [1] NA
min(new_vctr(NA_character_), na.rm = TRUE)
#> <vctrs_vctr[1]>
#> [1] NA

# Logical
min(logical())
#> Warning in min(logical()): no non-missing arguments to min; returning Inf
#> [1] Inf
min(new_vctr(logical()))
#> <vctrs_vctr[1]>
#> [1] NA

min(NA, na.rm = TRUE)
#> Warning in min(NA, na.rm = TRUE): no non-missing arguments to min; returning Inf
#> [1] Inf
min(new_vctr(NA), na.rm = TRUE)
#> <vctrs_vctr[1]>
#> [1] NA
```

<sup>Created on 2021-04-10 by the [reprex package](https://reprex.tidyverse.org) (v1.0.0)</sup>

### DIFF
--- a/R/type-vctr.R
+++ b/R/type-vctr.R
@@ -477,7 +477,7 @@ quantile.vctrs_vctr <- function(x, ..., type = 1, na.rm = FALSE) {
 
 vec_cast_or_na <- function(x, to, ...) {
   tryCatch(
-    vctrs_error_incompatible_type = function(...) vec_cast(rep(NA, length(x)), to),
+    vctrs_error_incompatible_type = function(...) vec_init(to, length(x)),
     vec_cast(x, to)
   )
 }
--- a/R/type-vctr.R
+++ b/R/type-vctr.R
@@ -475,10 +475,17 @@ quantile.vctrs_vctr <- function(x, ..., type = 1, na.rm = FALSE) {
   # nocov end
 }
 
+vec_cast_or_na <- function(x, to, ...) {
+  tryCatch(
+    vctrs_error_incompatible_type = function(...) vec_cast(rep(NA, length(x)), to),
+    vec_cast(x, to)
+  )
+}
+
 #' @export
 min.vctrs_vctr <- function(x, ..., na.rm = FALSE) {
   if (vec_is_empty(x)) {
-    return(vec_cast(Inf, x))
+    return(vec_cast_or_na(Inf, x))
   }
 
   # TODO: implement to do vec_arg_min()
@@ -486,6 +493,9 @@ min.vctrs_vctr <- function(x, ..., na.rm = FALSE) {
 
   if (isTRUE(na.rm)) {
     idx <- which.min(rank)
+    if (vec_is_empty(idx)) {
+      return(vec_cast_or_na(Inf, x))
+    }
   } else {
     idx <- which(vec_equal(rank, min(rank), na_equal = TRUE))
   }
@@ -496,7 +506,7 @@ min.vctrs_vctr <- function(x, ..., na.rm = FALSE) {
 #' @export
 max.vctrs_vctr <- function(x, ..., na.rm = FALSE) {
   if (vec_is_empty(x)) {
-    return(vec_cast(-Inf, x))
+    return(vec_cast_or_na(-Inf, x))
   }
 
   # TODO: implement to do vec_arg_max()
@@ -504,6 +514,9 @@ max.vctrs_vctr <- function(x, ..., na.rm = FALSE) {
 
   if (isTRUE(na.rm)) {
     idx <- which.max(rank)
+    if (vec_is_empty(idx)) {
+      return(vec_cast_or_na(-Inf, x))
+    }
   } else {
     idx <- which(vec_equal(rank, max(rank), na_equal = TRUE))
   }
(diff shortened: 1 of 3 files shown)


R/type-vctr.R
```diff
@@ -475,17 +475,27 @@ quantile.vctrs_vctr <- function(x, ..., type = 1, na.rm = FALSE) {
...
-    return(vec_cast(Inf, x))
+    return(vec_cast_or_na(Inf, x))
   }
 
   # TODO: implement to do vec_arg_min()
   rank <- xtfrm(x)
 
   if (isTRUE(na.rm)) {
     idx <- which.min(rank)
+    if (vec_is_empty(idx)) {
```
lionel-, line 496: Are the empty idx cases covered by unit tests?

lionel- (commented):
Seems reasonable until we figure out something better.

tests/testthat/test-type-vctr.R
```diff
@@ -584,3 +584,26 @@ test_that("xtfrm() returns a bare vector", {
 test_that("xtfrm() works with character subclass", {
   expect_identical(xtfrm(new_vctr(chr())), int())
 })
+
+test_that("Summary generics behave identically to base if na.rm = TRUE and all values are NA (#1357)", {
+  expect_warning(
+    expect_identical(
+      min(new_vctr(NA_real_), na.rm = TRUE),
+      new_vctr(min(NA_real_, na.rm = TRUE))
```
hadley, line 592: I think you should just inline `Inf` here, to make it easier to see what's going on (and then you don't need to suppress the warnings).

gorcha: Good call, shall do. Thanks!

R/type-vctr.R
gorcha: I don't think so, I'll add in some extra tests. Thanks!

gorcha:
Thanks for the feedback, I've made that `vec_init()` change and expanded the tests to cover the NA and empty vector cases for double, integer, character and logical.

lionel-:
Thanks for the nice PR @gorcha! You'll need to reset your remote master branch since I've pushed to it to resolve conflicts. To avoid this I recommend working from a branch that you can then delete after the patch is merged.

Issue #736, opened 2020-01-08
Consider sorting instead of hashing for uniqueness


hadley:
https://news.ycombinator.com/item?id=21985246

This would allow us to use radix sort of for integer indices, which would help us to match data.table performance when grouping in dplyr.

DavisVaughan:
See https://github.com/r-lib/vctrs/pull/1361#issuecomment-1261016775 for explanation of why we probably don't want switch all of the dictionary functions over to sorting

We did expose `vec_locate_sorted_groups()` (which uses radix ordering) for use in dplyr's `group_by()` https://github.com/tidyverse/dplyr/pull/6297, which was really the original motivation of this issue

PR #1361, opened 2021-04-13
Implement alternatives to the `vec_unique_*()` family

DavisVaughan:
Part of https://github.com/r-lib/vctrs/issues/736

This PR adds order based alternative implementations to the `vec_unique_*()` family of functions. I have not replaced them outright yet; it was useful for benchmarking to have both variants around at the same time.

- This required tweaking the internal vctrs_tuple type to have a `vec_proxy_compare()` method
- This required a `vec_proxy_compare()` method for `numeric_version`, which is also useful for use in `vec_compare()`
- I've updated the original sorting vs hashing benchmark document

Timings look pretty good, with the new sort method being about the same or a _little_ slower for small-ish data, but we make up for that by being much faster when there is large data with a large number of groups.

Link to the benchmark doc:
https://github.com/r-lib/vctrs/blob/4479b2f696dd5bdbc587e2a7470a8ea27b97566d/bench/sorting-vs-hashing.md

Also updated the ordering benchmark to include single threaded data table:
https://github.com/r-lib/vctrs/blob/ada71f75ca26de3544ffe40faccf885fb338e5e7/bench/order.md


lionel- (commented):
Since the sort-based implementation is rather small, would it be worth it to branch the implementation depending on the size of the input, so we'd keep the hash-based impl for small vectors?

DavisVaughan:
I'm not sure its worth it. For the small vectors where the hash-based impl is faster, the total time for either impl is generally in the microsecond range (normally with differences like: sort-based 450us, hash-based 280us). Is that worth maintaining two separate implementations for?

lionel-:
If it were for `unique()` I would think it is worth it. When vctrs gets used in funs more (or perhaps in a vctrs-powered REPL down the line), this sort of optimisations for these micro-operations should make a difference in user loops. Is the maintenance burden expected to be large for these implementations? I was imagining it's mostly the starting cost of figuring out the right threshold and setting up unit tests to cover both small and large vectors.

lionel-:
> If the input is ordered already, the order-based methods pick up on this and are extremely fast

oh yes, sounds like a good argument for going order-based. Looking at the results it generally seems reasonable to use the order-based implementation.

Please make the call on whether to branch the implementations or use a single one.

DavisVaughan:
Since we figured out how to fix the dictionary is full issue in https://github.com/r-lib/vctrs/pull/1363, I'm not 100% convinced we should do this

It does seem to be decently faster in some cases, but I think exposing `vec_locate_sorted_groups()` is a nice alternative to switching the dictionary functions over to the ordering approach, because that takes advantage of the fact that we want the output to be ordered (or dont care about order). The approach used here basically calls `vec_locate_sorted_groups()` and then has to re-order by first appearance, which kind of sucks to have to do.

I don't think the dictionary family would ever fully go away (it is required for `vctrs:::vec_proxy_order.list()` for example), so our time is probably better spent optimizing the dictionary approach further as required.

Switching to sorting over hashing also has the issue of making `vec_proxy_order()` a requirement, where previously only `vec_proxy_equal()` was needed. This means that `vec_unique()` couldn't be used on non-orderable classes like `clock::year_month_weekday()`, which seems undesirable (see `clock:::vec_proxy_compare.clock_year_month_weekday()`).

PR #1363, opened 2021-04-14
Fix dictionary function overflow with long input

### LINKED ISSUE #1133: dplyr issue with dictionary size
https://github.com/tidyverse/dplyr/issues/5291

DavisVaughan:
Closes https://github.com/r-lib/vctrs/issues/1133

I am very happy to report that I figured out how to fix #1133. It turns out that we just needed to consistently use `uint32_t` for the `size` and `hash` value results.

As mentioned in #1133, when `x` has size `1e9`, this resulted in our `ceil2()` call of `ceil2(vec_size(x) / 0.77)` overflowing, since it tried to compute `2 ^ 31 = 2147483648`, which is 1 larger than `INT_MAX`. Since `ceil2()` takes and returns `int32_t`, it overflowed. Really it should take and return a `uint32_t`, since that is what the `size` member of the dictionary is stored as. This prevents us from overflowing, since the max `uint32_t` value is `2 ^ 32 - 1`.

Additionally, if a vector has a size greater than `2147483648 * .77 = 1653562409`, we'd have an additional issue where `ceil2(vec_size(x) / 0.77)` would try to theoretically compute and return `2 ^ 32`, which is 1 larger than `UINT_MAX`. To fix this, I've capped the `vec_size(x) / 0.77` value to `INT_MAX`, meaning that the maximum key size that can be computed is `2 ^ 31`. Since we only support short vectors right now, this should never be an issue, it just means that the maximum load factor will be greater than 77% for those vectors.

Here is a reprex for the original issue, which works now:

```r
library(vctrs)

N <- 1e9
K <- 1e2L

set.seed(108)

x <- sample(sprintf("id%03d",1:K), N, TRUE)

tictoc::tic()
zz <- vec_group_loc(x)
tictoc::toc()
#> 30.779 sec elapsed
```

I also rented an EC2 instance with 60gb of RAM to test out this much beefier example of using a vector with the maximum "short" length, and a _lot_ of groups. It also works now (but it takes a loooong time to run). The second commit here regarding capping the key size was required to allow this to work.

```r
library(vctrs)

N <- .Machine$integer.max
K <- 1e8L

set.seed(108)

x <- sample(1:K, N, TRUE)

zz <- vec_group_loc(x)
```

### DIFF
--- a/src/dictionary.c
+++ b/src/dictionary.c
@@ -13,7 +13,7 @@ struct vctrs_arg args_haystack;
 
 
 // http://graphics.stanford.edu/~seander/bithacks.html#RoundUpPowerOf2
-int32_t ceil2(int32_t x) {
+uint32_t ceil2(uint32_t x) {
   x--;
   x |= x >> 1;
   x |= x >> 2;
@@ -79,7 +79,7 @@ static struct dictionary* new_dictionary_opts(SEXP x, struct dictionary_opts* op
     // of at most 77%. We round up to power of 2 to ensure quadratic probing
     // strategy works.
     // Rprintf("size: %i\n", size);
-    R_len_t size = ceil2(vec_size(x) / 0.77);
+    uint32_t size = ceil2(vec_size(x) / 0.77);
     size = (size < 16) ? 16 : size;
 
     d->key = (R_len_t*) R_alloc(size, sizeof(R_len_t));
@@ -492,7 +492,7 @@ SEXP vctrs_count(SEXP x) {
   int* p_val = INTEGER(val);
 
   for (int i = 0; i < n; ++i) {
-    int32_t hash = dict_hash_scalar(d, i);
+    uint32_t hash = dict_hash_scalar(d, i);
 
     if (d->key[hash] == DICT_EMPTY) {
       dict_put(d, hash, i);
@@ -508,7 +508,7 @@ SEXP vctrs_count(SEXP x) {
   int* p_out_val = INTEGER(out_val);
 
   int i = 0;
-  for (int hash = 0; hash < d->size; ++hash) {
+  for (uint32_t hash = 0; hash < d->size; ++hash) {
     if (d->key[hash] == DICT_EMPTY)
       continue;
 
@@ -544,7 +544,7 @@ SEXP vctrs_duplicated(SEXP x) {
   int* p_val = INTEGER(val);
 
   for (int i = 0; i < n; ++i) {
-    int32_t hash = dict_hash_scalar(d, i);
+    uint32_t hash = dict_hash_scalar(d, i);
 
     if (d->key[hash] == DICT_EMPTY) {
       dict_put(d, hash, i);
@@ -558,7 +558,7 @@ SEXP vctrs_duplicated(SEXP x) {
   int* p_out = LOGICAL(out);
 
   for (int i = 0; i < n; ++i) {
-    int32_t hash = dict_hash_scalar(d, i);
+    uint32_t hash = dict_hash_scalar(d, i);
     p_out[i] = p_val[hash] != 1;
   }
 
--- a/src/dictionary.c
+++ b/src/dictionary.c
@@ -174,7 +174,13 @@ uint32_t dict_key_size(SEXP x) {
     stop_internal("dict_key_size", "Dictionary functions do not support long vectors.");
   }
 
-  uint32_t size = (uint32_t)(x_size / 0.77);
+  const double load_adjusted_size = x_size / 0.77;
+
+  if (load_adjusted_size > UINT32_MAX) {
+    stop_internal("dict_key_size", "Can't safely cast load adjusted size to a `uint32_t`.");
+  }
+
+  uint32_t size = (uint32_t)load_adjusted_size;
   size = size > INT_MAX ? INT_MAX : size;
   size = u32_ceil2(size);
   size = (size < 16) ? 16 : size;
--- a/src/dictionary.c
+++ b/src/dictionary.c
@@ -13,6 +13,7 @@ struct vctrs_arg args_haystack;
 
 
 // http://graphics.stanford.edu/~seander/bithacks.html#RoundUpPowerOf2
+static inline
 uint32_t ceil2(uint32_t x) {
   x--;
   x |= x >> 1;
@@ -52,6 +53,8 @@ static struct dictionary* new_dictionary_params(SEXP x, bool partial, bool na_eq
   return new_dictionary_opts(x, &opts);
 }
 
+static inline uint32_t dict_key_size(SEXP x);
+
 static struct dictionary* new_dictionary_opts(SEXP x, struct dictionary_opts* opts) {
   int nprot = 0;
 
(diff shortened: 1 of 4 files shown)


DavisVaughan:
Another thing of note: I think that even if we switch over to sort based functions, we will still need a hash based equivalent of `vec_duplicate_id()` for lists, since that is used in `vec_proxy_order.list()` to return an integer proxy for ordering lists by appearance. So it is nice that both this and `vec_order_radix()` will work with vectors up to `INT_MAX` in size.

src/dictionary.c
```diff
@@ -160,6 +158,24 @@ void dict_put(struct dictionary* d, uint32_t hash, R_len_t i) {
...
+// of at most 77%. We round up to power of 2 to ensure quadratic probing
+// strategy works. Maximum power of 2 we can store in a uint32_t is 2^31,
+// as 2^32 is 1 greater than the max uint32_t value, so we clamp sizes that
+// would result in 2^32 to INT_MAX to ensure that our maximum ceiling value
+// is only 2^31. This will increase the load factor above 77% for `x` with
+// length greater than 1653562409 (2147483648 * .77), but it ensures that
+// it can run.
+static inline
+uint32_t dict_key_size(SEXP x) {
+  uint32_t size = (uint32_t)(vec_size(x) / 0.77);
```
lionel-, line 171: While we are at it we should probably make that code robust to a switch to long vector sizes:

- Check that the division doesn't produce `NaN` by overflow.
- Check that the division result (a double since literal floats are always doubles and integers are promoted to the float type) is representable as a `u32`, i.e. it is equal or less to `2^53`.

src/dictionary.c
```diff
@@ -52,6 +53,8 @@ static struct dictionary* new_dictionary_params(SEXP x, bool partial, bool na_eq
   return new_dictionary_opts(x, &opts);
 }
 
+static inline uint32_t dict_key_size(SEXP x);
```
lionel-, line 56: Move to decl file.

src/dictionary.c
```diff
@@ -13,7 +13,8 @@ struct vctrs_arg args_haystack;
 
 
 // http://graphics.stanford.edu/~seander/bithacks.html#RoundUpPowerOf2
-int32_t ceil2(int32_t x) {
+static inline
+uint32_t ceil2(uint32_t x) {
```
lionel-, line 17: Rename to `u32_ceil2()`?

lionel- (approved):
Awesome! This makes it very clear that we should be extra careful about the semantics of numeric operations performed in vctrs.

src/dictionary.c
DavisVaughan: For long vectors, I guess the largest size would be the max `ptrdiff_t` value, usually `9223372036854775807`? If you divide that by `.77`, that won't produce a `NaN`. Did you have something else in mind?

> representable as a u32, i.e. it is equal or less to 2^53.

Did you mean `2^32 - 1`? The max u32 value? If we were using u64 hash values, then yea I agree the max double value where all integerish values smaller than it are precisely representable as an integer is `2^53`, but I'm not sure we want that here.

What do we do in the case where we get a long vector that this implementation can't handle? Error?

lionel-: (1) oops you're right long vectors can't be that long. (2) You're also right, I was confused and thinking about u64 while writing u32.

An error seems appropriate. If that becomes a problem we can add a 64 bits hashing code path later on.

src/dictionary.c
```diff
@@ -160,6 +158,37 @@ void dict_put(struct dictionary* d, uint32_t hash, R_len_t i) {
...
+// as 2^32 is 1 greater than the max uint32_t value, so we clamp sizes that
+// would result in 2^32 to INT_MAX to ensure that our maximum ceiling value
+// is only 2^31. This will increase the load factor above 77% for `x` with
+// length greater than 1653562409 (2147483648 * .77), but it ensures that
+// it can run.
+static inline
+uint32_t dict_key_size(SEXP x) {
+  const R_len_t x_size = vec_size(x);
+
+  if (x_size > R_LEN_T_MAX) {
```
lionel-, line 173: Needs a comment that this is in prevision of a switch to long vector sizes.

src/dictionary.c
```diff
@@ -160,6 +158,37 @@ void dict_put(struct dictionary* d, uint32_t hash, R_len_t i) {
...
+    stop_internal("dict_key_size", "Dictionary functions do not support long vectors.");
```
lionel-, line 174: `r_stop_internal()`

src/dictionary.c
```diff
@@ -160,6 +158,37 @@ void dict_put(struct dictionary* d, uint32_t hash, R_len_t i) {
...
+  }
+
+  const double load_adjusted_size = x_size / 0.77;
```
lionel-, line 177: I would not add `const` keywords by default.

src/dictionary.c
```diff
@@ -160,6 +158,37 @@ void dict_put(struct dictionary* d, uint32_t hash, R_len_t i) {
...
+
+  if (load_adjusted_size > UINT32_MAX) {
+    stop_internal("dict_key_size", "Can't safely cast load adjusted size to a `uint32_t`.");
+  }
+
+  uint32_t size = (uint32_t)load_adjusted_size;
+  size = size > INT_MAX ? INT_MAX : size;
```
lionel-, line 184: As mentioned on Slack we should also check that the final `size` is larger than the length of the vector to be defensive. Open addressing doesn't support load factors > 100%.

src/dictionary.c
```diff
@@ -7,13 +7,16 @@
...
+
 // Initialised at load time
 struct vctrs_arg args_needles;
 struct vctrs_arg args_haystack;
 
 
 // http://graphics.stanford.edu/~seander/bithacks.html#RoundUpPowerOf2
-int32_t ceil2(int32_t x) {
+static inline
+uint32_t u32_ceil2(uint32_t x) {
```
lionel-, line 19: An overflow check might also be nice here, i.e. `x != 0` before returning. This will overflow when `x` is greater than `2^31`. Should not happen with short vector sizes but better be defensive?

src/dictionary.c
DavisVaughan: In the moment it felt a bit clearer here, but generally I think I agree with you

src/dictionary.c
DavisVaughan: Oh neat!

I've added the check and a few comments about it

I've also ensured that if the input was `0`, it returns `2^0 = 1`. Previously it always returned `0` for that case, but that isn't a power of 2 and would be caught by the error check.

src/dictionary.c
DavisVaughan: The point of the clamping to `INT_MAX` is to ensure that we can't exceed a load factor of 100%.

This will all have to be updated when we change to long vectors anyways, and the `x_size > R_LEN_T_MAX` check will catch the long vectors before they cause any issues here.

Would a comment saying that this will need to be adjusted for long vectors suffice?

src/dictionary.c
DavisVaughan: b1f2354356b8a7cd31f2084cc9d1e9c3591fc241

src/dictionary.c
DavisVaughan: 91b0a9327b735779a1ad937fe9a45713369deb20

src/dictionary.c
DavisVaughan: 04e28ad8f52972f479f7dad31f3061337e13c8f9

src/dictionary.c
DavisVaughan: I went ahead and added the check e560cdf661b997be51e50294d0d0f910a63a9b52

lionel-: > The point of the clamping to INT_MAX is to ensure that we can't exceed a load factor of 100%.

Reducing the adjusted size will only increase the worst case load factor. Aren't we clamping to INT_MAX to avoid `ceil2()` overflow? It would be useful to have an inline comment about this here as well.

```c
if (x_size > size) { r_stop_internal("key size must be at least as large as x"); }
```

That sounds good thanks! Though I'd use "hash table size" rather than "key size".

```c
if (x_size > size) {
  r_stop_internal(, "Hash table size must be at least as large as input.");
}
```

PR #1375, opened 2021-04-27
Formally switch to radix ordering in `vec_order()`

### LINKED ISSUE #1302: Export vec_order_radix()
I noticed there was an internal function `vec_order_radix()` - is this going to be exported at some point, or is it remaining internal?

DavisVaughan:
Closes #1302 

- Renames `vec_order()` to new `vec_order_base()`
- Renames `vec_order_radix()` to `vec_order()`
- Folds `order-radix.R` into `order.R` (similar with tests, and C side)
- Big NEWS bullet outlining the changes
  - One of which is that I've added `...` to both `vec_order()` and `vec_sort()`, which feels appropriate given that we have so many optional arguments to them now, and could potentially add more.

Will likely wait quite a while to merge this PR, as it will immediately affect dplyr and slider


R/order.R
```diff
@@ -1,80 +1,274 @@
 #' Order and sort vectors
 #'
+#' @description
+#' `vec_order()` computes the order of `x`. For data frames, the order is
+#' computed along the rows by computing the order of the first column and
+#' using subsequent columns to break ties.
+#'
+#' `vec_sort()` sorts `x` by computing its order and using `vec_slice()` to
```
lionel-, line 8:
```suggestion
#' `vec_sort(x)` sorts `x`. It is equivalent to `vec_slice(x, vec_order(x))`.
```

R/order.R
```diff
@@ -1,80 +1,274 @@
...
+#'
+#' @description
+#' `vec_order_base()` orders vectors using [base::order()], but can handle
+#' more complex types, like data frames and [`vctrs_vctr`][vctr] objects, using
+#' vctrs principles.
+#'
+#' `vec_order_base()` is mainly provided for backwards compatibility with vctrs
+#' <= 0.3.7. New code should instead use [vec_order()], which has more
+#' capabilities. The main difference between the two is that `vec_order()`
+#' orders character vectors in the C locale (which is highly performance), while
```
lionel-, line 216:
```suggestion
#' orders character vectors in the C locale (which is highly performant), while
```

R/order.R
```diff
@@ -1,80 +1,274 @@
...
+                           chr_ordered = TRUE) {
+  if (!missing(...)) {
+    ellipsis::check_dots_empty()
+  }
+  .Call(vctrs_order_info, x, direction, na_value, nan_distinct, chr_transform, chr_ordered)
+}
+
+# ------------------------------------------------------------------------------
+
+#' Order vectors
```
lionel-, line 206:
```suggestion
#' Order vectors with base compatibility
```

lionel- (approved):
Since we are changing the default I think we should consider a C sort variant with letters swapped so that it's closer to POSIX sorting. It's probably nearly as efficient and will sort more usefully with English datasets (and other alphbetical languages).

lionel-:
Eventually we'll have character classes that attach a locale to a vector. Compared to sorting with the current locale, these will have the advantage of producing a consistent sort across platforms. The most obvious way to implement this class is with an order proxy.

So I'm wondering what is the interaction between `chr_transform`, which is presented as a way of setting a locale for sorting, and the order proxy? I think it would be helpful to clarify this and maybe implement a POC localised character class before cementing the design.

hadley:
We're going to argue that this change increases consistency with stringr and data.table; I think using some custom sort would be a step in the wrong direction.

DavisVaughan:
Superseded by #1435 which is a clean rebase of this on master

Issue #1304, opened 2020-12-08
vec_equal_na() doesn't work for bit64


mgirlich:
and consequently `vec_fill_missing()` also doesn't work

``` r
library(vctrs)
int <- c(1L, NA)
bit64 <- bit64::as.integer64(int)

vec_equal_na(int)
#> [1] FALSE  TRUE
vec_equal_na(bit64)
#> [1] FALSE FALSE

vec_fill_missing(int)
#> [1] 1 1
vec_fill_missing(bit64)
#> integer64
#> [1] 1    <NA>
```

<sup>Created on 2020-12-08 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

PR #1378, opened 2021-05-06
Revise integer64 proxy

DavisVaughan:
Closes https://github.com/r-lib/vctrs/issues/1304

This PR revises the proxy used for integer64 objects. It now uses a two column data frame of double vectors. The first one holds the "left" 32 bits and the second holds the "right" 32 bits.

The construction of the data frame tries to be a bit clever. To perform bit shifting, we have to convert the `int64_t` to a `uint64_t` at some point, because bit shifting isn't defined for signed types. When converting to `uint64_t`, we do so in such a way that we map `INT64_MIN -> 0` and `INT64_MAX -> UINT64_MAX` through a little arithmetic, relying on unsigned integer overflow. This has the very nice side effect of ensuring that the `left` and `right` 32 bits that we extract from the result maintain the same ordering as the original `x` when compared lexicographically, as vctrs does with data frames (Note that I am about 90% sure this is correct everywhere, but would appreciate thoughts on this method). (These ideas originally came from https://github.com/DavisVaughan/tidy64/blob/master/src/pack.h, but I was able to simplify that even further).

For arrays of integer64, we convert to a data frame ahead of time, and then reapply the proxy method, which will iterate over each integer64 column and proxy each one, resulting in df-cols that eventually get nicely flattened back out.

I've had to add another custom array -> data frame method, since `as.data.frame2()` won't work here. It unclasses the input, which we don't actually want here. We want each column of the resulting data frame to be integer64.

Note that there is no need to have a `vec_proxy()` method, as slicing the double storage of an integer64 generally works fine

``` r
# Behavior around 0
x <- bit64::as.integer64(c(-1, 0, 1))

vctrs:::integer64_proxy(x)
#>         left      right
#> 1 2147483647 4294967295
#> 2 2147483648          0
#> 3 2147483648          1

vctrs:::integer64_restore(vctrs:::integer64_proxy(x))
#> integer64
#> [1] -1 0  1


# Largest
x <- bit64::as.integer64("9223372036854775807") + -1:0

# 4294967295 = UINT32_MAX
vctrs:::integer64_proxy(x)
#>         left      right
#> 1 4294967295 4294967294
#> 2 4294967295 4294967295


# Smallest
x <- bit64::as.integer64("-9223372036854775807") + 0:1

vctrs:::integer64_proxy(x)
#>   left right
#> 1    0     1
#> 2    0     2
```

<sup>Created on 2021-05-07 by the [reprex package](https://reprex.tidyverse.org) (v1.0.0)</sup>


DavisVaughan:
Should also fix `vec_compare()` with integer64. Currently we take the rank and use that as the proxy, but you can't take the rank separately for each vector and them compare them.

``` r
vctrs::vec_compare(bit64::as.integer64(1), bit64::as.integer64(2))
#> [1] 0
```

<sup>Created on 2021-06-02 by the [reprex package](https://reprex.tidyverse.org) (v2.0.0)</sup>

src/type-integer64.c
```diff
@@ -0,0 +1,155 @@
...
+  r_obj* left = r_list_get(x, INTEGER64_PROXY_DF_LOCS_left);
+  r_obj* right = r_list_get(x, INTEGER64_PROXY_DF_LOCS_right);
+
+  const double* v_left = r_dbl_cbegin(left);
+  const double* v_right = r_dbl_cbegin(right);
+
+  r_ssize size = r_length(left);
+
+  r_obj* out = KEEP(r_alloc_double(size));
+  long long* v_out = (long long*) r_dbl_begin(out);
```
lionel-, line 94: Also a comment that refers to the comment above about UB.

lionel- (approved):
> Note that there is no need to have a vec_proxy() method, as slicing the double storage of an integer64 generally works fine

Makes sense. In general I've been thinking that the "vector" proxy of a type is different than the "arithmetic" proxy, and it would make sense to have both generics.

PR #1385, opened 2021-06-04
Fix memory leak of strings_posixt

### LINKED ISSUE #1384: Memory leak of `strings_posixt`.
There is a typo in the assignment into `strings` on this line:
https://github.com/r-lib/vctrs/blob/52157b7cc172105404b59d7c4d85802ed79be0f0/src/utils.c#L1670

Instead of `SET_STRING_ELT(strings, strings_posixlt)` there should be `SET_STRING_ELT(strings, strings_posixt)`.

On FastR, this typo causes memory issues as `strings_posixt` is unprotected and not referenced from any other object, therefore, it is eligible for garbage collection. On GNU-R, I am not able to reproduce this issue even with `gctorture()` - it seems that GNU-R has a different policy for garbage-collecting CHARSXP objects.

Akirathan:
This PR fixes #1384 - correctly insert `strings_posixt` into `strings` so that it is protected against garbage collection.

### DIFF
--- a/src/utils.c
+++ b/src/utils.c
@@ -1667,7 +1667,7 @@ void vctrs_init_utils(SEXP ns) {
   SET_STRING_ELT(strings, 4, strings_posixlt);
 
   strings_posixt = Rf_mkChar("POSIXt");
-  SET_STRING_ELT(strings, 5, strings_posixlt);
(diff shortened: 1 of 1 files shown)


lionel-:
Thanks!

PR #1402, opened 2021-06-22
Fix OOB detection in `r__intmax_subtract()`

### LINKED ISSUE #1399: Incorrect overflow/underflow detection in `intmax_subtract()`
https://github.com/r-lib/vctrs/blob/81e32472a575a212539bdc9f6b8447e429635f29/src/utils.h#L409

The second check here should be:

```c
(y < 0 && x > (INTMAX_MAX + y))) { 
            ^ right here
```

Which would align with the C CERT standards recommendations
https://wiki.sei.cmu.edu/confluence/display/c/INT32-C.+Ensure+that+operations+on+signed+integers+do+not+result+in+overflow

```c
#include <limits.h>
  
void func(signed int si_a, signed int si_b) {
  signed int diff;
  if ((si_b > 0 && si_a < INT_MIN + si_b) ||
      (si_b < 0 && si_a > INT_MAX + si_b)) {
    /* Handle error */
  } else {
    diff = si_a - si_b;
  }
 
  /* ... */
}
```

DavisVaughan:
Closes #1399 

Also removes our current intmax utilities in favor of the rlang ones

### DIFF
--- a/src/order-radix.c
+++ b/src/order-radix.c
@@ -1268,7 +1268,7 @@ void int_compute_range(const int* p_x,
    * - We need to go up to `intmax_t` to avoid intermediate overflow.
    * - `+ 1` to get an inclusive range on both ends.
    */
-  range = (uint32_t) intmax_add(intmax_subtract(x_max, x_min), 1);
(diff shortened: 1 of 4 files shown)


lionel-:
Can you fix upstream first please @DavisVaughan?

PR #1404, opened 2021-06-23
Revise `vec_detect_complete()` to treat rcrds like data frames

### LINKED ISSUE #1386: `vec_match()` should only use incomplete propagation with `"data.frame"`s
i.e. with rcrd types that have a data frame proxy, it should require fully missing rows to be considered missing

```r
> x <- new_rcrd(list(x = c(1, NA), y = c(NA_real_, NA_real_)))
> vec_match(x, x, na_equal = FALSE)
[1] NA NA
> vec_detect_complete(x)
[1]  TRUE FALSE
```

In this case, element 1 should still match, while element 2 should propagate an NA

This probably requires calling `vec_detect_complete()` up front on `needles` in the `na_equal = FALSE` case. Then replacing this `dict_is_missing()` call with `if (!v_complete[i])` https://github.com/r-lib/vctrs/blob/173efaf201b23f1a2e36d00b3b9ea4d77912f489/src/dictionary.c#L442

DavisVaughan:
Closes #1386 

This PR aligns the way `vec_detect_complete()` computes completeness for rcrds and data frames. Previously:

For data frames:
- A row is missing if all columns are missing
- A row is incomplete if any columns are missing

For rcrd data frame proxies:
- A row is missing if all columns are missing
- A row is incomplete if all columns are missing

Now, both consider a row as incomplete if _any_ columns are missing.

This makes C level coding around these conventions a bit easier, in particular with functions like `vec_match()` which try to detect completeness on the fly by iterating across the columns of the equality proxy looking to see if any columns are missing. Since rcrd types aren't treated specially anymore, the way this works now aligns with `vec_detect_complete()`.

I believe this also makes a bit more sense in general. I'd argue that the definition of "missing" is what you get from `vec_init(x)`, but the definition of "incomplete" is when any component of your observation has missing information. For example:

- From the vignette, the rational rcrd class has a numerator and denominator. If either one is missing, I’d argue that you have an incomplete record.
- A similar argument can be made for the decimal2 rcrd class, which stores the numbers to the left and right of the decimal as separate integers.
- In base R, the complex type is considered incomplete (by complete.cases()) if either the real or imaginary part is NA.
- I’d argue that if any component of the POSIXlt list is missing, you have an incomplete record because you can’t tell what time it actually is.
- I could make similar arguments for clock types (although in clock I worked hard to ensure that if any field was set to NA, then all were, so I didn’t have to deal with this)

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,9 @@
 # vctrs (development version)
 
+* `vec_detect_complete()` now computes completeness for `vctrs_rcrd` types in
+  the same way as data frames, which means that if any field is missing, the
+  entire record is considered incomplete (#1386).
+  
 * The `na_value` argument of `vec_order()` and `vec_sort()` now correctly
   respect missing values in lists (#1401).
 
--- a/R/complete.R
+++ b/R/complete.R
@@ -10,8 +10,8 @@
 #' rows that are partially complete (they have at least one non-missing value).
 #'
 #' @details
-#' A [record][new_rcrd] type vector is considered complete if any field is
-#' non-missing.
+#' A [record][new_rcrd] type vector is similar to a data frame, and is only
+#' considered complete if all fields are non-missing.
 #'
 #' @param x A vector
 #'
@@ -49,8 +49,4 @@ vec_slice_complete <- function(x) {
 
 vec_locate_complete <- function(x) {
   .Call(vctrs_locate_complete, x)
-}
-
-vec_proxy_complete <- function(x) {
-  .Call(vctrs_proxy_complete, x)
 }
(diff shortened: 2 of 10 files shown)


lionel- (approved):
> in clock I worked hard to ensure that if any field was set to NA, then all were, so I didn’t have to deal with this

If you had to go out of your way to work around the original design then it's a very good signal that the design wasn't correct.

DavisVaughan:
The other reason was to ensure that if someone sets one component to `NA`, then extracting a different component will contain that propagated `NA`.

```r
library(clock)

x <- year_month_day(2019, 1, 1)
x <- set_year(x, NA)
get_month(x)
#> [1] NA
```

PR #1406, opened 2021-06-24
Ensure that missing complex values are always grouped together

### LINKED ISSUE #1403: Proposal for `vec_proxy_order.complex()`
I believe that we need `vec_proxy_order.complex()` to ensure that "missing" complex values are always pushed to the front or back of the ordering results. Note that a complex value is considered missing if either the real or imaginary part is `NA` or `NaN`.

Currently, `vec_order_radix()` treats a complex vector like a data frame of two double vectors. This poses an issue because the missingness isn't considered jointly. It becomes possible for `1+1i` to be sorted before `2+NAi`, since the 1 is ordered before the 2 before missingness is considered. This is an issue for algorithms that expect missing / incomplete values (as returned by `vec_detect_complete()`) to be "first" after ordering with `na_value = "smallest"`.

I am proposing a `vec_proxy_order()` method for complex that would normalize the missingness up front so that if either the real or imaginary part contain a missing value, then both parts should contain the same missing value.

For `nan_distinct = TRUE` combined with the case where you have something like `NaN+NAi` or `NA+NaNi`, we would follow base R's lead of letting `NA` be more infectious than `NaN` - so those two values would both become `NA+NAi`.

Update: It is also possible that I could probably handle this in the complex type path directly inside vec-order-radix, without the need for a proxy.

``` r
library(vctrs)

x <- complex(
  real = c(NaN, NA, NA, NA, NaN, NaN, 1, 1, 1, 2),
  imaginary = c(NA, NA, NaN, 1, NaN, 1, NA, NaN, 1, NA)
)

# A little easier to see in a df
df <- data_frame(x = x, real = Re(x), imag = Im(x))
df
#>           x real imag
#> 1        NA  NaN   NA
#> 2        NA   NA   NA
#> 3        NA   NA  NaN
#> 4        NA   NA    1
#> 5  NaN+NaNi  NaN  NaN
#> 6  NaN+  1i  NaN    1
#> 7        NA    1   NA
#> 8    1+NaNi    1  NaN
#> 9    1+  1i    1    1
#> 10       NA    2   NA

# Note that if NA or NaN is present anywhere, it is considered missing
is.na(x)
#>  [1]  TRUE  TRUE  TRUE  TRUE  TRUE  TRUE  TRUE  TRUE FALSE  TRUE

# All missing obs are returned in appearance order, then obs without missings
# NOTE: 2+NAi is moved up before 1+1i
vec_slice(df, order(x, na.last = FALSE))
#>           x real imag
#> 1        NA  NaN   NA
#> 2        NA   NA   NA
#> 3        NA   NA  NaN
#> 4        NA   NA    1
#> 5  NaN+NaNi  NaN  NaN
#> 6  NaN+  1i  NaN    1
#> 7        NA    1   NA
#> 8    1+NaNi    1  NaN
#> 9        NA    2   NA
#> 10   1+  1i    1    1

# For both real and imag it puts NA first, followed by NaN, then real numbers
vec_slice(df, vec_order_radix(x, na_value = "smallest", nan_distinct = TRUE))
#>           x real imag
#> 1        NA   NA   NA
#> 2        NA   NA  NaN
#> 3        NA   NA    1
#> 4        NA  NaN   NA
#> 5  NaN+NaNi  NaN  NaN
#> 6  NaN+  1i  NaN    1
#> 7        NA    1   NA
#> 8    1+NaNi    1  NaN
#> 9    1+  1i    1    1
#> 10       NA    2   NA

# For both real and imag it puts NA/NaN first (with no distinction), then real numbers
# NOTE: This is problematic because `2+NAi` ends up after `1+1i` even though we requested NAs to be smallest
vec_slice(df, vec_order_radix(x, na_value = "smallest", nan_distinct = FALSE))
#>           x real imag
#> 1        NA  NaN   NA
#> 2        NA   NA   NA
#> 3        NA   NA  NaN
#> 4  NaN+NaNi  NaN  NaN
#> 5        NA   NA    1
#> 6  NaN+  1i  NaN    1
#> 7        NA    1   NA
#> 8    1+NaNi    1  NaN
#> 9    1+  1i    1    1
#> 10       NA    2   NA

vec_proxy_order.complex <- function(x, ...) {
  if (!anyNA(x)) {
    return(x)
  }
  
  real <- Re(x)
  imag <- Im(x)
  
  # Propagate infectious NA bidirectionally first
  na <- (is.na(real) & !is.nan(real)) | (is.na(imag) & !is.nan(imag))
  if (any(na)) {
    real[na] <- NA_real_
    imag[na] <- NA_real_
  }
  
  # Then bidirectionally propagate remaining NaN that
  # may have survived after NA propagation
  nan <- is.nan(real) | is.nan(imag)
  if (any(nan)) {
    real[nan] <- NaN
    imag[nan] <- NaN
  }
  
  complex(real = real, imaginary = imag)
}

# That makes these the same
identical(
  vec_slice(df, vec_order_radix(x, na_value = "smallest", nan_distinct = FALSE)),
  vec_slice(df, order(x, na.last = FALSE))
)
#> [1] TRUE

# When NaN and NA are present and `nan_distinct = TRUE`:
# - If NA is present, it is infectious and "wins" over anything else. These are put first in appearance order.
# - If NaN is present after NA propagation, they go next in appearance order.
# - Then non-missing complex numbers are returned.
vec_slice(df, vec_order_radix(x, na_value = "smallest", nan_distinct = TRUE))
#>           x real imag
#> 1        NA  NaN   NA
#> 2        NA   NA   NA
#> 3        NA   NA  NaN
#> 4        NA   NA    1
#> 5        NA    1   NA
#> 6        NA    2   NA
#> 7  NaN+NaNi  NaN  NaN
#> 8  NaN+  1i  NaN    1
#> 9    1+NaNi    1  NaN
#> 10   1+  1i    1    1
```

DavisVaughan:
Closes #1403 

Rather than adding a new proxy method, I've fixed this directly in `vec_order_radix()`, as I think that is more appropriate.

In #1403 I mentioned:

> For `nan_distinct = TRUE` combined with the case where you have something like `NaN+NAi` or `NA+NaNi`, we would follow base R's lead of letting `NA` be more infectious than `NaN` - so those two values would both become `NA+NAi`.

I have changed my mind since then. I'm now allowing `NA+NAi`, `NA+NaNi`, `NaN+NAi` and `NaN+NaNi` to all be distinct values, as I think that aligns better with what that argument is generally trying to achieve anyways. I don't think this is that important of a distinction, the main thing is that all 4 types of missingness get grouped together either before any or after all of the non-missing values have appeared.

### DIFF
--- a/src/order-radix.c
+++ b/src/order-radix.c
@@ -2623,6 +2623,9 @@ uint8_t dbl_extract_uint64_byte(uint64_t x, uint8_t shift) {
 
 // -----------------------------------------------------------------------------
 
+static inline
+r_complex_t cpl_normalise(r_complex_t x);
+
 /*
  * `cpl_order()` uses the fact that Rcomplex is really just a rcrd
  * type of two double vectors. It orders first on the real vector, and then on
@@ -2667,7 +2670,7 @@ void cpl_order(SEXP x,
 
   // Handle the real portion first
   for (r_ssize i = 0; i < size; ++i) {
-    p_x_chunk_dbl[i] = p_x_cpl[i].r;
+    p_x_chunk_dbl[i] = cpl_normalise(p_x_cpl[i]).r;
   }
 
   /*
@@ -2720,7 +2723,7 @@ void cpl_order(SEXP x,
   // Uses updated ordering to place it in sequential order.
   for (r_ssize i = 0; i < size; ++i) {
     const int loc = p_o[i] - 1;
-    p_x_chunk_dbl[i] = p_x_cpl[loc].i;
+    p_x_chunk_dbl[i] = cpl_normalise(p_x_cpl[loc]).i;
   }
 
   // Iterate over the group chunks from the first pass
@@ -2752,6 +2755,47 @@ void cpl_order(SEXP x,
     p_x_chunk_dbl += group_size;
     p_o += group_size;
   }
+}
+
+/*
+ * Normalises a complex value so that if one side is missing, both are. This
+ * ensures that all missing complex values are grouped together, no matter
+ * what type of missingness it is. NA and NaN can still be separated by
+ * `nan_distinct`, resulting in 4 different combinations of missingness. These
+ * 4 groups of missingness will still all be grouped together, either before
+ * or after any non-missing values have appeared.
+ * See issue #1403 for more information.
+ */
+static inline
+r_complex_t cpl_normalise(r_complex_t x) {
+  const double na = r_globals.na_dbl;
+  const double nan = R_NaN;
+
+  const enum vctrs_dbl_class r_type = dbl_classify(x.r);
+  const enum vctrs_dbl_class i_type = dbl_classify(x.i);
+
+  switch (r_type) {
+  case vctrs_dbl_number:
+    switch (i_type) {
+    case vctrs_dbl_number: return x;
+    case vctrs_dbl_missing: return (r_complex_t) {na, na};
+    case vctrs_dbl_nan: return (r_complex_t) {nan, nan};
+    }
+  case vctrs_dbl_missing:
+    switch (i_type) {
+    case vctrs_dbl_number: return (r_complex_t) {na, na};
+    case vctrs_dbl_missing: return x;
+    case vctrs_dbl_nan: return x;
+    }
+  case vctrs_dbl_nan:
+    switch (i_type) {
+    case vctrs_dbl_number: return (r_complex_t) {nan, nan};
+    case vctrs_dbl_missing: return x;
+    case vctrs_dbl_nan: return x;
+    }
+  }
+
+  never_reached("cpl_normalise");
 }
 
 // -----------------------------------------------------------------------------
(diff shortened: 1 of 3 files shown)


src/order-radix.c
```diff
@@ -2754,6 +2757,47 @@ void cpl_order(SEXP x,
...
+ * Normalises a complex value so that if one side is missing, both are. This
+ * ensures that all missing complex values are grouped together, no matter
+ * what type of missingness it is. NA and NaN can still be separated by
+ * `nan_distinct`, resulting in 4 different combinations of missingness. These
+ * 4 groups of missingness will still all be grouped together, either before
+ * or after any non-missing values have appeared.
+ * See issue #1403 for more information.
+ */
+static inline
+r_complex_t cpl_normalise(r_complex_t x) {
```
lionel-, line 2770: Maybe `cpl_normalise_na()`? I tend to avoid "normalise"/"standardise" without mentioning the context or standard against which an input is normalised.

lionel- (approved):

PR #1418, opened 2021-07-21
Drop unnused arg

hadley:


lionel- (approved):
The source file for `s3_register()` is now in rlang and has already diverged from the one in vctrs.

PR #1422, opened 2021-08-04
Add a `names<-` method for `vctrs_rcrd`

### LINKED ISSUE #1419: vec_set_names(names = NULL) creates invalid rcrd
``` r
library(vctrs)

rcrd <- new_rcrd(data.frame(a = 1:3))
unnamed_rcrd <- vec_set_names(rcrd, NULL)
field(rcrd, "a")
#> [1] 1 2 3
field(unnamed_rcrd, "a")
#> Error: Corrupt x: no names
waldo::compare(rcrd, unnamed_rcrd)
#> `names(old)` is character ('a')
#> `names(new)` is absent
```

<sup>Created on 2021-07-22 by the [reprex package](https://reprex.tidyverse.org) (v2.0.0.9000)</sup>

Do we need `validate_rcrd()` too, and perhaps `format.rcrd()` that delegates to pillar if it's installed?

DavisVaughan:
Closes #1419 

This is a no-op for `NULL` names, but otherwise errors. We should reconsider this in the future once we are ready to try and support names on rcrds, but for now this prevents `vec_set_names()` from corrupting rcrd types.


lionel- (approved):
oh I thought we had already added that exception!

PR #1428, opened 2021-08-17
Implement `list_drop_empty()`

### LINKED ISSUE #1395: Add `list_compact()`
When dealing with lists it is often necessary to remove `NULL` elements. While `purrr::compact(x)` does the job nicely it is pretty slow. I think something like `list_compact()` would be a great addition to `vctrs`.

It might make sense to allow the user to control whether only `NULL` should be removed and/or other empty elements, e.g.

```r
drop_null <- function(x) {
  vec_slice(x, !vapply(x, is.null, logical(1)))
}

vctrs_compact <- function(x) {
  vec_slice(x, list_sizes(x) != 0)
}

l <- list(a = NULL, b = integer(), c = 1L)
# only remove `NULL`
drop_null(l)

# remove empty elements
vctrs_compact(l)
```

## Benchmarks

``` r
library(purrr)
library(vctrs)

vctrs_compact <- function(x) {
  vec_slice(x, list_sizes(x) != 0)
}

# bigger lists
n1 <- 1e3
l <- vec_rep(list(1, NULL, 2), n1)

bench::mark(
  purrr = compact(l),
  vctrs = vctrs_compact(l)
)
#> # A tibble: 2 x 6
#> 1 purrr        7.23ms   8.02ms      124.   161.5KB    24.3 
#> 2 vctrs       90.47µs 116.08µs     8048.    51.6KB     6.29

# tight loops
l2 <- vec_rep(list(vec_rep(list(1, NULL, 2), 10L)), 1e3)
bench::mark(
  purrr = lapply(l2, compact),
  vctrs = lapply(l2, vctrs_compact)
)
#> # A tibble: 2 x 6
#> 1 purrr       107.5ms 110.77ms      8.80     211KB    19.4 
#> 2 vctrs        2.98ms   3.35ms    281.       211KB     9.98
```

<sup>Created on 2021-06-18 by the [reprex package](https://reprex.tidyverse.org) (v2.0.0)</sup>

DavisVaughan:
Closes #1395 

It is slightly more complex than you might expect because I make an effort to return `x` unchanged if there is nothing to compact (which I feel could be a somewhat common case).

I've added a `drop = c("missing", "empty")` argument that allows you to decide whether to just drop `NULL`s or to drop all size 0 elements.

``` r
library(purrr)
library(vctrs)

vctrs_compact <- function(x) {
  vec_slice(x, list_sizes(x) != 0)
}

n <- 1e3

# 3000 elements
l <- vec_rep(list(1, NULL, 2), n)

bench::mark(
  purrr = compact(l),
  vctrs_r = vctrs_compact(l),
  list_compact1 = list_compact(l, drop = "missing"),
  list_compact2 = list_compact(l, drop = "empty")
)
#> # A tibble: 4 × 6
#> 1 purrr           8.19ms   8.91ms      111.   132.9KB   325.  
#> 2 vctrs_r        87.57µs 108.65µs     8765.    59.8KB     6.25
#> 3 list_compact1  19.86µs  24.09µs    40319.    52.4KB    24.2 
#> 4 list_compact2  77.44µs   89.9µs    10749.    35.3KB     8.41

# no missing elements
l <- vec_rep(list(1, 2, 2), n)

bench::mark(
  purrr = compact(l),
  vctrs_r = vctrs_compact(l),
  list_compact1 = list_compact(l, drop = "missing"),
  list_compact2 = list_compact(l, drop = "empty")
)
#> # A tibble: 4 × 6
#> 1 purrr           6.55ms   7.59ms      133.    94.1KB    21.4 
#> 2 vctrs_r        88.13µs  96.65µs     9693.    58.8KB     8.59
#> 3 list_compact1   2.56µs   2.85µs   333328.        0B     0   
#> 4 list_compact2  56.39µs  57.54µs    16552.        0B     0
```

<sup>Created on 2021-08-17 by the [reprex package](https://reprex.tidyverse.org) (v2.0.0.9000)</sup>


R/compact.R
```diff
@@ -0,0 +1,33 @@
...
+#' @examples
+#' x <- list(1, NULL, integer(), 2)
+#'
+#' # By default, only `NULL` elements are dropped
+#' list_compact(x)
+#'
+#' # You can also drop any empty element
+#' list_compact(x, drop = "empty")
+list_compact <- function(x, ..., drop = c("missing", "empty")) {
+  if (!missing(...)) {
```
lionel-, line 26: Since we depend on dev rlang this can be `check_dots_empty0(...)`.

lionel- (commented):
I haven't reviewed the implementation yet but I wonder if that should be `vec_compact()` to generically remove missing elements. That would fit vctrs better and also provide an equivalent for `na.omit()`.

`drop` could take the values `"missing"` and `"incomplete"`.

If we go that route, `vec_compact()` becomes recursive over data frames. That is seems powerful but this also poses some questions.

- What is the meaning of incomplete in the recursive case? I guess any incomplete row in any packed data frame would qualify. 

- Would we want to allow `drop = "empty"` to remove any row that has an empty vector stored in a list-column?

- Then probably `drop` is an inclusive list, e.g. `drop = c("incomplete", "empty")` makes sense.

DavisVaughan:
I wonder if this is really 3 functions which we can make consistent with our current definitions of missing vs incomplete

1. `vec_drop_missing()` for dropping missing values
   - Drops only `NULL` elements in lists
   - Drops row if _all_ columns of that row are empty in data frames
   - Like `na.omit()`, which relies on `is.na()` for each col of the data frame

    ```r
    # Alias for consistency
    vec_detect_missing <- vec_equal_na
    
    vec_drop_missing <- function(x) {
      vec_slice(x, vec_detect_missing(x))
    }
    ```

2. `vec_drop_incomplete()` for dropping incomplete values
    - Drops only `NULL` elements in lists
    - Drops row if _any_ columns of that row are empty in data frames

    ```r
    vec_drop_incomplete <- function(x) {
      vec_slice(x, !vec_detect_complete(x))
    }
    ```

3. `list_compact()` / `list_drop_empty()` for dropping empty values from a list
    - Drops both `NULL` and empty elements like `integer()` from lists
    - I don't think dropping empty values makes much sense for anything besides a list
    - Theoretically works nicely because it uses the similarly named `list_sizes()`

    ```r
    list_compact <- function(x) {
      vec_slice(x, list_sizes(x) != 0)
    }
    ```

---

> Would we want to allow `drop = "empty"` to remove any row that has an empty vector stored in a list-column?

This is the one case not covered by the above set of functions, but I don't really see it being that common. If a user needed something like this, they could do:

```r
vec_slice(df, list_sizes(df$list_col) != 0L)
```

They could extend this to multiple columns as needed, and could maybe use `vec_proxy_equal()` to flatten df-cols if they really need to be fully generic.

I'm also not sure I like the idea of dropping any row that has an empty vector in a list-col. Like, the `2` in the `x` column isn't "empty", so it feels weird if something like `vec_compact()` were to drop that row because the value in `y` made the row partially empty.

``` r
library(tibble)

df <- tibble(
  x = c(1, NA, 2),
  y = list(NULL, 1, integer())
)

df
#> # A tibble: 3 × 2
#>       x y        
#>   <dbl> <list>   
#> 1     1 <NULL>   
#> 2    NA <dbl [1]>
#> 3     2 <int [0]>
```

lionel-:
I think you're right that it is better to have more functions than complex interactions of semantics in fewer functions. I like the 3 suggested functions and their names, especially `list_drop_empty()` which makes it very explicit what is going on. My only small concern is that it might be a bit non-obvious for a user to reach for `vec_drop_missing()` to compact `NULL` elements of a list. That is consistent with vctrs naming though.

@hadley What do you think of Davis' proposal?

hadley:
My sense is that it's better to keep the "list compaction" and "dropping missing values" pieces separate. I don't have any sense of the need for `vec_drop_missing()`/`vec_drop_incomplete()` except that I don't think you should implement unless you have seen a clear need for them (i.e. not just as a speculative API completion exercise).

I think improving the performance of `compact()` and giving the ability to drop either empty or NULL components would be very useful. But it's not clear to me that should live in vctrs, since we use it a lot for programming. I wonder if it should actually live in rlang, replacing the `compact()` implementation in `purrr-compat.R`. And then `purrr::compact()` would just call this function. 

This is tangled up with our need for a very small subset of purrr in many packages. It feels a bit weird to add a single list manipulation function to rlang, but maybe we could consider a few others that time have shown to be widely needed? Off the top of my head I can think of `list_modify()`, but maybe there are others?

Alternatively, maybe the way forward is to put more of mini-purrr into vctrs (i.e. efficiently implement all of the functions in `purrr-compat.R`) and then add vctrs to `use_tidy_dependencies()`?

lionel-:
My sense is that since `na.omit()` is an important R function it makes sense to have a vctrs equivalent. I sometimes reach for this function in my package programming and I remember being wary about the consequences of carrying the attribute attached to the result to unexpected places.

Also I think speculative API design is important in vctrs in order to ensure consistency. The work Davis has done in the vec-prefix document is useful in that regard. We don't have to implement all of these now but it helps to give a sense of what the API should look like.

I am a bit worried about implementing  a fast`list_drop_empty()` or `list_compact()` in rlang because we'll still need a vctrs variant for genericity. Unless we use `[` for genericity, in which case we only need a simple R wrapper that can live in utils.R or compat-purrr.R.

> Alternatively, maybe the way forward is to put more of mini-purrr into vctrs (i.e. efficiently implement all of the functions in purrr-compat.R) and then add vctrs to use_tidy_dependencies()?

I think so. We have several ongoing PRs/experiments to reimplement purrr functions in vctrs.

DavisVaughan:
I don't think that a low level `compact()` that drops empty values could live in rlang, since we determine emptiness by `vec_size(elt) == 0`

DavisVaughan:
I've reworked this PR to just implement `list_drop_empty()`, as that seems like the most important case right now for tidyr/purrr. I've opened #1449 to revisit `vec_drop_missing()` and `vec_drop_incomplete()` / `vec_slice_complete()` in the future if we feel like they merit an implementation

R/empty.R
```diff
@@ -0,0 +1,15 @@
+#' Drop empty elements from a list
+#'
+#' `list_drop_empty()` removes empty elements from a list. This includes `NULL`
+#' elements along with empty vectors, like `integer(0)`. This is equivalent to,
+#' but faster than, `vec_slice(x, list_sizes(x) != 0L)`.
+#'
+#' @param x A list.
+#'
```
lionel-, line 8: `vec_slice()` dependency

lionel- (approved):

PR #1430, opened 2021-08-18
Update to testthat 3e

DavisVaughan:
This PR updates vctrs to use testthat 3e

- `verify_output()` has become `expect_snapshot()`
- `expect_known_output()` has become `expect_snapshot()`
- Our stub for `verify_errors()` has become `expect_snapshot()`
  - I merged the cases where we had 2 separate tests for testing output vs testing error classes into one test block
- Removed now unused `try2()` and `try_cat()` and `expect_known_output_nobang()`

The nice thing about this is that our `testthat/` directory is simplified to just `.R` test files and a `_snaps/` folder. No more free floating text files.

---

This PR is a little incomplete. When `check()` runs on testthat 3e, messages are now printed out. This prints a ton of name repair messages that are not currently being handled, and we need https://github.com/r-lib/rlang/pull/1249 to handle them. We also need to convert `describe_repair()` to use `rlang::names_inform_repair()`, but that is going to cause a ton of merge conflicts with this PR because the rlang name repair message is _slightly_ different than the vctrs one, so all the snaps need to be updated.

So I propose merging this as-is, and then I'll immediately do PRs to:
- Use `names_inform_repair()` in `describe_repair()` and update all snaps for the changes
- Silence the name repair messages in the tests by using `local_options(rlib_name_repair_verbosity = "quiet")`

---

@lionel- I don't think you need to do a really in depth review. Most of the commits are doing the same thing, just in different files, so if you take a look at a few of them you should be able to grasp the general idea of the changes.


lionel- (approved):
🎉

PR #1432, opened 2021-08-19
Use `names_inform_repair()` for name repair messages

### LINKED ISSUE #1317: Global option to silence name repair messages

### LINKED ISSUE #1429: Switch `describe_repair()` from using `message()` to `inform()`
So name repair messages look less scary in RStudio (i.e. not in red / orange)

DavisVaughan:
Closes #1429 
Closes #1317 

I've updated the failing tests to use `expect_snapshot()`, which should be a little more forgiving if the rlang implementation changes

### DIFF
--- a/tests/testthat/_snaps/bind.md
+++ b/tests/testthat/_snaps/bind.md
@@ -203,7 +203,7 @@
       # Integers as rows
       suppressMessages(with_memory_prof(vec_rbind_list(ints)))
     Output
-      [1] 22.1KB
+      [1] 2.53KB
     Code
       suppressMessages(with_memory_prof(vec_rbind_list(named_ints)))
     Output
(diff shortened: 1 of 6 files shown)


tests/testthat/_snaps/bind.md
```diff
@@ -181,7 +203,7 @@
       # Integers as rows
       suppressMessages(with_memory_prof(vec_rbind_list(ints)))
     Output
-      [1] 2.53KB
+      [1] 22.1KB
```
DavisVaughan, line 206: This was a little surprising. This is generating 100 name repair messages and shouldn't be a very common case (it generally indicates the user has messed something up), but the code is a lot slower because of this too.

lionel- (approved):

PR #1435, opened 2021-08-20
Switch to radix ordering in `vec_order()`

### LINKED ISSUE #1302: Export vec_order_radix()

DavisVaughan:
Supersedes #1375
Closes #1302 

This is essentially a rebase of #1375 on master, but that got so outdated that it was too difficult to do it directly in that PR.

- Renames `vec_order()` to new `vec_order_base()`
- Renames `vec_order_radix()` to `vec_order()`
- Folds `order-radix.R` into `order.R` (similar with tests, and C side)
- Big NEWS bullet outlining the changes
  - One of which is that I've added `...` to both `vec_order()` and `vec_sort()`, which feels appropriate given that we have so many optional arguments to them now, and could potentially add more.


DavisVaughan:
@lionel- I don't think you need to review this, as you reviewed #1375 already. But I think we are ready to move forward with merging this in, then merging https://github.com/tidyverse/dplyr/pull/5942. Does that sound good to you?

lionel- (commented):
Should we consider renaming the `chr_transform` argument to `locale` for consistency with dplyr? Though I don't mind a slightly more technical name for `vec_order()`.

Should the default be `"en_us"` as in tidyverse packages? Or do we consider vctrs is used primarily for programming purposes and so the fast C locale should be the default?

DavisVaughan:
I think the technical name is more appropriate here, since it really can be any generic transform, like `tolower()`

I think we had landed on using the C locale in vctrs for its performance and to avoid entangling vctrs with stringi in any way, and at this point I'm still in favor of both of those reasons.

lionel- (approved):
🚀

PR #1438, opened 2021-08-25
Silence name repair messages during testing

DavisVaughan:
99% of the cases just needed `local_name_repair_quiet()`. The other 1% weren't really testing name repair, so I added names explicitly to prevent them from being repaired.


lionel-:
I think it would be less verbose and distracting to disable messages at top-level and reenable them locally when testing the messages.

PR #1453, opened 2021-09-21
Always materialize ALTREP vectors in `vec_chop()`

### LINKED ISSUE #1450: Should `vec_chop()` materialize ALTREP vectors?
We typically use `vec_chop()` in two ways:

1. Like `as.list()`, where we turn a vector into a list where each list element holds 1 element of the original vector
2. For chunking groups in `dplyr::group_by()` / `summarise()`

In both of these cases, we are guaranteed to touch every element of the vector.

It is possible to not touch every element, like `vec_chop(1:5, list(1, 2))`, but I don't think I've ever used that.

Currently, `vec_chop()` internally uses `vec_slice()` which slices ALTREP vectors using their `Extract_subset` method if it exists, which can return another ALTREP result (like with vroom). Performance can degrade significantly if we end up with many small ALTREP chunks, as we will eventually have to call `DATAPTR()` on each chunk and materialize it, which could be slow, like in https://github.com/tidyverse/dplyr/issues/6015.

General advice seems to be that you should go ahead and materialize the original ALTREP vector if you know you are going to touch every element of it (https://github.com/tidyverse/dplyr/issues/6015#issuecomment-921084048).

So to avoid these performance issues, I think we should consider having `vec_chop()` always avoid the special ALTREP `vec_slice()` path, and just go straight into the standard path that would call `DATAPTR()` and materialize the full vector (forcing downstream chunks to also be materialized).

DavisVaughan:
Closes #1450

We now signal to `vec_slice_base()` that we would like it to materialize any ALTREP vectors by skipping the special ALTREP handling and instead dropping into the standard slicing behavior, which naturally forces materialization. This has large performance benefits downstream from `vec_chop()`, since manipulating many small ALTREP vectors can have a large amount of overhead.

The original example from tidyverse/dplyr#6015 is below. Notice that the benchmark associated with this PR allocates more memory than before, which makes sense because it will now materialize the ALTREP columns mentioned in the `summarize()` expressions.

``` r
library(readr)
library(dplyr)

url <- url("https://media.githubusercontent.com/media/matanmazor/thesis/main/index/data/RC/RC.csv")

df <- read_csv(url)

gdf <- df %>%
  mutate(
    confidence = confidence / 1000,
    direction = ifelse(direction == 3,1,0)
  ) %>%
  group_by(subj_id, trial_id)

gdf
#> # A tibble: 1,488,000 × 15
#> # Groups:   subj_id, trial_id [24,000]
#>    subj_id trial_id trial_number timepoint detection direction signal response
#>      <dbl>    <dbl>        <dbl>     <dbl>     <dbl>     <dbl>  <dbl>    <dbl>
#>  1       1        1            1         1         1         0      1        0
#>  2       1        1            1         2         1         0      1        0
#>  3       1        1            1         3         1         0      1        0
#>  4       1        1            1         4         1         0      1        0
#>  5       1        1            1         5         1         0      1        0
#>  6       1        1            1         6         1         0      1        0
#>  7       1        1            1         7         1         0      1        0
#>  8       1        1            1         8         1         0      1        0
#>  9       1        1            1         9         1         0      1        0
#> 10       1        1            1        10         1         0      1        0
#> # … with 1,487,990 more rows, and 7 more variables: RT <dbl>, confidence <dbl>,
#> #   correct <dbl>, energyRight <dbl>, energyLeft <dbl>, energyDown <dbl>,
#> #   energyUp <dbl>

bench::mark(
  dplyr_summarise = summarise(
    gdf,
    detection = detection[timepoint == 1],
    RT = RT[timepoint == 1] - 700,
    .groups = "drop"
  )
)

# CRAN vctrs
#> # A tibble: 1 × 6
#> 1 dplyr_summarise    14.9s    14.9s    0.0672    44.4MB    0.134

# This PR
#> # A tibble: 1 × 6
#> 1 dplyr_summarise    172ms    187ms      5.34     103MB     5.34
```

<sup>Created on 2021-09-21 by the [reprex package](https://reprex.tidyverse.org) (v2.0.0.9000)</sup>

### DIFF
--- a/src/slice-chop.c
+++ b/src/slice-chop.c
@@ -192,7 +192,10 @@ static SEXP chop(SEXP x, SEXP indices, struct vctrs_chop_info info) {
     }
 
     // Always materialize ALTREP vectors when chopping to avoid inefficiently
-    // creating a large amount of small ALTREP objects (#1450)
+    // creating a large amount of small ALTREP objects that are used downstream.
+    // This is a heuristic and we should also be on the lookout for cases where
+    // we chop to create a small amount of large ALTREP objects that are
+    // quickly discarded (#1450).
     SEXP elt = PROTECT(vec_slice_base(
       info.proxy_info.type,
       proxy,
--- a/src/slice-chop.c
+++ b/src/slice-chop.c
@@ -191,7 +191,14 @@ static SEXP chop(SEXP x, SEXP indices, struct vctrs_chop_info info) {
       ++(*info.p_index);
     }
 
-    SEXP elt = PROTECT(vec_slice_base(info.proxy_info.type, proxy, info.index));
+    // Always materialize ALTREP vectors when chopping to avoid inefficiently
+    // creating a large amount of small ALTREP objects (#1450)
+    SEXP elt = PROTECT(vec_slice_base(
+      info.proxy_info.type,
+      proxy,
+      info.index,
+      VCTRS_MATERIALIZE_true
+    ));
 
     if (names != R_NilValue) {
       SEXP elt_names = PROTECT(slice_names(names, info.index));
--- a/src/slice.c
+++ b/src/slice.c
@@ -73,7 +73,7 @@ SEXP fns_vec_slice_dispatch_integer64 = NULL;
   return out
 
 #define SLICE(RTYPE, CTYPE, DEREF, CONST_DEREF, NA_VALUE)                   \
-  if (ALTREP(x)) {                                                          \
+  if (!materialize && ALTREP(x)) {                                          \
     SEXP alt_subscript = PROTECT(compact_materialize(subscript));           \
     SEXP out = ALTVEC_EXTRACT_SUBSET_PROXY(x, alt_subscript, R_NilValue);   \
     UNPROTECT(1);                                                           \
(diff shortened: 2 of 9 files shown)


src/slice.h
```diff
@@ -4,9 +4,18 @@
 extern SEXP syms_vec_slice_dispatch_integer64;
 extern SEXP fns_vec_slice_dispatch_integer64;
 
+enum vctrs_materialize {
```
lionel-, line 7: I think we started out with british spelling in this repo.

src/slice.c
```diff
@@ -397,7 +401,7 @@ SEXP vec_slice_impl(SEXP x, SEXP subscript) {
         Rf_setAttrib(out, R_DimNamesSymbol, names);
       }
     } else {
-      out = PROTECT_N(vec_slice_base(info.type, data, subscript), &nprot);
+      out = PROTECT_N(vec_slice_base(info.type, data, subscript, VCTRS_MATERIALIZE_false), &nprot);
```
lionel-, line 404: yup much better with the boolean enum.

src/slice-chop.c
```diff
@@ -191,7 +191,14 @@ static SEXP chop(SEXP x, SEXP indices, struct vctrs_chop_info info) {
       ++(*info.p_index);
     }
 
-    SEXP elt = PROTECT(vec_slice_base(info.proxy_info.type, proxy, info.index));
+    // Always materialize ALTREP vectors when chopping to avoid inefficiently
```
lionel-, line 194: We should be on the lookout for potential performance improvements in the other direction (i.e. small number of large chops which eventually get discarded).

lionel- (approved):

src/slice.h
DavisVaughan: We actually use the Z in quite a few places:

- ALTREP rle inspect method `"vctrs_altrep_rle (len=%d, materialized=%s)\n"`
- `altrep_rle_string_Materialize()`
- `compact_materialize()`

I think I'm just going to leave it for now and we can change them all over later to be consistent

PR #1457, opened 2021-09-22
Update for upcoming testthat changes related to condition handling

DavisVaughan:
These didn't show up until now because only the dev version of vctrs uses testthat 3e


lionel- (approved):
Thanks!

PR #1458, opened 2021-09-24
Export `vec_rank()`

DavisVaughan:


R/rank.R
```diff
@@ -87,9 +86,7 @@ vec_rank <- function(x,
                      na_value = "largest",
                      nan_distinct = FALSE,
                      chr_transform = NULL) {
```
lionel-, line 88: It's still weird to me that we have this seemingly arbitrary `chr_transform` argument (why only chr? why doesn't the caller transform?) instead of something more suggestive that this is really about locales.

lionel- (approved):

PR #1460, opened 2021-09-28
Rename `chr_transform` to `chr_proxy_collate`

DavisVaughan:
See conversation at https://github.com/r-lib/vctrs/pull/1458#discussion_r717334782


lionel- (approved):
I think it's better :)

PR #1461, opened 2021-09-28
Rework `na_propagate` as `incomplete` in `vec_rank()`

DavisVaughan:


R/rank.R
```diff
@@ -36,9 +33,13 @@
...
 #'
-#' @param na_propagate A single logical specifying whether or not missing
-#'   values should be propagated. If `TRUE`, all missing values are given
-#'   the rank `NA`.
+#' @param incomplete Treatment of [incomplete][vec_detect_complete] values.
+#'
+#'   - `"rank"`: Give incomplete values a numeric rank, which is affected
+#'   by `na_value` and `nan_distinct`.
+#'
+#'   - `"na"`: Give incomplete values a rank of `NA`, regardless of the value
```
lionel-, line 41: I don't understand this part. Won't incomplete rows be affected by `na_value` if they are treated as `NA`?

R/rank.R
```diff
@@ -36,9 +33,13 @@
 #'   non-duplicate value will have a rank incremented by `1`, effectively
 #'   removing any gaps in the ranking.
 #'
-#' @param na_propagate A single logical specifying whether or not missing
-#'   values should be propagated. If `TRUE`, all missing values are given
-#'   the rank `NA`.
+#' @param incomplete Treatment of [incomplete][vec_detect_complete] values.
+#'
+#'   - `"rank"`: Give incomplete values a numeric rank, which is affected
```
lionel-, line 38: Using the term "rows" instead of "values" would probably be clearer?

lionel- (approved):

R/rank.R
DavisVaughan: They aren't treated like `NA`. We don't rank them at all, instead giving them a value of `NA` in the result.

DavisVaughan: We actually rank them when `incomplete = "rank"` is set. Which is when `na_value` and `nan_distinct` comes into play, because that determines how missing values rank vs non-missing values, and vs `NaN` values

R/rank.R
DavisVaughan: I stayed away from "rows" because you can rank atomic vectors too and I think this documentation needs to make sense in that case too

DavisVaughan: I could use "observations"?

DavisVaughan:
@lionel- I've revised the documentation of `incomplete`. Is it any clearer now?

R/rank.R
```diff
@@ -36,9 +33,15 @@
 #'   non-duplicate value will have a rank incremented by `1`, effectively
 #'   removing any gaps in the ranking.
 #'
-#' @param na_propagate A single logical specifying whether or not missing
-#'   values should be propagated. If `TRUE`, all missing values are given
-#'   the rank `NA`.
+#' @param incomplete Treatment of [incomplete][vec_detect_complete]
```
lionel-, line 36: Maybe:

> Treatment of missing observations (in the [incomplete][vec_detect_complete] sense).

And then use missing in the rest of the text? I think it's hard to reason about the doc with the term incomplete. Especially with non-rows values. I would not have first guessed I should use this argument for `NA` values. Of course they are incomplete, but in my mind they are missing and completeness is a property I associate with rows.

Maybe we should hold this renaming? Something doesn't feel right. I'd be tempted to rename to `missing` instead of `incomplete`, and then use documentation to disambiguate.

DavisVaughan: Okay we can hold off on merging this immediately, I don't have a perfect solution yet either

Note that `na_value` also says:

> Treatment of missing values.

lionel-:
We could merge as is with just this change:

```
@param incomplete Treatment of [incomplete][vec_detect_complete] values
```

->

```
@param incomplete Treatment of missing and [incomplete][vec_detect_complete] observations
```

PR #1387, opened 2021-06-08
WIP: `vec_matches()`

DavisVaughan:

PR #1468, opened 2021-10-01
Expose `struct poly_df_data` in `poly-op.h`

DavisVaughan:
For #1387

This is a pretty useful data structure to pass around, since it holds the pointers to each column


src/poly-op.h
```diff
@@ -9,6 +9,11 @@ poly_binary_int_fn_ptr new_poly_p_equal_na_equal(enum vctrs_type type);
 typedef bool (*poly_unary_bool_fn_ptr)(const void* x, r_ssize i);
 poly_unary_bool_fn_ptr new_poly_p_is_missing(enum vctrs_type type);
 
+struct poly_df_data {
+  enum vctrs_type* col_types;
+  const void** col_ptrs;
```
lionel-, line 14: For consistency we'd now write this right?

```suggestion
  enum vctrs_type* v_col_type;
  const void** v_col_ptr;
```

lionel- (approved):

PR #1471, opened 2021-10-04
Use an argument label when casting and recycling `size`

### LINKED ISSUE #1470: Improve error message for vec_assert()
I am starting to work on https://github.com/r-lib/testthat/issues/1423 and noticed `testthat::expect_vector()`, and was wondering if it might be useful for the intended `expect_shape()`; I tried:

```
testthat::expect_vector(matrix(1:10, 2, 5), ptype = integer(), size = c(2, 5))
```

And was met with:

```
Error: `matrix(1:10, 2, 5)` must be a vector with type <integer>.
Instead, it has type <integer[,5]>.
```

Thinking I might just need a prototype with the right number of columns, I tried:

```
testthat::expect_vector(matrix(1:10, 2, 5), ptype = rbind(integer(5)), size = c(2, 5))
```

Which gives the more puzzling error:

```
Error: Can't recycle input of size 2 to size 1.
```

Both of these errors are coming from `vctrs::vec_assert()`. From a glance at the code, it looks like `length(size) == 1L` is assumed; this is not _totally_ explicit in the documentation, and anyway a simple test would allow for a clearer failure here.

Happy to file a PR but from a glance at the code, I am afraid I won't use idiomatic internal functions; I would write something as simple as

```
stopifnot(length(size) == 1L)
```

If that's enough, let me know & I'll file the PR.

DavisVaughan:
Closes #1470 

@MichaelChirico With this change you get the more informative:

``` r
testthat::expect_vector(matrix(1:10, 2, 5), ptype = rbind(integer(5)), size = c(2, 5))
#> Error: Can't recycle `size` (size 2) to size 1.
```

### DIFF
--- a/R/assert.R
+++ b/R/assert.R
@@ -91,7 +91,12 @@ vec_assert <- function(x, ptype = NULL, size = NULL, arg = as_label(substitute(x
 
   if (!is_null(size)) {
     size <- vec_cast(size, integer(), x_arg = "size")
-    size <- vec_recycle(size, 1L, x_arg = "size")
+
+    n_size <- length(size)
+    if (n_size != 1L) {
+      abort(glue::glue("`size` must be length 1, not length {n_size}."))
+    }
+
     x_size <- vec_size(x)
     if (!identical(x_size, size)) {
       stop_assert_size(x_size, size, arg)
--- a/R/assert.R
+++ b/R/assert.R
@@ -61,7 +61,7 @@
 #'   class, its [vec_ptype()] is compared to that of `x` with
 #'   `identical()`. Otherwise, its [typeof()] is compared to that of
 #'   `x` with `==`.
-#' @param size A single integer size to compare against.
+#' @param size A single integer size against which to compare.
 #' @param arg Name of argument being checked. This is used in error
 #'   messages. The label of the expression passed as `x` is taken as
 #'   default.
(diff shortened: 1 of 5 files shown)


DavisVaughan:
Actually I don't think there is any need to recycle at all here, so we can go with something more in line with your original suggestion. I've also updated the docs.

``` r
testthat::expect_vector(matrix(1:10, 2, 5), ptype = rbind(integer(5)), size = c(2, 5))
#> Error: `size` must be length 1, not length 2.
```

R/assert.R
```diff
@@ -61,7 +61,7 @@
 #'   class, its [vec_ptype()] is compared to that of `x` with
 #'   `identical()`. Otherwise, its [typeof()] is compared to that of
 #'   `x` with `==`.
-#' @param size Size to compare against
+#' @param size A single integer size to compare against.
 #' @param arg Name of argument being checked. This is used in error
```
MichaelChirico, line 65: against which to compare, if we're being pedantic 😃

MichaelChirico (approved):
LGTM, thanks!

tests/testthat/test-assert.R
```diff
@@ -218,6 +218,12 @@ test_that("assertion failures are explained", {
   expect_snapshot(error = TRUE, vec_assert(data.frame(x = 1, y = 2), data.frame(x = "foo", y = 2)))
 })
 
+test_that("vec_assert() validates `size` (#1470)", {
+  expect_snapshot(error = TRUE, vec_assert(1, size = c(2, 3)))
```
lionel-, line 222: Can you please use the `expect_snapshot((expect_error()))` approach? This way the rlang print method gets actioned and we test the whole error prefix including call.

lionel- (approved):

PR #1472, opened 2021-10-05
Implement `vec_locate_matches()`

DavisVaughan:
Closes https://github.com/r-lib/vctrs/pull/1387, but look there to see the commit history that got us here

This PR implements `vec_locate_matches()` to support https://github.com/tidyverse/dplyr/pull/5910.

`vec_locate_matches()` is similar to `vec_match()`, but detects _all_ matches by default, and can match on conditions other than equality (like `>=` and `<`). There are also various other arguments to limit or adjust exactly which kinds of matches are returned. Here is an example:

``` r
library(vctrs)

x <- c("a", "b", "a", "c", "d")
y <- c("d", "b", "a", "d", "a", "e")

# For each value of `x`, find all matches in `y`
# - The "c" in `x` doesn't have a match, so it gets an NA location by default
# - The "e" in `y` isn't matched by anything in `x`, so it is dropped by default
vec_locate_matches(x, y)
#>   needles haystack
#> 1       1        3
#> 2       1        5
#> 3       2        2
#> 4       3        3
#> 5       3        5
#> 6       4       NA
#> 7       5        1
#> 8       5        4
```

This is a big PR, so I'll provide a bit of guidance below.

## Recommended reading order

- Algorithm description below

- NEWS bullet

- `match.R` to get a feel for the function

- Look at the messages generated by the `stop_*()` and `warn_*()` helpers in `match.R`

- Read `match.c` up to and through `vec_locate_matches()`

- At this point, read `df_detect_complete_by_col()` to see how we collect completeness information per column

- At this point, read `df_joint_xtfrm_by_col()` to understand how it is computing joint ranks by column
   - This should take you to `vec_joint_xtfrm()` in `match-joint.c`, which gives a good explanation of how this works and what it returns.
   - Within `vec_joint_xtfrm()` is `vec_joint_proxy_order()`, which specially handles list-columns here.
   - Within `vec_joint_xtfrm()`, you'll see the comparison functions from `match-compare.h`, which have been designed to compute comparisons that align with the way that `vec_order()` orders elements.

- Hopefully now you understand that `needles` and `haystack` have been simplified to columns of integer vectors which represent the joint ranks per column.

- Continue on to `df_locate_matches()`. Try to ignore everything about "container ids" for now, and read through until you get to `df_locate_matches_recurse()`. This is the workhorse for finding and recording information about the matches. Hopefully there are enough notes to guide you through it.

- Now look at `expand_compact_indices()` to understand how the "compact" information recorded in `df_locate_matches_recurse()` gets expanded into the final result. Again try to ignore most notes about "container ids" for now.

- Now it is worth looking at the container id generation. Start with `compute_nesting_container_info()`, which does some optimizations to avoid computing container ids if possible. Then look at `compute_nesting_container_ids()`, which is the workhorse for computing the container ids. Essentially container ids are an integer column of ids that are used to break the `haystack` up into non-overlapping containers based on how the elements of `haystack` are ordered (more on this below). Each element of `needles` checks for matches in all of the `haystack` containers. The matches are collected across the containers at the end in `expand_compact_indices()` and are reordered if required.

- Then look at `df_locate_matches_with_containers()`, which is the only place the container ids are really used. It does the actual separation of containers and hands off each container to `df_locate_matches_recurse()` to find the matches in that container.

- That should be all of the C algorithmic code.

- Read over tests as needed, but there are a lot.

### Overview and `==`

The simplest (approximate) way to think about the algorithm that `df_locate_matches_recurse()` uses is that it sorts both inputs, and then starts at the midpoint in `needles` and uses a binary search to find each needle in `haystack`. Since there might be multiple of the same needle, we find the location of the lower and upper duplicate of that needle to handle all duplicates of that needle at once. Similarly, if there are duplicates of a matching `haystack` value, we find the lower and upper duplicates of the match.

If the condition is `==`, that is pretty much all we have to do. For each needle, we then record 3 things: the location of the needle, the location of the lower match in the haystack, and the match size (i.e. `loc_upper_match - loc_lower_match + 1`). This later gets expanded in `expand_compact_indices()` into the actual output.

After recording the matches for a single needle, we perform the same procedure on the LHS and RHS of that needle (remember we started on the midpoint needle). i.e. from `[1, loc_needle-1]` and `[loc_needle+1, size_needles]`, again taking the midpoint of those two ranges, finding their respective needle in the haystack, recording matches, and continuing on to the next needle.

When we have a data frame with multiple columns, we add a layer of recursion to this. For the first column, we find the locations of the lower/upper duplicate of the current needle, and we find the locations of the lower/upper matches in the haystack. If we are on the final column, we record the matches, otherwise, we pass these on to another call to `df_locate_matches_recurse()`, bumping the column index and using these lower/upper bounds as the starting bounds for the next column.

I think an example would be useful here, so below I step through this process for a few iterations:

<details>

``` r
library(tibble)

# these are sorted already for simplicity
needles <- tibble(x = c(1, 1, 2, 2, 2, 3), y = c(1, 2, 3, 4, 5, 3))
haystack <- tibble(x = c(1, 1, 2, 2, 3), y = c(2, 3, 4, 4, 1))

needles
#> # A tibble: 6 × 2
#>       x     y
#>   <dbl> <dbl>
#> 1     1     1
#> 2     1     2
#> 3     2     3
#> 4     2     4
#> 5     2     5
#> 6     3     3

haystack
#> # A tibble: 5 × 2
#>       x     y
#>   <dbl> <dbl>
#> 1     1     2
#> 2     1     3
#> 3     2     4
#> 4     2     4
#> 5     3     1

## Column 1, iteration 1

# start at midpoint in needles
# this corresponds to x==2
loc_mid_needles <- 3L

# finding all x==2 values in needles gives us:
loc_lower_duplicate_needles <- 3L
loc_upper_duplicate_needles <- 5L

# finding matches in haystack give us:
loc_lower_match_haystack <- 3L
loc_upper_match_haystack <- 4L

# compute LHS/RHS bounds for next needle
lhs_loc_lower_bound_needles <- 1L # original lower bound
lhs_loc_upper_bound_needles <- 2L # lower_duplicate-1

rhs_loc_lower_bound_needles <- 6L # upper_duplicate+1
rhs_loc_upper_bound_needles <- 6L # original upper bound

# We still have a 2nd column to check. So recurse and pass on the current
# duplicate and match bounds to start the 2nd column with.

## Column 2, iteration 1

# midpoint of [3, 5]
# value y==4
loc_mid_needles <- 4L

loc_lower_duplicate_needles <- 4L
loc_upper_duplicate_needles <- 4L

loc_lower_match_haystack <- 3L
loc_upper_match_haystack <- 4L

# last column, so record matches
# - this was location 4 in needles
# - lower match in haystack is at loc 3
# - match size is 2

# Now handle LHS and RHS of needle midpoint
lhs_loc_lower_bound_needles <- 3L # original lower bound
lhs_loc_upper_bound_needles <- 3L # lower_duplicate-1

rhs_loc_lower_bound_needles <- 5L # upper_duplicate+1
rhs_loc_upper_bound_needles <- 5L # original upper bound

## Column 2, iteration 2 (using LHS bounds)

# midpoint of [3,3]
# value of y==3
loc_mid_needles <- 3L

loc_lower_duplicate_needles <- 3L
loc_upper_duplicate_needles <- 3L

# no match! no y==3 in haystack for x==2
# lower-match will always end up > upper-match in this case
loc_lower_match_haystack <- 3L
loc_upper_match_haystack <- 2L

# no LHS or RHS needle values to do, so we are done here

## Column 2, iteration 3 (using RHS bounds)

# same as above, range of [5,5], value of y==5, which has no match in haystack

## Column 1, iteration 2 (LHS of first x needle)

# Now we are done with the x needles from [3,5], so move on to the LHS and RHS
# of that. Here we would do the LHS:

# midpoint of [1,2]
loc_mid_needles <- 1L

# ...

## Column 1, iteration 3 (RHS of first x needle)

# midpoint of [6,6]
loc_mid_needles <- 6L

# ...
```

</details>

In the real code, rather than comparing the double values of the columns directly, we replace each column with "joint ranks" computed between column x of `needles` and column x of `haystack`. It is approximately like doing `vec_rank(vec_c(needles$x, haystack$x), type = "dense")`, then splitting the resulting ranks back up into their corresponding needle/haystack columns. This keeps the recursion code simpler, because we only have to worry about comparing integers.

### Non-equi conditions and containers

At this point we can talk about non-equi conditions like `<` or `>=`. The general idea is pretty simple, and just builds on the above algorithm. For example, start with the `x` column from needles/haystack above:

``` r
needles$x
#> [1] 1 1 2 2 2 3

haystack$x
#> [1] 1 1 2 2 3
```

If we used a condition of `<=`, then we'd do everything the same as before:

- Midpoint in needles is location 3, value `x==2`
- Find lower/upper duplicates in needles, giving locations `[3, 5]`
- Find lower/upper exact match in haystack, giving locations `[3, 4]`

At this point we need to "adjust" the haystack match bounds to account for the condition. Since `haystack` is ordered, our "rule" for `<=` is to keep the lower match location the same, but extend the upper match location to the upper bound, so we end up with `[3, 5]`. We know we can extend the upper match location because every value after the exact match should be less than the needle. Then we just record the matches and continue on normally.

This approach is really nice, because we only have to exactly match the `needle` in `haystack`. We don't have to compare each needle against every value in `haystack`, which would take a massive amount of time.

However, it gets slightly more complex with data frames with multiple columns. Let's go back to our original `needles` and `haystack` data frames and apply the condition `<=` to each column. Here is another worked example, which shows a case where our "rule" falls apart on the second column

<details>

``` r
needles
#> # A tibble: 6 × 2
#>       x     y
#>   <dbl> <dbl>
#> 1     1     1
#> 2     1     2
#> 3     2     3
#> 4     2     4
#> 5     2     5
#> 6     3     3

haystack
#> # A tibble: 5 × 2
#>       x     y
#>   <dbl> <dbl>
#> 1     1     2
#> 2     1     3
#> 3     2     4
#> 4     2     4
#> 5     3     1

# `condition = c("<=", "<=")`

## Column 1, iteration 1

# x == 2
loc_mid_needles <- 3L

loc_lower_duplicate_needles <- 3L
loc_upper_duplicate_needles <- 5L

# finding exact matches in haystack give us:
loc_lower_match_haystack <- 3L
loc_upper_match_haystack <- 4L

# because haystack is ordered we know we can expand the upper bound automatically
# to include everything past the match. i.e. needle of x==2 must be less than
# the haystack value at loc 5, which we can check by seeing that it is x==3.
loc_lower_match_haystack <- 3L
loc_upper_match_haystack <- 5L

## Column 2, iteration 1

# needles range of [3, 5]
# y == 4
loc_mid_needles <- 4L

loc_lower_duplicate_needles <- 4L
loc_upper_duplicate_needles <- 4L

# finding exact matches in haystack give us:
loc_lower_match_haystack <- 3L
loc_upper_match_haystack <- 4L

# lets try using our rule, which tells us we should be able to extend the upper
# bound:
loc_lower_match_haystack <- 3L
loc_upper_match_haystack <- 5L

# but the haystack value of y at location 5 is y==1, which is not less than y==4
# in the needles! looks like our rule failed us.
```

</details>

If you read through the above example, you'll see that the rule didn't work here. The problem is that while `haystack` is ordered (by `vec_order()`s standards), each column isn't ordered _independently_ of the others. Instead, each column is ordered within the "group" created by previous columns. Concretely, `haystack` here has an ordered `x` column, but if you look at `y` by itself, it isn't ordered (because of that 1 at the end). That is what causes the rule to fail.

``` r
haystack
#> # A tibble: 5 × 2
#>       x     y
#>   <dbl> <dbl>
#> 1     1     2
#> 2     1     3
#> 3     2     4
#> 4     2     4
#> 5     3     1
```

To fix this, we need to create haystack "containers" where the values within each container are all totally ordered. For `haystack` that would create 2 containers and look like:

``` r
haystack[1:4,]
#> # A tibble: 4 × 2
#>       x     y
#>   <dbl> <dbl>
#> 1     1     2
#> 2     1     3
#> 3     2     4
#> 4     2     4

haystack[5,]
#> # A tibble: 1 × 2
#>       x     y
#>   <dbl> <dbl>
#> 1     3     1
```

This is essentially what `computing_nesting_container_ids()` does. You can actually see these ids with the helper `compute_nesting_container_info()`:

``` r
haystack2 <- haystack
haystack2$x <- as.integer(haystack2$x)
haystack2$y <- as.integer(haystack2$y)

info <- compute_nesting_container_info(haystack2, condition = c("<=", "<="), multiple = "all")

# container ids break haystack into [1, 4] and [5, 5]
info[[2]]
#> [1] 0 0 0 0 1
```

So the idea is that for each needle, we look in each haystack container and find all the matches, then we aggregate all of the matches once at the end. `df_locate_matches_with_containers()` has the job of iterating over the containers.

Computing containers can be expensive, but luckily it doesn't happen very often in normal usage.
- If there are all `==` conditions, we don't need containers (i.e. any equi join)
- If there is only 1 non-equi condition and no conditions after it, we don't need containers (i.e. most rolling joins)
- Otherwise the typical case where we need containers is if we have something like `date >= lower, date <= upper`. Even so, the computation cost scales with the number of columns in haystack you compute containers with (here 2), and it only really slows down around 4 columns or so, which I haven't ever seen an example of.


DavisVaughan:
TODO: Figure out why this errors (probably something do to with the unspecified type)

``` r
library(vctrs)

needles <- data.frame(x = logical(), y = logical())
haystack <- data.frame(x = NA, y = NA)

vec_locate_matches(needles, haystack)
#> Error:
#> ! Internal error in `vec_joint_proxy_order()`: `x` and `y` should have the same type.
```

DavisVaughan:
It's possible we should make `incomplete = "match"` always match missing values with equality. I was using this for the interval code and was using `condition = c(">", "<")` which doesn't match missing values because they aren't "or equal to" conditions, but I definitely still wanted them to match.

DavisVaughan:
Question: Since a cross join (generated by `condition = NULL`) is now exactly the same as supplying two vectors of `1`s with `condition = "=="`, should we remove that feature? It always felt a bit out of place, and was only present to support dplyr's cross join feature. But this seems simple enough that we could push the burden of creating these vectors onto dplyr. It would just be `x = vec_rep(1L, vec_size(x)); y = vec_rep(1L, vec_size(y)); condition = "=="` if we detect we are doing a cross join.

PR #1482, opened 2021-10-27
`obj_print*()` gain max argument

### LINKED ISSUE #1355: Obey max argument and max.print option in obj_print()?
We're already obeying `getOption("max.print")` via `print(format(x))`, but we're still calling `format()` on the full vctr. We could do a better job by providing the necessary infrastructure in `obj_print()` and passing `max` to `obj_print_footer()` .

I'll experiment with `pillar::num()` and `pillar::char()` .

``` r
print(vctrs::new_vctr(letters), max = 10)
#> <vctrs_vctr[26]>
#>  [1] a b c d e f g h i j k l m n o p q r s t u v w x y z
options(max.print = 5)
print(vctrs::new_vctr(letters))
#> <vctrs_vctr[26]>
#> [1] a b c d e
#>  [ reached getOption("max.print") -- omitted 21 entries ]

print(letters, max = 10)
#>  [1] "a" "b" "c" "d" "e" "f" "g" "h" "i" "j"
#>  [ reached getOption("max.print") -- omitted 16 entries ]
print(letters)
#> [1] "a" "b" "c" "d" "e"
#>  [ reached getOption("max.print") -- omitted 21 entries ]
```

<sup>Created on 2021-04-08 by the [reprex package](https://reprex.tidyverse.org) (v1.0.0)</sup>

krlmlr:
This argument is also accepted by `base::print()`, but undocumented.

Closes #1355.


lionel-:
I don't understand why we don't slice in the generic?

krlmlr:
@lionel-: IIRC, to slice in the generic, we would have to pass both the sliced vector and the original lengths to the detail methods. Do you prefer this solution?

lionel-:
Don't we pass the whole vector?

krlmlr:
Yes, right now we pass the whole vector.

Are you proposing the following logic:

- we slice in the generic
- the methods only ever see a truncated version of the vector
- `obj_print_footer()` is no longer responsible for printing the informative line
- the generic prints the informative line after all methods have been called

?

I rather like that.

lionel-:
yup exactly

R/print-str.R
```diff
@@ -58,6 +104,23 @@ obj_print_footer.default <- function(x, ...) {
...
+
+  stopifnot(is_integerish(max, 1L, finite = TRUE), max >= 0, max < 2147483648)
+  max <- as.integer(max)
+
+  if (max > max_print) {
+    # Avoid truncation in case we're forwarding to print()
+    local_options(max.print = max, .frame = frame)
+  }
+
+  structure(max, max_print = max_print)
```
lionel-, line 121: Should return a list.

lionel-: Why do we need both `max` and `max_print`? Could we take `max = getOption("max.print")` in the generic instead, to simplify things?

R/print-str.R
```diff
@@ -58,6 +104,23 @@ obj_print_footer.default <- function(x, ...) {
...
+  if (is.null(max)) {
+    max <- max_print
+  }
+
+  stopifnot(is_integerish(max, 1L, finite = TRUE), max >= 0, max < 2147483648)
+  max <- as.integer(max)
+
+  if (max > max_print) {
+    # Avoid truncation in case we're forwarding to print()
+    local_options(max.print = max, .frame = frame)
```
lionel-, line 118: 1. Is there any downside to unconditionally setting `max.print` to simplify things?

2. Can you please try and move the `local_options()` to the generic and rename this function to `as_max_print()`. This would simplify things.

Issue #1147, opened 2020-06-14
Compare CHARSXP instead of using strcmp()


krlmlr:
because the former is significantly faster. Nanoseconds.

Requires preinstalling all strings used in the C code, we have infrastructure for this.

``` r
N <- 10000

t1 <- Rcpp::cppFunction('
  bool cmp_str(Rcpp::CharacterVector c1) {
    for (int i = 0; i < c1.length(); ++i) {
      if (strcmp(CHAR(c1[i]), "propagate") != 0) return false;
    }
    return true;
  }
')

t2 <- Rcpp::cppFunction('
  bool cmp_str(Rcpp::CharacterVector c1) {
    static const Rcpp::String string_propagate = "propagate";
    static SEXP sexp_propagate = string_propagate.get_sexp();
    for (int i = 0; i < c1.length(); ++i) {
      if (c1[i] != sexp_propagate) return false;
    }
    return true;
  }
')

propagate <- rep("propagate", N)

bench::mark(
  t1(propagate),
  t2(propagate)
)
#> # A tibble: 2 x 6
#>   expression         min   median `itr/sec` mem_alloc `gc/sec`
#>   <bch:expr>    <bch:tm> <bch:tm>     <dbl> <bch:byt>    <dbl>
#> 1 t1(propagate)  245.6µs  279.5µs     3300.    2.49KB        0
#> 2 t2(propagate)   41.9µs   46.7µs    19436.    2.49KB        0

f1 <- Rcpp::cppFunction('
  bool cmp_str(Rcpp::CharacterVector c1) {
    for (int i = 0; i < c1.length(); ++i) {
      if (!strcmp(CHAR(c1[i]), "propagate")) return true;
    }
    return false;
  }
')

f2 <- Rcpp::cppFunction('
  bool cmp_str(Rcpp::CharacterVector c1) {
    static const Rcpp::String string_propagate = "propagate";
    static SEXP sexp_propagate = string_propagate.get_sexp();
    for (int i = 0; i < c1.length(); ++i) {
      if (c1[i] == sexp_propagate) return true;
    }
    return false;
  }
')

propagat <- rep("propagat", N)

bench::mark(
  f1(propagat),
  f2(propagat)
)
#> # A tibble: 2 x 6
#>   expression        min   median `itr/sec` mem_alloc `gc/sec`
#>   <bch:expr>   <bch:tm> <bch:tm>     <dbl> <bch:byt>    <dbl>
[...]

PR #1486, opened 2021-10-29
`r_new_shared_string()` and collecting all string constants used in `strcmp()`

krlmlr:
For #1147.

Generated with:

```sh
egrep -Roh 'strcmp[(][^,]+, "[^"]+"[)]' src | grep -o '".*"' | sort | uniq | gsed -r 's/"(.*)"/extern SEXP strs_\1;/' | pbcopy
egrep -Roh 'strcmp[(][^,]+, "[^"]+"[)]' src | grep -o '".*"' | sort | uniq | gsed -r 's/"(.*)"/SEXP strs_\1 = NULL;/' | pbcopy
egrep -Roh 'strcmp[(][^,]+, "[^"]+"[)]' src | grep -o '".*"' | sort | uniq | gsed -r 's/"(.*)"/  strs_\1 = r_new_shared_string("\1");/' | pbcopy
```


DavisVaughan:
We also already have a character vector called `strings` that gets loaded on package load. Each element holds a useful string constant that we use elsewhere (similar to the ones added here) https://github.com/r-lib/vctrs/pull/1486/files#diff-3e04e056edbdc0f772b0ffd1c93cef1187209f37276d40d2a281d9b989398a73R1688-R1693

lionel-:
Thanks Kirill, but on second review this doesn't seem necessary as most `strcmp()` are for checking scalar arguments. Also many of those validation routines should be replaced with `r_arg_match()`. If you see an `strcmp()` in a loop, can you please send a specific patch that switches to pointer comparison?

PR #1488, opened 2021-11-11
Directly insert `""` whenever we encounter a `NA_integer_` subscript

DavisVaughan:
Looking over the `NA` name repair code again, it became obvious that we can greatly simplify it by never inserting `NA_character_` as a name to begin with. Since we can customize the `NA_VALUE` in `SLICE()`, I just made a new wrapper that inserts the empty string instead.


lionel- (approved):
Nice!

PR #1492, opened 2021-11-11
Mention why we have an `as.character()` method for list-of

### LINKED ISSUE #1490: Remove `as.character.vctrs_list_of()`
I am not sure why we have this, and it doesn't follow any of our principles.

I don't think we should do what `as.character()` does for bare lists, so I think we should just remove this

``` r
library(vctrs)

x <- list_of(1L, 2:3)

as.character(x)
#> [1] "<int>" "<int>"

# I don't think we should do this
as.character(as.list(x))
#> [1] "1"   "2:3"
```

<sup>Created on 2021-11-11 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

DavisVaughan:
Closes #1490 

It turns out we have this for compatibility with the RStudio Viewer, i.e. tidyverse/tidyr#654. So I'm adding a note about that.

A little over a year ago, the IDE switched to checking for a `format()` method first before calling `as.character()` and it would have used that instead if a `format()` method existed. We do have a `format.vctrs_list_of()` method. But that IDE feature is a little broken https://github.com/rstudio/rstudio/issues/10073, plus we'd probably need to keep this to support older versions of the IDE anyways.

### DIFF
--- a/R/type-list-of.R
+++ b/R/type-list-of.R
@@ -139,6 +139,7 @@ as.list.vctrs_list_of <- function(x, ...) {
 }
 #' @export
 as.character.vctrs_list_of <- function(x, ...) {
+  # For compatibility with the RStudio Viewer. See tidyverse/tidyr#654.
   map_chr(x, function(elt) paste0("<", vec_ptype_abbr(elt), ">"))
 }
 


lionel- (approved):
Feel free to push this sort of changes directly to main :)

DavisVaughan:
Mainly I just wanted to leave a documentation trail for this one 😅

PR #1506, opened 2022-01-25
Preliminary support for error calls

lionel-:
This adds `call` argument to:

- `vec_cast()` and `vec_ptype2()`
- `vec_default_cast()` and `vec_default_ptype2()`
- `vec_assert()`
- `vec_as_names()`
- `stop_` constructors like `stop_incompatible_type()`

Default ptype2 and cast methods should automatically pass the error call to the stop function if they pass `...` to `vec_default_ptype2()` and `vec_default_cast()`, as they should.

I've decided not to use the `local_error_call()` pattern in the end. My current feeling is that it's too easy to get it wrong, with internal errors reported as caller errors.

At the C level, we pass `call` through a new `struct r_lazy`. This way we don't evaluate `call` arguments unless needed. This struct is flexible. It sometimes carry a `call` symbol and a frame environment, and other times it carries a forced `call` value.

There's still a lot of places where the error call is not forwarded, but I think this should be tackled progressively as we encounter the cases. For this release I plan to do some more work on subscript errors for tidyselect and look into what tidyr needs for call reporting (e.g. the bind operations).

### DIFF
--- a/R/cast.R
+++ b/R/cast.R
@@ -8,6 +8,7 @@
 #'
 #' @includeRmd man/faq/developer/links-coercion.Rmd
 #'
+#' @inheritParams rlang::args_error_context
 #' @param x Vectors to cast.
 #' @param ... For `vec_cast_common()`, vectors to cast. For
 #'   `vec_cast()`, `vec_cast_default()`, and `vec_restore()`, these
@@ -62,11 +63,16 @@
 #'
 #' # Cast to common type
 #' vec_cast_common(factor("a"), factor(c("a", "b")))
-vec_cast <- function(x, to, ..., x_arg = "", to_arg = "") {
+vec_cast <- function(x,
+                     to,
+                     ...,
+                     x_arg = "",
+                     to_arg = "",
+                     call = caller_env()) {
   if (!missing(...)) {
     check_ptype2_dots_empty(...)
   }
-  return(.Call(vctrs_cast, x, to, x_arg, to_arg))
+  return(.Call(ffi_cast, x, to, x_arg, to_arg, environment()))
   UseMethod("vec_cast", to)
 }
 vec_cast_dispatch <- function(x, to, ..., x_arg = "", to_arg = "") {
--- a/R/cast.R
+++ b/R/cast.R
@@ -82,9 +82,21 @@ vec_cast_dispatch <- function(x, to, ..., x_arg = "", to_arg = "") {
 vec_cast_no_fallback <- function(x, to) {
   vec_cast_common_params(x = x, .to = to, .df_fallback = DF_FALLBACK_none)$x
 }
-vec_cast_dispatch_native <- function(x, to, ..., x_arg = "", to_arg = "") {
-  fallback_opts <- match_fallback_opts(...)
-  .Call(vctrs_cast_dispatch_native, x, to, fallback_opts, x_arg, to_arg)
+vec_cast_dispatch_native <- function(x,
+                                     to,
+                                     ...,
+                                     x_arg = "",
+                                     to_arg = "",
+                                     call = caller_env()) {
+  .Call(
+    ffi_cast_dispatch_native,
+    x,
+    to,
+    match_fallback_opts(...),
+    x_arg,
+    to_arg,
+    environment()
+  )
 }
 
 #' @export
--- a/R/cast.R
+++ b/R/cast.R
@@ -108,16 +108,39 @@ vec_cast_common_fallback <- function(..., .to = NULL) {
 #' @rdname vec_default_ptype2
 #' @inheritParams vec_cast
 #' @export
-vec_default_cast <- function(x, to, ..., x_arg = "", to_arg = "") {
+vec_default_cast <- function(x,
+                             to,
+                             ...,
+                             x_arg = "",
+                             to_arg = "",
+                             call = caller_env()) {
   if (is_asis(x)) {
-    return(vec_cast_from_asis(x, to, x_arg = x_arg, to_arg = to_arg))
+    return(vec_cast_from_asis(
+      x,
+      to,
+      x_arg = x_arg,
+      to_arg = to_arg,
+      call = call
+    ))
   }
   if (is_asis(to)) {
-    return(vec_cast_to_asis(x, to, x_arg = x_arg, to_arg = to_arg))
+    return(vec_cast_to_asis(
+      x,
+      to,
+      x_arg = x_arg,
+      to_arg = to_arg,
+      call = call
+    ))
   }
 
   if (inherits(to, "vctrs_vctr") && !inherits(to, c("vctrs_rcrd", "vctrs_list_of"))) {
-    return(vctr_cast(x, to, x_arg = x_arg, to_arg = to_arg))
+    return(vctr_cast(
+      x,
+      to,
+      x_arg = x_arg,
+      to_arg = to_arg,
+      call = call
+    ))
   }
 
   opts <- match_fallback_opts(...)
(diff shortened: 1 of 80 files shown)


lionel-:
cc @hadley @DavisVaughan

R/cast.R
```diff
@@ -62,11 +63,16 @@
...
+                     to,
+                     ...,
+                     x_arg = "",
+                     to_arg = "",
+                     call = caller_env()) {
   if (!missing(...)) {
     check_ptype2_dots_empty(...)
   }
-  return(.Call(vctrs_cast, x, to, x_arg, to_arg))
+  return(.Call(ffi_cast, x, to, x_arg, to_arg, environment()))
```
DavisVaughan, line 75: Oh I may have answered my question about `call` vs `frame`. I was going to ask if you should be passing `call` through rather than `environment()`, but is the idea that `environment()` is the "frame" and you can look up the call from the `call` object that is present in that frame?

src/fill.c
```diff
@@ -246,19 +246,23 @@ void stop_bad_direction() {
...
   if (x == R_NilValue) {
     return INFINITE_FILL;
   }
 
-  x = PROTECT(vec_cast(x, vctrs_shared_empty_int, args_max_fill, args_empty));
+  x = KEEP(vec_cast(x,
+                    vctrs_shared_empty_int,
+                    args_max_fill,
+                    args_empty,
+                    r_lazy_null));
```
DavisVaughan, line 258: So eventually we will go back through places like this and pass the call through?

src/names.c
```diff
@@ -34,27 +34,35 @@ SEXP vec_as_names(SEXP names, const struct name_repair_opts* opts) {
...
 }
 
 // [[ register() ]]
-SEXP vctrs_as_names(SEXP names, SEXP repair, SEXP repair_arg, SEXP quiet) {
-  if (!r_is_bool(quiet)) {
-    Rf_errorcall(R_NilValue, "`quiet` must a boolean value.");
+r_obj* ffi_as_names(r_obj* names,
+                    r_obj* repair,
+                    r_obj* ffi_repair_arg,
+                    r_obj* ffi_quiet,
```
DavisVaughan, line 47: Neat naming convention idea, since you have `bool quiet` below and that is the "better" name

src/cast.c
```diff
@@ -17,11 +17,17 @@ SEXP vec_cast_switch_native(const struct cast_opts* opts,
...
+r_obj* ffi_cast(r_obj* x,
+                r_obj* to,
+                r_obj* ffi_x_arg,
+                r_obj* ffi_to_arg,
+                r_obj* frame) {
+  struct vctrs_arg x_arg = vec_as_arg(ffi_x_arg);
+  struct vctrs_arg to_arg = vec_as_arg(ffi_to_arg);
 
-  return vec_cast(x, to, &x_arg, &to_arg);
+  struct r_lazy call = { .x = syms_call, .env = frame };
```
DavisVaughan, line 28: Do you need `(struct r_lazy)`? I saw that in other places

src/names.c
```diff
@@ -833,12 +849,17 @@ void stop_name_repair() {
...
+    .shelter = r_null,
     .type = 0,
-    .fn = R_NilValue,
-    .arg = arg,
-    .quiet = quiet
+    .fn = r_null,
+    .name_repair_arg = name_repair_arg,
+    .quiet = quiet,
+    .call = call
   };
```
DavisVaughan, line 863: I guess we just expect that someone else protects the elements of an `struct r_lazy`? Like, it doesn't look like the `shelter` here protects the elements of that `call` object at all.

src/names.c
```diff
@@ -939,12 +961,12 @@ void vctrs_init_names(SEXP ns) {
   syms_set_rownames_fallback = Rf_install("set_rownames_fallback");
   syms_set_names_fallback = Rf_install("set_names_fallback");
   syms_as_universal_names = Rf_install("as_universal_names");
-  syms_validate_unique_names = Rf_install("validate_unique");
+  syms_check_unique_names = Rf_install("validate_unique");
```
DavisVaughan, line 964: Sync name on the R side?

src/slice.c
```diff
@@ -483,14 +483,22 @@ SEXP vec_init(SEXP x, R_len_t n) {
...
-SEXP vctrs_init(SEXP x, SEXP n) {
-  n = PROTECT(vec_cast(n, vctrs_shared_empty_int, args_n, args_empty));
-  vec_assert_size(n, 1, args_n);
-  R_len_t n_ = r_int_get(n, 0);
-
-  SEXP out = vec_init(x, n_);
-
-  UNPROTECT(1);
+r_obj* ffi_init(r_obj* x, r_obj* ffi_n, r_obj* ffi_frame) {
+  struct r_lazy frame = { .x = ffi_frame, .env = r_null };
```
DavisVaughan, line 487: Is there any reason you didn't just add `call = caller_env()` to `vec_init()`?

tests/testthat/_snaps/assert.md
```diff
@@ -12,7 +12,7 @@
     Code
       vec_assert(lgl(), factor())
     Condition
-      Error in `vec_assert()`:
+      Error:
```
DavisVaughan, line 15: So the error is considered to come from the "top level" now, so that is why we don't get the function name in the error?

tests/testthat/_snaps/error-call.md
```diff
@@ -0,0 +1,175 @@
+# failing common type reports correct error call
+
+    Code
+      (expect_error(my_function()))
+    Output
+      <error/vctrs_error_incompatible_type>
+      Error:
```
DavisVaughan, line 7: Why doesn't it say `in my_function():` here? There are a few below where I expected that

R/cast.R
lionel-: yup in this case I'm using `frame` to create a `struct r_lazy` for `call`.

src/fill.c
lionel-: yeah, that's a lot of repetitive work and perhaps not worth it in some cases. In the same way that in some places we haven't passed an `arg` either.

src/slice.c
DavisVaughan: If you were to add `call` to `vec_init()`, do you really need the r_lazy struct idea? Why doesn't just passing through the `call` work?

src/names.c
lionel-: yes the caller protects it. In this case the caller is the R function.

src/names.c
lionel-: Eventually.

src/cast.c
DavisVaughan: Yea I just don't know how strict the compilers are, to me it feels reasonable to not have to force that cast

src/slice.c
lionel-: Yes, I think it'd be weird to add call arguments to the vector manipulation functions. So I've only done it in vctrs operations that validate or check inputs. But this is still evolving. In particular, perhaps the tidyr and dplyr ports will show us that call arguments are needed.

tests/testthat/_snaps/assert.md
lionel-: Yes, in this case the error call is actually `eval()` from the testthat REPL, and this is hidden by rlang.

tests/testthat/_snaps/error-call.md
lionel-: oh yes these still need to be fixed. It's already an improvement since it no longer shows stopper calls.

src/slice.c
lionel-: > If you were to add call to vec_init(), do you really need the r_lazy struct idea? Why doesn't just passing through the call work?

It's just for low-level functions to avoid evaluating error-related arguments unless needed.

tests/testthat/_snaps/error-call.md
lionel-: This is now better, thanks!

PR #1507, opened 2022-01-26
Pass error call in subscript errors

lionel-:


lionel-:
A nice benefit of the `r_lazy` variable is that it can be used to get the relevant environment for internal errors: https://github.com/r-lib/vctrs/pull/1507/commits/6657e4b63fa1a9ecb7899c5500db3c059317eb88.

R/subscript.R
```diff
@@ -29,26 +30,34 @@ vec_as_subscript <- function(i,
...
   .Call(
-    vctrs_as_subscript,
+    ffi_as_subscript,
     i = i,
     logical = logical,
     numeric = numeric,
     character = character,
-    arg = arg
+    arg = arg,
+    call = environment()
```
DavisVaughan, line 44: `frame`?

R/subscript.R
```diff
@@ -29,26 +30,34 @@ vec_as_subscript <- function(i,
...
   .Call(
-    vctrs_as_subscript_result,
+    ffi_as_subscript_result,
     i = i,
     logical = logical,
     numeric = numeric,
     character = character,
-    arg = arg
+    arg = arg,
+    call = environment()
```
DavisVaughan, line 60: `frame`?

src/subscript-loc.c
```diff
@@ -222,336 +302,293 @@ r_obj* dbl_as_location(r_obj* subscript,
...
+
+  if (ffi_n == r_null && r_typeof(subscript) == R_TYPE_character) {
+    n = r_length(subscript);
+  } else {
+    if (r_is_object(ffi_n) || r_typeof(ffi_n) != R_TYPE_integer) {
+      ffi_n = vec_cast(ffi_n,
+                       vctrs_shared_empty_int,
+                       args_empty,
+                       args_empty,
+                       r_lazy_null);
```
DavisVaughan, line 353: Don't you have the `frame` here?

tests/testthat/_snaps/error-call.md
```diff
@@ -173,3 +173,54 @@
...
         * "..." at location 1.
 
+# subscript validation reports correct error calls
+
+    Code
+      (expect_error(my_function()))
+    Output
+      <error/rlang_error>
+      Error in `vctrs::num_as_location()`:
+      ! `missing` must be one of "propagate" or "error".
```
DavisVaughan, line 183: So it knows to report the vctrs function here rather than "in `my_function()`"? I guess that is right?

lionel-: yes this is the correct error call. See https://github.com/r-lib/vctrs/pull/1507#issuecomment-1022272169 for the implementation.

Issue #1150, opened 2020-06-15
Avoid eager evaluation of arg in vec_as_location*() and vec_recycle()


krlmlr:
for pretty error messages in tibble.

Do we need to replace `.Call()` by something different?

It would be great to move the `check_dots_empty()` check to C code too.

``` r
system.time(vctrs::vec_as_location(1, 1L, arg = Sys.sleep(1)))
#>    user  system elapsed 
#>   0.023   0.000   1.025
```

<sup>Created on 2020-06-15 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

lionel-:
This would require a new kind of `vctrs_arg` object that takes an environment and a symbol name.

krlmlr:
The way I see it we could use a technique similar to [`9ef15b6` (#998)](https://github.com/r-lib/rlang/pull/998/commits/9ef15b6238f0178a23a17175a2fef3c3097e26d2) at the front-end. Are you suggesting we need to introduce a new internal structure to pass along the environment and the name where to look up `arg`?

krlmlr:
Would you review a PR?

krlmlr:
Simpler reprex:

``` r
library(vctrs)

vec_as_subscript(1, arg = { print("oof"); deparse(quote(i)) })
#> [1] "oof"
#> [1] 1
```

<sup>Created on 2021-07-20 by the [reprex package](https://reprex.tidyverse.org) (v2.0.0)</sup>

The `deparse()` call is slow, it should be run only if there is an actual error.

PR #1508, opened 2022-01-27
Evaluate args lazily

krlmlr:
This removes calls to `vec_as_arg()` when the `frame` argument is already available. More complicated cases follow in separate PRs. 

Part of #1150.


lionel-:
Thanks!

PR #1509, opened 2022-01-27
More lazy evaluation of args

### LINKED ISSUE #1150: Avoid eager evaluation of arg in vec_as_location*() and vec_recycle()

krlmlr:
The remaining instances of `vec_as_arg()` seem very tricky and not worth it.

Closes #1150.

### DIFF
--- a/R/dictionary.R
+++ b/R/dictionary.R
@@ -260,7 +260,7 @@ vec_match <- function(needles,
                       needles_arg = "",
                       haystack_arg = "") {
   check_dots_empty0(...)
-  .Call(vctrs_match, needles, haystack, na_equal, needles_arg, haystack_arg)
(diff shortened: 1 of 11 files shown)


lionel-:
Thanks. This will be useful to switch to using `caller_arg()` systematically without incurring a performance cost.

PR #1511, opened 2022-01-28
Follow-up to lazy arg

lionel-:
Follow-up to #1484.

- Use common buffer filling implementation in wrapper args and lazy args.
- Use new `struct r_lazy` in lazy args

cc @krlmlr

### DIFF
--- a/src/rlang-dev.h
+++ b/src/rlang-dev.h
@@ -1,0 +1,36 @@
+#ifndef VCTRS_RLANG_DEV_H
+#define VCTRS_RLANG_DEV_H
+
+
+struct r_lazy {
+  r_obj* x;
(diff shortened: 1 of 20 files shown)


src/rlang-dev.h
```diff
@@ -0,0 +1,36 @@
+#ifndef VCTRS_RLANG_DEV_H
+#define VCTRS_RLANG_DEV_H
```
DavisVaughan, line 2: Is the idea that this is a holding place for ideas that might move to rlang?

lionel-: Exactly, it seems like `struct r_lazy` would be a good addition. I mostly added the file to solve a header deps issue.

PR #1518, opened 2022-02-10
Check list of vectors

lionel-:
Three new functions:

```r
export(vec_check_list)
export(vec_check_list_of_vectors)
export(vec_is_list_of_vectors)
```

I chose "check" instead of "assert" for consistency with the naming scheme we are adopting in other packages.

These checkers will be useful for the tidyr error discussed on Slack and are also used in `list_sizes()`.

```r
f <- function(x) vec_check_list_of_vectors(x)

f(env())
#> Error in `f()`:
#> ! `x` must be a list, not an environment.

f(list(1, env()))
#> Error in `f()`:
#> ! `x[[2]]` must be a vector, not an environment.

f(list(1, foo = env()))
#> Error in `f()`:
#> ! `x$foo` must be a vector, not an environment.
```

The numeric or character subscript is mentioned in the error message using a new `vctrs_arg` subclass. The constructor of this subclass is a bit different to other ctors in that it allocates on the heap, returns a pointer instead of a struct, and requires protection via a new `shelter` member. It uses C polymorphism to achieve that.

Heap allocation seemed better in this case to hide the data struct from the caller. I don't mind imposing the caller to allocate the data on the stack when it's a commonly used type like `struct r_lazy`, but I think it's more user friendly to hide it when it's an ad hoc struct. Also in principle the performance hit of heap allocation (gc triggers in particular) seems to matter less when the arg is used in functions that loop over data.

The interface seems neat to me. If the ctor returns a `struct vctrs_arg`, no need to protect. If it returns a pointer, protect its shelter.

cc @hadley


DESCRIPTION
```diff
@@ -29,6 +29,7 @@ BugReports: https://github.com/r-lib/vctrs/issues
 Depends: 
     R (>= 3.3)
 Imports:
+    cli (>= 3.1.1.9000),
```
DavisVaughan, line 32: I feel like this is worth a new bullet, saying something like "vctrs now imports cli to help with generation of useful error messages."

Normally I wouldn't mention something like this, but since vctrs is nearly "free", this seems worth mentioning

R/assert.R
```diff
@@ -203,6 +203,14 @@ vec_is_vector <- function(x) {
 #' * `x` is a bare list with no class.
 #' * `x` is a list explicitly inheriting from `"list"`.
 #'
+#' `vec_is_list_of_vectors()` checks in addition that all elements of
+#' `x` are vectors.
```
DavisVaughan, line 207:
```suggestion
#' `vec_is_list_of_vectors()` additionally checks that all elements of
#' `x` are vectors.
```

R/assert.R
```diff
@@ -218,6 +226,32 @@ vec_is_vector <- function(x) {
...
+vec_is_list_of_vectors <- function(x) {
+  .Call(ffi_is_list_of_vectors, x)
+}
+
+#' @rdname vec_is_list
+#' @export
+vec_check_list <- function(x,
+                           arg = caller_arg(x),
+                           call = caller_env()) {
+  .Call(ffi_check_list, x, environment())
```
DavisVaughan, line 240: Should this also be `invisible()` if the next one is?

R/assert.R
```diff
@@ -218,6 +226,32 @@ vec_is_vector <- function(x) {
...
+#' @rdname vec_is_list
+#' @export
+vec_is_list_of_vectors <- function(x) {
+  .Call(ffi_is_list_of_vectors, x)
+}
+
+#' @rdname vec_is_list
+#' @export
+vec_check_list <- function(x,
+                           arg = caller_arg(x),
```
DavisVaughan, line 238: I have gone ALL IN on requiring empty `...` between required and optional args in functions like these, and personally feel like it is the right way forward in general because it creates safer and more readable user/developer code.

DavisVaughan (commented):
I think my only worry is that these check functions won't get used because they might not _exactly_ match the error message that we want to throw - i.e. the package calling these functions might want to customize certain pieces of the error. I don't have a specific example in mind but it has historically been hard for me to create "generic" check functions because the error message I want to throw is highly context dependent.

The only other alternative I can think of is instead creating a function called `list_first_non_vector()` that returns `NA_integer_` if all elements are vectors, and otherwise returns an index to the first non-vector. Then the caller can use that information in their error message (and if the input was named they can do `names(input)[[index]]` themselves)

src/arg.c
```diff
@@ -140,6 +140,61 @@ r_ssize lazy_arg_fill(void* data_, char* buf, r_ssize remaining) {
...
+  if (i >= n) {
+    r_stop_internal("subscript_arg_fill", "`i` can't be greater than `vec_size(x)`.");
+  }
+
+  int len;
+  if (r_has_name_at(names, i)) {
+    len = snprintf(buf, remaining, "$%s", r_chr_get_c_string(names, i));
+  } else {
+    len = snprintf(buf, remaining, "[[%td]]", i + 1);
+  }
```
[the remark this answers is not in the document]
lionel-: I might also use it in dots when I add `arg` to `vec_cast_common()` and the ptype variant.

lionel-:
Now uses the naming scheme discussed on Slack. If `list_` functions are supplied a non-list, this is an internal error instead of a user error.

PR #1523, opened 2022-02-14
Check for negative and missing `n` in `new_data_frame()`

### LINKED ISSUE #1477: `new_data_frame(n = <negative>)` should fail
``` r
vctrs::new_data_frame(list(x=1), n = -1L)
#> [1] x
```

<sup>Created on 2021-10-25 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

From https://github.com/r-lib/vctrs/issues/949#issuecomment-950255378

### LINKED ISSUE #1500: `vec_compare()` with zero column data frames should treat them as equal
i.e. I think we are safe to remove this line
https://github.com/r-lib/vctrs/blob/f9afbff431da9d0b711db2b54be4c3ee925c6e5b/src/compare.c#L120

and the "correct" behavior will just fall out

``` r
# should be c(0L, 0L)?
vec_compare(data_frame(.size = 2), data_frame(.size = 2))
#> Error: `x` and `y` are not comparable: data frame with zero columns
```

DavisVaughan:
Closes #1477

### DIFF
--- a/src/type-data-frame.c
+++ b/src/type-data-frame.c
@@ -159,10 +159,19 @@ static R_len_t df_size_from_list(SEXP x, SEXP n) {
 
 static R_len_t df_size_from_n(SEXP n) {
   if (TYPEOF(n) != INTSXP || Rf_length(n) != 1) {
-    Rf_errorcall(R_NilValue, "`n` must be an integer of size 1");
+    r_abort("`n` must be an integer of size 1.");
   }
 
-  return r_int_get(n, 0);
+  R_len_t out = r_int_get(n, 0);
+
+  if (out == r_globals.na_int) {
+    r_abort("`n` can't be missing.");
+  }
+  if (out < 0) {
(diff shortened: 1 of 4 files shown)


src/type-data-frame.c
```diff
@@ -159,10 +159,19 @@ static R_len_t df_size_from_list(SEXP x, SEXP n) {
 
 static R_len_t df_size_from_n(SEXP n) {
   if (TYPEOF(n) != INTSXP || Rf_length(n) != 1) {
-    Rf_errorcall(R_NilValue, "`n` must be an integer of size 1");
+    r_abort("`n` must be an integer of size 1.");
   }
 
-  return r_int_get(n, 0);
+  R_len_t out = r_int_get(n, 0);
```
DavisVaughan, line 165: We will need a larger refresh of this file at some point, but not thinking about that yet

PR #1525, opened 2022-02-14
Add `.arg` argument to `vec_ptype_common()` and `vec_cast_common()`

lionel-:
```r
manip <- function(xs, arg = caller_arg(xs)) {
  vec_ptype_common(!!!xs, .arg = arg)
}

manip(list(1, list()), arg = "foo")
#> Error in `manip()`:
#> ! Can't combine `foo[[1]]` <double> and `foo[[2]]` <list>.

manip(list(x = 1, y = list()), arg = "foo")
#> Error in `manip()`:
#> ! Can't combine `foo$x` <double> and `foo$y` <list>.
```

### DIFF
--- a/src/bind.c
+++ b/src/bind.c
@@ -107,6 +107,7 @@ static SEXP vec_rbind(SEXP xs,
                               ptype,
                               DF_FALLBACK_DEFAULT,
                               S3_FALLBACK_true,
+                              args_empty,
                               r_lazy_null);
   PROTECT_N(xs, &n_prot);
 
--- a/src/bind.c
+++ b/src/bind.c
@@ -64,6 +64,7 @@ static SEXP vec_rbind(SEXP xs,
                                   ptype,
                                   DF_FALLBACK_DEFAULT,
                                   S3_FALLBACK_true,
+                                  args_empty,
                                   r_lazy_null);
   PROTECT_N(ptype, &n_prot);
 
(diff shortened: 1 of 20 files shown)


hadley:
Looks good!

src/bind.c
```diff
@@ -376,6 +378,7 @@ static SEXP vec_cbind(SEXP xs, SEXP ptype, SEXP size, struct name_repair_opts* n
                                               ptype,
                                               DF_FALLBACK_DEFAULT,
                                               S3_FALLBACK_false,
+                                              args_empty,
```
DavisVaughan, line 381: So I guess `arg` could be added to `vec_rbind()`, `vec_cbind()`, and `vec_unchop()` too? At least in theory

src/size-common.c
```diff
@@ -40,7 +40,7 @@ static
 SEXP vctrs_size2_common(SEXP x, SEXP y, struct counters* counters, void* data);
 
 r_ssize vec_size_common(r_obj* xs, r_ssize absent) {
-  r_obj* common = KEEP(reduce(R_NilValue, args_empty, xs, &vctrs_size2_common, NULL));
+  r_obj* common = KEEP(reduce(R_NilValue, args_empty, NULL, xs, &vctrs_size2_common, NULL));
```
DavisVaughan, line 43: And here too in `vec_size_common()`. I feel like this one would make sense. Maybe `vec_recycle_common()` too since I often do things like `args <- vec_cast_common(!!!args); args <- vec_recycle_common(!!!args)` back to back and it would be nice to be able to pass an arg name through both

src/bind.c
lionel-: Yes in theory, though I think I'd prefer to introduce `df_cast_common()` first if needed.

src/size-common.c
lionel-: yeah `vec_size_common()` and `vec_recycle_common()` need custom error contexts as well, I'll do that.

PR #1527, opened 2022-02-15
Add `caller_arg()` defaults to input validators

lionel-:
This might cause unhelpful arg tags being displayed in error messages. I think we should make this change now while vctrs users are also focusing on proper error calls for rlang 1.0, so that both argument tags and error calls can be updated at the same time.

cc @hadley @DavisVaughan @krlmlr

### DIFF
--- a/tests/testthat/_snaps/assert.md
+++ b/tests/testthat/_snaps/assert.md
@@ -179,14 +179,14 @@
     Output
       <error/vctrs_error_cast_lossy>
       Error in `vec_assert()`:
-      ! Can't convert from `size` <double> to `integer()` <integer> due to loss of precision.
+      ! Can't convert from `size` <double> to <integer> due to loss of precision.
       * Locations: 1
     Code
       (expect_error(vec_assert(1, size = "x")))
     Output
       <error/vctrs_error_incompatible_type>
       Error in `vec_assert()`:
-      ! Can't convert `size` <character> to match type of `integer()` <integer>.
+      ! Can't convert `size` <character> to <integer>.
 
 # list_all_vectors() works
 
--- a/tests/testthat/_snaps/assert.md
+++ b/tests/testthat/_snaps/assert.md
@@ -179,14 +179,14 @@
     Output
       <error/vctrs_error_cast_lossy>
       Error in `vec_assert()`:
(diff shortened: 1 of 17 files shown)


tests/testthat/_snaps/assert.md
```diff
@@ -179,14 +179,14 @@
     Output
       <error/vctrs_error_cast_lossy>
       Error in `vec_assert()`:
-      ! Can't convert from `size` <double> to <integer> due to loss of precision.
+      ! Can't convert from `size` <double> to `integer()` <integer> due to loss of precision.
```
DavisVaughan, line 182: I don't love this one. I think it was much nicer when we just had `to = ""` which resulted in no arg being displayed here

tests/testthat/_snaps/cast.md
```diff
@@ -12,15 +12,15 @@
...
+      ! Can't convert `foo` <double> to match type of `""` <character>.
 
 # cast errors create helpful messages (#57, #225)
 
     Code
       vec_cast(1.5, 10L)
     Condition
       Error:
-      ! Can't convert from <double> to <integer> due to loss of precision.
+      ! Can't convert from `1.5` <double> to `10L` <integer> due to loss of precision.
```
DavisVaughan, line 23: I think what I don't love about this is that the message makes me think I am trying to cast to the _value_ of `10L`, when really I am trying to cast to the _type_ of `10L` which is <integer>

DavisVaughan: Oh it looks like "to match type of `10`" is used below, so may it is just the lossy cast message that needs an update?

DavisVaughan: Maybe:

```
Can't convert `1.5` <double> to match type of `10L` <integer> due to loss of precision.
```

Which would be more consistent with the general cast error anyways

tests/testthat/_snaps/assert.md
DavisVaughan: See next comment about updating the lossy cast message

lionel-: These are just unit tests, they don't reflect how `vec_cast()` is supposed to be used.

PR #1528, opened 2022-02-15
Improve pluralisation in OOB errors

lionel-:
- Fix a pluralisation bug.
- Use cli for pluralisation and collapsing.
- Use more compact info bullet instead of two bullets. It uses the same sort of structure as ``"`foo` must be a bar, not a baz."``.

Before:

```r
vec_assign(1, 2, 0)
#> Error:
#> ! Can't assign to elements that don't exist.
#> ✖ Location 2 doesn't exist.
#> ℹ There are only 1 element.
```

After

```r
vec_assign(1, 2, 0)
#> Error:
#> ! Can't assign to elements that don't exist.
#> ℹ Location 2 doesn't exist, there is only 1 element.
```

### DIFF
--- a/R/subscript-loc.R
+++ b/R/subscript-loc.R
@@ -490,7 +490,8 @@ cnd_body_vctrs_error_subscript_oob_location <- function(cnd, ...) {
 
   # TODO: Switch to `format_inline()` and format bullets lazily through rlang
   cli::format_error(c(
-    "i" = "{cli::qty(n_loc)} Location{?s} {oob_enum} do{?esn't/n't} exist, there {cli::qty(n)} {?is/are} only {elt}."
+    "i" = "{cli::qty(n_loc)} Location{?s} {oob_enum} do{?esn't/n't} exist.",
+    "i" = "There {cli::qty(n)} {?is/are} only {elt}."
   ))
 }
 cnd_body_vctrs_error_subscript_oob_name <- function(cnd, ...) {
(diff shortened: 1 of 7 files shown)


R/subscript-loc.R
```diff
@@ -483,19 +482,15 @@ cnd_body_vctrs_error_subscript_oob_location <- function(cnd, ...) {
...
-      "There are only {cnd$size} {elt[[1]]}.",
-      "There are only {cnd$size} {elt[[2]]}.",
-    ))
+  n_loc <- length(oob)
+  n <- cnd$size
+  elt <- cnd_subscript_element_cli(n, cnd)
+
+  # TODO: Switch to `format_inline()` and format bullets lazily through rlang
+  cli::format_error(c(
+    "i" = "{cli::qty(n_loc)} Location{?s} {oob_enum} do{?esn't/n't} exist, there {cli::qty(n)} {?is/are} only {elt}."
```
hadley, line 493:
```suggestion
    "i" = "{cli::qty(n_loc)} Location{?s} {oob_enum} do{?esn't/n't} exist; there {cli::qty(n)} {?is/are} only {elt}."
```

hadley (approved):

R/subscript-loc.R
lionel-: As discussed on Slack, now uses two bullets as before, and the "past the end" phrasing.

PR #1529, opened 2022-02-15
`AsIs` lists are treated as lists

### LINKED ISSUE #1463: `vec_is_list()` should recognize `I(list())` as a list
This will probably require teaching `class_type()` about the `"AsIs"` class type so we can check for it here:
https://github.com/r-lib/vctrs/blob/81e32472a575a212539bdc9f6b8447e429635f29/src/type-info.c#L122

Check that everything needed to support `unnest()` with AsIs lists works as well
https://github.com/tidyverse/tidyr/issues/1162

This is a bit of an ad hoc solution for this one class type, as we are trying to treat it like a wrapper / mixin where the original class retains all of its original behavior.

lionel-:
Closes #1463

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,7 @@
 # vctrs (development version)
 
+* `vec_is_list()` now returns `TRUE` for `AsIs` lists (#1463).
+
 * `vec_assert()`, `vec_ptype2()`, `vec_cast()`, and `vec_as_location()`
   now use `caller_arg()` to infer a default `arg` value from the
   caller.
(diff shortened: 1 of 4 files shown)


lionel-:
With this fix

```r
df <- data.frame(a = 1:2, b = I(list(3:4, 5:6)))
tidyr::unnest(df, "b")
#> # A tibble: 4 × 2
#>       a     b
#>   <int> <int>
#> 1     1     3
#> 2     1     4
#> 3     2     5
#> 4     2     6
```

PR #1531, opened 2022-02-15
Pass error call through in `vec_init()`

DavisVaughan:
This builds on top of previous `vec_init()` work. We now pass the error call all the way through to checks like `vec_check_vector()` and the check that ensures that `n` is positive. We will also eventually pass the error call through to the `n > R_LEN_T_MAX` check that I will add in a follow up PR

### DIFF
--- a/src/bind.c
+++ b/src/bind.c
@@ -124,7 +124,7 @@ static SEXP vec_rbind(SEXP xs,
   }
 
   PROTECT_INDEX out_pi;
-  SEXP out = vec_init(proxy, n_rows);
+  SEXP out = vec_init(proxy, n_rows, r_lazy_null);
   PROTECT_WITH_INDEX(out, &out_pi);
   ++n_prot;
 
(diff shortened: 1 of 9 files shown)


lionel- (approved):
Looks good, thanks for helping out with this!

If you don't mind, the next step would be to take a `struct vec_init_opts` that bundles `x` and `n` args (see `vec_slice_opts()` for an example). And then you can use `vec_format_arg()` from #1533 in your abort calls to include the relevant args in error messages.

lionel-:
oops sorry I forgot about this. I went with another approach in the end, `vec_init()` is hard-coded for now:

```c
  vec_check_vector(x, vec_args.x, lazy_calls.vec_init);

  if (n < 0) {
    r_abort_lazy_call(lazy_calls.vec_init,
                      "%s must be a positive integer.",
                      r_c_str_format_error_arg("n"));
  }
```

PR #1533, opened 2022-02-16
Add error helpers on the C side

lionel-:
- `vec_arg_format()` which takes a vctrs arg and returns a vmax-protected C string formatted by cli via `r_format_error_arg()`.

- `r_c_str_format_error_arg()` which is like `r_format_error_arg()` but takes a C string. The result is also vmax-protected.

- `r_friendly_type_of()` in `rlang-dev.c` which returns a friendly type as a vmax-protected string. I've also updated the compat file that supports it.

This should make it easier to create well formatted errors from C, using `r_abort_call()` and `r_abort_lazy_call()`.


lionel-:
Merging now but feel free to review the API @DavisVaughan

DavisVaughan (commented):
LGTM

PR #1534, opened 2022-02-16
Improve size checking

### LINKED ISSUE #1437: `vec_init(x, n)` should error more informatively about long vector support
This seems to be one of the most common places where these errors come up, and they are fairly confusing

``` r
library(vctrs)

vec_init(1, n = .Machine$integer.max + 1)
#> Error: Can't convert from `n` <double> to <integer> due to loss of precision.
#> • Locations: 1
```

I think internally we should just use `r_arg_as_ssize(n, "n")` in `vctrs_init()` rather than `vec_cast()` (I don't expect the user to input much here besides an integer or integerish double so casting seems overkill). Then we can error immediately after that if `n > R_LEN_T_MAX` stating that we don't allow long vectors right now

lionel-:
- New `vec_as_ssize()` adapted from `r_arg_as_ssize()`. It takes `arg` and `call` arguments. It now casts the input to double and checks for `R_SSIZE_MAX`.

- New `vec_as_short_length()` which calls `vec_as_ssize()` and checks for `INT_MAX` and negative values. (ssize can be negative). When we switch to long vector support, we'll use a new `vec_as_length()` wrapper.

- `vec_init()` uses `vec_as_short_length()`, closes #1437.

- To support the casting, `vec_cast_e()` now catches `vctrs_error_incompatible_type` instead of lossy cast errors. To make this work, lossy cast errors are now a subclass of cast errors. This changes the internal structure a bit but a quick CRAN search suggest this shouldn't cause issues (to be confirmed in revdep checks).

### DIFF
--- a/R/conditions.R
+++ b/R/conditions.R
@@ -24,7 +24,7 @@
 #' # Most of the time, `maybe_lossy_cast()` returns its input normally:
 #' maybe_lossy_cast(
 #'   c("foo", "bar"),
-#'   NULL,
+#'   NA,
 #'   "",
 #'   lossy = c(FALSE, FALSE),
 #'   x_arg = "",
@@ -34,7 +34,7 @@
 #' # If `lossy` has any `TRUE`, an error is thrown:
 #' try(maybe_lossy_cast(
 #'   c("foo", "bar"),
-#'   NULL,
+#'   NA,
 #'   "",
 #'   lossy = c(FALSE, TRUE),
 #'   x_arg = "",
@@ -45,7 +45,7 @@
 #' allow_lossy_cast(
 #'   maybe_lossy_cast(
 #'     c("foo", "bar"),
-#'     NULL,
+#'     NA,
 #'     "",
 #'     lossy = c(FALSE, TRUE),
 #'     x_arg = "",
@@ -469,7 +469,9 @@ maybe_lossy_cast <- function(result, x, to,
     )
   )
 }
-stop_lossy_cast <- function(x, to, result,
+stop_lossy_cast <- function(x,
+                            to,
+                            result,
                             locations = NULL,
                             ...,
                             loss_type,
@@ -479,10 +481,8 @@ stop_lossy_cast <- function(x, to, result,
                             message = NULL,
                             class = NULL,
                             call = caller_env()) {
-  stop_vctrs(
-    message,
+  stop_incompatible_cast(
     x = x,
-    y = to,
     to = to,
     result = result,
     locations = locations,
@@ -512,7 +512,7 @@ conditionMessage.vctrs_error_cast_lossy <- function(c) {
 #' @export
 cnd_header.vctrs_error_cast_lossy <- function(cnd, ...) {
   x_label <- format_arg_label(vec_ptype_full(cnd$x), cnd$x_arg)
-  to_label <- format_arg_label(vec_ptype_full(cnd$to), cnd$to_arg)
+  to_label <- format_arg_label(vec_ptype_full(cnd$y), cnd$y_arg)
   loss_type <- loss_type(cnd$loss_type)
   glue::glue("Can't convert from {x_label} to {to_label} due to loss of {loss_type}.")
 }
@@ -536,12 +536,12 @@ loss_type <- function(x) {
 # Used in maybe_warn_deprecated_lossy_cast()
 new_error_cast_lossy <- function(x, to, loss_type, x_arg = "", to_arg = "") {
   error_cnd(
-    "vctrs_error_cast_lossy",
+    c("vctrs_error_cast_lossy", "vctrs_error_incompatible_type"),
     x = x,
-    to = to,
+    y = to,
     loss_type = loss_type,
     x_arg = x_arg,
-    to_arg = to_arg
+    y_arg = to_arg
   )
 }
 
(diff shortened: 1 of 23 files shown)


R/conditions.R
```diff
@@ -536,12 +536,12 @@ loss_type <- function(x) {
...
+    c("vctrs_error_cast_lossy", "vctrs_error_incompatible_type"),
     x = x,
-    to = to,
+    y = to,
     loss_type = loss_type,
     x_arg = x_arg,
-    to_arg = to_arg
+    y_arg = to_arg
   )
 }
```
DavisVaughan, line 546: It seems like it may be worth it to have `new_error_incompatible_type()` that this could use? To try to prevent them from ever getting out of sync?

src/rlang-dev.c
```diff
@@ -1,8 +1,19 @@
...
+
+const char* r_friendly_type_of_length(r_obj* x) {
+  return r_friendly_type_of_opts(x, true, true);
+}
+
+const char* r_friendly_type_of_opts(r_obj* x,
+                                    bool value,
+                                    bool length) {
+  r_obj* call = KEEP(r_parse("friendly_type_of(x, value = y, length = z)"));
+  r_obj* ffi_out = KEEP(r_eval_with_xyz(call, x, r_lgl(value), r_lgl(length), vctrs_ns_env));
```
DavisVaughan, line 16: Do the `r_lgl()`s need protection?

src/size.c
```diff
@@ -233,29 +233,99 @@ r_obj* vec_recycle_fallback(r_obj* x,
...
+                     struct r_lazy call) {
+  if (r_is_object(n)) {
+    struct cast_opts cast_opts = {
+      .x = n,
+      .to = r_globals.empty_dbl,
+      .x_arg = p_arg,
+      .call = call
+    };
+    ERR err = NULL;
+    n = vec_cast_e(&cast_opts, &err);
```
DavisVaughan, line 278: The lack of protection here makes me nervous. In `invalid`, I think `vec_arg_format()` can allocate right? Which is called before `r_friendly_type_of_length(n)` is called, which could cause issues?

src/size.c
```diff
@@ -233,29 +233,99 @@ r_obj* vec_recycle_fallback(r_obj* x,
...
+    }
+    if (out != floor(out)) {
+      r_abort_lazy_call(call,
+                        "%s must be a whole number, not a fractional number.",
+                        vec_arg_format(p_arg));
+    }
+
+    if (out > R_SSIZE_MAX) {
+      r_abort_lazy_call(call,
+                        "%s is too large a number and long vectors are not supported.",
```
DavisVaughan, line 304: I don't think this is about long vectors. I think this exceeds even the long vector size.

src/size.c
```diff
@@ -233,29 +233,99 @@ r_obj* vec_recycle_fallback(r_obj* x,
...
+                        vec_arg_format(p_arg));
+    }
+
+    return (r_ssize) floor(out);
```
DavisVaughan, line 308: No need to floor again?

src/size.c
```diff
@@ -233,29 +233,99 @@ r_obj* vec_recycle_fallback(r_obj* x,
...
+  }
+
+  case R_TYPE_integer: {
+    if (r_length(n) != 1) {
+      goto invalid;
+    }
+    int out = (r_ssize) r_int_get(n, 0);
```
DavisVaughan, line 315: Probably keep this as `int` for the check against `r_globals.na_int` then `(r_ssize)` cast at the end, like the double branch?

src/size.c
```diff
@@ -233,29 +233,99 @@ r_obj* vec_recycle_fallback(r_obj* x,
...
+
+    if (out == r_globals.na_int) {
+      goto invalid;
+    }
+
+    return out;
+  }
+
+  invalid:
+  default:
```
DavisVaughan, line 325: Neat that you can `goto` a spot that is in between cases of a switch statement

src/size.c
```diff
@@ -233,29 +233,99 @@ r_obj* vec_recycle_fallback(r_obj* x,
...
+    r_abort_lazy_call(call,
+                      "%s must be a single number, not %s.",
+                      vec_arg_format(p_arg),
+                      r_friendly_type_of_length(n));
```
DavisVaughan, line 329: This is where I was worried about the lack of protection on `n`

R/conditions.R
lionel-: yes but that requires too many changes in conditions.R

src/size.c
lionel-: do you mean we should differentiate between short-max < long-max < too-large?

DavisVaughan: Yes, I mean that this error is about the number being too large for R altogether. It doesn't need to mention long vectors because that implies that if we _did_ support long vectors then we could handle an `out` of this size, which is not true.

src/rlang-dev.c
DavisVaughan: Is this because `r_lgl()` uses `R_ScalarLogical()`, which doesn't allocate at all?

I didn't know this was the definition of that, which is neat https://github.com/wch/r-source/blob/79298c499218846d14500255efd622b5021c10ec/src/include/Rinlinedfuns.h#L992-L998

src/size.c
lionel-: What do you suggest we do? You recommended to mention lack of long vector support in the linked issue, I thought this was the reason for improving the error message.

I'm not particularly worried about implying that gigantic lengths could be supported if with long vectors. With unlimited memory we support sizes up to `R_SSIZE_MAX`.

lionel-: Sorry it must be late, this is for sizes beyond ssize-max, which are not supported. And we still mention long vector support in the short-length path.

DavisVaughan: > You recommended to mention lack of long vector support in the linked issue

And you do that! You do it in `vec_as_short_length()` which I think is the right place to mention long vectors

But here in `vec_as_ssize()` it seems like `%s is too large a number` is enough and is the most appropriate message

PR #1535, opened 2022-02-16
Take `.call` and `.arg` arguments in `vec_size_common()` and `vec_recycle_common()`

lionel-:
Branched from #1534.

### DIFF
--- a/tests/testthat/_snaps/size.md
+++ b/tests/testthat/_snaps/size.md
@@ -61,13 +61,13 @@
       (expect_error(vec_size_common(.size = "foo")))
     Output
       <error/rlang_error>
-      Error:
+      Error in `vec_size_common()`:
       ! `.size` must be a single number, not a string.
     Code
       (expect_error(vec_size_common(.size = 1:2)))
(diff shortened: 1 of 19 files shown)


tests/testthat/_snaps/size.md
```diff
@@ -61,22 +61,31 @@
...
       <error/rlang_error>
-      Error in `vec_size_common()`:
+      Error:
       ! `.size` must be a single number, not a string.
     Code
       (expect_error(vec_size_common(.size = 1:2)))
     Output
       <error/rlang_error>
-      Error in `vec_size_common()`:
+      Error:
```
DavisVaughan, line 70: I think this was expected right? It no longer reports itself because of that `.call = caller_env()` argument?

DavisVaughan (approved):

tests/testthat/_snaps/size.md
lionel-: well this is a case of internal error.

Thanks for being patient with me today :D

PR #1544, opened 2022-02-24
Use `R_PRI_SSIZE` in favor of `ld` format specifier

DavisVaughan:
Hopefully this avoids the warnings on Windows

### DIFF
--- a/src/arg.c
+++ b/src/arg.c
@@ -197,7 +197,7 @@ r_ssize subscript_arg_fill(void* p_data_, char* buf, r_ssize remaining) {
   r_ssize n = p_data->n;
 
   if (i >= n) {
-    r_stop_internal("`i = %ld` can't be greater than `vec_size(x) = %ld`.",
+    r_stop_internal("`i = %" R_PRI_SSIZE "` can't be greater than `vec_size(x) = %" R_PRI_SSIZE "`.",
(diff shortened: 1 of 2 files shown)


DavisVaughan:
I just manually confirmed that this does resolve the warnings of the following form on both 32-bit and 64-bit Windows:

```c
  "C:/rtools40/mingw64/bin/"gcc  -I"C:/PROGRA~1/R/R-40~1.0/include" -DNDEBUG -I./rlang         -O2 -Wall  -std=gnu99 -mfpmath=sse -msse2 -mstackrealign -c arg.c -o arg.o
   arg.c: In function 'counter_arg_fill':
   arg.c:270:43: warning: format '%ld' expects argument of type 'long int', but argument 4 has type 'r_ssize' {aka 'long long int'} [-Wformat=]
          len = snprintf(buf, remaining, "[[%ld]]", i + 1);
                                            ~~^     ~~~~~
                                            %lld
   arg.c:276:43: warning: format '%ld' expects argument of type 'long int', but argument 4 has type 'r_ssize' {aka 'long long int'} [-Wformat=]
          len = snprintf(buf, remaining, "..%ld", i + 1);
                                            ~~^   ~~~~~
                                            %lld
```

PR #1558, opened 2022-05-05
Rewrite `vec_joint_proxy_order()` with a more conservative heuristic

DavisVaughan:
There are two issues that can arise with the way `vec_joint_proxy_order()` works, both of which have to do with the fact that the "dependent" joint proxy is only ever computed for vectors that inherit from `"list"` explicitly.

- sfc vectors, which are list-columns without an explicit `"list"` class, pass through the proxy method and return a list, which is not a supported type in `vec_order()`. Ideally this would just forward to `vec_proxy_order.list()` through a `vec_proxy_order.sfc()` method, but then you'd also need the "dependent" joint proxy path to run, and it currently won't do that because it only runs for lists.
- bignum vectors, like biginteger, have a `vec_proxy_order()` method defined for them that depends on the data in the vector itself - it computes the ranks and uses that as the order proxy. This means that this also needs to run through the "dependent" joint proxy path (i.e. combine `x` and `y` together, take the proxy, split them apart)

I have fixed this by changing the way `vec_joint_proxy_order()` decides whether or not to combine `x` and `y` together before taking the proxy. We are now _much_ more conservative. The only way we ever take the proxies "independently" (i.e. don't combine them together) is if we are working with a base R atomic type (except list) or a base R S3 type that we support in vctrs (factor, date, etc). Otherwise we always fallback to a "dependent" joint proxy that combines them, because we have no way to know if the `vec_proxy_order()` method is dependent on the data or not.

That automatically fixed bignum, and then I just had to add a `vec_proxy_order.sfc()` method to fix sf. It should be more compatible with other classes in the long run as well.

The downside is that it:
- Is slower than the independent method because you have to combine `x` and `y` and then re-split them
- Limits the maximum size to `vec_size(x) + vec_size(y) <= INT_MAX`

But this only ever comes into play when a user defined S3 type is used, so I'm hoping that is acceptable.

Here are some examples of sfc and bignum problems which have been fixed with this:

```r
library(sf)
#> Linking to GEOS 3.9.1, GDAL 3.4.0, PROJ 8.1.1; sf_use_s2() is TRUE
library(vctrs)
​
x <- c(
  st_geometry(st_point(c(0, 0))),
  st_geometry(st_point(c(0, 1))),
  st_geometry(st_point(c(2, 1)))
)
​
y <- c(
  st_geometry(st_point(c(0, 1))),
  st_geometry(st_point(c(0, 0))),
  st_geometry(st_point(c(0, 3))),
  st_geometry(st_point(c(0, 1)))
)
​
x
#> Geometry set for 3 features 
#> Geometry type: POINT
#> Dimension:     XY
#> Bounding box:  xmin: 0 ymin: 0 xmax: 2 ymax: 1
#> CRS:           NA
#> POINT (0 0)
#> POINT (0 1)
#> POINT (2 1)
y
#> Geometry set for 4 features 
#> Geometry type: POINT
#> Dimension:     XY
#> Bounding box:  xmin: 0 ymin: 0 xmax: 0 ymax: 3
#> CRS:           NA
#> POINT (0 1)
#> POINT (0 0)
#> POINT (0 3)
#> POINT (0 1)
​
# Geometry sets are `sfc`s, or "sf columns", which are documented as list columns
# but they don't explicitly have a "list" class
class(x)
#> [1] "sfc_POINT" "sfc"
​
# The old way `left_join()` and friends worked
vec_match(x,y)
#> [1]  2  1 NA
​
# The new way
vec_locate_matches(x,y)
#> Error: This type is not supported by `vec_order()`.
```

```r
library(vctrs)
library(bignum)

x <- biginteger(5)
y <- biginteger(6)

# uh oh
vec_locate_matches(x, y)
#>   needles haystack
#> 1       1        1

vec_proxy_order(x)
#> [1] 1
vec_proxy_order(y)
#> [1] 1

bignum:::vec_proxy_order.bignum_biginteger
#> function (x, ...) 
#> {
#>     c_biginteger_rank(x)
#> }
#> <bytecode: 0x7fe647e42dc8>
#> <environment: namespace:bignum>
```

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,14 @@
 # vctrs (development version)
 
+* `vec_locate_matches()` now uses a more conservative heuristic when taking the
+  joint ordering proxy. This allows it to work correctly with sf's sfc vectors
+  and the classes from the bignum package (#1558).
+  
+* An sfc method for `vec_proxy_order()` was added to better support the sf
+  package. These vectors are generally treated like list-columns even though
+  they don't explicitly have a `"list"` class, and the `vec_proxy_order()`
+  method now forwards to the list method to reflect that (#1558).
+
 * `vec_proxy_compare()` now works correctly for raw vectors wrapped in `I()`.
   `vec_proxy_order()` now works correctly for raw and list vectors wrapped in
   `I()` (#1557).
@@ -11,7 +20,6 @@
 
 * Fixed memory protection issues related to common type
   determination (#1551, tidyverse/tidyr#1348).
-
 
 # vctrs 0.4.0
 
--- a/R/type-sf.R
+++ b/R/type-sf.R
@@ -128,6 +128,12 @@ vec_cast.data.frame.sf = function(x, to, ...) {
 	df_cast(x, to, ...)
 }
 
+vec_proxy_order.sfc <- function(x, ...) {
+  # These are list columns, so they need to use the order-by-appearance proxy
+  # that is defined by `vec_proxy_order.list()`
+  x <- unstructure(x)
+  vec_proxy_order(x)
+}
 
 # take conservative approach of requiring equal CRS and precision
 common_crs = function(x, y) {
--- a/R/zzz.R
+++ b/R/zzz.R
@@ -89,6 +89,9 @@
       s3_register("vctrs::vec_cast", "sf.data.frame")
       s3_register("vctrs::vec_cast", "data.frame.sf")
     }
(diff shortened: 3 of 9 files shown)


DavisVaughan:
A revdepcheck of dplyr with this installed shows that it fixes the sfnetworks failure that was due to the sfc problem, and caused no other issues, so I feel pretty good about it.

MontanaCabana:
please take me off this distribution list

lionel-:
@MontanaCabana Only you can do that. Login and click the Unwatch button at the top.

Issue #1549, opened 2022-03-25
vec_match() fails on class "Date" and "IDate" with obscure error


beansrowning:
I've noticed in several of my projects where I sometimes use data.table and dplyr together that an obscure error can arise if you attempt to perform a mutating join on a Date formated column with an inferred IDate/Date column as read in by `data.table::fread()`

```
Error: Can't join on `x$date` x `y$date` because of incompatible types.
i `x$date` is of type <date>>.
i `y$date` is of type <date>>.
Run `rlang::last_error()` to see where the error occurred.
```

This error was unintelligible to me and took a while to debug. I was able to browse into dplyr:::join_rows() and saw the offending line was the [vec_match() call on line 7](https://github.com/tidyverse/dplyr/blob/6d22c0f0e1b0d945f2fff28cecb71d9cd548098b/R/join-rows.R#L7), yet the join columns of both were reported to be of type "Date" via the error and print.tbl_df:

![image](https://user-images.githubusercontent.com/25888380/160167593-c17cff99-cdd4-41c8-b581-13dcb8c978cb.png)
![image](https://user-images.githubusercontent.com/25888380/160167622-e44f8b05-4975-4d00-b346-bd52afd315e5.png)

I finally realized the column in the tibble originally sourced from fread() was of class "IDate", "Date". Which seems to trip vec_match()

![image](https://user-images.githubusercontent.com/25888380/160165371-da8250a6-ab33-43c0-8239-c0a4423077ee.png)
![image](https://user-images.githubusercontent.com/25888380/160165422-07687802-08b6-4735-a751-7b8ec607f5a7.png)

I'm not sure what solution makes sense to make this more explicit to the end user, but as it stands it's hard to reason through.  

### Reprex
``` r
library(dplyr)
#> 
#> Attaching package: 'dplyr'
#> The following objects are masked from 'package:stats':
#> 
#>     filter, lag
#> The following objects are masked from 'package:base':
#> 
#>     intersect, setdiff, setequal, union
library(data.table)
#> 
#> Attaching package: 'data.table'
#> The following objects are masked from 'package:dplyr':
#> 
#>     between, first, last
library(readr)

a <- tibble(
  a = letters,
  b = seq(Sys.Date(), Sys.Date() + 25, "days")
)

print(a)
#> # A tibble: 26 x 2
#>    a     b         
#>    <chr> <date>    
#>  1 a     2022-03-25
#>  2 b     2022-03-26
#>  3 c     2022-03-27
[...]

PR #1571, opened 2022-06-08
Don't inherit `vec_ptype_abbr()` and `vec_ptype_full()` methods

lionel-:
Part of #1549.

```r
x <- Sys.Date()
y <- data.table::as.IDate(x)

vec_ptype2(x, y)
#> Error:
#> ! Can't combine `x` <date> and `y` <IDate>.
```

cc @hadley


R/ptype-abbr-full.R
```diff
@@ -32,15 +32,25 @@
...
   check_dots_empty0(...)
-  UseMethod("vec_ptype_full")
+
+  # Data frames dispatch to the `data.frame` method by default to get
+  # the inner types format
+  method <- s3_method_specific(
+    x,
+    "vec_ptype_full",
+    ns = "vctrs",
+    df_default = TRUE
```
DavisVaughan, line 42: You use this for full but not abbr?

R/ptype-abbr-full.R
```diff
@@ -32,15 +32,25 @@
...
 }
 
 #' @export
 #' @rdname vec_ptype_full
 vec_ptype_abbr <- function(x, ..., prefix_named = FALSE, suffix_shape = TRUE) {
   check_dots_empty0(...)
 
-  abbr <- vec_ptype_abbr_dispatch(x)
+  method <- s3_method_specific(x, "vec_ptype_abbr", ns = "vctrs")
+  abbr <- method(x, ...)
```
DavisVaughan, line 53: I don't think this is quite right yet for data frames. The `vec_ptype_abbr()` default will call `vec_ptype_full()` and abbreviate it, which can result in some ugly results for things like data tables, which don't have a `vec_ptype_abbr()` method (unlike tibbles, provided by the tibble package).

``` r
library(vctrs)

df <- tibble::tibble(x = 1, y = 2, z = 3)

# tibble:::vec_ptype_full.tbl_df does not exist.
# uses the data frame fallback in `vec_ptype_full()` because `df_default = T`
vec_ptype_full(df)
#> [1] "tbl_df<\n  x: double\n  y: double\n  z: double\n>"

# tibble:::vec_ptype_abbr.tbl_df exists so this is nice
vec_ptype_abbr(df)
#> [1] "tibble[,3]"

library(data.table)

dt <- data.table(x = 1, y = 2, z = 3)

# again uses the data frame fallback
vec_ptype_full(dt)
#> [1] "data.table<\n  x: double\n  y: double\n  z: double\n>"

# there is no data table method for vec_ptype_abbr, so it falls through to
# vec_ptype_abbr.default, which abbreviates the vec_ptype_full result in
# an ugly way
vec_ptype_abbr(dt)
#> [1] "dxdydzd>[,3]"
```

<sup>Created on 2022-06-08 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

R/ptype-abbr-full.R
```diff
@@ -32,15 +32,25 @@
 #' cat(vec_ptype_abbr(1:10))
 vec_ptype_full <- function(x, ...) {
   check_dots_empty0(...)
-  UseMethod("vec_ptype_full")
```
DavisVaughan, line 35: I swear we needed to keep `UseMethod()` in there at the end of the function because of something to do with R CMD check, even though it can't be called. Like how `vec_ptype()` is set up:

```r
> vec_ptype
function (x, ..., x_arg = "", call = caller_env()) 
{
    check_dots_empty0(...)
    return(.Call(ffi_ptype, x, x_arg, environment()))
    UseMethod("vec_ptype")
}
```

R/ptype-abbr-full.R
DavisVaughan: Would be great to see a data frame test or two added as well once this gets fixed

DavisVaughan (changes requested):
If it is ok with you, I'll let you make the changes to the data frame bits and then I'd like to take another look

R/ptype-abbr-full.R
lionel-: I think it's so that roxygen2 knows it's a generic.

R/ptype-abbr-full.R
lionel-: yes the data.frame method was guarded by `inherits_only()` already.

R/ptype-abbr-full.R
lionel-: Good catch. But isn't this a problem in general with the `is.object` branch of this default "abbr" method that first calls "full" and then abbreviate it?

lionel-: I don't see any broken test when I change the method to:

```r
vec_ptype_abbr.default <- function(x, ...) {
  if (is.object(x)) {
    type <- class(x)[[1]]
  } else if (is_vector(x)) {
    type <- vec_ptype_abbr_bare(x, ...)
  } else {
    abort("Not a vector.")
  }
  unname(abbreviate(type, 8))
}
```

We should probably do that.

lionel-:
Worth noting that this change breaks `NextMethod()` which could still be useful to reach the default methods. If needed, we could:

- Export the default methods.
- Add a `dispatch` or `default` argument to the generics to disable dispatch.

I did a quick CRAN search and did not find any use of `NextMethod()`, only recursive calls to the generics. There's already too many methods to look through them all so I'll just use revdepcheck to be sure.

Regardless, I think it's probably a good idea to add that `dispatch` argument to the generics.

R/ptype-abbr-full.R
```diff
@@ -74,26 +81,55 @@ vec_ptype_full.default <- function(x, ...) {
...
-      abbreviate(typeof(x))
-    )
+    type <- vec_ptype_abbr_bare(x, ...)
   } else {
     abort("Not a vector.")
   }
+  unname(abbreviate(type, 8))
 }
 
+vec_ptype_full_data_frame <- function(x, ...) {
```
DavisVaughan, line 96: Are the dots useful for anything here? If not, can we remove them?

src/utils.c
```diff
@@ -479,6 +479,19 @@ SEXP vctrs_s3_find_method(SEXP generic, SEXP x, SEXP table) {
   return s3_find_method(r_chr_get_c_string(generic, 0), x, table);
 }
 
+// [[ register() ]]
+r_obj* ffi_s3_get_method(r_obj* generic, r_obj* class, r_obj* table) {
+  if (!r_is_string(generic)) {
+    r_abort("`generic` must be a string");
```
DavisVaughan, line 485:
```suggestion
    r_stop_internal("`generic` must be a string");
```

?

src/utils.c
```diff
@@ -479,6 +479,19 @@ SEXP vctrs_s3_find_method(SEXP generic, SEXP x, SEXP table) {
...
+  }
+  if (!r_is_string(class)) {
+    r_abort("`class` must be a string");
```
DavisVaughan, line 488:
```suggestion
    r_stop_internal("`class` must be a string");
```

R/utils.R
```diff
@@ -168,6 +168,39 @@ s3_find_method <- function(x, generic, ns = "base") {
...
+    is_string(ns)
+  )
+  table <- ns_methods(ns_env(ns))
+  .Call(ffi_s3_get_method, generic, class, table)
+}
+
+s3_method_specific <- function(x,
+                               generic,
+                               ns = "base",
+                               df_default = FALSE,
```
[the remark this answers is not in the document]
lionel-: No opinion here, I'll remove it.

tests/testthat/_snaps/ptype-abbr-full.md
```diff
@@ -0,0 +1,24 @@
...
+    Output
+      [1] "vctrs_foobar<\n  x: double\n  y: character\n  z: logical\n>"
+
+---
+
+    Code
+      dt <- data.table::data.table(x = 1, y = 2, z = 3)
+      vec_ptype_abbr(dt)
+    Output
+      [1] "data.tbl[,3]"
```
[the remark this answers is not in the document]
lionel-: yup I thought abbreviating unknown data frames was more consistent with the documented behaviour of `vec_ptype_abbr()`, and more useful regarding its intended usage.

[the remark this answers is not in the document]
lionel-: Should we add a data.table method for the sake of our dt users? I don't imagine the data.table team will be interested in maintaining tidyverse integration in their package, so we could do it here in vctrs. In this case, I suppose we'd abbreviate to `"dt"`?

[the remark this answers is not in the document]
DavisVaughan: Seems reasonable since we have ptype2/cast methods for them

Maybe pull out a `type-data-table.R` file starting from here?
https://github.com/r-lib/vctrs/blob/e86bb9a4b57459eb56099c0b52af37877a9bf515/R/type-misc.R#L126

I also noticed we have some tests that were skipped because we weren't Suggesting data table until now, maybe we can update that too
https://github.com/r-lib/vctrs/blob/e86bb9a4b57459eb56099c0b52af37877a9bf515/tests/testthat/test-type-misc.R#L33

All of that together might be enough for a separate PR?

R/ptype-abbr-full.R
lionel-: I wrote these internal methods as if they were not internal.

PR #1578, opened 2022-06-23
Fix inconsistent reporting of conflicting inputs

### LINKED ISSUE #1570: Names are incorrect in `vec_ptype_common()` error messages when `NULL` is involved with splicing
If you use `NULL` inputs in combination with splicing or with the `{name} := value` behavior, then the error messages don't show the right names. I have a feeling the two incorrect cases below have the same root cause.

``` r
library(vctrs)

# Shows problem with `x` and `z` - this is good
vec_ptype_common(
  a = "foo",
  b = "bar",
  y = NULL,
  z = 1
)
#> Error:
#> ! Can't combine `a` <character> and `z` <double>.

args1 <- list(a = "foo", b = "bar")
args2 <- list(y = NULL, z = 1)

# Problem with `y` and `z` - this is wrong
vec_ptype_common(
  !!!args1,
  !!!args2
)
#> Error:
#> ! Can't combine `y` <character> and `z` <double>.

y_name <- "y"
z_name <- "z"

# Problem with `..4` and `z` - this is very wrong
vec_ptype_common(
  !!!args1,
  "{y_name}" := NULL,
  "{z_name}" := 1
)
#> Error:
#> ! Can't combine `..4` <character> and `z` <double>.
```

<sup>Created on 2022-06-07 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

lionel-:
Closes #1570

Following these changes, since we now flatten spliced inputs into simple lists of arguments, we can simplify the arg counter code and remove any logic that supported splice box walking. I'll look into this another time though.

### DIFF
--- a/tests/testthat/_snaps/type.md
+++ b/tests/testthat/_snaps/type.md
@@ -101,7 +101,7 @@
       vec_ptype_common(TRUE, !!!list(1, 2), "foo")
     Condition
       Error:
-      ! Can't combine `..2` <double> and `..5` <character>.
+      ! Can't combine `..2` <double> and `..4` <character>.
 
 ---
 
@@ -109,7 +109,7 @@
       vec_ptype_common(1, !!!list(TRUE, FALSE), "foo")
     Condition
       Error:
-      ! Can't combine `..1` <double> and `..5` <character>.
+      ! Can't combine `..1` <double> and `..4` <character>.
 
 ---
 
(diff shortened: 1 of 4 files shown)


tests/testthat/_snaps/type.md
```diff
@@ -125,7 +125,7 @@
       vec_ptype_common(foo = TRUE, !!!list(bar = 1, "foo"))
     Condition
       Error:
-      ! Can't combine `bar` <double> and `..3` <character>.
+      ! Can't combine `foo` <double> and `..3` <character>.
```
DavisVaughan, line 128: This is mildly confusing

lionel-: Agreed, this is a separate issue though. In some cases we could just refer to the latest input. It can get complicated because input 1 and 2 might create a new type, and then there is nothing to refer to if that new type is not compatible with input 3.

PR #1580, opened 2022-06-29
Add vector recycling rules doc page

### LINKED ISSUE #1579: Documentation topic that describes recycling rules
So that in other packages we could write (e.g.)

```R
#' @returns A character vector with size found by [recycling][vctrs_recycling_rules] `string`, `pattern`, 
  and `replacement`.
```

DavisVaughan:
Closes #1579


R/recycle.R
```diff
@@ -1,31 +1,125 @@
...
+#' throw an error if recycling is not possible. See [vec_size()] for the precise
+#' definition of size.
 #'
 #' @inheritParams rlang::args_error_context
 #'
 #' @param x A vector to recycle.
-#' @param ...
+#' @param ... Depending on the function used:
 #'   * For `vec_recycle_common()`, vectors to recycle.
 #'   * For `vec_recycle()`, these dots should be empty.
```
DavisVaughan, line 124: I added this because if you look at `?vec_recycle` right now, the bullets aren't aligned, they look nested. I feel like their is a roxygen2 issue open for this, but I can't remember

R/recycle.R
```diff
@@ -1,31 +1,125 @@
-#' Vector recycling
+#' Recycling rules used by r-lib and the tidyverse
+#'
+#' @description
```
DavisVaughan, line 3: We also talk about these rules in this vignette but I didn't make changes there. I think it is consistent with what I have here https://vctrs.r-lib.org/articles/type-size.html#common-sizes-recycling-rules

R/recycle.R
```diff
@@ -1,31 +1,125 @@
...
+#' # This is an error
+#' vec_recycle_common(1:3, 4:7)
+#' #> Error:
+#' #> ! Can't recycle `..1` (size 3) to match `..2` (size 4).
+#' ```
 #'
-#' @section Recycling rules:
-#' The common size of two vectors defines the recycling rules, and can be
-#' summarise with the following table:
+#' @section Differences with base R:
```
DavisVaughan, line 53: There is an appendix section in this vignette that talks about this too. I drew some examples from there, but again left it unchanged https://vctrs.r-lib.org/articles/type-size.html#appendix-recycling-in-base-r

R/recycle.R
```diff
@@ -1,31 +1,126 @@
...
+#' - Vectors of size 1 will be recycled to the size of any other vector
+#'
+#' - Otherwise, all vectors must have the same size
+#'
+#' @section Examples:
+#'
+#' Vectors of size 1 are recycled to the size of any other vector:
+#'
+#' ```
+#' vec_recycle_common(1:3, 1)
```
lionel-, line 17: Maybe use a familiar function like `tibble()` or `mutate()` to demonstrate the rules? The audience for this doc page might be large right?

R/recycle.R
```diff
@@ -1,31 +1,126 @@
...
+#' # Length 0 vector overrides any other length
+#' 1 + numeric()
+#' #> numeric(0)
+#' 1:3 + numeric()
+#' #> numeric(0)
+#' ```
+#'
+#' These rules come from the [R Language
+#' Definition](https://cran.r-project.org/doc/manuals/r-release/R-lang.html#Recycling-rules)
+#' and are mainly focused on arithmetic. Outside of arithmetic, these rules are
```
lionel-, line 86: And vector combination / assembling? E.g. `cbind()`.

R/recycle.R
```diff
@@ -1,31 +1,126 @@
...
+#' data.frame(1:2, 1:3)
+#' #> Error in base::data.frame(..., stringsAsFactors = stringsAsFactors):
+#' #> arguments imply differing number of rows: 2, 3
+#' ```
+#'
+#' @name vector_recycling_rules
+#' @keywords internal
+NULL
+
+#' Vector recycling
```
lionel-, line 112: If you use a tidyverse function in examples above, this could become a vctrs backend section?

lionel- (approved):

R/recycle.R
DavisVaughan: I'm not quite sure what you mean

R/recycle.R
lionel-: But kind of:

```
cbind(1:2, 1:3)
     [,1] [,2]
[1,]    1    1
[2,]    2    2
[3,]    1    3
Warning message:
In cbind(1:2, 1:3) :
  number of rows of result is not a multiple of vector length (arg 1)
```

R/recycle.R
lionel-: You'd mention `vec_recycle_common()` here instead of above.

DavisVaughan: Oh so a `@section` that describes how `vec_recycle_common()` is the engine that powers this for r-lib and the tidyverse

R/bind.R
```diff
@@ -180,11 +180,12 @@ vec_rbind <- fn_inline_formals(vec_rbind, ".name_repair")
 
 #' @export
 #' @rdname vec_bind
-#' @param .size If, `NULL`, the default, will determine the number of
-#'   rows in `vec_cbind()` output by using the standard recycling rules.
+#' @param .size If, `NULL`, the default, will determine the number of rows in
+#'   `vec_cbind()` output by using the standard [recycling
```
hadley, line 184:
```suggestion
#'   `vec_cbind()` output by using the tidyverse [recycling
```

?

R/recycle.R
```diff
@@ -1,31 +1,164 @@
...
+#' @description
+#' Recycling describes the concept of repeating elements of one vector to match
+#' the size of another vector. There are two rules that guide the recycling
+#' process used by packages in r-lib and the tidyverse:
+#'
+#' - Vectors of size 1 will be recycled to the size of any other vector
+#'
+#' - Otherwise, all vectors must have the same size
+#'
+#' @section Examples:
```
hadley, line 12:
```suggestion
#' Recycling describes the concept of repeating elements of one vector to match
#' the size of another. There are two rules that underlying the "tidyverse" recycling
#' rules:
#'
#' - Vectors of size 1 will be recycled to the size of any other vector
#'
#' - Otherwise, all vectors must have the same size
#'
#' @section Examples:
```

R/recycle.R
```diff
@@ -1,31 +1,164 @@
...
+#'
+#' ```
+#' tibble(x = 1:3, y = 1L)
+#' #> # A tibble: 3 × 2
+#' #>       x     y
+#' #>   <int> <int>
+#' #> 1     1     1
+#' #> 2     2     1
+#' #> 3     3     1
+#' ```
```
hadley, line 24:
```suggestion
#' ```{r}
#' tibble(x = 1:3, y = 1L)
#' ```
```

?

R/recycle.R
hadley: And then maybe say "These are stricter than the recycling rules generally used by base R, which are ... ", moving the content from below.

R/recycle.R
```diff
@@ -1,31 +1,164 @@
...
+#' The recycling rules described here are stricter than the ones generally used
+#' by base R. With base R, the rules are usually:
+#'
+#' - If any vector is length 0, the output will be length 0
+#'
+#' - Otherwise, the output will be length `max(length_x, length_y)`, and a
+#'   warning will be thrown if the length of the longer vector is not an integer
+#'   multiple of the length of the shorter vector.
+#'
+#' ```
```
hadley, line 84: I think I'd drop the examples here. I think it makes sense to explores the inconsistencies in the vignette, but I'd keep it brief here since this is user facing. But we could say "See `vignette("...")` for more details".

hadley (approved):

R/recycle.R
DavisVaughan: I've dropped the examples and linked to the vignette, but I've kept this as a small separate section at the very bottom of the doc. I think including the base R rules between our rules and the Examples section is a little confusing, and it looks better to me to include them at the very end

PR #1586, opened 2022-07-07
Implement `vec_any_missing()`

### LINKED ISSUE #544: Implement `vec_any_missing()`
As a faster alternative to `any(vec_equal_na(x))`, with the ability to early exit as soon as any `NA` values are found. Should be pretty easy to implement, and similar to `vec_equal_na()`

### LINKED ISSUE #1563: Implement `list_drop_missing()`
For #1562 and because it seems useful generally

Also useful for dropping `NULL`s before coalescing in `coalesce()`

### LINKED ISSUE #1585: What should `vec_equal_na()` return with data frames with N rows and 0 cols?
It currently returns `TRUE` for each of the `N` rows, implying it is "missing". Maybe this is right? Hard to say.

We need to ensure `vec_any_missing()` is aligned with it, and that both get the 0 row, 0 col case right, which I think should definitely be `FALSE` for `vec_any_missing()` (because `vec_any_missing(integer())` and friends returns `FALSE`).

``` r
library(vctrs)

# 0 rows, 0 cols
df <- data_frame(.size = 0)
vec_equal_na(df)
#> logical(0)
vec_any_missing(df)
#> [1] FALSE

# N rows, 0 cols (What to return here?)
df <- data_frame(.size = 5)
vec_equal_na(df)
#> [1] TRUE TRUE TRUE TRUE TRUE
vec_any_missing(df)
#> [1] FALSE

# The `vec_equal()` page says `vec_equal_na()` preserves this invariant, so
# maybe vec_any_missing(df) should return `TRUE` with N rows, 0 cols.
# (This could just mean that `vec_equal()` is wrong too?)
vec_equal(df, vec_init(df), na_equal = TRUE)
#> [1] TRUE TRUE TRUE TRUE TRUE
```

DavisVaughan:
Closes #1563 
Closes #544 
Closes #1585

I am going to document and export this in a different PR. In that one, I'll also add the `vec_detect_missing()` alias for `vec_equal_na()` and document it on the same page as `vec_any_missing()`. The alias/renaming seemed like a decent bit of changes, so it felt right to leave it out of this.

---

The implementation here is fairly straightforward for atomic types. I've built it on top of a new C level `vec_first_missing()`, which might be useful in other scenarios (it was already useful for the data frame method).

I think we generally want to avoid allocations here, so the data frame implementation of this uses a poly-op helper and works rowwise using that (rather than working colwise and tracking missing values in a separate vector we have to allocate). That also allows us to short circuit at the first missing row.

Since the poly-op helper generally isn't as fast as working colwise, I've added a little trick that helps quite a bit in the two most common scenarios:
- Data frames with no missing rows
- Data frames with missing rows but no partially incomplete rows, where a partially incomplete row is like `data_frame(x = NA, y = 1)`

The trick is to look at the first column (which is guaranteed to be a simple atomic vector from df-flattening) and to call `vec_first_missing()` on that, which is extremely fast. If there aren't missing values in the first column, there aren't any missing rows in the entire data frame, so we can exit (that's the first scenario above). If it does hit a missing value, then we set `i` (the poly-op rowwise loop counter) to start from that row. If there are no partially incomplete rows, then it will immediately return after processing that row, meaning we only have to use the poly-op tooling on 1 row (that's the second scenario above).

I don't think it will be very common to have partially incomplete rows. This is an algorithmic function, and probably won't be very user facing. It probably won't be common to call this on (for example) a data frame that was read in from Excel, where you might have missing cells spread throughout the data. And even if this was the case, the slow case would only be hit if the _first_ column had a missing value, but the rest of the row wasn't missing.

I _do_ think it will be common to call this on data frames that have fully missing rows. I can see this function being used on rcrd vectors with data frame proxies, and we've said that a missing value for a rcrd has to correspond to a fully missing row in the data frame proxy. So optimizing for the fully missing row case works nicely here.

## Benchmarks

In the "fast" common cases, the df method is as fast as the atomic vector method

``` r
# No missing values.
# Basically the same speed between df and vector.
x <- rep(1, 1e6)
df <- data_frame(x = x, y = x, z = x, w = x)

bench::mark(
  vec_any_missing(df),
  vec_any_missing(x)
)
#> # A tibble: 2 × 6
#> 1 vec_any_missing(df)    436µs    647µs     1552.        0B     2.00
#> 2 vec_any_missing(x)     424µs    613µs     1518.        0B     0

# Obviously very fast if there is an NA at the beginning.
# df method is a little slower because it has to check all 4 columns.
x <- c(NA, rep(1, 1e6))
df <- data_frame(x = x, y = x, z = x, w = x)

bench::mark(
  vec_any_missing(df),
  vec_any_missing(x)
)
#> # A tibble: 2 × 6
#> 1 vec_any_missing(df)   4.37µs   4.89µs   173722.        0B        0
#> 2 vec_any_missing(x)    1.04µs   1.18µs   751957.        0B        0

# Missing value halfway through
x <- c(rep(1, 1e6/2), NA, rep(1, 1e6/2))
df <- data_frame(x = x, y = x, z = x, w = x)

bench::mark(
  vec_any_missing(df),
  vec_any_missing(x)
)
#> # A tibble: 2 × 6
#> 1 vec_any_missing(df)    201µs    211µs     4273.        0B        0
#> 2 vec_any_missing(x)     196µs    199µs     4500.        0B        0
```

Now investigate performance in the "worst" cases, where we have a partially incomplete row somewhere, but no truly missing rows. I think usage on data like this is going to be pretty rare, but it honestly still isn't that bad. `vec_equal_na()` is impressively fast here though.

``` r
# Partially incomplete row halfway through.
# Rows after this one use the slower poly-op rowwise method.
x <- c(rep(1, 1e6/2), NA, rep(1, 1e6/2))
df <- data_frame(x = x, y = x, z = x, w = seq_along(x) + 0L)

bench::mark(
  vec_any_missing(df), 
  vec_equal_na(df),
  check = FALSE,
  iterations = 1000
)
#> # A tibble: 2 × 6
#> 1 vec_any_missing(df)   1.91ms   2.17ms      448.        0B      0  
#> 2 vec_equal_na(df)      2.06ms   2.74ms      350.    11.4MB     35.0

# Worst case possible.
# Partially incomplete row is the first row!
# Rows after this one use the slower poly-op rowwise method.
x <- c(NA, rep(1, 1e6))
df <- data_frame(x = x, y = x, z = x, w = seq_along(x) + 0L)

bench::mark(
  vec_any_missing(df), 
  vec_equal_na(df),
  check = FALSE,
  iterations = 1000
)
#> # A tibble: 2 × 6
#> 1 vec_any_missing(df)   3.37ms   3.66ms      264.        0B      0  
#> 2 vec_equal_na(df)      2.09ms   2.73ms      337.    11.4MB     32.9

# But note that if the column with no missing values is up front, then
# it is fast again
x <- c(NA, rep(1, 1e6))
df <- data_frame(x = seq_along(x) + 0L, y = x, z = x, w = x)

bench::mark(
  vec_any_missing(df), 
  vec_equal_na(df),
  check = FALSE,
  iterations = 1000
)
#> # A tibble: 2 × 6
#> 1 vec_any_missing(df) 250.22µs    266µs     3475.        0B      0  
#> 2 vec_equal_na(df)      1.68ms    2.6ms      363.    11.4MB     36.4
```


src/missing.c
```diff
@@ -260,6 +260,138 @@ r_ssize list_col_equal_na(r_obj* x,
 
 // -----------------------------------------------------------------------------
 
+// [[ register() ]]
+r_obj* ffi_vec_any_missing(r_obj* x) {
+  return r_lgl(vec_any_missing(x));
+}
+
+// [[ include("missing.h") ]]
```
lionel-, line 268: By the way I think we no longer need include statements, since we decided that any non static functions must be exported from the companion header file.

src/missing.c
```diff
@@ -260,6 +260,138 @@ r_ssize list_col_equal_na(r_obj* x,
...
+r_ssize vec_first_missing(r_obj* x) {
+  r_obj* proxy = KEEP(vec_proxy_equal(x));
+  r_ssize out = proxy_first_missing(proxy);
+  FREE(1);
+  return out;
+}
+
+static inline
+r_ssize proxy_first_missing(r_obj* proxy) {
```
lionel-, line 282: It's starting to feel like we need macros to help define this sort of polymorphic operations. Though maybe the implementations are just so slightly different that they can't be automatically generated?

lionel- (approved):
I'm a bit concerned about the rowwise iteration but I trust you if you think this operation will mainly be used in programmatic contexts with fully missing rows. What are the expected applications by the way?

Also, one context in which partially missing rows may occur programmatically is after a join.

DavisVaughan:
> What are the expected applications by the way?

I often find myself wanting to use `vec_any_missing()` to see if there are any missing values before doing some combination of `vec_equal_na()` + `vec_slice()` or `vec_assign()`.

This is a good example:
https://github.com/tidyverse/dplyr/blob/aa98ecd0cf43a707dc066e489d464dab04a18b19/R/nth-value.R#L97-L106

I would switch this to

`if (na_rm && vec_any_missing(x)) {`

DavisVaughan:
I think it would very weird to see this used on a general data frame, like `vec_any_missing(mtcars)`.

I expect the data frame `vec_any_missing()` case to come up with vectors that have df-proxies, like:

```r
ymd <- clock::year_month_day(c(NA, 2019), c(NA, 12)) # this has a df-proxy
first(ymd, na_rm = TRUE)
```

Where `first(na_rm = TRUE)` would do the `vec_any_missing()` check I described above.

src/missing.c
DavisVaughan: > any non static functions must be exported from the companion header file.

I didn't know/remember we decided that but that works for me 😆

DavisVaughan: And exported functions get an `ffi_` prefix so we don't need `register()` either? (im going to remove it)

lionel-: That makes sense.

PR #1589, opened 2022-07-13
Ensure that `vec_count(sort = "count")` uses a stable sort when there are ties

### LINKED ISSUE #1588: `vec_count()` is non-deterministic across sessions with the default sort
The order of values returned by `vec_count()` when `sort = "count"` or `sort = "none"` is non-deterministic across sessions. Within the same R session it's consistent, but the results change when running in a new session. Is this intended? If so, I think it'd be useful to document.

Reprex for the default sort:

``` r
set.seed(123)
head(vctrs::vec_count(dplyr::starwars$name))
#>                  key count
#> 1      Ratts Tyerell     1
#> 2        Poe Dameron     1
#> 3    Raymus Antilles     1
#> 4 Beru Whitesun lars     1
#> 5   Anakin Skywalker     1
#> 6              Watto     1
```

<sup>Created on 2022-07-13 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

``` r
set.seed(123)
head(vctrs::vec_count(dplyr::starwars$name))
#>               key count
#> 1 Dexter Jettster     1
#> 2       Sly Moore     1
#> 3          R4-P17     1
#> 4             Rey     1
#> 5         Tarfful     1
#> 6      Rugor Nass     1
```

<sup>Created on 2022-07-13 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

And for `sort = "none"`:

``` r
set.seed(123)
head(vctrs::vec_count(dplyr::starwars$name, sort = "none"))
#>              key count
#> 1         Greedo     1
#> 2     Jango Fett     1
#> 3 Wilhuff Tarkin     1
#> 4       Han Solo     1
#> 5    Saesee Tiin     1
#> 6        Sebulba     1
```

<sup>Created on 2022-07-13 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

``` r
set.seed(123)
head(vctrs::vec_count(dplyr::starwars$name, sort = "none"))
#>                  key count
#> 1 Beru Whitesun lars     1
#> 2                BB8     1
#> 3         Jango Fett     1
#> 4               Finn     1
#> 5                Rey     1
#> 6               Yoda     1
```

<sup>Created on 2022-07-13 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

DavisVaughan:
Closes #1588 

With `sort = "count"`, we need to sort by descending `count` AND by ascending `loc` (where `loc` is the original location in `x`). This ensures that when there are ties in the `count`, we still get a stable sort (i.e. if key A and key B both appear 5 times, but we saw key A first, then key A should appear first in the output)

Here is a minimal reprex of the original issue:

``` r
library(vctrs)
x <- c(1, 2, 1, 5, 2, 5)
vec_count(x)
#>   key count
#> 1   1     2
#> 2   5     2
#> 3   2     2
```

Which is fixed in this PR:

```r
library(vctrs)
x <- c(1, 2, 1, 5, 2, 5)
vec_count(x)
#>   key count
#> 1   1     2
#> 2   2     2
#> 3   5     2
```

This was doubly bad for character vectors, because their hash value can change from one R session to another, which is what caused #1588.

In the case of `sort = "none"`, I think it is fine that the result doesn't have an order that is guaranteed to be the same from one R session to another. I've documented the fact that this can occur though.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,9 @@
 # vctrs (development version)
 
+* `vec_count(sort = "count")` now uses a stable sorting method. This ensures
+  that different keys with the same count are sorted in the order that they
+  originally appeared in (#1588).
+
 * Lossy cast error conditions now show the correct message when
   `conditionMessage()` is called on them (#1592).
 


R/dictionary.R
```diff
@@ -46,41 +47,39 @@
...
 
-reset_rownames <- function(x) {
-  rownames(x) <- NULL
-
-  is_df <- map_lgl(x, is.data.frame)
-  x[is_df] <- lapply(x[is_df], reset_rownames)
-
-  x
-}
-
```
DavisVaughan, line 83: I don't know what this was doing, but I have a feeling it had to do with some bug we used to have with `vec_slice()`, which has probably been fixed? We don't use this anywhere else, and it didn't break any tests to remove it.

src/dictionary.c
```diff
@@ -535,42 +536,47 @@ SEXP vctrs_count(SEXP x) {
...
-  SEXP out_key = PROTECT_N(Rf_allocVector(INTSXP, d->used), &nprot);
-  SEXP out_val = PROTECT_N(Rf_allocVector(INTSXP, d->used), &nprot);
-  int* p_out_key = INTEGER(out_key);
-  int* p_out_val = INTEGER(out_val);
+  SEXP out_loc = PROTECT_N(Rf_allocVector(INTSXP, d->used), &nprot);
+  int* p_out_loc = INTEGER(out_loc);
+
+  // Reuse `count` storage, which will be narrowed
+  SEXP out_count = count;
+  int* p_out_count = p_count;
```
DavisVaughan, line 558: I was able to optimize this while I was here to avoid an allocation of size `d->used`

We can reuse the storage of `count` for the output `out_count`, and just resize it at the end

lionel- (approved):

PR #1590, opened 2022-07-15
Early exit for empty index in `vec_assign()`

### LINKED ISSUE #1408: Fast path for `vec_assign(x, integer())`?
Similarly to #1400 I think it would make sense to have a fast path for `vec_assign(x, integer())`

The motivation is also similar: e.g. when using `vec_assign()` in a `coalesce()` like fashion `vec_assign(x, is.na(x), 1L)`.

## Benchmark

``` r
library(rlang)
library(vctrs)

x <- 0L + 1:10e6

bench::mark(
  vec_assign = {vec_assign(x, integer(), integer())},
  check = FALSE
)
#> # A tibble: 1 x 6
#> 1 vec_assign    5.1ms   5.53ms      165.    38.2MB     172.
```

<sup>Created on 2021-06-25 by the [reprex package](https://reprex.tidyverse.org) (v2.0.0)</sup>

mgirlich:
Closes #1408.

For example this is relevant for `coalesce()` like uses, e.g.  `vec_assign(x, is.na(x), 1L)`.

I only added this optimisation for `vec_assign_opts()` and not for other functions in `slice-assign.h` e.g. `vec_proxy_assign_opts()`.  I only saw `vec_proxy_assign_opts()` being called in `unchop()` and `vec_c()` but it don't think one would encounter an empty index there.

Do you want an explicit test that `vec_assign(x, integer(), -1L)` works? It is currently tested "indirectly" via `vec_slice(x, FALSE) <- 2L`.

``` r
library(rlang)
library(vctrs)

n <- 10e6
x <- 0L + 1:n
idx <- vec_rep(FALSE, n)

bench::mark(
  false_n = vec_assign(x, idx, -1L),
  false = vec_assign(x, FALSE, -1L),
  empty = vec_assign(x, integer(), -1L),
  check = FALSE,
  min_iterations = 20
)

# Before
#> # A tibble: 3 × 6
#> 1 false_n     13.98ms  14.58ms      62.8    38.1MB     68.1
#> 2 false        5.07ms   5.49ms     179.     38.1MB    172. 
#> 3 empty        5.03ms   5.59ms     177.     38.1MB    171.

# After
#> # A tibble: 3 × 6
#> 1 false_n      8.37ms   9.01ms      109.     3.5KB      0  
#> 2 false        1.74µs   1.86µs   371996.        0B     37.2
#> 3 empty        2.11µs   2.29µs   341040.        0B      0
```

<sup>Created on 2022-07-15 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

### DIFF
--- a/src/slice-assign.c
+++ b/src/slice-assign.c
@@ -27,6 +27,11 @@ r_obj* vec_assign_opts(r_obj* x,
                                        KEEP(vec_names(x)),
                                        &location_opts));
 
+  if (vec_size(index) == 0) {
+    FREE(2);
+    return x;
+  }
+
   // Cast and recycle `value`
   value = KEEP(vec_cast(value, x, opts.value_arg, opts.x_arg, opts.call));
   value = KEEP(vec_check_recycle(value, vec_size(index), opts.value_arg, opts.call));


lionel-:
> Do you want an explicit test that vec_assign(x, integer(), -1L) works?

Yes please, it never hurts to add one more test :-)

src/slice-assign.c
```diff
@@ -27,6 +27,11 @@ r_obj* vec_assign_opts(r_obj* x,
                                        KEEP(vec_names(x)),
                                        &location_opts));
 
+  if (vec_size(index) == 0) {
+    FREE(2);
+    return x;
+  }
+
   // Cast and recycle `value`
   value = KEEP(vec_cast(value, x, opts.value_arg, opts.x_arg, opts.call));
```
DavisVaughan, line 36: I think the cast and recycle checks need to run before we can think about early exiting

DavisVaughan (commented):
I'm actually tempted to put this optimization deep within the assignment macros, like `ASSIGN_INDEX()`, `ASSIGN_COMPACT()`, the corresponding BARRIER macros, and the corresponding "shaped" ones for arrays. Otherwise I think it is very hard to get the placement of the optimization right, because we aren't always exactly sure what `index` is (is it compact, or not?) and we have all kinds of other checks on `x`/`proxy`/`value` that we want to ensure are run before we decide to early exit. 

Basically I think the early exit should be done as close to the generation of the copy as possible (i.e. the `vec_clone_referenced()` call)

lionel-:
Closing for the reasons explained in https://github.com/r-lib/vctrs/pull/1636#issuecomment-1252300105

hadley:
Reopening, because I'd like to reconsider in light of https://github.com/tidyverse/tidyr/pull/1392 — the performance gains seem meaningful, and it seems much more reasonable to handle this in vctrs rather than having to wrap every use of `vec_assign()`.

PR #1591, opened 2022-07-18
Early exit for only `TRUE` in `vec_slice()`

mgirlich:
Useful e.g. in `vec_slice(x, !is.na(x))`

``` r
library(rlang)
library(vctrs)

n <- 10e6
x <- 0L + 1:n
idx <- vec_rep(TRUE, n)

bench::mark(
  false_n = vec_slice(x, idx),
  check = FALSE,
  min_iterations = 20
)
# Before
#> # A tibble: 1 × 6
#> 1 false_n      29.6ms   30.6ms      28.9    76.3MB     30.3

# After
#> # A tibble: 1 × 6
#> 1 false_n      7.39ms   7.91ms      121.    3.73KB        0
```

<sup>Created on 2022-07-18 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

### DIFF
--- a/src/slice-assign.c
+++ b/src/slice-assign.c
@@ -27,6 +27,11 @@ r_obj* vec_assign_opts(r_obj* x,
                                        KEEP(vec_names(x)),
                                        &location_opts));
 
+  if (is_compact(index)) {
+    index = compact_materialize(index);
+  }
+  KEEP(index);
+
   // Cast and recycle `value`
   value = KEEP(vec_cast(value, x, opts.value_arg, opts.x_arg, opts.call));
   value = KEEP(vec_check_recycle(value, vec_size(index), opts.value_arg, opts.call));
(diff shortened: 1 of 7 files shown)


DavisVaughan:
I'm not quite sure we'd want to optimize for this case yet. I don't love early exits like these in general because they add one more special code path that we have to keep in mind whenever we make any changes in the future (even if the performance is nice)

I think a better pattern for your example is

```r
if (vec_any_missing(x)) {
  x <- vec_slice(x, !vec_equal_na(x))
}
```

Then you don't need the early exit because you only do "work" when you know there is at least 1 missing value

lionel-:
Closing for the reasons explained in https://github.com/r-lib/vctrs/pull/1636#issuecomment-1252300105

Issue #1606, opened 2022-08-11
`vec_as_subscript2()` doesn't throw lossy cast errors correctly


DavisVaughan:
``` r
library(vctrs)

vec_as_subscript(1.5)
#> Error:
#> ! Must subset elements with a valid subscript vector.
#> ✖ Can't convert from <double> to <integer> due to loss of precision.

vec_as_subscript2(1.5)
#> Error:
#> ! Must extract element with a single valid subscript.
#> ✖ Subscript has the wrong type `double`.
#> ℹ It must be logical, numeric, or character.
```

It looks like it was supposed to use a lossy cast body here:
https://github.com/r-lib/vctrs/blob/544cc704c78d1ebac6b6de84ae12024ee97e1778/R/subscript.R#L104-L106

But it looks like `parent` is actually always `NULL`? It somehow isn't being set in the `err` object. Commenting out line 106 doesn't break any tests, so I'm assuming this just isn't tested

PR #1607, opened 2022-08-11
Improve `vec_as_location2()` and `vec_as_subscript2()` usage of `call`

### LINKED ISSUE #1605: `vec_as_location()` doesn't always use `call`
``` r
vctrs::vec_as_location2("foo", 1, names = "x", call = quote(abc()))
#> Error in `vec_as_location2_result()`:
#> ! Can't subset elements that don't exist.
#> ✖ Element `foo` doesn't exist.
```

<sup>Created on 2022-08-11 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

DavisVaughan:
Closes #1605

### DIFF
--- a/R/subscript-loc.R
+++ b/R/subscript-loc.R
@@ -239,7 +239,7 @@ vec_as_location2_result <- function(i,
   # FIXME: Use result approach in internal implementation?
   err <- NULL
   i <- tryCatch(
-    vec_as_location(i, n, names = names, arg = arg),
+    vec_as_location(i, n, names = names, arg = arg, call = call),
     vctrs_error_subscript_type = function(err) {
       err <<- err
       i
--- a/tests/testthat/_snaps/subscript.md
+++ b/tests/testthat/_snaps/subscript.md
@@ -139,3 +139,23 @@
       x Subscript has the wrong type `logical`.
       i It must be numeric or character.
 
+# vec_as_subscript2() retains the call when throwing vec_as_subscript() errors (#1605)
+
+    Code
+      vec_as_subscript2(1L, numeric = "error", call = call("foo"))
+    Condition
+      Error in `foo()`:
+      ! Must extract element with a single valid subscript.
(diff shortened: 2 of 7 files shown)


R/subscript-loc.R
```diff
@@ -239,7 +239,7 @@ vec_as_location2_result <- function(i,
   # FIXME: Use result approach in internal implementation?
   err <- NULL
   i <- tryCatch(
-    vec_as_location(i, n, names = names, arg = arg),
+    vec_as_location(i, n, names = names, arg = arg, call = call),
     vctrs_error_subscript_type = function(err) {
```
DavisVaughan, line 243: Even though we tryCatch this, it can throw other relevant errors that need the `call`

tests/testthat/_snaps/subscript.md
```diff
@@ -139,3 +139,33 @@
...
+
+---
+
+    Code
+      vec_as_subscript2(1.5, call = call("foo"))
+    Condition
+      Error in `foo()`:
+      ! Must extract element with a single valid subscript.
+      x Subscript has the wrong type `double`.
+      i It must be logical, numeric, or character.
```
DavisVaughan, line 160: This should be a different error. But that is #1606 and we should tackle it in a different PR because it looks messy

tests/testthat/_snaps/subscript.md
```diff
@@ -139,3 +139,33 @@
...
+
+# vec_as_subscript2() retains the call when erroring on logical input (#1605)
+
+    Code
+      vec_as_subscript2(TRUE, call = call("foo"))
+    Condition
+      Error in `foo()`:
+      ! Must extract element with a single valid subscript.
+      x Subscript has the wrong type `logical`.
+      i It must be logical, numeric, or character.
```
DavisVaughan, line 170: This error is misleading, but that is #1608

lionel- (approved):

PR #1609, opened 2022-08-12
Add a `.call` argument to `vec_rbind()` and `vec_cbind()`

### LINKED ISSUE #1597: Add `call` arguments to `vec_cbind()`/`vec_rbind()`
For https://github.com/tidyverse/dplyr/issues/6309

DavisVaughan:
Closes #1597 

Fairly straightforward because we already passed a call through, it just wasn't a user supplied one.

For the tests, I extended a few existing tests where it seemed important that the call was being passed through (like common type or size tests). I also updated some to `expect_snapshot(error = TRUE` style

### DIFF
--- a/src/bind.c
+++ b/src/bind.c
@@ -362,12 +362,12 @@ r_obj* cbind_names_to(bool has_names,
 
 
 // [[ register(external = TRUE) ]]
-r_obj* ffi_cbind(r_obj* ffi_call, r_obj* op, r_obj* args, r_obj* env) {
+r_obj* ffi_cbind(r_obj* ffi_call, r_obj* op, r_obj* args, r_obj* frame) {
   args = r_node_cdr(args);
 
-  struct r_lazy call = { .x = env, .env = r_null };
+  struct r_lazy call = { .x = syms_dot_call, .env = frame };
 
-  r_obj* xs = KEEP(rlang_env_dots_list(env));
+  r_obj* xs = KEEP(rlang_env_dots_list(frame));
   r_obj* ptype = r_node_car(args); args = r_node_cdr(args);
   r_obj* size = r_node_car(args); args = r_node_cdr(args);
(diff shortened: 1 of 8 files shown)


src/bind.c
```diff
@@ -3,12 +3,12 @@
 #include "decl/bind-decl.h"
 
 // [[ register(external = TRUE) ]]
-r_obj* ffi_rbind(r_obj* ffi_call, r_obj* op, r_obj* args, r_obj* env) {
+r_obj* ffi_rbind(r_obj* ffi_call, r_obj* op, r_obj* args, r_obj* frame) {
   args = r_node_cdr(args);
 
-  struct r_lazy call = { .x = env, .env = r_null };
+  struct r_lazy call = { .x = syms_dot_call, .env = frame };
```
DavisVaughan, line 9: Fairly confident that the environment we get from `.External2()` is the same one we'd get from calling `environment()` and passing that through

tests/testthat/test-bind.R
```diff
@@ -378,9 +391,18 @@ test_that("performance: Row binding with df-cols doesn't duplicate on every assi
 # cols --------------------------------------------------------------------
 
 test_that("vec_cbind() reports error context", {
-  expect_snapshot({
-    (expect_error(vec_cbind(foobar(list()))))
-    (expect_error(vec_cbind(a = 1:2, b = int())))
+  expect_snapshot(error = TRUE, {
```
lionel-, line 394: I think in a package like vctrs we'll still want to monitor error classes. With this change, the class no longer appears in snapshots. Can you revert to the previous pattern and also use it in the other tests please?

DavisVaughan: Sounds good, I've changed back to the previous pattern and re-requested a review

lionel- (approved):
Thanks!

Issue #1595, opened 2022-07-21
`num_as_location()` needs more options to support `dplyr::slice()`


hadley:
I think it needs `oob = "ignore", and `missing = "ignore"`

so we have:

```R
num_as_location(c(NA), 10, missing = "ignore")
#> integer(0)
num_as_location(c(-20, 20), 10, oob = "ignore")
#> integer(0)
```

From https://github.com/tidyverse/dplyr/issues/5242

hadley:
Possibly the same as #1168

PR #1615, opened 2022-08-12
Implement `num_as_location(oob = "remove")`

DavisVaughan:
Part of #1595 

We also need `missing = "remove"`, but I will do that in another PR because these code paths are already highly complex due to the number of options

### DIFF
--- a/src/subscript-loc.c
+++ b/src/subscript-loc.c
@@ -133,7 +133,7 @@ r_obj* int_as_location(r_obj* subscript,
   // positive indices need to go through and `int_filter_zero()`.
   r_ssize n_zero = 0;
 
-  r_ssize n_extend = 0;
+  r_ssize n_oob = 0;
 
   for (r_ssize i = 0; i < loc_n; ++i, ++data) {
     int elt = *data;
@@ -158,10 +158,11 @@ r_obj* int_as_location(r_obj* subscript,
         case LOC_ZERO_IGNORE: break;
         }
       } else if (abs(elt) > n) {
-        if (opts->loc_oob == LOC_OOB_ERROR) {
-          stop_subscript_oob_location(subscript, n, opts);
+        switch (opts->loc_oob) {
+        case LOC_OOB_ERROR: stop_subscript_oob_location(subscript, n, opts);
+        case LOC_OOB_EXTEND:
+        case LOC_OOB_REMOVE: ++n_oob; break;
         }
-        ++n_extend;
       }
     }
   }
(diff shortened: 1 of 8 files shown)


src/subscript-loc.c
```diff
@@ -158,10 +158,11 @@ r_obj* int_as_location(r_obj* subscript,
...
         }
       } else if (abs(elt) > n) {
-        if (opts->loc_oob == LOC_OOB_ERROR) {
-          stop_subscript_oob_location(subscript, n, opts);
+        switch (opts->loc_oob) {
+        case LOC_OOB_ERROR: stop_subscript_oob_location(subscript, n, opts);
+        case LOC_OOB_EXTEND:
+        case LOC_OOB_REMOVE: ++n_oob; break;
         }
-        ++n_extend;
```
DavisVaughan, line 164: I strongly prefer the switch statement in these kinds of paths. If we add a new `LOC_OOB_*` option then the compiler will throw a warning that we haven't handled that case, so we can't forget about it

src/subscript-loc.c
```diff
@@ -207,9 +213,19 @@ r_obj* int_invert_location(r_obj* subscript,
 
     j = -j;
     if (j > n) {
-      struct location_opts updated_opts = *opts;
-      updated_opts.subscript_opts.action = SUBSCRIPT_ACTION_NEGATE;
-      stop_subscript_oob_location(subscript, n, &updated_opts);
+      switch (opts->loc_oob) {
+      case LOC_OOB_REMOVE: {
+        continue;
+      }
```
DavisVaughan, line 219: To support this behavior:

``` r
library(dplyr)

df <- tibble(x = 1:5)

slice(df, -6)
#> # A tibble: 5 × 1
#>       x
#>   <int>
#> 1     1
#> 2     2
#> 3     3
#> 4     4
#> 5     5

slice(df, c(-2, -4))
#> # A tibble: 3 × 1
#>       x
#>   <int>
#> 1     1
#> 2     3
#> 3     5
```

<sup>Created on 2022-08-12 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

src/subscript-loc.c
```diff
@@ -171,11 +172,16 @@ r_obj* int_as_location(r_obj* subscript,
   }
   KEEP(subscript);
 
-  if (n_extend > 0) {
-    int_check_consecutive(subscript, n, n_extend, opts);
+  if (n_oob > 0 && opts->loc_oob == LOC_OOB_EXTEND) {
```
lionel-, line 175: Use a `switch` inside `if (n_oob > 0)`?

lionel- (approved):

PR #1616, opened 2022-08-12
Implement `.unpack` argument for `df_list()`

DavisVaughan:
We've talked about having a way to disable auto-unpacking in `df_list()` before. We finally have a need for this in dplyr, where `consecutive_id()` and `n_distinct()` are implemented like this:

```r
> dplyr:::consecutive_id
function(...) {
  check_dots_unnamed()
  data <- vctrs::data_frame(..., .name_repair = "minimal")

  out <- vec_identify_runs(data)
  attr(out, "n") <- NULL

  out
}
```

We don't necessarily want to unpack here. It doesn't actually end up affecting these functions, but I could see how it could be undesired in other cases.

I think this is an extremely low level thing to tweak though, so I don't want to expose it in `data_frame()`. For now I've just exposed it in `df_list()`, and I'll replace the `data_frame()` call above with:

```r
data <- vctrs::df_list(..., .name_repair = "minimal", .unpack = FALSE)
data <- vctrs::new_data_frame(data)
```


tests/testthat/_snaps/type-data-frame.md
```diff
@@ -190,6 +190,22 @@
...
+      Error in `df_list()`:
+      ! `.unpack` must be a logical value.
+
+---
+
+    Code
+      df_list(.unpack = c(TRUE, FALSE))
+    Condition
+      Error in `df_list()`:
+      ! `.unpack` must be a logical value.
```
DavisVaughan, line 207: Suboptimal error here that we should fix in the rlang lib `r_arg_as_bool()` function

lionel- (approved):
I like that `df_list()` doesn't return an invalid object like `data_frame()` would if it didn't unpack unnamed arguments.

DavisVaughan:
@lionel- why would `data_frame()` return an invalid object?

lionel-:
Because the df-col would be unnamed. Am I missing something?

DavisVaughan:
Oh, I guess it depends on how you define "invalid"

I see `""` names as being valid, but not particularly useful, names on a data frame, because you can't match against them 

``` r
df <- vctrs::data_frame(1, .name_repair = "minimal")
df
#>    
#> 1 1
names(df)
#> [1] ""
```

I wouldn't call them invalid, at least not from a developer perspective, because we use them heavily in dplyr now with `consecutive_id()` and `n_distinct()`. i.e. this is what `consecutive_id()` is going to be, and the whole point is that you just want to turn `...` into a data frame (for `vec_identify_runs()`) but you don't care about the names so we just let minimal name repair give it `""` names.

```r
consecutive_id <- function(...) {
  check_dots_unnamed()
  data <- vctrs::df_list(..., .name_repair = "minimal", .unpack = FALSE)
  data <- vctrs::new_data_frame(data)

  out <- vec_identify_runs(data)
  attr(out, "n") <- NULL

  out
}
```

From a user perspective I agree that these are basically useless and we should avoid generating them by default

PR #1619, opened 2022-08-17
Add `"select"` and `"relocate"` subscript actions

### LINKED ISSUE #1596: Allow "select" subscript action
For tidyverse/dplyr#6200.

DavisVaughan:
Closes #1596

Note that `"relocate"` does the same thing as `"rename"` in that it never uses the `"past the end"` terminology. I don't think that ever makes sense for relocating.

### DIFF
--- a/R/subscript.R
+++ b/R/subscript.R
@@ -346,7 +346,9 @@ cnd_subscript_element_cli <- function(n, cnd, capital = FALSE) {
 }
 
 subscript_actions <- c(
-  "subset", "extract", "assign", "rename", "remove", "negate"
+  "select", "subset", "extract",
+  "assign", "rename", "relocate",
+  "remove", "negate"
 )
 cnd_subscript_action <- function(cnd, assign_to = TRUE) {
   action <- cnd$subscript_action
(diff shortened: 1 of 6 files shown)


R/subscript.R
```diff
@@ -360,10 +362,10 @@ cnd_subscript_action <- function(cnd, assign_to = TRUE) {
   }
 
   if (!is_string(action, subscript_actions)) {
-    abort(paste0(
-      "Internal error: `cnd$subscript_action` must be one of ",
-      "`subset`, `extract`, `assign`, `rename`, `remove`, or `negate`."
-    ))
+    cli::cli_abort(
+      "`cnd$subscript_action` must be one of {.or {.arg {subscript_actions}}}.",
+      .internal = TRUE
```
DavisVaughan, line 367: I know about the `.internal` bug with `cli_abort()` that requires the dev version of cli to get the right `.frame`, but I'm not going to require dev cli here because I don't want vctrs to be blocked by it. We don't snapshot this or anything, so it shouldn't really matter

R/subscript.R
```diff
@@ -346,7 +346,9 @@ cnd_subscript_element_cli <- function(n, cnd, capital = FALSE) {
 }
 
 subscript_actions <- c(
-  "subset", "extract", "assign", "rename", "remove", "negate"
+  "select", "subset", "extract",
+  "assign", "rename", "relocate",
+  "remove", "negate"
```
lionel-, line 351: wow relocate and extract stand out here

lionel- (approved):

PR #1621, opened 2022-08-19
Export vec_rank()

### LINKED ISSUE #1600: Export vec_rank()

krlmlr:
Closes #1600.


lionel-:
I was expecting @DavisVaughan to do this given that he implemented `vec_rank()` and didn't export it right away on purpose (there might still be some work to do before export).

krlmlr:
We can merge this whenever we're ready, I didn't change any of the semantics.

R/rank.R
```diff
@@ -1,8 +1,3 @@
-# TODO: Use this NEWS bullet when we export `vec_rank()`
-#
-# * New `vec_rank()` to compute various types of sample ranks.
-
-
 #' Compute ranks
 #'
 #' `vec_rank()` computes the sample ranks of a vector. For data frames, ranks
```
DavisVaughan, line 3: Note that these docs reference `Like [vec_order_radix()]`, i.e. the unexported `vec_order_radix()` function that we do have public documentation for, it is just hidden away with `@keyword internal`.

I think this is fine, when we eventually make `vec_order()` use `vec_order_radix()`, this will just get updated to point to `vec_order()` instead.

R/rank.R
```diff
@@ -53,7 +48,7 @@
 #' - [vec_order_radix()]
 #' - [vec_slice()]
 #'
-#' @noRd
+#' @export
 #' @examples
 #' x <- c(5L, 6L, 3L, 3L, 5L, 3L)
 #'
```
DavisVaughan, line 54: The other thing worth commenting on is that right now `vec_rank()` and `vec_order()` are inconsistent because `vec_order()` uses the system locale and `vec_rank()` uses the C locale

This would eventually get resolved when `vec_order()` switches to using `vec_order_radix()`, which we can't do until after dplyr 1.1.0 is out.

My original plan was to switch `vec_order()` over AND export `vec_rank()` at the same time, in a release that comes after dplyr 1.1.0 is out. That way they are both consistent with each other. @lionel- do you have any thoughts about this? If we delay exporting `vec_rank()` for this reason, we can still use it in dplyr like we currently do with `vec_order_base()` and `vec_order_radix()`. I think I prefer this for consistency reasons if we aren't dead set on exporting `vec_rank()` right now (it seems mostly motivated by dplyr)
https://github.com/tidyverse/dplyr/blob/eaf9b69e2d15b58048d963f04c061f8de296d616/R/zzz.r#L13-L21

CC @hadley

lionel-: I think this is complex enough that we might want to discuss it at the dplyr meeting? I'm not sure I understand the issue in its entirety.

hadley: Can't we just allow them to be inconsistent for a little bit? I see only a handful of non-us packages that are using `vec_order()`.

DavisVaughan: Letting them be inconsistent for a little while is probably fine, since we are the main users of this. I'll merge this in today

PR #1625, opened 2022-08-26
Pass `.call` through `data_frame()` and `df_list()`

### LINKED ISSUE #1610: Add `call` to `df_list()` and maybe `data_frame()`
https://github.com/r-lib/vctrs/issues/1597#issuecomment-1200500946

DavisVaughan:
Closes #1610

Using the following signatures based on how we feel they are used in practice:
- `data_frame(.call = current_env())`
- `df_list(.call = caller_env())`


lionel- (approved):
I'm a bit concerned about giving `df_list()` and `data_frame()` different defaults. I think I'd use `call = current_env()` in `df_list()` too.

DavisVaughan:
It seems reasonable to default to `call = current_env()` if we are at all unsure about it, because it doesn't change the call at all by default but still allows developers to pass on their own call

PR #1626, opened 2022-08-26
Add `call` argument to `vec_locate_matches()`

### LINKED ISSUE #1611: Add `call` to `vec_locate_matches()`
Probably default to `call = caller_env()` because this seems like a pure backend function

DavisVaughan:
Closes #1611 

- I ended up going with `call = current_env()` as the default. I do think that this function could be used by advanced users at the top level, in the same way that `match()` is used.

- `vec_ptype2_params()` and `vec_cast_params()` were missing a `call` argument, so I added that in

Other than that, this was a straightforward if tedious PR. Each commit is pretty self-contained

### DIFF
--- a/src/decl/match-decl.h
+++ b/src/decl/match-decl.h
@@ -21,7 +21,8 @@ r_obj* vec_locate_matches(r_obj* needles,
                           bool nan_distinct,
                           r_obj* chr_proxy_collate,
                           struct vctrs_arg* needles_arg,
-                          struct vctrs_arg* haystack_arg);
+                          struct vctrs_arg* haystack_arg,
+                          struct r_lazy call);
 
 static
 r_obj* df_locate_matches(r_obj* needles,
--- a/src/decl/match-decl.h
+++ b/src/decl/match-decl.h
@@ -225,7 +225,8 @@ void stop_matches_incomplete(r_ssize i,
 static inline
 void stop_matches_multiple(r_ssize i,
                            struct vctrs_arg* needles_arg,
-                           struct vctrs_arg* haystack_arg);
+                           struct vctrs_arg* haystack_arg,
+                           struct r_lazy call);
 
 static inline
 void warn_matches_multiple(r_ssize i,
--- a/src/decl/match-decl.h
+++ b/src/decl/match-decl.h
@@ -231,4 +231,5 @@ void stop_matches_multiple(r_ssize i,
 static inline
 void warn_matches_multiple(r_ssize i,
                            struct vctrs_arg* needles_arg,
-                           struct vctrs_arg* haystack_arg);
+                           struct vctrs_arg* haystack_arg,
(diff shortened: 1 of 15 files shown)


src/decl/match-decl.h
```diff
@@ -160,13 +166,13 @@ r_obj* expand_compact_indices(const int* v_o_haystack,
...
-                              struct vctrs_arg* haystack_arg);
+                              struct vctrs_arg* haystack_arg,
+                              struct r_lazy call);
 
 static
 r_obj* compute_nesting_container_info(r_obj* haystack,
                                       r_ssize size_haystack,
-                                      const enum vctrs_ops* v_ops,
-                                      struct vctrs_arg* haystack_arg);
+                                      const enum vctrs_ops* v_ops);
```
DavisVaughan, line 175: `haystack_arg` wasn't being used here, so I removed it as part of this PR

src/dictionary.c
```diff
@@ -330,6 +330,8 @@ SEXP vctrs_id(SEXP x) {
 // [[ register() ]]
 SEXP vctrs_match(SEXP needles, SEXP haystack, SEXP na_equal,
                  SEXP frame) {
+  struct r_lazy call = { .x = frame, .env = r_null };
```
DavisVaughan, line 333: Some changes here in `dictionary.c` and `interval.c` are due to `vec_ptype2_params()` and `vec_cast_params()` getting a `call` argument

src/match.c
```diff
@@ -83,11 +83,15 @@ r_obj* ffi_locate_matches(r_obj* needles,
...
-  const enum vctrs_multiple c_multiple = parse_multiple(multiple);
+                          r_obj* haystack_arg,
+                          r_obj* frame) {
+  struct r_lazy call = { .x = syms_call, .env = frame };
+  struct r_lazy internal_call = { .x = frame, .env = r_null };
+
+  const struct vctrs_incomplete c_incomplete = parse_incomplete(incomplete, internal_call);
+  const struct vctrs_no_match c_no_match = parse_no_match(no_match, internal_call);
+  const struct vctrs_remaining c_remaining = parse_remaining(remaining, internal_call);
+  const enum vctrs_multiple c_multiple = parse_multiple(multiple, internal_call);
```
DavisVaughan, line 94: I've made a judgement call that these invalid argument issues should report `vec_locate_matches()` as the source of the error. I feel like _most_ of the time these arguments aren't directly exposed to the user. And even if they are, a suboptimal error here isn't the worst thing.

DavisVaughan:
I'm not sure why the coverage GHA check fails. It seems to think the `match.c` line number that needs to be reported in the internal error has changed, which is odd

lionel- (approved):
Might need to move the internal errors out of snapshots to fix the failure on Windows.

PR #1631, opened 2022-08-29
Correctly error on `"ignore"`d oob negative values

### LINKED ISSUE #1614: `num_as_location()` internal error with `oob = "extend"` and `negative = "ignore"`
``` r
num_as_location(-5, 4, oob = "extend", negative = "ignore")
#> Error in `num_as_location()`:
#> ! int_check_consecutive
#> ℹ In file 'subscript-loc.c' at line 321.
#> ℹ This is an internal error in the vctrs package, please report it to the package authors.
```

Essentially the negative value isn't handled, but looks OOB so it is counted as something to "extend" by

### LINKED ISSUE #1630: `num_as_location()` `negative = "invert", oob = "extend"` error shouldn't report positive values as oob
``` r
library(vctrs)

num_as_location(c(-6L, 6L, 7L), n = 5L, negative = "invert", oob = "extend")
#> Error:
#> ! Can't negate elements past the end.
#> ℹ Locations 6, 6, and 7 don't exist.
#> ℹ There are only 5 elements.

#> Backtrace:
#>     ▆
#>  1. └─vctrs::num_as_location(...)
#>  2.   └─vctrs (local) `<fn>`()
#>  3.     └─vctrs:::stop_subscript_oob(...)
#>  4.       └─vctrs:::stop_subscript(...)
#>  5.         └─rlang::abort(...)
```

<sup>Created on 2022-08-29 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

DavisVaughan:
Closes #1614 
Closes #1630 

Slightly more invasive change than the other `num_as_location()` changes so far. I've restructured the for-loop to be one series of flat `if/else` statements. The main change that comes from this (beyond looking a little nicer IMO) is that the OOB behavior for negative values becomes more explicit because it gets its own `switch()` statement under `LOC_NEGATIVE_IGNORE`. This makes sense to me because it has slightly different behavior from the normal OOB behavior.

### DIFF
--- a/R/subscript-loc.R
+++ b/R/subscript-loc.R
@@ -481,11 +481,16 @@ cnd_body.vctrs_error_subscript_oob <- function(cnd, ...) {
 cnd_body_vctrs_error_subscript_oob_location <- function(cnd, ...) {
   i <- cnd$i
 
-  # In case of negative indexing
-  i <- abs(i)
-
   # In case of missing locations
   i <- i[!is.na(i)]
+
+  if (cnd_subscript_action(cnd) == "negate") {
+    # Only report negative indices
+    i <- i[i < 0L]
+  }
+
+  # In case of negative indexing
+  i <- abs(i)
 
   oob <- i[i > cnd$size]
   oob_enum <- vctrs_cli_vec(oob)
--- a/src/subscript-loc.c
+++ b/src/subscript-loc.c
@@ -142,27 +142,32 @@ r_obj* int_as_location(r_obj* subscript,
       if (opts->missing == SUBSCRIPT_MISSING_ERROR) {
         stop_subscript_missing(subscript, opts);
       }
-    } else {
-      if (elt < 0) {
-        switch (opts->loc_negative) {
-        case LOC_NEGATIVE_INVERT: return int_invert_location(subscript, n, opts);
-        case LOC_NEGATIVE_ERROR: stop_location_negative(subscript, opts);
-        case LOC_NEGATIVE_IGNORE: break;
-        }
+    } else if (elt == 0) {
+      switch (opts->loc_zero) {
+      case LOC_ZERO_REMOVE: ++n_zero; break;
+      case LOC_ZERO_ERROR: stop_location_zero(subscript, opts);
+      case LOC_ZERO_IGNORE: break;
       }
-
-      if (elt == 0) {
-        switch (opts->loc_zero) {
-        case LOC_ZERO_REMOVE: ++n_zero; break;
-        case LOC_ZERO_ERROR: stop_location_zero(subscript, opts);
-        case LOC_ZERO_IGNORE: break;
+    } else if (elt < 0) {
+      switch (opts->loc_negative) {
+      case LOC_NEGATIVE_INVERT: return int_invert_location(subscript, n, opts);
+      case LOC_NEGATIVE_ERROR: stop_location_negative(subscript, opts);
+      case LOC_NEGATIVE_IGNORE: {
+        if (abs(elt) > n) {
+          switch (opts->loc_oob) {
+          case LOC_OOB_ERROR: stop_subscript_oob_location(subscript, n, opts);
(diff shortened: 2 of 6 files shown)


R/subscript-loc.R
```diff
@@ -481,12 +481,17 @@ cnd_body.vctrs_error_subscript_oob <- function(cnd, ...) {
...
   # In case of missing locations
   i <- i[!is.na(i)]
 
+  if (cnd_subscript_action(cnd) == "negate") {
+    # Only report negative indices
+    i <- i[i < 0L]
+  }
+
+  # In case of negative indexing
+  i <- abs(i)
```
DavisVaughan, line 493: This is actually #1630, but is also important here because it uses the same error

src/subscript-loc.c
```diff
@@ -142,27 +142,32 @@ r_obj* int_as_location(r_obj* subscript,
...
         stop_subscript_missing(subscript, opts);
       }
-    } else {
-      if (elt < 0) {
-        switch (opts->loc_negative) {
-        case LOC_NEGATIVE_INVERT: return int_invert_location(subscript, n, opts);
-        case LOC_NEGATIVE_ERROR: stop_location_negative(subscript, opts);
-        case LOC_NEGATIVE_IGNORE: break;
-        }
+    } else if (elt == 0) {
```
DavisVaughan, line 145: Nothing changed with the 0 case, I just lifted it up to be the 2nd condition here (after `NA`)

src/subscript-loc.c
```diff
@@ -142,27 +142,32 @@ r_obj* int_as_location(r_obj* subscript,
...
+        if (abs(elt) > n) {
+          switch (opts->loc_oob) {
+          case LOC_OOB_ERROR: stop_subscript_oob_location(subscript, n, opts);
+          case LOC_OOB_EXTEND: stop_subscript_negative_oob_location(subscript, n, opts);
+          case LOC_OOB_REMOVE: ++n_oob; break;
+          }
         }
+        break;
+      }
+      }
```
DavisVaughan, line 165: This section now fully handles a negative index, rather than letting other code handle the details of the `LOC_NEGATIVE_IGNORE` case (which is what resulted in the original issue here)

src/subscript-loc.c
```diff
@@ -142,27 +142,32 @@ r_obj* int_as_location(r_obj* subscript,
...
-        case LOC_OOB_REMOVE: ++n_oob; break;
+    } else if (elt < 0) {
+      switch (opts->loc_negative) {
+      case LOC_NEGATIVE_INVERT: return int_invert_location(subscript, n, opts);
+      case LOC_NEGATIVE_ERROR: stop_location_negative(subscript, opts);
+      case LOC_NEGATIVE_IGNORE: {
+        if (abs(elt) > n) {
+          switch (opts->loc_oob) {
+          case LOC_OOB_ERROR: stop_subscript_oob_location(subscript, n, opts);
+          case LOC_OOB_EXTEND: stop_subscript_negative_oob_location(subscript, n, opts);
```
DavisVaughan, line 159: These purposefully use different error messages

With `oob = "error", negative = "ignore"`, we want to report both negative and positive oob locations

With `oob = "extend", negative = "ignore"`, we only want to report negative oob locations (because positive oob locations might not actually be oob, they might be valid `"extend"` indices)

I've added tests for both of these cases

src/subscript-loc.c
```diff
@@ -142,27 +142,32 @@ r_obj* int_as_location(r_obj* subscript,
...
+          case LOC_OOB_REMOVE: ++n_oob; break;
+          }
         }
+        break;
+      }
+      }
+    } else if (elt > n) {
```
DavisVaughan, line 166: No longer needs `abs(elt) > n` because the negative case is fully covered in the previous branch

lionel- (approved):
I haven't closely analysed your changes to the loop body but they seem to make sense. The new errors in the unit tests look good and consistent!

PR #1633, opened 2022-08-30
Implement `missing = "remove"`

### LINKED ISSUE #1489: Should `vec_as_location()` treat `""` and `NA_character_` indices like `NA_integer_`?
Since in R, `""` and `NA_character_` are never matched as names

``` r
library(vctrs)

x <- c(1, 2)
names(x) <- c("", NA_character_)

# Treated like `NA_integer_`?
x[""]
#> <NA> 
#>   NA
x[NA_character_]
#> <NA> 
#>   NA

# Seems to actually match
vec_slice(x, "")
#>   
#> 1
vec_slice(x, NA_character_)
#> <NA> 
#>    2

# Source of this:
vec_as_location("", 2L, names(x))
#> [1] 1
vec_as_location(NA_character_, 2L, names(x))
#> [1] 2
```

<sup>Created on 2021-11-11 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

### LINKED ISSUE #1595: `num_as_location()` needs more options to support `dplyr::slice()`

DavisVaughan:
Closes #1595 
Closes #1489

Another fairly complex PR unfortunately, but I've tried to keep each commit self contained again

- I tackled #1489 while I was here, because it was staring me in the face as I modified `chr_as_location()`

- `num_as_location2()` and `vec_as_location2()` did not gain this argument. They have to select exactly 1 element so it doesn't make sense there

I'll mark some other important things inline

### DIFF
--- a/src/subscript-loc.c
+++ b/src/subscript-loc.c
@@ -581,7 +581,7 @@ enum num_loc_zero parse_loc_zero(r_obj* x,
 
 static
 void stop_subscript_arg_missing(struct r_lazy call) {
-  r_abort_call(call.env, "`missing` must be one of \"propagate\" or \"error\".");
+  r_abort_call(call.env, "`missing` must be one of \"propagate\", \"remove\", or \"error\".");
 }
 static
 void stop_bad_negative(struct r_lazy call) {
--- a/src/subscript-loc.c
+++ b/src/subscript-loc.c
@@ -255,6 +255,8 @@ r_obj* int_invert_location(r_obj* subscript,
     int j = *data;
 
     if (j == r_globals.na_int) {
+      // Following base R by erroring on `missing = "propagate"`, e.g. `1[c(NA, -1)]`.
+      // Doesn't make sense to invert an `NA`, so we can't meaningfully propagate.
       switch (opts->missing) {
       case SUBSCRIPT_MISSING_PROPAGATE: stop_location_negative_missing(subscript, opts);
       case SUBSCRIPT_MISSING_REMOVE: continue;
--- a/src/subscript-loc.c
+++ b/src/subscript-loc.c
@@ -163,13 +163,16 @@ r_obj* int_as_location(r_obj* subscript,
   r_ssize n_zero = 0;
 
   r_ssize n_oob = 0;
+  r_ssize n_missing = 0;
 
   for (r_ssize i = 0; i < loc_n; ++i, ++data) {
     int elt = *data;
 
     if (elt == r_globals.na_int) {
-      if (opts->missing == SUBSCRIPT_MISSING_ERROR) {
-        stop_subscript_missing(subscript, opts);
+      switch (opts->missing) {
+      case SUBSCRIPT_MISSING_PROPAGATE: break;
+      case SUBSCRIPT_MISSING_REMOVE: ++n_missing; break;
+      case SUBSCRIPT_MISSING_ERROR: stop_subscript_missing(subscript, opts);
       }
     } else if (elt == 0) {
       switch (opts->loc_zero) {
@@ -204,7 +207,12 @@ r_obj* int_as_location(r_obj* subscript,
   r_keep_loc subscript_shelter;
   KEEP_HERE(subscript, &subscript_shelter);
 
-  if (n_zero) {
+  if (n_missing > 0) {
+    subscript = int_filter_missing(subscript, n_missing);
+    KEEP_AT(subscript, subscript_shelter);
+  }
+
+  if (n_zero > 0) {
     subscript = int_filter_zero(subscript, n_zero);
     KEEP_AT(subscript, subscript_shelter);
   }
@@ -246,8 +254,13 @@ r_obj* int_invert_location(r_obj* subscript,
     int j = *data;
 
     if (j == r_globals.na_int) {
-      stop_location_negative_missing(subscript, opts);
+      switch (opts->missing) {
+      case SUBSCRIPT_MISSING_PROPAGATE: stop_location_negative_missing(subscript, opts);
+      case SUBSCRIPT_MISSING_REMOVE: continue;
+      case SUBSCRIPT_MISSING_ERROR: stop_location_negative_missing(subscript, opts);
+      }
     }
+
     if (j >= 0) {
       if (j == 0) {
         switch (opts->loc_zero) {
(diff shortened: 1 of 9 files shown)


src/subscript-loc.c
```diff
@@ -62,18 +62,27 @@ r_obj* lgl_as_location(r_obj* subscript,
                        const struct location_opts* opts) {
   r_ssize subscript_n = r_length(subscript);
 
-  if (opts->missing == SUBSCRIPT_MISSING_ERROR && lgl_any_na(subscript)) {
-    stop_subscript_missing(subscript, opts);
-  }
```
DavisVaughan, line 67: Now tackled directly in the `subscript_n == n` and `subscript_n == 1` branches

src/subscript-loc.c
```diff
@@ -92,23 +101,43 @@ r_obj* lgl_as_location(r_obj* subscript,
   if (subscript_n == 1) {
     int elt = r_lgl_get(subscript, 0);
 
-    r_obj* out;
+    r_ssize recycle_size = n;
```
DavisVaughan, line 104: This is specific to handling names later on. In the two cases when we return `integer()`, we recycle names to size `0` rather than size `n`

This is one of the names handling things that feels more complicated than it is worth

src/subscript-loc.c
```diff
@@ -92,23 +101,43 @@ r_obj* lgl_as_location(r_obj* subscript,
...
-      r_int_fill(out, r_globals.na_int, n);
+      switch (opts->missing) {
+      case SUBSCRIPT_MISSING_PROPAGATE: {
+        out = r_alloc_integer(n);
+        KEEP_AT(out, out_shelter);
+        r_int_fill(out, r_globals.na_int, n);
+        break;
+      }
+      case SUBSCRIPT_MISSING_REMOVE: {
+        out = vctrs_shared_empty_int;
```
DavisVaughan, line 119: i.e.

```r
vec_as_location(NA, n = 5, missing = "remove")
```

src/subscript-loc.c
```diff
@@ -217,8 +254,13 @@ r_obj* int_invert_location(r_obj* subscript,
     int j = *data;
 
     if (j == r_globals.na_int) {
-      stop_location_negative_missing(subscript, opts);
+      switch (opts->missing) {
+      case SUBSCRIPT_MISSING_PROPAGATE: stop_location_negative_missing(subscript, opts);
+      case SUBSCRIPT_MISSING_REMOVE: continue;
```
DavisVaughan, line 259: As discussed:

```r
# remains an error
num_as_location(c(-1, NA), n = 2, missing = "propagate", negative = "invert")

# this is allowed
num_as_location(c(-1, NA), n = 2, missing = "remove", negative = "invert")
```

src/subscript-loc.c
```diff
@@ -256,25 +298,34 @@ r_obj* int_invert_location(r_obj* subscript,
 }
 
 static
-r_obj* int_filter_zero(r_obj* subscript,
-                       r_ssize n_zero) {
-  r_ssize loc_n = vec_size(subscript);
-  const int* data = r_int_cbegin(subscript);
+r_obj* int_filter(r_obj* subscript, r_ssize n_filter, int value) {
```
DavisVaughan, line 301: Generalized this a bit for `int_filter_missing()` and updated the style

src/subscript-loc.c
```diff
@@ -385,31 +436,52 @@ r_obj* chr_as_location(r_obj* subscript,
...
 
   for (r_ssize k = 0; k < n; ++k) {
-    if (p[k] != r_globals.na_int) {
-      continue;
+    const r_obj* elt = ip[k];
+
+    if (elt == r_strs.empty) {
+      // `""` never matches, even if `names` contains a `""` name
+      stop_subscript_empty(subscript, opts);
+    }
```
DavisVaughan, line 454: i.e. if the user supplies `""` as an index in `vec_as_location()`, that is unconditionally an error because it can't match anything (#1489)

src/subscript-loc.c
lionel-: I think we're going to assign names to that shared object, below?

tests/testthat/test-subscript-loc.R
```diff
@@ -401,6 +401,91 @@ test_that("can disallow missing values", {
   })
 })
 
+test_that("can alter logical missing value handling (#1595)", {
+  x <- c(TRUE, NA, FALSE, NA)
```
lionel-, line 405: Probably best to use a named vector in the logical and integer cases too, as you did in the character case.

src/subscript-loc.c
```diff
@@ -92,23 +101,43 @@ r_obj* lgl_as_location(r_obj* subscript,
...
+      }
+      }
     } else if (elt) {
-      out = KEEP(r_alloc_integer(n));
+      out = r_alloc_integer(n);
+      KEEP_AT(out, out_shelter);
       r_int_fill_seq(out, 1, n);
     } else {
-      return vctrs_shared_empty_int;
+      out = vctrs_shared_empty_int;
```
lionel-, line 132: Also need to create a new object here.

src/subscript-loc.c
lionel-: Add a comment that we're just following base R here? And that it does not really make sense to invert an `NA` value so we can't meaningfully propagate, and we'd rather not drop missing values where base R doesn't.

lionel- (approved):

src/subscript-loc.c
DavisVaughan: Good catch!

tests/testthat/test-subscript-loc.R
DavisVaughan: That revealed a few more places where this comes up
https://github.com/r-lib/vctrs/pull/1633/commits/d3ca2ad8e2eae8b2d53cabaf3b46b87c3db8ae49

I've prepped for the merge of r-lib/rlang#1471, ill leave this open until we can rebase on that to fix these tests
https://github.com/r-lib/vctrs/pull/1633/commits/c1c6f2c3d7b60201aadb57976322e8ad44bc7975

PR #1638, opened 2022-09-01
Remove `logical` argument of `vec_as_subscript2()`

### LINKED ISSUE #1608: `vec_as_subscript2()` has a misleading error when given logical input with `logical = "cast"`
Look at the first message:

> Subscript has the wrong type `logical`.

> It must be logical, numeric, or character.

``` r
library(vctrs)

vec_as_subscript2(TRUE, logical = "cast")
#> Error:
#> ! Must extract element with a single valid subscript.
#> ✖ Subscript has the wrong type `logical`.

vec_as_subscript2(TRUE, logical = "error")
#> Error:
#> ! Must extract element with a single valid subscript.
#> ✖ Subscript has the wrong type `logical`.
#> ℹ It must be numeric or character.
```

Also, why do we even expose a `logical` argument? Shouldn't it always be error? We internally error right before exiting if we have a logical argument.

lionel-:
Closes #1608

### DIFF
--- a/R/subscript.R
+++ b/R/subscript.R
@@ -70,9 +70,9 @@ vec_as_subscript2 <- function(i,
                               call = caller_env()) {
   check_dots <- function(..., logical = "error") {
     if (!is_string(logical, "error")) {
-      lifecycle::deprecate_stop(
-        "0.4.1.9000",
-        "vctrs::vec_as_subscript2(logical = 'no longer supports \"cast\"')"
+      abort(
+        "`vctrs::vec_as_subscript2(logical = 'cast')` is deprecated.",
+        call = caller_env()
       )
     }
     check_dots_empty0(...)
--- a/R/subscript.R
+++ b/R/subscript.R
@@ -64,17 +64,25 @@ vec_as_subscript_result <- function(i,
 #' @export
 vec_as_subscript2 <- function(i,
                               ...,
-                              logical = c("cast", "error"),
                               numeric = c("cast", "error"),
                               character = c("cast", "error"),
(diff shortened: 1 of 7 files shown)


lionel-:
Will need to update tidyselect and tibble. And keep the argument around in vctrs for some while in case users update vctrs but not tidyselect or tibble.

R/subscript.R
```diff
@@ -64,36 +64,42 @@ vec_as_subscript_result <- function(i,
...
                               call = caller_env()) {
-  check_dots_empty0(...)
+  check_dots <- function(..., logical = "error") {
+    if (!is_string(logical, "error")) {
+      abort(
+        "`vctrs::vec_as_subscript2(logical = 'cast')` is deprecated.",
+        call = caller_env()
+      )
+    }
+    check_dots_empty0(...)
```
DavisVaughan, line 78: Do we want to tweak the `call` of `check_dots_empty0()` to also be `caller_env()`?

R/subscript.R
```diff
@@ -126,7 +131,6 @@ vec_as_subscript2_result <- function(i,
   if (typeof(i) == "logical") {
```
DavisVaughan, line 131: Can you remove this section now? That's what I was hoping for

DavisVaughan (approved):

R/subscript.R
lionel-: yup good catch

R/subscript.R
lionel-: Thanks, now simplified.

PR #1639, opened 2022-09-01
Simplify and fix scalar subscript errors

### LINKED ISSUE #1606: `vec_as_subscript2()` doesn't throw lossy cast errors correctly

lionel-:
Branched from #1638.
Fixes #1606.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -30,6 +30,7 @@ Depends:
     R (>= 3.3)
 Imports:
     cli (>= 3.2.0),
+    lifecycle (>= 1.0.1),
     glue,
     rlang (>= 1.0.4)
 Suggests: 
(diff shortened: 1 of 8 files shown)


lionel-:
oops, forgot to change the base branch before deleting it. The PR got closed and can't be reopened, so I just pushed the commit directly to main.

PR #1646, opened 2022-09-07
Move `vec_order()` and `vec_sort()` to radix-backed implementation

DavisVaughan:
Follow up to https://github.com/r-lib/vctrs/pull/1621, https://github.com/r-lib/vctrs/pull/1521

- Removes old `vec_order()` and `vec_sort()` code
- Renames `vec_order_radix()` -> `vec_order()`
- Renames `vec_sort_radix()` -> `vec_sort()`
- Keeps `vec_order_base()`, but it is unexported. Might be useful for us in the future

I've bumped the dev version and will do a dplyr PR against this PR that updates dev dplyr to use `vec_order()` everywhere (rather than `vec_order_radix()`).

I'll also check revdeps and report back `"36b7efe8-7d9b-4201-9083-cdb15a3bd01e"`


DavisVaughan:
Closing in favor of this plan:

* vctrs 0.5.0 still has system-locale backed `vec_order()`
* vctrs 0.5.0 has internal `vec_order_radix()`
* dplyr 1.1.0 grabs `vctrs:::vec_order_radix()` and uses it. dplyr no longer calls `vec_order()`
* vctrs 0.6.0 moves `vec_order()` to radix backed implementation (which should not affect dplyr at all!)
* dplyr 1.2.0 requires vctrs 0.6.0 and uses `vec_order()`
* Future vctrs release removes `vec_order_radix()` after we feel that users have updated to dplyr 1.2.0

That way we never publicly expose a new function to clutter the api, and dplyr is still safe

PR #1649, opened 2022-09-13
Remove experimental `n` argument of `vec_restore()`

### LINKED ISSUE #650: Consider letting `vec_restore()` take an R_len_t `n` rather than a SEXP

lionel-:
It was only used to restore data frames from bare lists, which was kind of an abstraction leak. Instead, we now initialise bare lists to data frame before calling `vec_restore()`.

Closes #650
cc @hadley

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,12 @@
 # vctrs (development version)
 
+* The experimental `n` argument of `vec_restore()` has been
+  removed. It was only used to inform on the size of data frames in
+  case a bare list is restored. It is now expected that bare lists be
+  initialised to data frame so that the size is carried through row
+  attributes. This makes the generic simpler and fixes some
+  performance issues (#650).
+
 * The `anyNA()` method for `vctrs_vctr` (and thus `vctrs_list_of`) now
   supports the `recursive` argument (#1278).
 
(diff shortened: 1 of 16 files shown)


DavisVaughan (approved):
Looks great! In retrospect, ensuring that `df_slice()` returns a data frame now seems somewhat obvious to me

lionel-:
yeah I was wondering why we didn't think of that earlier.

PR #1651, opened 2022-09-13
Unclass clonable proxies to prevent further dispatch

### LINKED ISSUE #1129: vec_restore() and vec_rbind()
I think I need some help here.

I have defined `vec_restore()` for the tsibble object, which solves the attribute updating issue when `vec_slice()` a tsibble. But it breaks `bind_rows()` that uses `vec_rbind()` internally.

In the tsibble code, I need to make sure that there's no `NA` in my index column, and `validate_index()` does the job. I have no idea when `NA` has been introduced in the `vec_restore()`. When I debug through my `validate_index()`, it will not trigger that error because of no `NA`.

``` r
library(tsibble)
library(vctrs)
library(lubridate)
#> 
#> Attaching package: 'lubridate'
#> The following object is masked from 'package:tsibble':
#> 
#>     interval
#> The following objects are masked from 'package:base':
#> 
#>     date, intersect, setdiff, union
idx_day <- seq.Date(ymd("2017-01-01"), ymd("2017-01-20"), by = 4)
tsbl <- tsibble(
  date = idx_day,
  value = rnorm(5),
  index = date
)
tsbl
#> # A tsibble: 5 x 2 [4D]
#>   date         value
#>   <date>       <dbl>
#> 1 2017-01-01  0.965 
#> 2 2017-01-05  0.0740
#> 3 2017-01-09 -1.35  
#> 4 2017-01-13 -0.257 
#> 5 2017-01-17 -0.814
new_data(tsbl)
#> # A tsibble: 1 x 1 [4D]
#>   date      
#>   <date>    
#> 1 2017-01-21

vec_rbind(tsbl, new_data(tsbl))
#> Error: Column `date` (index) must not contain `NA`.
```

<sup>Created on 2020-05-29 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

lionel-:
Branched from #1648.
Closes #1129.

### DIFF
--- a/R/type2.R
+++ b/R/type2.R
@@ -216,14 +216,24 @@ can_fall_back <- function(x, y) {
   }
 
   # Suboptimal: Prevent bad interaction with proxy-assign
-  has_no_proxy(x) && has_no_proxy(y)
+  !is_proxied(x) && !is_proxied(y)
 }
-has_no_proxy <- function(x) {
+
+# Don't compare data for performance
+is_proxied <- function(x) {
   proxy <- vec_proxy(x)
 
-  # Don't compare data for performance
-  identical(typeof(x), typeof(proxy)) &&
-    identical(attributes(x), attributes(proxy))
+  if (!identical(typeof(x), typeof(proxy))) {
+    return(FALSE)
+  }
+
+  if (inherits(x, "list")) {
+    class(x) <- NULL
+  } else if (inherits(x, "data.frame")) {
+    class(x) <- "data.frame"
+  }
+
+  !identical(attributes(x), attributes(proxy))
 }
 
 new_common_class_fallback <- function(x, fallback_class) {
--- a/src/proxy.c
+++ b/src/proxy.c
@@ -3,14 +3,19 @@
 #include "decl/proxy-decl.h"
 
 
-// [[ register() ]]
 r_obj* vec_proxy(r_obj* x) {
   return vec_proxy_2(x, false);
 }
-
-// [[ register() ]]
 r_obj* vec_proxy_recurse(r_obj* x) {
   return vec_proxy_2(x, true);
+}
+
+r_obj* ffi_vec_proxy(r_obj* x, r_obj* recurse) {
+  if (r_as_bool(recurse)) {
+    return vec_proxy_2(x, true);
+  } else {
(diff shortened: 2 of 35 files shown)


R/type2.R
```diff
@@ -216,14 +216,24 @@ can_fall_back <- function(x, y) {
...
-    identical(attributes(x), attributes(proxy))
+  if (!identical(typeof(x), typeof(proxy))) {
+    return(FALSE)
+  }
+
+  if (inherits(x, "list")) {
+    class(x) <- NULL
+  } else if (inherits(x, "data.frame")) {
+    class(x) <- "data.frame"
+  }
```
DavisVaughan, line 234: Maybe a comment about why we do this for these two types in particular? Linking to a relevant PR/issue?

src/proxy.c
```diff
@@ -32,6 +32,24 @@ r_obj* vec_proxy_2(r_obj* x, bool recurse) {
 
   case VCTRS_TYPE_s3: {
     r_obj* out = vec_proxy_invoke(x, info.proxy_method, recurse);
```
DavisVaughan, line 34: I think `out` needs to be protected now, because you might call `r_clone()`, which allocates?

tests/testthat/test-bind.R
```diff
@@ -1093,6 +1093,34 @@ test_that("vec_rbind() only restores one time", {
   ))
 })
 
+test_that("vec_rbind() only restores one time (data frame case)", {
+  restored <- list()
+
+  local_methods(
+    vec_ptype2.vctrs_foobar.vctrs_foobar = function(x, y, ...) foobar(df_ptype2(x, y, ...)),
+    vec_cast.vctrs_foobar.vctrs_foobar = function(x, to, ...) foobar(df_cast(x, to, ...)),
+    vec_proxy.vctrs_foobar = function(x, ...) x,
```
DavisVaughan, line 1102: So I guess the alternative to this PR is to require that data frame subclasses provide a `vec_proxy()` method that just strips the class, which I then realized is what you advised here https://github.com/r-lib/vctrs/issues/1129#issuecomment-635991682

I agree that this PR is a good improvement, even if a little awkward

DavisVaughan (approved):

PR #1653, opened 2022-09-13
Support `anyNA(recursive = TRUE)` for `vctrs_vctr`

### LINKED ISSUE #1278: Different behaviour of anyNA() for lists and vctrs_list_of
### What's the problem?
Function `anyNA()` returns different results when called on `list` or `vctrs_list_of`, even though both have the same values. I suspect it is due to overriding this method for `vctrs_vctr` class and ignoring `recursive` parameter.

### What do I expect?
I expect `anyNA.vctrs_list_of()` to return the same values as do `anyNA.list()` now.

### Steps to reproduce
```r
t <- list(1:4, c(2, NA, 5))
anyNA(t)  # returns FALSE
anyNA(t, recursive = TRUE)  # returns TRUE

tt <- as_list_of(t)
anyNA(tt)  # returns FALSE
anyNA(tt, recursive = TRUE)  # returns FALSE, should return TRUE
```

### Which version do I use?
I used `vctrs v0.3.2` and I suspect that no other info is necessary here (if I'm wrong, I'll gladly provide more info).

lionel-:
Closes #1278

### DIFF
--- a/R/type-vctr.R
+++ b/R/type-vctr.R
@@ -460,12 +460,10 @@ na_remove <- function(x, type) {
 }
 
 #' @export
-anyNA.vctrs_vctr <- if (getRversion() >= "3.2") {
-  function(x, recursive = FALSE) {
-    any(is.na(x))
-  }
-} else {
-  function(x) {
+anyNA.vctrs_vctr <- function(x, recursive = FALSE) {
+  if (recursive && vec_is_list(x)) {
+    any(map_lgl(x, anyNA))
+  } else {
     any(is.na(x))
   }
 }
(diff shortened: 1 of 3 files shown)


R/type-vctr.R
```diff
@@ -460,12 +460,10 @@ na_remove <- function(x, type) {
...
 #' @export
-anyNA.vctrs_vctr <- if (getRversion() >= "3.2") {
-  function(x, recursive = FALSE) {
-    any(is.na(x))
-  }
-} else {
-  function(x) {
+anyNA.vctrs_vctr <- function(x, recursive = FALSE) {
+  if (recursive && vec_is_list(x)) {
+    any(map_lgl(x, anyNA))
```
DavisVaughan, line 465: Should it pass on `recursive`?

lionel-: yup good catch

PR #1657, opened 2022-09-14
Add call and arg arguments to `vec_rep()` and friends

### LINKED ISSUE #1303: Add x_arg/times_arg arguments to vec-rep and vec-rep-each
That could have been useful in `uncount()`, where `times` is really `weights`
https://github.com/tidyverse/tidyr/pull/1070#discussion_r537527020

lionel-:
Closes #1303.

### DIFF
--- a/R/rep.R
+++ b/R/rep.R
@@ -28,6 +28,8 @@
 #'   length. This means that `vec_unrep()` works on data frames by compressing
 #'   repeated rows.
 #'
+#' @inheritParams rlang::args_error_context
+#' @inheritParams rlang::args_dots_empty
 #' @param x A vector.
 #' @param times
 #'   For `vec_rep()`, a single integer for the number of times to repeat
@@ -36,7 +38,6 @@
 #'   For `vec_rep_each()`, an integer vector of the number of times to repeat
 #'   each element of `x`. `times` will be [recycled][vector_recycling_rules] to
 #'   the size of `x`.
-#' @inheritParams rlang::args_error_context
 #' @param x_arg,times_arg Argument names for errors.
 #'
 #' @return
@@ -87,9 +88,11 @@ NULL
 #' @export
 vec_rep <- function(x,
                     times,
+                    ...,
                     call = current_env(),
                     x_arg = "x",
                     times_arg = "times") {
+  check_dots_empty0(...)
   .Call(ffi_vec_rep, x, times, environment())
 }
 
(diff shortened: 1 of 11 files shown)


R/rep.R
```diff
@@ -83,14 +85,22 @@ NULL
 
 #' @rdname vec-rep
 #' @export
-vec_rep <- function(x, times) {
-  .Call(ffi_vec_rep, x, times)
+vec_rep <- function(x,
+                    times,
```
DavisVaughan, line 89: Strongly recommend adding empty `...` in both of these functions to require named optional args

R/rep.R
```diff
@@ -36,6 +36,8 @@
 #'   For `vec_rep_each()`, an integer vector of the number of times to repeat
 #'   each element of `x`. `times` will be [recycled][vector_recycling_rules] to
 #'   the size of `x`.
+#' @inheritParams rlang::args_error_context
```
DavisVaughan, line 39: I thought we always put `@inheritParams` tags above all other `@param` tags (i.e. `@inheritParams` would be the first tag listed)

src/rep.c
```diff
@@ -1,11 +1,17 @@
+#include "rlang.h"
 #include "vctrs.h"
 #include "type-data-frame.h"
 #include "decl/rep-decl.h"
+#include "vec.h"
```
DavisVaughan, line 5:
```suggestion
#include "vec.h"
#include "decl/rep-decl.h"
```

I think we always want decl/ headers last?

src/rep.c
```diff
@@ -379,9 +440,4 @@ r_obj* new_unrep_data_frame(r_obj* key, r_obj* times, r_ssize size) {
 
 // -----------------------------------------------------------------------------
 
-void vctrs_init_rep(r_obj* ns) {
-  args_times_ = new_wrapper_arg(NULL, "times");
-}
-
-static
-struct vctrs_arg* const p_args_times = &args_times_;
+void vctrs_init_rep(r_obj* ns) { }
```
DavisVaughan, line 443: I'm fine with removing this altogether now. We can add it back if we need it

DavisVaughan (approved):

R/rep.R
lionel-: oh yeah

src/rep.c
lionel-: I think it's ok to leave it there, and then it's less work if we need it again. I.e. with ideal tooling every file would get a init hook automatically.

Issue #1641, opened 2022-09-03
Should vec_unchop()/list_unchop() have `error_call` argument?


hadley:

DavisVaughan:
Probably, and probably `vec_c()`

I imagine them all defaulting to `call = current_env()`, as these are somewhat likely to be called at the top level by advanced users

hadley:
And probably `error_arg`, at least for `list_unchop()`

hadley:
And would be nice if 

``` r
vctrs::list_unchop(list(1, 2, "a"), ptype = integer())
#> Error:
#> ! Can't convert <character> to <integer>.
```

was something like

``` r
vctrs::list_unchop(list(1, 2, "a"), ptype = integer())
#> Error:
#> ! Can't convert x[[3]] from <character> to <integer>.
```

DavisVaughan:
Ironically the error is a little better if you provide indices

```r
vctrs::list_unchop(
  list(1, 2, "a"), 
  list(1, 2, 3),
  ptype = integer()
)
#> Error:
#> ! Can't convert `..3` <character> to <integer>.
```

It looks like `vec_c()` casts the inputs to the common type one at a time, while `list_unchop()` (when it has indices) uses `vec_cast_common()` to cast them all up front
https://github.com/r-lib/vctrs/blob/facb752cd3da921a0d4c2a0d139df4a036d1c13e/src/c.c#L131

Ideally I think `vec_c()` would also do the cast up front, which would allow it to throw a positional error like this, but currently casting can drop names, so we will have to be a little careful about it.

PR #1660, opened 2022-09-14
Add `.call` to `vec_c()`

DavisVaughan:
Part of #1641 

Might have merge conflicts with #1648 so I'll wait until that is merged to merge this one.

I'll do a follow up PR to add `call` to `list_unchop()`, but we needed this first.


tests/testthat/_snaps/bind.md
```diff
@@ -371,7 +371,7 @@
       )
     Output
       <error/vctrs_error_incompatible_type>
-      Error:
+      Error in `vec_rbind()`:
```
DavisVaughan, line 374: Nice side effect

lionel- (approved):

PR #1662, opened 2022-09-15
Another pass at call and arg arguments for `vec_as_names()`

lionel-:
Still not perfect but a good improvement.

### DIFF
--- a/R/names.R
+++ b/R/names.R
@@ -158,7 +158,7 @@
 vec_as_names <- function(names,
                          ...,
                          repair = c("minimal", "unique", "universal", "check_unique"),
-                         repair_arg = "",
+                         repair_arg = caller_arg(repair),
                          quiet = FALSE,
                          call = caller_env()) {
   check_dots_empty0(...)
--- a/tests/testthat/_snaps/conditions.md
+++ b/tests/testthat/_snaps/conditions.md
@@ -124,6 +124,7 @@
       x These names are duplicated:
         * "x" at locations 1, 2, and 3.
         * "y" at locations 4 and 5.
+      i Use argument `"check_unique"` to specify repair strategy.
     Code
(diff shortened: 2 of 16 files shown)


tests/testthat/_snaps/conditions.md
```diff
@@ -124,6 +124,7 @@
       x These names are duplicated:
         * "x" at locations 1, 2, and 3.
         * "y" at locations 4 and 5.
+      i Use argument `"check_unique"` to specify repair strategy.
```
DavisVaughan, line 127: @lionel-, This doesn't look correct, right?

R/names.R
```diff
@@ -158,7 +158,7 @@
 vec_as_names <- function(names,
                          ...,
                          repair = c("minimal", "unique", "universal", "check_unique"),
-                         repair_arg = "",
+                         repair_arg = caller_arg(repair),
```
DavisVaughan, line 161: I'm not sure this is right, I feel like it was correct as `""` because we don't want to reference `"minimal"` and friends in the output, we want to say what the name of `repair` was in the top level function we call `vec_as_names()` from

lionel-: I'm not sure I understand what you mean regarding references "minimal". This is looking for the name of `repair` in the caller.

tests/testthat/_snaps/conditions.md
lionel-: oh I see what you mean now. Hmmmm

R/names.R
lionel-: I think we need to take `repair_arg = NULL`, and then treat two different cases:

- If `NULL`, the repair strategy is fixed. Wrong values are _internal_ errors that need to use the internal call. We also don't provide recommendations about changing the repair strategy.

- If a string, the repair strategy is provided by the user. Wrong values are _user_ errors and we do want to provide recommendations about repair strategy.

DavisVaughan: I think that sounds right!

PR #1665, opened 2022-09-16
Distinguish between internal and user `repair_arg` errors

lionel-:
Follow up to #1662

### DIFF
--- a/src/names.c
+++ b/src/names.c
@@ -39,7 +39,6 @@ r_obj* ffi_vec_as_names(r_obj* names,
                                                              quiet,
                                                              call);
   KEEP(repair_opts.shelter);
-  repair_opts.frame = frame;
 
   r_obj* out = vec_as_names(names, &repair_opts);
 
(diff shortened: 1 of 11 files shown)


src/names.c
```diff
@@ -820,13 +866,13 @@ struct name_repair_opts new_name_repair_opts(r_obj* name_repair,
     .fn = r_null,
     .name_repair_arg = name_repair_arg,
     .quiet = quiet,
-    .call = call
+    .call = call,
```
DavisVaughan, line 869: Purposeful?

DavisVaughan: I don't see the new `.frame` struct element being used anywhere?

DavisVaughan (approved):

src/names.c
lionel-: It is used in `new_repair_error_info()` to create an internal error call.

DavisVaughan: Yes, but I don't see where you ever set the `frame` element when you create a new `struct name_repair_opts`

lionel-: oh yeah what happened to that arg. Hmm I'll just call `peek_frame()`

lionel-: but what if `vec_as_names()` was called from C :-(
This stuff is too hard.

src/names.c
```diff
@@ -22,30 +22,82 @@ r_obj* vec_as_names(r_obj* names, const struct name_repair_opts* opts) {
...
+
+    // This won't be set when `vec_as_names()` is called from C
+    if (p_opts->frame) {
+      out.input_error_call = p_opts->frame;
+      r_list_poke(out.shelter, 3, out.input_error_call);
+    } else {
+      // Create fake `vec_as_names()` call for the C case
+      out.input_error_call = r_call(r_sym("vec_as_names"));
+    }
+    r_list_poke(out.shelter, 3, out.input_error_call);
```
DavisVaughan, line 87: Looks like you are double poking

DavisVaughan (approved):
LGTM now besides the double poke i mentioned

PR #1668, opened 2022-09-16
Clean up location documentation

DavisVaughan:
When working on the location functions, I noticed they had a few issues and could use with a general spruce up

- Mentioned `num_as_location/2()` in the description and return value sections
- Fixed a double inheritance issue in the `negative` parameter documentation
- Tweaked the oob param docs to specifically mention that only positive values can be extended
- Changed the parameters to a bulleted list form, which I find much easier to scan easily when reading the docs (both in the `.R` file and in the help pane)


lionel- (approved):
I like this style of enum documentation, thanks!

PR #1670, opened 2022-09-19
Fix error message when combining outer name with single inner name

### LINKED ISSUE #522: Better error message for vec_c() outer name failure
In this case the problem is that there are inner names and outer names, not that the vector has length > 1

``` r
vctrs::vec_c(x = c(xx = 1))
#> Error: Can't merge the outer name `x` with a vector of length > 1.
```

<sup>Created on 2019-08-03 by the [reprex package](https://reprex.tidyverse.org) (v0.2.1)</sup>

lionel-:
Closes #522

### DIFF
--- a/src/names.c
+++ b/src/names.c
@@ -589,11 +589,18 @@ r_obj* apply_name_spec(r_obj* name_spec, r_obj* outer, r_obj* inner, r_ssize n) 
   default:
     name_spec = r_as_function(name_spec, ".name_spec");
     break;
-  case R_TYPE_null:
-    r_abort("Can't merge the outer name `%s` with a vector of length > 1.\n"
+  case R_TYPE_null: {
+    const char* reason;
+    if (n > 1) {
+      reason = "a vector of length > 1";
+    } else {
+      reason = "a named vector";
+    }
+    r_abort("Can't merge the outer name `%s` with %s.\n"
             "Please supply a `.name_spec` specification.",
-            r_str_c_string(outer));
-  }
+            r_str_c_string(outer),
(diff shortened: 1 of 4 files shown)


src/names.c
```diff
@@ -589,11 +589,18 @@ r_obj* apply_name_spec(r_obj* name_spec, r_obj* outer, r_obj* inner, r_ssize n)
...
     break;
-  case R_TYPE_null:
-    r_abort("Can't merge the outer name `%s` with a vector of length > 1.\n"
+  case R_TYPE_null: {
+    const char* reason;
+    if (n > 1) {
+      reason = "a vector of length > 1";
+    } else {
+      reason = "a named vector";
+    }
```
DavisVaughan, line 598: Here I'm just explicitly documenting that the grid of possibilities is `outer` + one of:
- Length 1 unnamed vector (allowed)
- Length 1 named vector (`"a named vector"` error)
- Length >1 unnamed vector (`"a vector of length >1"` error)
- Length >1 named vector (`"a vector of length >1"` error)

And the following length 0 cases don't ever go through this error path (may or may not be worth a test, up to you)

``` r
# Length 0 unnamed vector
vctrs::vec_c(x = integer())
#> named integer(0)

# Length 0 named vector
vctrs::vec_c(x = rlang::set_names(integer(), character()))
#> named integer(0)
```

DavisVaughan (approved):

src/names.c
lionel-: The first test case might have been missing, I've added it in case.

PR #1677, opened 2022-09-24
More name repair strings

### LINKED ISSUE #1629: Proposal for `repair = "unique_quiet"` and `repair = "universal_quiet"`
I'd like to propose the addition of 2 name repair strings to the standard list of `vec_as_names()`:

``` r
vec_as_names(
  ...,
  repair = c("minimal", "unique", "universal", "check_unique"),
  quiet = FALSE,
  ...
)
```

I'm proposing we add:

* `unique_quiet`
* `universal_quiet`

which would have the effect of requesting a specific level of name repair AND setting `quiet = TRUE`.

I know this might sound sort of gross, but hear me out :sweat_smile: The existing `check_unique` is already a mix of a repair strategy and how to report on repair.

I'm writing this as the maintainer of readr, vroom, readxl, and googlesheets4, all of which do name repair, with various degrees of historical baggage, but all of which do now expose a `name_repair` argument. Some of these packages call `vctrs::vec_as_names()` directly (e.g. vroom), but others do not (readxl and googlesheets4 currently access name repair via tibble).

The ability to suppress name repair messages is a common and understandable feature request from users.

These feel like my current options for addressing this:

* Documentation: Show user how to write and use a custom name repair function that does not message.

  ``` r  
  quiet_repair <- function(x) {
    vctrs::vec_as_names(x, repair = "unique", quiet = TRUE)
  }

  read_csv(I("x,x\n1,2\n3,4"), name_repair = quiet_repair)
  ```
  
* Documentation: Show user how to use the `rlib_name_repair_verbosity` global option to suppress name repair messaging.
  
  ``` r
  withr::with_options(
    list(rlib_name_repair_verbosity = "quiet"),
    read_csv(I("x,x\n1,2\n3,4"))
  )
  ```

* Feature: Actually change the signature of all relevant user-facing functions in readr, vroom, readxl, and googlesheets4 to expose a new, 2nd argument re: name repair, that controls verbosity. One example:
  
  ``` r
  read_excel <- function(...
    .name_repair = "unique",            # <-- this is already there
    .name_repair_verbosity = "verbose"  # <-- this would be new
  )
  ```
  
* Feature: I could implement my proposal in each of these packages, by intercepting `name_repair` and, if I see `minimal_quiet` or `universal_quiet`, use whatever means I have available to make it so (implementation would differ across the packages).

I don't love any of these options. To quote @DavisVaughan:

> I do want to make it somewhat hard to silence [name repair messages], but right now I think it is a little too hard

The first 2 solutions (documentation, user does extra work) feel too hard on the user.

The third solution (new argument everywhere) feels like a lot of work/change for such a small behavioural tweak.

The fourth solution (special handling for the new strings in certain packages) feels like it breaks a tidyverse rule re: consistency.

I think `name_repair = "unique_quiet"` and `name_repair = "universal_quiet"` could be really handy for giving users more control with the single `name_repair` argument commonly seen in user-facing functions. To be clear, I think we should never default to these.

jennybc:
Closes #1629 

Before I go further, two big questions are emerging:

1. Should `unique_quiet` and `universal_quiet` be added to every `repair`/`[.]?name_repair()` argument?
    - I settled on YES if the function exposes `quiet` and NO otherwise. I'm not sure this is where we want to end up, though.
    - My decision might seem a little perverse, in that it means we don't offer this simple method of requesting `quiet = TRUE` for exactly those functions where you also can't specify `quiet` yourself. But it's because all of these functions effectively have `quiet = FALSE` hard-wired into them, which is presumably why they don't have `quiet` as an argument.
1. How should `repair` and `quiet` interact when the new strings are specified?
    - When `quiet = FALSE` (because it's the default) and `.name_repair = "*_quiet"`, it seems natural to silently ignore/override `quiet` and set `quiet = TRUE`.
    - When `quiet = FALSE` (because it was explicitly specified) and `.name_repair = "*_quiet"`, it's less clear what to do. Should we error/warn/message? vctrs is not currently set up for this, so the real question is if we care enough to do that work.
    - If we emit any sort of error/warning/message, it would be implemented in the R function `validate_name_repair_arg(repair =)`, which calls the C function `vctrs_validate_name_repair_arg()`, both of which would need to learn about jointly validating `repair` and `quiet`. (Depending on the answer to the first question, there might be even more work.)

(I haven't even started on documentation changes or NEWS bullet.)

(I **have** checked that even at this point, when I install vctrs from this PR, I can use `"unique_quiet"` and `"universal_quiet"` to get silent name repair in readxl, googlesheets4, readr, and vroom with zero code changes there and it's GLORIOUS.)

---

Some analysis I've done that might be handy as we discuss this

Functions that expose `repair`/ `.name_repair` and `quiet`:
  * `vec_names2(repair =, quiet = FALSE)` calls `validate_name_repair_arg(repair =)`
    - `vec_repair_names()` sends `repair` and `quiet` to `vec_names2()`
  * `vec_as_names(repair =, quiet = FALSE)` calls C function `ffi_vec_as_names()` and sends both `repair` and `quiet`. Should this function call `validate_name_repair_arg(repair =)` before going to C, similar to `vec_names2()`?
  
Functions that expose `repair`/ `[.]?name_repair` but do NOT expose `quiet`, basically go straight to C, where (eventually) `quiet = false` is hard-wired:
  * `vec_rbind()` calls C function `ffi_rbind()` and eventually `validate_bind_name_repair()`
  * `vec_cbind()` calls C function `ffi_cbind()` and eventually `validate_bind_name_repair()`
  * `vec_c()` calls C function `ffi_vec_c()`
  * `list_unchop()` calls C function `ffi_list_unchop()`
    - `vec_interleave()` calls `list_unchop()`
  * `data_frame()` calls C function `ffi_data_frame()`
  * `df_list()` calls C function `ffi_df_list()`

Callers of the C function `vec_as_names()`:

* `as_df_row_impl()`
* `vec_cbind()`
* `list_unchop()`
* `vec_c_opts()`
* `ffi_vec_as_names()`
* `df_list()`
* `df_repair_names()`

### DIFF
--- a/R/bind.R
+++ b/R/bind.R
@@ -176,6 +176,7 @@ NULL
 vec_rbind <- function(...,
                       .ptype = NULL,
                       .names_to = rlang::zap(),
+                      # TODO: confirm that this should be unchanged
                       .name_repair = c("unique", "universal", "check_unique"),
                       .name_spec = NULL,
                       .error_call = current_env()) {
@@ -194,6 +195,7 @@ vec_rbind <- fn_inline_formals(vec_rbind, ".name_repair")
 vec_cbind <- function(...,
                       .ptype = NULL,
                       .size = NULL,
+                      # TODO: confirm that this should be unchanged
                       .name_repair = c("unique", "universal", "check_unique", "minimal"),
                       .error_call = current_env()) {
   .External2(ffi_cbind, .ptype, .size, .name_repair)
--- a/R/bind.R
+++ b/R/bind.R
@@ -176,7 +176,6 @@ NULL
 vec_rbind <- function(...,
                       .ptype = NULL,
                       .names_to = rlang::zap(),
-                      # TODO: confirm that this should be unchanged
                       .name_repair = c("unique", "universal", "check_unique"),
                       .name_spec = NULL,
                       .error_call = current_env()) {
@@ -195,7 +194,6 @@ vec_rbind <- fn_inline_formals(vec_rbind, ".name_repair")
 vec_cbind <- function(...,
                       .ptype = NULL,
                       .size = NULL,
-                      # TODO: confirm that this should be unchanged
                       .name_repair = c("unique", "universal", "check_unique", "minimal"),
                       .error_call = current_env()) {
   .External2(ffi_cbind, .ptype, .size, .name_repair)
--- a/src/names.c
+++ b/src/names.c
@@ -849,7 +849,6 @@ r_obj* vec_proxy_set_names(r_obj* x, r_obj* names, const enum vctrs_owned owned)
 r_obj* vctrs_validate_name_repair_arg(r_obj* arg) {
   struct name_repair_opts opts = new_name_repair_opts(arg,
                                                       r_lazy_null,
-                                                      // think about this 'true'
                                                       true,
                                                       r_lazy_null);
   if (opts.type == NAME_REPAIR_custom) {
--- a/src/names.c
+++ b/src/names.c
@@ -17,6 +17,12 @@ r_obj* vec_as_names(r_obj* names, const struct name_repair_opts* opts) {
   case NAME_REPAIR_unique: return vec_as_unique_names(names, opts->quiet);
   case NAME_REPAIR_universal: return vec_as_universal_names(names, opts->quiet);
   case NAME_REPAIR_check_unique: return check_unique_names(names, opts);
+  // I have to handle these elements of the enum to please the compiler, but the
+  // functions that call vec_as_names() generally have `quiet = false`
+  // hard-wired and, on the R side, don't advertise support for unique_quiet and
+  // universal_quiet. What to do?
+  case NAME_REPAIR_unique_quiet: return vec_as_unique_names(names, true);
+  case NAME_REPAIR_universal_quiet: return vec_as_universal_names(names, true);
   case NAME_REPAIR_custom: return vec_as_custom_names(names, opts);
   }
   r_stop_unreachable();
@@ -843,6 +849,7 @@ r_obj* vec_proxy_set_names(r_obj* x, r_obj* names, const enum vctrs_owned owned)
 r_obj* vctrs_validate_name_repair_arg(r_obj* arg) {
   struct name_repair_opts opts = new_name_repair_opts(arg,
                                                       r_lazy_null,
+                                                      // think about this 'true'
                                                       true,
                                                       r_lazy_null);
   if (opts.type == NAME_REPAIR_custom) {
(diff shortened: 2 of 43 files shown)


R/bind.R
```diff
@@ -176,6 +176,7 @@ NULL
 vec_rbind <- function(...,
                       .ptype = NULL,
                       .names_to = rlang::zap(),
+                      # TODO: confirm that this should be unchanged
```
jennybc, line 179: I left a comment like this in all the places that _could_ gain `"unique_quiet"` and `"universal_quiet"`, but have not *yet*. These functions don't expose `quiet` currently and `quiet = false` is basically hard-wired into them in various ways.

src/names.c
```diff
@@ -17,6 +17,12 @@ r_obj* vec_as_names(r_obj* names, const struct name_repair_opts* opts) {
   case NAME_REPAIR_unique: return vec_as_unique_names(names, opts->quiet);
   case NAME_REPAIR_universal: return vec_as_universal_names(names, opts->quiet);
   case NAME_REPAIR_check_unique: return check_unique_names(names, opts);
+  // I have to handle these elements of the enum to please the compiler, but the
```
lionel-, line 20: I think it's ok to give precedence to the enum setting. And it's also ok to support options that are not advertised.

lionel- (commented):
> Should unique_quiet and universal_quiet be added to every repair/[.]?name_repair() argument?

Since we don't have `_verbose` variants that could override the hardcoded `quiet` value, I think it's better not to advertise the `_quiet` options in the function signatures. That would confusingly suggest the default is verbose when it is not.

On the other hand, if there are functions where `quiet` is set to `TRUE`, or where `quiet` is exposed to the user, then it makes sense to include the quiet options in the signature.

> When quiet = FALSE (because it was explicitly specified) and .name_repair = "*_quiet", it's less clear what to do.

I think we should give precedence to `repair`. I see `quiet` as the preference of the developer, and `repair` as the preference of the user. I think it's fine to ignore cases where `quiet` is exposed to the user and they supply contradictory preferences. We just consistently give precedence to `repair`.

src/utils.c
```diff
@@ -1693,6 +1695,12 @@ void vctrs_init_utils(SEXP ns) {
   strings_check_unique = Rf_mkChar("check_unique");
   SET_STRING_ELT(strings2, 10, strings_check_unique);
 
+  strings_unique_quiet = Rf_mkChar("unique_quiet");
+  SET_STRING_ELT(strings2, 23, strings_unique_quiet);
+
+  strings_universal_quiet = Rf_mkChar("universal_quiet");
+  SET_STRING_ELT(strings2, 24, strings_universal_quiet);
+
```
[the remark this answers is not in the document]
lionel-: Nope you did it exactly as I would have done it, just appending to avoid making mistakes.

Worth noting that in the future all of this will be moved to `globals.c`, see https://github.com/r-lib/vctrs/blob/a0ee85860d94459bf83194e036138f3160329cdc/src/globals.c#L64-L66

jennybc:
Thanks @lionel-. So I'm hearing that I should finish this off in basically its current form. Your answers to the "big questions" imply that we don't need to get into the matters that would increase the footprint of this PR.

jennybc:
@lionel- I think this is done.

jennybc:
My PR seems to include some line break changes in `.Rd` files unrelated to this PR. Let me know if you want to get those out of the PR.

lionel-:
> My PR seems to include some line break changes in .Rd files unrelated to this PR. Let me know if you want to get those out of the PR.

Yes because we've settled on @DavisVaughan's version of these files for now. But glad to see I'm not the only one with the other version! (We're trying to figure out the cause of these diffs)

jennybc:
@lionel- I believe the off-target diffs are gone now.

lionel-:
Thanks Jenny!

PR #1681, opened 2022-09-26
Use `vec_any_missing()` internally

DavisVaughan:
I had marked a few places to use `vec_any_missing()` where it made sense from a code cleanliness or performance perspective


lionel- (approved):
Nice

Issue #1496, opened 2021-11-16
Quadratic memory consumption in `vec_unchop()` for `list_of` prototype


mgirlich:
I was wondering why the new implementation of `unnest_wider()` is so slow and memory hungry for tibble columns. It turned out to be an issue with `vec_unchop()`:

``` r
library(vctrs)


make_list_of <- function(n) {
  df <- tibble::tibble(
    x = new_list_of(vec_chop(1:n), ptype = integer())
  )
  vec_chop(df)
}

df_list1 <- make_list_of(1e3)
df_list2 <- make_list_of(2e3)
df_list4 <- make_list_of(4e3)
df_list8 <- make_list_of(8e3)

ptype <- vec_ptype(df_list1[[1]])

bench::mark(
  df1 = vec_unchop(df_list1, ptype = ptype),
  df2 = vec_unchop(df_list2, ptype = ptype),
  df4 = vec_unchop(df_list4, ptype = ptype),
  df8 = vec_unchop(df_list8, ptype = ptype),
  check = FALSE
)
#> Warning: Some expressions had a GC in every iteration; so filtering is disabled.
#> # A tibble: 4 × 6
#>   expression      min   median `itr/sec` mem_alloc `gc/sec`
#>   <bch:expr> <bch:tm> <bch:tm>     <dbl> <bch:byt>    <dbl>
#> 1 df1         60.76ms  64.61ms    15.3      15.4MB     21.0
#> 2 df2        149.47ms 156.88ms     6.42     61.3MB     17.7
#> 3 df4           428ms 452.35ms     2.21    244.6MB     12.2
#> 4 df8           1.32s    1.32s     0.758   977.4MB     17.4
```

<sup>Created on 2021-11-16 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

DavisVaughan:
I think this is mainly due to the fact that df-assign has to proxy and restore the output container at every iteration. i.e. recursive proxy/restore would really help here https://github.com/r-lib/vctrs/issues/1107

Compare against just combining list-ofs, with no data frame involved:

``` r
library(vctrs)

make_list_of <- function(n) {
  new_list_of(as.list(1:n), ptype = integer())
}

df_list1 <- make_list_of(1e3)
df_list2 <- make_list_of(2e3)
df_list4 <- make_list_of(4e3)
df_list8 <- make_list_of(8e3)

ptype <- vec_ptype(df_list1[[1]])

bench::mark(
[...]

PR #1686, opened 2022-09-27
Improve performance of list-of implementation

DavisVaughan:
A few improvements to the list-of implementation including:
- Avoiding calling `!!!` on a list-of. This has to go through `rlang:::rlang_as_list()` to unstructure the list, which can be expensive if called from the cast method repeatedly
- Faster internal `new_list_of0()` that avoids a few checks that don't need when passing lists around internally
- Avoiding dispatch of `as_list_of()` in the self-self cast method

A benchmark using a simplified version of the example from #1496 (this only focuses on the list-of part, not the list-col-in-a-tibble part)

``` r
library(vctrs)

make_list_of <- function(n) {
  vec_chop(new_list_of(vec_chop(1:n), ptype = integer()))
}

list1 <- make_list_of(1e3)
list2 <- make_list_of(2e3)
list4 <- make_list_of(4e3)
list8 <- make_list_of(8e3)

ptype <- vec_ptype(list1[[1]])

bench::mark(
  list1 = list_unchop(list1, ptype = ptype),
  list2 = list_unchop(list2, ptype = ptype),
  list4 = list_unchop(list4, ptype = ptype),
  list8 = list_unchop(list8, ptype = ptype),
  min_iterations = 200,
  check = FALSE
)

# Main
#> # A tibble: 4 × 6
#> 1 list1        39.2ms   44.6ms     22.1     94.6KB     16.7
#> 2 list2        64.6ms   83.1ms     12.1     70.5KB     18.3
#> 3 list4       131.8ms    146ms      6.58   140.9KB     20.0
#> 4 list8       277.7ms  350.7ms      2.86   281.5KB     17.4

# This PR
#> # A tibble: 4 × 6
#> 1 list1        24.2ms   28.9ms     34.0     75.3KB     17.9
#> 2 list2        49.4ms   53.4ms     18.4     70.5KB     19.4
#> 3 list4       100.1ms  105.8ms      9.27   140.9KB     19.4
#> 4 list8       202.6ms    214ms      4.62   281.5KB     19.4
```

### DIFF
--- a/R/type-list-of.R
+++ b/R/type-list-of.R
@@ -65,13 +65,7 @@ new_list_of <- function(x = list(), ptype = logical(), ..., class = character())
 }
 
 new_list_of0 <- function(x, ptype, ..., class = character()) {
-  if (length(class) == 0L) {
-    class <- "vctrs_list_of"
-  } else {
-    class <- c(class, "vctrs_list_of")
-  }
-
-  new_vctr(x, ..., ptype = ptype, class = class)
+  new_vctr(x, ..., ptype = ptype, class = c(class, "vctrs_list_of"))
 }
 
 #' @export
--- a/R/type-list-of.R
+++ b/R/type-list-of.R
@@ -20,15 +20,7 @@
 #' vec_c(list_of(1, 2), list_of(FALSE, TRUE))
 list_of <- function(..., .ptype = NULL) {
   args <- list2(...)
-
-  ptype <- vec_ptype_common(!!!args, .ptype = .ptype)
-  if (is.null(ptype)) {
-    abort("Could not find common type for elements of `x`.")
-  }
-
-  args <- vec_cast_common(!!!args, .to = ptype)
-
-  new_list_of(args, ptype)
+  list_as_list_of(args, ptype = .ptype)
 }
 
 #' @export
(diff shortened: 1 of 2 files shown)


R/type-list-of.R
```diff
@@ -38,17 +30,18 @@ as_list_of <- function(x, ...) {
 }
 
 #' @export
-as_list_of.vctrs_list_of <- function(x, .ptype = NULL, ...) {
+as_list_of.vctrs_list_of <- function(x, ..., .ptype = NULL) {
```
DavisVaughan, line 33: I think this `.ptype` arg was in the wrong spot. `as_list_of.list()` has it after the `...` and I think that is right. We can change back if revdeps show that it breaks anything but this feels better to me

R/type-list-of.R
```diff
@@ -68,7 +61,17 @@ new_list_of <- function(x = list(), ptype = logical(), ..., class = character())
     abort("`ptype` must have size 0.")
   }
 
-  new_vctr(x, ..., ptype = ptype, class = c(class, "vctrs_list_of"))
+  new_list_of0(x = x, ptype = ptype, ..., class = class)
+}
+
+new_list_of0 <- function(x, ptype, ..., class = character()) {
+  if (length(class) == 0L) {
```
lionel-, line 68: This feels like a micro-optimisation. Does it make a significant difference?

lionel- (approved):

R/type-list-of.R
DavisVaughan: Doesn't seem to affect the benchmark in my opening comment so I'll remove

PR #1693, opened 2022-09-28
Don't compute the common ptype twice

DavisVaughan:
The second call wasn't actually expensive because `ptype` wasn't ever `NULL` at this point, but it still seems weird that it was being called since we computed the common `ptype` already.


src/c.c
```diff
@@ -56,10 +56,9 @@ r_obj* vec_c_opts(r_obj* xs,
   if ((is_data_frame(ptype) && fallback_opts->s3 == S3_FALLBACK_true) ||
```
DavisVaughan, line 56: Up on line 28 (out of sight here) we compute the common ptype already, and nothing changes between there and when we compute it again here, so I concluded that we were "computing" it again for no reason

lionel- (approved):

PR #1694, opened 2022-09-28
Use `vec_cast_common()` in `vec_c()` to get positional errors

### LINKED ISSUE #1690: `vec_c()` should use `vec_cast_common()`
Rather than individual calls to `vec_cast()`
https://github.com/r-lib/vctrs/blob/c0bf5ab90f5a5ca45d73ca1a9727e9b4d0b25768/src/c.c#L142

This will allow `vec_c()` to give positional error messages on cast failures, as requested in https://github.com/r-lib/vctrs/issues/1641#issuecomment-1247239942

But we need to be careful about retaining names that can currently be lost during the cast https://github.com/r-lib/vctrs/issues/1689

DavisVaughan:
Closes #1690

### DIFF
--- a/src/c.c
+++ b/src/c.c
@@ -24,7 +24,9 @@ r_obj* vec_c_opts(r_obj* xs,
     .fallback = *fallback_opts
   };
 
+  r_obj* orig_xs = xs;
   r_obj* orig_ptype = ptype;
+
   ptype = KEEP(vec_ptype_common_opts(xs, orig_ptype, &ptype_opts));
 
   if (ptype == r_null) {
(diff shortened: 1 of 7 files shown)


src/c.c
```diff
@@ -111,8 +119,12 @@ r_obj* vec_c_opts(r_obj* xs,
     init_compact_seq(p_loc, counter, size, true);
 
     if (assign_names) {
+      // FIXME: Common cast may have lost names, so must pull names from
+      // original `xs` https://github.com/r-lib/vctrs/issues/623
+      r_obj* orig_x = r_list_get(orig_xs, i);
+
       r_obj* outer = xs_is_named ? r_chr_get(xs_names, i) : r_null;
-      r_obj* inner = KEEP(vec_names(x));
+      r_obj* inner = KEEP(vec_names(orig_x));
```
DavisVaughan, line 127: Seems like this is really the only tricky part

src/c.c
```diff
@@ -61,6 +63,12 @@ r_obj* vec_c_opts(r_obj* xs,
     ptype = KEEP(vec_ptype_common_opts(xs, ptype, &ptype_opts));
   }
 
+  struct cast_common_opts cast_opts = {
+    .call = error_call,
+    .fallback = *fallback_opts
+  };
+  xs = KEEP(vec_cast_common_opts(xs, ptype, &cast_opts));
```
DavisVaughan, line 70: This does mean that we allocate another list the same size as `xs` which we didn't do before, but that seems worth it to let the `vec_cast_common_opts()` loop handle all the details of getting the positional errors right

PR #1699, opened 2022-09-29
Retain names in `list_unchop()`

### LINKED ISSUE #1689: `list_unchop(indices =)` can drop names of `x`
``` r
list_unchop(list(c(x=1L)), indices = list(1), ptype = double())
#> [1] 1
```

i.e. when we go through the actual `list_unchop()` path (not the `vec_c()` path) AND `vec_cast_common()` has to do some actual work by casting - we can lose names because `vec_cast()` can drop names.

I think we should continue to use `vec_cast_common()` because it gives positional error messages (rather than calling `vec_cast()` on each element), but we need to be a little more careful about retaining the original input names until https://github.com/r-lib/vctrs/issues/623 can be fixed.

Noticed while looking at https://github.com/r-lib/vctrs/issues/1641#issuecomment-1247251797

### LINKED ISSUE #1690: `vec_c()` should use `vec_cast_common()`

DavisVaughan:
Closes #1689

I mentioned that I was planning on switching both `list_unchop()` and `vec_c()` to using `vec_cast_common()` rather than casting in the loop, but I no longer like this plan.

- I didn't realize how easy it was to get positional errors with `new_subscript_arg()`. It isn't really that bad!
- This also allows us to get positional errors for recycling issues
- This sets us up nicely for adding `error_arg`
- This avoids the extra allocation caused by `vec_cast_common()` duplicating `xs`

I'm going to apply this to fix #1690 too in place of #1694


tests/testthat/test-c.R
```diff
@@ -439,11 +439,7 @@ test_that("named empty vectors force named output (#1263)", {
   expect_named(list_unchop(list(x, x), list(int(), int())), chr())
   expect_named(list_unchop(list(x, 1L), list(int(), 1)), "")
 
-  # FIXME: `vec_cast_common()` dropped names
-  # https://github.com/r-lib/vctrs/issues/623
-  expect_failure(
-    expect_named(list_unchop(list(x, 1), list(int(), 1)), "")
-  )
+  expect_named(list_unchop(list(x, 1), list(int(), 1)), "")
```
DavisVaughan, line 442: Nice side effect

tests/testthat/test-slice-chop.R
```diff
@@ -307,6 +307,22 @@ test_that("NULLs are ignored when unchopped with other vectors", {
   expect_identical(list_unchop(list("a", NULL, "b"), list(2, integer(), 1)), c("b", "a"))
 })
 
+test_that("can use a `NULL` element with a corresponding index", {
+  # FIXME: Probably not quite right, but not entirely clear what it should be:
+  # - Maybe `unspecified(2)`?
+  # - Or should `NULL`s even be allowed in `list_unchop()`?
+  expect_null(list_unchop(list(NULL), indices = list(1:2)))
```
DavisVaughan, line 314: Not entirely sure about this result, but it seems rare enough to not care too much right now

lionel-: hmm I tend to think we should respect the invariant that the output size is equal to the indices size.

You mean an unclassed unspecified vector right? That seems reasonable.

tests/testthat/test-slice-chop.R
```diff
@@ -452,6 +468,21 @@ test_that("outer names can be merged with inner names", {
...
+
+  # With name spec
+  name_spec <- "{outer}_{inner}"
+  expect_named(list_unchop(list(foo = c(a = 1)), ptype = integer(), name_spec = name_spec), "foo_a")
+  expect_named(list_unchop(list(foo = c(a = 1)), ptype = integer(), name_spec = name_spec, indices = list(1)), "foo_a")
+
+  # When `x` elements are recycled, names are also recycled
+  x <- list(c(a = 1), c(b = 2))
+  indices <- list(1:2, 3:4)
+  expect_named(list_unchop(x, indices = indices, ptype = integer()), c("a", "a", "b", "b"))
```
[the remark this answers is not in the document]
lionel-: I'll be happy when this is fixed...

Maybe we should take a look for 0.5.0?

lionel- (approved):

tests/testthat/test-slice-chop.R
DavisVaughan: I opened https://github.com/r-lib/vctrs/issues/1702 because I think this is also tied to what we should return in the empty case, like `vec_c()` with no inputs and `list_unchop(list(), indices = list())`

tests/testthat/test-slice-chop.R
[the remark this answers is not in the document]
DavisVaughan: It would be nice but it might be a sizable project

PR #1701, opened 2022-09-30
Make `list_of()` lax

### LINKED ISSUE #1161: Should be able to combine list and list_of<>

lionel-:
Closes #1161
Closes #1231

This no longer fails:

- list-of-chr + list = list

Furthermore, to ensure consistent behaviour when lists are thrown into the mix:

- list-of-int + list-of-chr = list

Otherwise incompatible list-ofs would cause an error except when a list is added to the set of elements to combine.

Given these semantics, it seems that `[[<-` should become lenient as well and upcoerce to a list in case of incompatible type. Does that make sense?


hadley (approved):
I think it's fine for `[[` to remain strict. That's consistent with factors:

``` r
x <- factor(c("a", "b"))
vctrs::vec_slice(x, 1) <- "c"
#> Error in `vec_slice<-`:
#> ! Can't convert from `"c"` <character> to <factor<38051>> due to loss of generality.
#> • Locations: 1
```

<sup>Created on 2022-09-30 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

R/type-list-of.R
```diff
@@ -212,6 +227,19 @@ vec_cast.vctrs_list_of.vctrs_list_of <- function(x, to, ...) {
   list_as_list_of(x, ptype = ptype)
 }
 
+#' @export
+vec_cast.list.vctrs_list_of <-function(x, to, ...) {
+  vec_data(x)
```
DavisVaughan, line 232: I think you should use the same thing in the `as.list.vctrs_list_of()` method, for consistency.

---

Also, maybe pull in `rlang:::vec_unstructure()` because we really don't need to pay for the `vec_assert()` and `vec_proxy()` calls from `vec_data()` - we just want to strip all attributes except names.

R/type-list-of.R
```diff
@@ -212,6 +227,19 @@ vec_cast.vctrs_list_of.vctrs_list_of <- function(x, to, ...) {
...
+}
+#' @export
+vec_cast.vctrs_list_of.list <-function(x, to, ...) {
+  list_as_list_of(
+    x,
+    attr(to, "ptype"),
+    error_call = caller_env()
```
DavisVaughan, line 239: I see how if you remove this you get `vec_cast.vctrs_list_of.list()` in the error:

```r
> vec_cast(list(1), list_of("x"))
Error in `vec_cast.vctrs_list_of.list()`:
! Can't convert `..1` <double> to <character>.
Run `rlang::last_trace()` to see where the error occurred.
```

But I'd really rather not pay for calling `caller_env()` every time, especially because it doesn't give you the right call

``` r
library(vctrs)

# Don't see `fn()`
fn <- function() vec_cast(list(1), list_of("x"))
fn()
#> Error:
#> ! Can't convert `..1` <double> to <character>.

fn <- function() vec_cast(1, "x")
fn()
#> Error in `fn()`:
#> ! Can't convert `1` <double> to <character>.
```

Could we do this since `call` is in the generic? It seems to give the right call and shouldn't be evaluated unless it is required.

```r
vec_cast.vctrs_list_of.list <-function(x, to, ..., call = caller_env()) {
  list_as_list_of(
    x,
    attr(to, "ptype"),
    error_call = call
  )
}
```

R/type-list-of.R
```diff
@@ -188,11 +188,26 @@ vec_ptype2.vctrs_list_of <- function(x, y, ..., x_arg = "", y_arg = "") {
 }
```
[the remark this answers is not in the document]
DavisVaughan: If you do this and close that issue, it would be great to add a test like this to ensure `[<-` and `vec_assign()` are more in sync

``` r
library(vctrs)

x <- list_of(1L, .ptype = integer())

`[<-`(x, 1, list(2L))
#> <list_of<integer>[1]>
#> [[1]]
#> [1] 2
vec_assign(x, 1, list(2L))
#> <list_of<integer>[1]>
#> [[1]]
#> [1] 2
```

R/type-list-of.R
```diff
@@ -188,11 +188,26 @@ vec_ptype2.vctrs_list_of <- function(x, y, ..., x_arg = "", y_arg = "") {
...
 #' @export
-vec_ptype2.vctrs_list_of.vctrs_list_of <- function(x, y, ...) {
-  ptype <- vec_ptype2(attr(x, "ptype"), attr(y, "ptype"))
+vec_ptype2.vctrs_list_of.vctrs_list_of <- function(x, y, ..., x_arg = "", y_arg = "") {
+  frame <- current_env()
+  ptype <- tryCatch(
+    vec_ptype2(attr(x, "ptype"), attr(y, "ptype"), x_arg = x_arg, y_arg = y_arg),
+    vctrs_error_incompatible_type = function(cnd) {
+      return_from(frame, list())
+    }
```
DavisVaughan, line 197: I'm a little worried about the performance of all this. Can we apply the same `identical()` check that we did in the cast method (see suggestion)? It seems to shave off about 20% or so. I can get the example down to 5 seconds by also providing a `vec_ptype.vctrs_list_of()` method (because right now it is doing `vec_slice(x, NULL)` on each input every time) but we can do that in another PR


```r
x <- list_of(.ptype = double())
xs <- rep(list(x), 1e5)

bench::mark(vec_ptype_common(!!!xs))

# Current PR
# A tibble: 1 × 13
  expression                   min median `itr/sec` mem_alloc `gc/sec` n_itr
  <bch:expr>              <bch:tm> <bch:>     <dbl> <bch:byt>    <dbl> <int>
1 vec_ptype_common(!!!xs)    10.3s  10.3s    0.0974     782KB     12.6     1
# … with 6 more variables: n_gc <dbl>, total_time <bch:tm>, result <list>,
#   memory <list>, time <list>, gc <list>

# With identical() check
# A tibble: 1 × 13
  expression                   min median `itr/sec` mem_alloc `gc/sec` n_itr
  <bch:expr>              <bch:tm> <bch:>     <dbl> <bch:byt>    <dbl> <int>
1 vec_ptype_common(!!!xs)     8.5s   8.5s     0.118     782KB     12.8     1
# … with 6 more variables: n_gc <dbl>, total_time <bch:tm>, result <list>,
#   memory <list>, time <list>, gc <list>
```

```suggestion
  x_ptype <- attr(x, "ptype", exact = TRUE)
  y_ptype <- attr(y, "ptype", exact = TRUE)

  if (identical(x_ptype, y_ptype)) {
    x
  } else {
    tryCatch(
      expr = {
        ptype <- vec_ptype2(x_ptype, y_ptype, x_arg = x_arg, y_arg = y_arg)
        new_list_of0(x = list(), ptype = ptype)
      },
      vctrs_error_incompatible_type = function(cnd) {
        list()
      }
    )
  }
```

tests/testthat/test-type-list-of.R
```diff
@@ -189,3 +196,31 @@ test_that("vec_ptype2(<list_of<>>, NA) is symmetric (#687)", {
   expect_identical(vec_ptype2(lof, NA), vec_ptype(lof))
   expect_identical(vec_ptype2(NA, lof), vec_ptype(lof))
 })
+
+test_that("list_of() coerces to list() and list_of()", {
```
DavisVaughan, line 200:
```suggestion
test_that("list_of() coerces to list() and list_of() (#1701)", {
```

DavisVaughan (approved):

DavisVaughan:
I am somewhat strongly against making `[[<-` lax. It is like `vec_assign()` and `vec_assign2()`, which should always retain the type of the input.

lionel-:
> I am somewhat strongly against making [[<- lax. It is like vec_assign() and vec_assign2(), which should always retain the type of the input.

Regarding `[[<-`, I think it's natural for integer vectors to be upcoerced to double vectors when assigning a decimal number. There is also the issue of coercing unspecified vectors.

R/type-list-of.R
lionel-: We are not paying for it every time, it's evaluated lazily since it's passed as argument.

Good point regarding the generic argument. I'll take a look and make sure we pass `call` through `...` down to methods.

R/type-list-of.R
lionel-: `vec_unstructure()` dispatches 4 times. If we're looking for efficiency, I'll just add `list_of_unstructure()`.

R/type-list-of.R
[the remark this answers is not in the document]
lionel-: Let's do this in another PR.

PR #1703, opened 2022-09-30
No need for `_quiet` enum options

DavisVaughan:
Follow up to #1677

At the C level, we create the `name_repair_opts` struct through either:
- `new_name_repair_opts()`, which handles translating the `"unique_quiet"` and `"universal_quiet"` strings into existing enum options
- Manual creation of the struct, with a hardcoded enum value and control of `quiet`

This means the enum options themselves don't need to know about `_quiet`

### DIFF
--- a/src/names.c
+++ b/src/names.c
@@ -17,12 +17,6 @@ r_obj* vec_as_names(r_obj* names, const struct name_repair_opts* opts) {
   case NAME_REPAIR_unique: return vec_as_unique_names(names, opts->quiet);
   case NAME_REPAIR_universal: return vec_as_universal_names(names, opts->quiet);
   case NAME_REPAIR_check_unique: return check_unique_names(names, opts);
-  // At the time when unique_quiet and universal_quiet were added, no function
-  // that calls the C function vec_as_names() actually accepts these strings at
-  // the R level, because these functions enforce `quiet = false`.
-  // But we still have to handle every case for the enum.
-  case NAME_REPAIR_unique_quiet: return vec_as_unique_names(names, true);
-  case NAME_REPAIR_universal_quiet: return vec_as_universal_names(names, true);
   case NAME_REPAIR_custom: return vec_as_custom_names(names, opts);
   }
   r_stop_unreachable();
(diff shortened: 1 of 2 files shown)


jennybc (approved):
LGTM

However, while I was poking around some more, I noticed that the `*_quiet` options have become (silently) available even in functions where they are not advertised as supported in the function signature. This has nothing to do with PR, i.e. it's true in `main` (and in this PR). Is that OK? If that's not OK, it feels like there's a missing layer of `arg_match()`-type logic in many places. Again, not related to this PR, but something I discovered while looking at all this again.

``` r
devtools::load_all("~/rrr/vctrs")
#> ℹ Loading vctrs

df_list(x = 1, x = 2, .name_repair = "universal")
#> New names:
#> • `x` -> `x...1`
#> • `x` -> `x...2`
#> $x...1
#> [1] 1
#> 
#> $x...2
#> [1] 2

df_list(x = 1, x = 2, .name_repair = "universal_quiet")
#> $x...1
#> [1] 1
#> 
#> $x...2
#> [1] 2

data_frame(x = 1, x = 2, .name_repair = "unique")
#> New names:
#> • `x` -> `x...1`
#> • `x` -> `x...2`
#>   x...1 x...2
#> 1     1     2

data_frame(x = 1, x = 2, .name_repair = "unique_quiet")
#>   x...1 x...2
#> 1     1     2

data_frame(x = 1, x = 2, .name_repair = "wut")
#> Error in `data_frame()`:
#> ! `.name_repair` can't be "wut". See `?vctrs::vec_as_names`.

#> Backtrace:
#>     ▆
#>  1. ├─vctrs::data_frame(x = 1, x = 2, .name_repair = "wut")
#>  2. └─rlang::abort(message = message, call = call)
```

<sup>Created on 2022-09-30 with [reprex v2.0.2.9000](https://reprex.tidyverse.org)</sup>

lionel-:
@jennybc yup that's the behaviour I had expected

lionel-:
Though if these functions do not expose `quiet =` then the new quiet options should be exposed in `repair` I think.

jennybc:
> Though if these functions do not expose `quiet =` then the new quiet options should be exposed in repair I think.

Do you want a new PR for this? I understand this to mean the signatures of some R functions should change (reveal the new possibilities) and also docs, as appropriate.

lionel-:
That'd be great thanks.

PR #1705, opened 2022-09-30
Prevent `$<-` assignment with `NULL` from shortening list-ofs

### LINKED ISSUE #1704: For `<list_of>`, `$<-` needs to handle `NULL` like `[[<-`
i.e., don't drop the element

``` r
library(vctrs)

x <- list_of(x = 1, y = 2, z = 3)

x$y <- NULL
x
#> <list_of<double>[2]>
#> $x
#> [1] 1
#> 
#> $z
#> [1] 3

x[["z"]] <- NULL
x
#> <list_of<double>[2]>
#> $x
#> [1] 1
#> 
#> $z
#> NULL
```

<sup>Created on 2022-09-30 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

DavisVaughan:
Closes #1704

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,8 @@
 # vctrs (development version)
 
+* The list-of method for `$<-` no longer shortens lists when assigning `NULL`
+  (#1704).
+
 * `validate_list_of()` has been removed. It hasn't proven to be practically
   useful, and isn't used by any packages on CRAN (#1697).
 
--- a/R/type-list-of.R
+++ b/R/type-list-of.R
@@ -155,6 +155,13 @@ as.character.vctrs_list_of <- function(x, ...) {
 
 #' @export
 `$<-.vctrs_list_of` <- function(x, i, value) {
+  if (is.null(value)) {
+    # Setting to NULL via `$<-` shortens the list! Example:
+    # `$<-`(list(a = 1), "a", NULL)
+    x[i] <- list(value)
+    return(x)
+  }
+
   value <- vec_cast(value, attr(x, "ptype"))
   NextMethod()
 }
--- a/tests/testthat/test-type-list-of.R
+++ b/tests/testthat/test-type-list-of.R
@@ -97,13 +97,19 @@ test_that("[<-, [[<- and $<- coerce their input", {
   x[[2]] <- NULL
   expect_equal(x, list_of(x = 0, y = NULL, z = 1, w = 0))
 
+  # #1704
+  x$z <- NULL
+  expect_equal(x, list_of(x = 0, y = NULL, z = NULL, w = 0))
+
   expect_error(x[[2]] <- list(20), class = "vctrs_error_incompatible_type")
   expect_error(x$y <- list(20), class = "vctrs_error_incompatible_type")
 
-  x[3:4] <- list(NULL)
-  expect_equal(x, list_of(x = 0, y = NULL, z = NULL, w = NULL))
+  x[["a"]] <- 1
 
-  expect_equal(is.na(x), c(FALSE, TRUE, TRUE, TRUE))
+  x[4:5] <- list(NULL)
+  expect_equal(x, list_of(x = 0, y = NULL, z = NULL, w = NULL, a = NULL))
+
+  expect_equal(is.na(x), c(FALSE, TRUE, TRUE, TRUE, TRUE))
 })
 
 test_that("assingment can increase size of vector", {


lionel- (commented):
Do we need support for `zap()`?

Should this be implemented at the `vctrs_vctr` level?

DavisVaughan:
> Do we need support for zap()?

Do we ever want `$<-` or `[[<-` to be able to shorten lists? Being size stable feels like nice properties of these functions.

I'd prefer implementing something like `loc <- subscript_invert("a", i = 5, names = c("a", "b"))`, `vec_slice(x, loc)` to remove elements.

Or maybe:

```r
chr_as_location(i, names, ..., invert = FALSE/TRUE, oob = c("error", "remove", "extend"))

loc <- chr_as_location("a", names = c("a", "b"), invert = TRUE)
vec_slice(x, loc)
```

DavisVaughan:
> Should this be implemented at the vctrs_vctr level?

I guess that depends on whether or not we'd want to enforce this "avoid shortening lists" behavior on all vctrs_vctr list subclasses, but it does feel kind of reasonable to do that? We'd do it for both `$<-` and `[[<-` I think.

lionel-:
In R, lists (and to a lesser extent vectors in general) are also the type for dictionary-like objects. That's why `$<-` (and name-based `[[<-` for consistency) should be able to remove objects from lists.

Then the question becomes what happens with location-based `[[<-`. Do we allow removing elements for consistency or do we treat it differently? For tibbles it behaves the same way.

I think it's fine to keep preventing atomic vectors from being shortened though.

DavisVaughan:
Do you mind if we merge this as-is and continue conversation in an issue for vctrs 1.0.0? I feel like at the very least we won't want `NULL` to be the way we zap elements, so this should be ok for now to at least align `$<-` and `[[<-`. I can open it and move all relevant conversation there.

lionel-:
I'm not sure about all this. I like the idea of deleting elements with `zap()`, but I worry that it's going to be confusing to have list and list-of behave differently. I think I'd rather follow base R behaviour here, with `NULL` for deletion.

lionel-:
By the way, if we remove this sort of syntax for name deletion, we'll need something easier than `x[-match(name, names(x))]`. The problem is that while we can do `x[-10]`, we can't do `x[-"name"]`. So removing `$<-` deletion entirely without an alternative doesn't appeal to me.

DavisVaughan:
@lionel- that's why I mentioned `chr_as_location("a", names = c("a", "b"), invert = TRUE)` as a way to generate a location vector that would delete `"a"`.

I would never do `x[-10]` programmatically either, because of the issue with `x[-integer()]` (I kind of which `num_as_location()` had a way to force negative inversion for that case)

DavisVaughan:
I don't think it would be unreasonable to provide `vec_remove()` as an easy way to remove elements (named or positional)

lionel-:
> that's why I mentioned chr_as_location("a", names = c("a", "b"), invert = TRUE) as a way to generate a location vector that would delete "a".

This is not a practical alternative to `x$a <- NULL` or `zap()` though. We need something syntactic.

> I would never do x[-10] programmatically either,

My comments are mostly about analytic usages. By the way, since we return list-ofs in tidyr, I think we should keep list and list-of behaviour in line. To me that suggests we shouldn't stray too far away from base R in our `$` and `[[` methods. Also, since we make it very easy to lose the class, I think it'd be a bit unsafe to programmatically rely on special behaviour such as `NULL` preservation in `$<-`.

DavisVaughan:
I guess I can live with `[[<-` and `$<-` inheriting base R rules (including `NULL` deletion and expansion like `x <- list(); x$y <- 1`)

And then `vec_assign2()` would be stricter, i.e. it wouldn't do `NULL` deletion with lists, and it would use `vec_as_location2()` so you can't expand the vector either. That would make `vec_assign2()` completely size stable.

So maybe I'll close this issue an open another one that actually remove the special `NULL` behavior from our `[[<-` method?

lionel-:
That sounds good. I guess I'm slowly coming to terms with the idea that vec-assign/assign2 behaves differently than user-facing methods. Maybe at some point it'd make sense to export helpers to take care of these things and of subset-assign coercions.

DavisVaughan:
> I guess I'm slowly coming to terms with the idea that vec-assign/assign2 behaves differently than user-facing methods.

Same, even though I didn't want that to be the case 😢

DavisVaughan:
See https://github.com/r-lib/vctrs/issues/1704#issuecomment-1295304348

PR #1707, opened 2022-09-30
Clean up slice fallback and remove bit64 slice hacks

DavisVaughan:
I originally intended to just remove the old bit64 hacks around slicing with `NA_integer_`, but trying to do so exposed some strange bugs with the way our fallback slicing was working. I'll add some inline comments below.

The first commit cleans up the fallback slicing approach. The second commit removes the bit64 hacks.

### DIFF
--- a/src/slice-chop.c
+++ b/src/slice-chop.c
@@ -296,6 +296,10 @@ static SEXP chop_shaped(SEXP x, SEXP indices, struct vctrs_chop_info info) {
 }
 
 static SEXP chop_fallback(SEXP x, SEXP indices, struct vctrs_chop_info info) {
+  // TODO: Do we really care about micro performance that much here? Can we just
+  // merge with `chop_shaped_fallback()` now that `vec_slice_fallback()`
+  // handles shaped an unshaped vectors?
+
   // Evaluate in a child of the global environment to allow dispatch
   // to custom functions. We define `[` to point to its base
   // definition to ensure consistent look-up. This is the same logic
@@ -329,6 +333,7 @@ static SEXP chop_fallback(SEXP x, SEXP indices, struct vctrs_chop_info info) {
 
     SEXP elt = PROTECT(Rf_eval(call, env));
 
+    // Same logic as `vec_slice_fallback()`
     if (!vec_is_restored(elt, x)) {
       elt = vec_restore(elt, x, vec_owned(elt));
     }
@@ -349,7 +354,6 @@ static r_obj* chop_fallback_shaped(r_obj* x, r_obj* indices, struct vctrs_chop_i
       ++(*info.p_index);
     }
 
-    // `vec_slice_fallback()` will also `vec_restore()` for us
     r_obj* elt = PROTECT(vec_slice_fallback(x, info.index));
 
     SET_VECTOR_ELT(info.out, i, elt);
--- a/src/slice-chop.c
+++ b/src/slice-chop.c
@@ -310,16 +310,8 @@ static SEXP chop_fallback(SEXP x, SEXP indices, struct vctrs_chop_info info) {
   Rf_defineVar(syms_i, info.index, env);
 
   // Construct call with symbols, not values, for performance.
-  // TODO - Remove once bit64 is updated on CRAN. Special casing integer64
-  // objects to ensure correct slicing with `NA_integer_`.
-  SEXP call;
-  if (is_integer64(x)) {
-    call = PROTECT(Rf_lang3(syms.vec_slice_dispatch_integer64, syms_x, syms_i));
-    Rf_defineVar(syms.vec_slice_dispatch_integer64, fns.vec_slice_dispatch_integer64, env);
-  } else {
-    call = PROTECT(Rf_lang3(syms_bracket, syms_x, syms_i));
-    Rf_defineVar(syms_bracket, fns_bracket, env);
-  }
+  SEXP call = PROTECT(Rf_lang3(syms_bracket, syms_x, syms_i));
+  Rf_defineVar(syms_bracket, fns_bracket, env);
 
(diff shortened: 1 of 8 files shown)


src/slice-chop.c
```diff
@@ -296,6 +296,10 @@ static SEXP chop_shaped(SEXP x, SEXP indices, struct vctrs_chop_info info) {
 }
 
 static SEXP chop_fallback(SEXP x, SEXP indices, struct vctrs_chop_info info) {
+  // TODO: Do we really care about micro performance that much here? Can we just
+  // merge with `chop_shaped_fallback()` now that `vec_slice_fallback()`
+  // handles shaped an unshaped vectors?
```
DavisVaughan, line 301: This `chop_fallback()` only runs on S3/S4 objects that don't have a `vec_proxy()` method defined for them (i.e. `vec_requires_fallback()` is `true`).

I can't imagine we care about the performance of those cases _that_ much. I think I'd rather simplify this code by a lot by merging it with `chop_fallback_shaped()`, which just allows `vec_slice_fallback()` to handle everything related to setting up the environment and doing `vec_is_restored()`. The less places we have that `vec_is_restored()` check lying around, the better IMO.

src/slice.c
```diff
@@ -207,34 +207,39 @@ r_obj* df_slice(r_obj* x, r_obj* subscript) {
...
+  );
+}
+static
+r_obj* bracket_shaped_dispatch(r_obj* x, r_obj* subscript) {
+  return vctrs_dispatch2(
+    syms.bracket_shaped_dispatch, fns.bracket_shaped_dispatch,
+    syms_x, x,
+    syms_i, subscript
+  );
+}
```
DavisVaughan, line 225: We now have these nice two parallel functions. `bracket_dispatch()` calls `x[i]` and `bracket_shaped_dispatch()` just calls `x[i, , , drop = FALSE]` with as many `,` as needed.

src/slice.c
```diff
@@ -207,34 +207,39 @@ r_obj* df_slice(r_obj* x, r_obj* subscript) {
...
 
 r_obj* vec_slice_fallback(r_obj* x, r_obj* subscript) {
```
DavisVaughan, line 227: The point of `vec_slice_fallback()` is now to completely handle the fallback slice case - which is basically how we have been trying to use it this whole time.

It first hands off to `[` and then optionally calls `vec_restore()` if it looks like we needed to. This allows us to localize the "maybe restore" logic into one place, so we don't forget to do it whenever we use `vec_slice_fallback()` elsewhere.

src/slice.c
```diff
@@ -381,11 +375,15 @@ bool vec_is_restored(r_obj* x, r_obj* to) {
...
+  // Class is restored if it contains any other attributes than names, dim, or
+  // dimnames. We might want to add support for data frames later on.
   r_obj* node = attrib;
   while (node != r_null) {
-    if (r_node_tag(node) == r_syms.names) {
+    r_obj* tag = r_node_tag(node);
+
+    if (tag == r_syms.names ||
+        tag == r_syms.dim ||
+        tag == r_syms.dim_names) {
```
DavisVaughan, line 386: `vec_is_restored()` had this bug where it said "you don't need to restore" if the object had `dim` or `dimname` attributes, which definitely isn't true. Those are two of the attributes that base R `[` methods will propagate, like names (but unlike the class).

tests/testthat/test-slice.R
```diff
@@ -272,6 +272,18 @@ test_that("vec_slice() unclasses input before calling `vec_restore()`", {
...
+  # x <- structure(1:4, class = "vctrs_foobar")
+  # x[1]
+  # So we shouldn't be special casing shaped objects.
+  # I believe that if your `[` method is going to call `vec_slice()` (i.e. you
+  # are requesting native vctrs slicing), then you need to declare a
+  # `vec_proxy()` method as well to tell vctrs what it needs to be natively
+  # slicing.
+
   local_methods(
     `[.vctrs_foobar` = function(x, i, ...) vec_slice(x, i)
```
DavisVaughan, line 288: Ok, so lets talk about this. This infloop check is for:
- A class that has clearly opted into using vctrs, because they call `vec_slice()`
- But isn't using `new_vctr()`, which has a `vec_proxy()` method defined
- And isn't defining their own `vec_proxy()` method

In that case, `vec_slice()` tries to fall back to the `[` method, which causes the infloop.

I don't think that avoiding this infloop is our responsibility. The way we tried to do this was by calling `unclass()` in that original `vec_slice_fallback()` function, but as I showed earlier that defeats the whole purpose of the function.

It it also 100% possible to do this already with non-shaped vectors. i.e. this is with CRAN vctrs:

```r
library(vctrs)
`[.vctrs_foobar` <- function(x, i, ...) vec_slice(x, i)
x <- structure(1:4, class = "vctrs_foobar")
# infloop
x[1]
```

I think that if you are telling vctrs that you want `[` to forward to `vec_slice()`, then you _have_ to tell vctrs what it should be natively slicing by providing a `vec_proxy()` method (if you aren't extending vctrs_vctr).

tests/testthat/test-slice.R
```diff
@@ -351,6 +363,7 @@ test_that("vec_restore() is called after slicing data frames", {
 
 test_that("additional subscripts are forwarded to `[`", {
   local_methods(
+    vec_proxy.vctrs_foobar = function(x, ...) x,
     `[.vctrs_foobar` = function(x, i, ...) vec_index(x, i, ...)
```
DavisVaughan, line 367: Otherwise produces the infloop from before, but we are just trying to test `vec_index()` properties here (used by vctrs_vctr) so providing a proxy method like vctrs_vctr to avoid the infloop seems reasonable.

Issue #1673, opened 2022-09-21
`vec_rep_each()` early exit issue


DavisVaughan:
``` r
vec_rep_each(NA, 0)
#> <unspecified> [0]
```

From calling `vec_ptype()` rather than `vec_slice(x, NULL)` or following up with `vec_ptype_finalize()`.

https://github.com/r-lib/vctrs/blob/0a219ba242ee368df63bc3876f06f9a6220e85c0/src/rep.c#L99

Update: should probably be `vec_slice()` because we'd want named vectors to keep their names, and `vec_ptype()` should theoretically drop names

PR #1708, opened 2022-09-30
Ensure `vec_rep_each(times = 0)` early exit works properly

DavisVaughan:
Closes https://github.com/r-lib/vctrs/issues/1673

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,8 @@
 # vctrs (development version)
 
+* `vec_rep_each(times = 0)` now works correctly with logical vectors that are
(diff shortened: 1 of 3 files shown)


lionel- (approved):
Is this a case where we could have used `vec_ptype_final()`? Linked to the issue of whether `vec_ptype()` returns namedness.

DavisVaughan:
I think `vec_slice(x, NULL)` is right because:
- The normal code path would have used `vec_slice()`
- I'm fairly confident names aren't part of the ptype

Issue #1692, opened 2022-09-28
`vec_c()` and `list_unchop()` should have `error_arg` arguments


DavisVaughan:
Requested in #1641 

Defaults of `vec_c(.error_arg = "")` and `list_unchop(error_arg = caller_arg(x))`

I think we should fix #1689 and #1690 first to maximize the benefit of this

PR #1709, opened 2022-10-01
Add `.error_arg` to `vec_c()`

DavisVaughan:
Part of #1692

### DIFF
--- a/src/bind.c
+++ b/src/bind.c
@@ -285,7 +285,15 @@ void df_c_fallback(r_obj* out,
       FREE(1);
     } else if (vec_is_common_class_fallback(ptype_col)) {
       r_obj* xs_col = KEEP(list_pluck(xs, i));
-      r_obj* out_col = vec_c_fallback(ptype_col, xs_col, name_spec, name_repair, error_call);
+
+      r_obj* out_col = vec_c_fallback(
+        ptype_col,
+        xs_col,
+        name_spec,
+        name_repair,
+        vec_args.empty,
+        error_call
+      );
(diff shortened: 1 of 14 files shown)


src/bind.c
```diff
@@ -285,7 +285,15 @@ void df_c_fallback(r_obj* out,
       FREE(1);
     } else if (vec_is_common_class_fallback(ptype_col)) {
       r_obj* xs_col = KEEP(list_pluck(xs, i));
-      r_obj* out_col = vec_c_fallback(ptype_col, xs_col, name_spec, name_repair, error_call);
+
+      r_obj* out_col = vec_c_fallback(
```
lionel-, line 289: By the way AFAIK Emacs can't indent that and will fight against this type of indentation. I'm fine with it but we used indent-at-paren until recently. If we generally switch to that style though, I'm going to have a hard time editing vctrs sources :-(

Maybe that has changed since I looked a few years ago. Do you know if this style of indentation has a known name?

lionel- (approved):

src/bind.c
DavisVaughan: It matches the R level tidyverse style guide https://style.tidyverse.org/syntax.html#long-lines and is what I find most readable 😢 , I like that it moves everything closer to the left hand side so I can read top to bottom rather than having to move my eyes left->right and back so much.

If it is not possible in Emacs we can use indent-at-paren (I actually don't think I ever used that 😬 ) but it would be great to try to see if Emacs can do this!

PR #1711, opened 2022-10-03
Support non-string args in `vec_cast()`

### LINKED ISSUE #1581: Can we improve on the hardcoded `arg` name in this error?
```r
vctrs::vec_cast(1, "x", x_arg = 1)
#> Error in `vctrs::vec_cast()`:
#> ! `arg` must be a string.
```

Ideally it would reference `x_arg` here.

This happens because `arg` is hardcoded here:
https://github.com/r-lib/vctrs/blob/79f5318b886a8316b594477a84d1dd8ab3373ece/src/arg.c#L136

But a lot of times `(*data).x` is a symbol with the right arg name, so maybe we can check to see if that is a symbol, and if so use `r_sym_c_string()` to turn it into a name we can use in the error?

lionel-:
Closes #1581


DavisVaughan:
Slightly related to https://github.com/r-lib/vctrs/issues/575

lionel-:
I think #575 is about locations whereas this issue is about error args.

PR #1712, opened 2022-10-03
Add `error_arg` to `list_unchop()`

### LINKED ISSUE #1692: `vec_c()` and `list_unchop()` should have `error_arg` arguments

DavisVaughan:
Closes #1692

### DIFF
--- a/R/slice-chop.R
+++ b/R/slice-chop.R
@@ -98,7 +98,7 @@ list_unchop <- function(x,
                         ptype = NULL,
                         name_spec = NULL,
                         name_repair = c("minimal", "unique", "check_unique", "universal"),
-                        error_arg = caller_arg(x),
+                        error_arg = "x",
                         error_call = current_env()) {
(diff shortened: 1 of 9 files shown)


R/slice-chop.R
```diff
@@ -98,6 +98,7 @@ list_unchop <- function(x,
                         ptype = NULL,
                         name_spec = NULL,
                         name_repair = c("minimal", "unique", "check_unique", "universal"),
+                        error_arg = caller_arg(x),
```
lionel-, line 101: Since the error call is set to `current_env()`:

```suggestion
                        error_arg = "x",
```

lionel- (approved):

R/slice-chop.R
DavisVaughan: Oh good point!

PR #1714, opened 2022-10-04
Implement `list_all_size()` and `list_check_all_size()`

### LINKED ISSUE #1582: Give `list_all_vectors()` and `list_check_all_vectors()` `ptype` and `size` arguments?
Because sometimes you want to check that a list is homogeneous in its type, or that all elements are the same size

https://github.com/tidyverse/dplyr/pull/6300#discussion_r903532014
https://github.com/tidyverse/dplyr/pull/6300#discussion_r903527937

The `ptype` arg might be a little tricky, because `vec_assert(ptype =)` is a little hard to formalize, and would need to be translated to C

DavisVaughan:
Closes #1582


R/assert.R
```diff
@@ -194,30 +194,34 @@ vec_is_vector <- function(x) {
   .Call(vctrs_is_vector, x)
 }
 
-#' Is the object a list?
+#' List checks
```
DavisVaughan, line 197: Gave the docs here a general refresh as it was getting a bit unwieldy

src/assert.c
```diff
@@ -25,10 +25,10 @@ void vec_check_size(r_obj* x,
                     r_ssize size,
                     struct vctrs_arg* arg,
                     struct r_lazy call) {
-  r_ssize x_size = vec_size(x);
+  r_ssize x_size = vec_size_params(x, arg, call);
 
   if (x_size != size) {
-    stop_assert_size(x_size, size, arg);
+    stop_assert_size(x_size, size, arg, call);
```
DavisVaughan, line 31: The R side of `stop_assert_size()` has a `call` argument, but we weren't taking advantage of it.

I'm not sure `vec_check_size()` was actually being used anywhere, which is part of the reason we haven't seen this

src/size.c
```diff
@@ -4,17 +4,20 @@
...
+r_ssize vec_size_params(r_obj* x,
+                        struct vctrs_arg* p_arg,
+                        struct r_lazy call) {
   struct vec_error_opts err = {
-    .p_arg = vec_args.x,
-    .call = lazy_calls.vec_size
+    .p_arg = p_arg,
+    .call = call
   };
   return vec_size_opts(x, &err);
```
DavisVaughan, line 22: I've introduced `vec_size_params()` here. It proved to be a rather useful variant of `vec_size()`. I know we mentioned `vec_size4()` being another option, but this just felt like the right name for this in the context of everything else around it in this file.

src/size.c
```diff
@@ -110,6 +113,48 @@ r_obj* list_sizes(r_obj* x, const struct vec_error_opts* opts) {
...
+  struct vctrs_arg* p_x_arg = new_subscript_arg(vec_args.x, xs_names, xs_size, &i);
+  KEEP(p_x_arg->shelter);
+
+  bool out = true;
+
+  for (; i < xs_size; ++i) {
+    r_obj* x = v_xs[i];
+
+    // Scalar list elements throw an error internal to `list_all_size()`
+    r_ssize x_size = vec_size_params(x, p_x_arg, lazy_calls.list_all_size);
```
DavisVaughan, line 146: Even though scalar elements are "internal issues", we at least show that the problem came from `list_all_size()`, which felt right

src/assert.c
```diff
@@ -85,3 +85,44 @@ r_obj* ffi_list_check_all_vectors(r_obj* x, r_obj* frame) {
...
+  r_ssize xs_size = r_length(xs);
+  r_obj* xs_names = r_names(xs);
+  r_obj* const* v_xs = r_list_cbegin(xs);
+
+  struct vctrs_arg* p_x_arg = new_subscript_arg(p_arg, xs_names, xs_size, &i);
+  KEEP(p_x_arg->shelter);
+
+  for (; i < xs_size; ++i) {
+    vec_check_size(v_xs[i], size, p_x_arg, call);
+  }
```
DavisVaughan, line 125: For this one, since we have the `arg` and `call` right there, I just pass them through to `vec_check_size()`. This means that scalar elements use the user supplied arg/call even though in `list_all_size()` we considered that an internal issue.

This seemed like the right tradeoff to me since:
- We already had those user arguments available to us in `list_check_all_size()`, so we may as well use them
- This makes `list_check_all_size()` a bit more generically useful

src/size.c
lionel-: I would use `vec_size_3()` since the goal is to get a feel for this naming scheme. I don't think we should mass rename until we're sure about the scheme, and using the new scheme only in blank files isn't practical. So we should at least use it for new functions.

In this case we might want to come up with a suffix that signals error context parameters. `vec_size_with_ctxt()`?

tests/testthat/_snaps/assert.md
```diff
@@ -243,6 +243,79 @@
...
+# list_check_all_size() works
+
+    Code
+      (expect_error(list_check_all_size(list(1:2, 1:3), 2)))
+    Output
+      <error/vctrs_error_assert_size>
+      Error:
+      ! `list(1:2, 1:3)[[2]]` must have size 2, not size 3.
+    Code
+      (expect_error(list_check_all_size(list(1:2, 1:3), 2, arg = "arg", call = call(
```
lionel-, line 255: By default I'd rather see a wrapper test than an explicit arg/call test. It's fine to have both of course.

src/assert.c
lionel-: I don't understand what is the trade off here. This all seems to be behaving according to specs?

lionel- (approved):
For this sort of features I'd prefer to see some wrapper tests since that's the main use case.

src/size.c
DavisVaughan: I went with `vec_size_3()` to use the emerging naming scheme

PR #1719, opened 2022-10-05
Add restart for incompatible errors

lionel-:
Only for ptype2 in this POC. The restart resumes from `vec_default_ptype2()`.

I think this requires subclassing ptype2 and cast errors, so that different restarting handlers may be registered for either type of incompatible type errors.

Uses ordered factors for testing and illustration (will need to change once relaxed). This required some changes in the native ptype2 implementation because it didn't call `vec_default_ptype2()`. I think this is specific to this native implementation. At the R level in particular, we are careful to call the default method so that fallback measures can take place, and a CRAN search shows this is the case in downstream deps too.


DavisVaughan:
Can you help me understand where else this would be useful if we are planning on relaxing ordered/factor coercion rules anyways?

Right now I'm struggling to see the benefits and it feels like the added complexity outweighs the practical benefits 😞

lionel-:
While we are able to relax the ordered classes before release, this might not be the case in the future for another application.

A more direct benefit is that this should allow me to simplify some fallback mechanisms for coercions.

Are you sure the added complexity is prohibitive? I thought it was a fairly straightforward change.

DavisVaughan:
One more question, so assuming that we had this kind of thing in place already, ggplot2 in theory could have written `with_ordered_restart()` themselves and used that? 

i.e. we provide support for a `vctrs_restart_ptype2` restart, but this becomes a hook that other package authors or users can use as an escape hatch from too strict (to them) type rules by signaling a restart?

I feel like I would be ok with that as an internal feature if it lets you fix some fallback mechanisms, but I don't love the idea of people just being able to tweak the ptype2/cast hierarchy when they don't like it

hadley (approved):
It seems useful to have this an escape hatch that we can use if needed. I doubt other folks will make use of it.

NEWS.md
```diff
@@ -1,5 +1,14 @@
 # vctrs (development version)
 
+* Common type and cast errors now inherit from `"vctrs_error_ptype2"`
+  and `"vctrs_error_cast"` respectively. They are still both
+  subclasses from `"vctrs_error_incompatible_type"` (which used to be
+  their most specific class and is now a parent class).
+
+* `vec_ptype2()` errors can now be restarted via
+  `"vctrs_restart_ptype2"`. The restart resumes execution
+  from `vec_default_ptype2()`.
```
DavisVaughan, line 10: If you triple colon the restart, might not be worth a bullet. Maybe treat as an internal feature for now?

R/type2.R
```diff
@@ -219,6 +224,14 @@ can_fall_back <- function(x, y) {
   has_no_proxy(x) && has_no_proxy(y)
 }
 has_no_proxy <- function(x) {
+  if (inherits(x, "vctrs:::common_class_fallback")) {
+    return(TRUE)
+  }
+
+  if (!is_null(s3_get_method(class(x)[[1]], "vec_proxy", ns = "vctrs"))) {
+    return(FALSE)
+  }
```
DavisVaughan, line 233: If it is true that there is no `vec_proxy()` method, then isn't that enough to determine that there isn't a proxy?

i.e., do you even need to do the `vec_proxy()` call below?

Or maybe `has_no_proxy()` is not named exactly right? Maybe it is more like `proxy_is_identical_to_x()`? i.e. if there is a `vec_proxy()` method then we'd like to check that the `proxy` type exactly matches `x`.

R/type2.R
```diff
@@ -409,3 +422,41 @@ vec_is_subtype <- function(x, super, ..., x_arg = "", super_arg = "") {
 vec_implements_ptype2 <- function(x) {
   .Call(vctrs_implements_ptype2, x)
 }
+
+# Example usage of ptype2 restart. As there is no cast restart (yet),
+# it requires `x` and `y` to be convertible to the restarted ptype.
+with_ordered_restart <- function(expr) {
```
DavisVaughan, line 428: Agreed that this should probably live in a test helper or in `type-factor.R`?

Or I guess it may just get removed with the relaxation PR and was just a POC?

R/type2.R
```diff
@@ -409,3 +422,41 @@ vec_is_subtype <- function(x, super, ..., x_arg = "", super_arg = "") {
...
+  withCallingHandlers(
+    expr,
+    vctrs_error_ptype2 = function(cnd) {
+      x <- cnd[["x"]]
+      y <- cnd[["y"]]
+
+      restart <- FALSE
+
+      if (is.ordered(x)) {
```
DavisVaughan, line 437: I guess this allows inheritance back into the ptype2 hierarchy. i.e. if someone made a ordered subclass it could hit this. Is `inherits_only(x, c("ordered", "factor"))` better?

R/type2.R
```diff
@@ -409,3 +422,41 @@ vec_is_subtype <- function(x, super, ..., x_arg = "", super_arg = "") {
...
+        x <- as.character(x)
+      }
+      if (is.ordered(y)) {
+        restart <- TRUE
+        y <- as.character(y)
+      }
+
+      # Don't recurse and let ptype2 error keep its course
+      if (!restart) {
+        return(zap())
```
DavisVaughan, line 448: Is `zap()` your own convention? I assume you could really return anything?

R/type2.R
```diff
@@ -409,3 +422,41 @@ vec_is_subtype <- function(x, super, ..., x_arg = "", super_arg = "") {
...
+      }
+
+      # Recurse with character methods and restart with the result
```
DavisVaughan, line 451:
```suggestion
      # Retry with character methods and restart with the result
```

Is it recurse or retry? In my head it is retry - like, you failed with ordered but now you want to _retry_ with character

R/type2.R
```diff
@@ -409,3 +422,41 @@ vec_is_subtype <- function(x, super, ..., x_arg = "", super_arg = "") {
...
+    expr,
+    vctrs_error_ptype2 = function(cnd) {
+      x <- cnd[["x"]]
+      y <- cnd[["y"]]
+
+      restart <- FALSE
+
+      if (is.ordered(x)) {
+        restart <- TRUE
+        x <- as.character(x)
```
DavisVaughan, line 439: I guess the way this is set up allows for `<ordered>` `x` values to gain access to _all_ of the `<character>` ptype2 methods.

What I'm trying to say is that even though we designed this for the case of: `<ordered> + <factor> = error` is retried as `<character> + <factor> = <character>`, it is really set up so that `<ordered> + <any>` can be retried as `<character> + <any>`.

It would be up to the author of this handler to check for `is.ordered(x) && is.factor(y)` if they want it to be more specific

src/cast.h
```diff
@@ -13,6 +13,19 @@ struct cast_opts {
   struct fallback_opts fallback;
 };
 
+// FIXME: Should we merge these two structs?
```
DavisVaughan, line 16: Seems like it. Probably a remnant of when cast/ptype2 were more distinct

R/type2.R
```diff
@@ -409,3 +422,41 @@ vec_is_subtype <- function(x, super, ..., x_arg = "", super_arg = "") {
...
+      maybe_restart("vctrs_restart_ptype2", ptype)
+    }
+  )
+}
+
+maybe_restart <- function(restart, ...) {
+  if (!is_null(findRestart(restart))) {
+    invokeRestart(restart, ...)
+  }
+}
```
DavisVaughan, line 462: Should rlang provide this?

DavisVaughan (approved):
I see that you have opened another PR that pulls most of this over. I'll go look at that too

R/type2.R
lionel-: It depends on the goals of this method. For ggplot I think you'd want inheritance to be maximally compatible.

R/type2.R
lionel-: Yes returning `zap()` is a reminder that the result is ignored and allows switching to `try_fetch()` seamlessly.

R/type2.R
lionel-: I think "recurse" is an algorithmic description and "retry" is a semantic description. The word "recurse" draws attention to the fact that you need a base case.

R/type2.R
lionel-: I think we can wait for `tryInvokeRestart()` to hit the last supported R version.

R/type2.R
lionel-: `ordered + any` is now retried as `factor + any` to preserve factorness if possible. I think simplicity is more important than conservatiness for this handler, so it doesn't try to prevent inheriting factor methods.

R/type2.R
lionel-: > If it is true that there is no vec_proxy() method, then isn't that enough to determine that there isn't a proxy?

Looks like removing the logic below doesn't break any tests, so it seems sufficient.

lionel-:
Following Davis' suggestions, the ordered-restart handler now retries with factor methods instead of character methods. This preserves factorness where possible.

This requires restarting cast methods as well since this causes ordered inputs to be cast to potentially incompatible factor inputs. So we now have a cast restart too.

src/cast.c
```diff
@@ -126,35 +126,47 @@ r_obj* vec_cast_switch_native(const struct cast_opts* opts,
...
+  r_obj* ffi_to_arg = KEEP(vctrs_arg(p_to_arg));
+
   r_obj* ffi_call = KEEP(r_lazy_eval(call));
   r_obj* out = vctrs_eval_mask8(syms.vec_default_cast,
                                 syms_x, x,
                                 syms_to, to,
-                                syms_x_arg, p_x_arg,
-                                syms_to_arg, p_to_arg,
+                                syms_x_arg, ffi_x_arg,
+                                syms_to_arg, ffi_to_arg,
```
DavisVaughan, line 148: Was this broken before?

DavisVaughan: Oh no I see we were doing this earlier and oddly naming it with `r_obj* p_x_arg`

tests/testthat/helper-restart.R
```diff
@@ -0,0 +1,52 @@
...
+    expr,
+    vctrs_error_incompatible_type = function(cnd) {
+      x <- cnd[["x"]]
+      y <- cnd[["y"]]
+
+      restart <- FALSE
+
+      if (is.ordered(x)) {
+        restart <- TRUE
+        x <- factor(as.character(x), levels = levels(x))
```
DavisVaughan, line 16: Annoying that `as.factor()` doesn't strip orderedness, which in theory is really all you need to do here

lionel-: yeah I was surprised

lionel-:
> What are potentially incompatible factor inputs?

Ordered to factor. Need to override `vec_cast(ordered(), to = factor())` with the cast restart.

tests/testthat/helper-restart.R
DavisVaughan: Even more horrifying is that it isn't a generic 🙃

DavisVaughan:
@lionel- oh I didn't see a test for that directed cast case, but that makes sense. Would you mind if I added one?

lionel-:
It's tested indirectly in the `vec_rbind()` case, you're welcome to add a direct test.

DavisVaughan:
Oh I see. I forget we removed the direct `<ordered> -> <factor>` cast method when we made cast/ptype2 more symmetric

PR #1727, opened 2022-10-10
Make `c()` fallback consistent across c, unchop, and rbind

lionel-:
- Apply `c()` fallback consistently in all 3 combinations code paths.
- Apply `c()` fallback after restoration. Fixes internal error with record vectors.

PR #1728, opened 2022-10-11
Fix recursive case of `base::c()` fallback

lionel-:
Follow-up to #1727. Use same approach as in recursive restoration to support record vectors that contain fields for which we fall back to `c()`.


lionel-:
Now included in #1727.

PR #1733, opened 2022-10-19
Simplify data frame fallbacks

lionel-:
This PR simplifies the bare-type (1) and same-type (2) fallbacks for data frames.

1. We now always fall back to bare data frames or tibbles in ptype2. Previously we only did it in `vec_rbind()`, now we also do it in `vec_c()` and `list_unchop()`. This is part of our move towards a generalised bare-type fallback. This allows us to remove some internal state and options and makes the concatenation paths more consistent.

2. Simplify the same-type fallback mechanism to avoid any dispatch to `[` and `[<-`. We now simply restore attributes after df-ptype2 and df-cast, if they match. This means that df subclasses are no longer able to do pre and post checks, but if that's a problem these classes should implement vctrs methods.

There is no impact on revdeps (vctrs + tidyr + dplyr). The two new failures are false positives AFAICT.


R/type2.R
```diff
@@ -125,55 +125,34 @@ vec_default_ptype2 <- function(x,
...
-  }
 
-  if (opts$s3_fallback && can_fall_back_2(x, y)) {
-    common <- common_class_suffix(x, y)
-    if (length(common)) {
-      return(new_common_class_fallback(x, common))
+    if (identical(non_df_attrib(x), non_df_attrib(y))) {
+      attributes(out) <- c(df_attrib(out), non_df_attrib(x))
     }
+
```
DavisVaughan, line 148: The reason I'm not seeing a bare-df fallback case here (unlike the cast path in https://github.com/r-lib/vctrs/pull/1733/files#diff-d24441095488b3e181d91060e43aa207b655e9070928ffc20ab5d9351d3ef2c7R212) is because it is already handled in `vec_ptype2_df_fallback()` right?

tests/testthat/test-type-dplyr.R
```diff
@@ -163,8 +163,6 @@ test_that("can cbind rowwise data frames", {
 })
 
 test_that("no common type between rowwise and grouped data frames", {
```
DavisVaughan, line 165: Should we update this test description?

DavisVaughan (commented):
Gosh I love how much simpler this is. Thanks so much for working on this, I know it was a nightmare to work though! It looks SO much better

R/type-data-frame.R
```diff
@@ -312,146 +308,29 @@ vec_ptype2.data.frame.data.frame <- function(x, y, ...) {
...
-  )
-}
-df_fallback_as_df <- function(x) {
+as_base_df <- function(x) {
   if (inherits(x, "tbl_df")) {
-    new_data_frame(x, class = c("tbl_df", "tbl", "data.frame"))
+    new_data_frame(x, class = c("tbl_df", "tbl"))
   } else {
     new_data_frame(x)
   }
```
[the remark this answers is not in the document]
lionel-: I think this is only meant to be called with data frames, never lists.

R/type2.R
lionel-: This creates a bare df common type (bare-type fallback) and restores attributes if possible (same-type callback).

PR #1735, opened 2022-10-20
Use chained errors in subscript checks

lionel-:


tests/testthat/_snaps/error-call.md
```diff
@@ -218,8 +218,9 @@
     Output
       <error/vctrs_error_subscript_type>
       Error in `my_function()`:
-      ! Must subset elements with a valid subscript vector.
-      x Can't convert from `my_arg` <double> to <integer> due to loss of precision.
+      ! Can't subset elements with `my_arg`.
```
hadley, line 221: What does the chaining buy us here? Couldn't this be:

```
! Can't subset elements with `my_arg`.
i Can't convert from <double> to <integer> due to loss of precision.
```

hadley (commented):
I started reviewing this overall it seems like quite a lot of work for a results that's feels to me like a net neutral (or even a slight decline, since most of the errors now need extra line). So I think I must be missing some context or motivation.

PR #1737, opened 2022-10-21
Update OOB errors

lionel-:
Branched from #1736 

* Use standard subscript error header in OOB errors.
* Use actual vs expected phrasing.


tests/testthat/_snaps/slice.md
```diff
@@ -108,17 +110,17 @@
       vec_slice(set_names(letters), c("foo", "bar"))
     Condition
       Error in `vec_slice()`:
-      ! Can't subset elements that don't exist.
-      x Elements `foo` and `bar` don't exist.
+      ! Can't subset elements with `i`.
+      x Can't find elements `foo` and `bar`.
```
lionel-, line 114: This is a case where chained errors might fit better because there are multiple bullets "can't / must" bullets describing a problem at different levels.

Same for the "must" cross bullet above.

tests/testthat/_snaps/conditions.md
```diff
@@ -22,24 +22,24 @@
     Output
       <error/vctrs_error_subscript_oob>
       Error in `vec_slice()`:
-      ! Can't subset elements that don't exist.
-      x Element `foo` doesn't exist.
+      ! Can't subset elements.
```
hadley, line 25: Do we really need "Can't subset elements" in these three errors?

tests/testthat/_snaps/error-call.md
```diff
@@ -207,8 +207,8 @@
     Output
       <error/vctrs_error_subscript_oob>
       Error in `my_function()`:
-      ! Can't subset elements past the end.
-      i Location 10 doesn't exist.
+      ! Can't subset elements with `10`.
+      x Location must be less than or equal to 2, not 10.
```
hadley, line 211: Could this be one line like `Can't find location 10; input is length 2`?

hadley (commented):
I think this is an improvement, but it feels like we can still do even better, so I'd like to push this off until the next release so we can spend more time on it, and do it once.

tests/testthat/_snaps/error-call.md
lionel-: This is slightly problematic because these errors are used for subsetting across tibble rows and cols. So it's hard to find generic formulations that work well for all cases.

tests/testthat/_snaps/conditions.md
lionel-: The thing is that this is this header that carries the `arg` information, if any. That will be the case when called from `select()` and tibble's subset methods.

tests/testthat/_snaps/error-call.md
lionel-: We could say "input has 2 elements" or "input has 3 rows". This is how we currently phrase it in an additional info bullet, but I guess you're saying this information should be closer?

Another difficult aspect is how to handle the case where multiple locations are not found without creating a too long line. To address this, I try to consistently mention multiple locations in their own bullet in this PR. But then it's hard to keep the length info close.

PR #1747, opened 2022-11-16
Implement `appearance` order for `vec_locate_sorted_groups()`

DavisVaughan:


src/order.c
```diff
@@ -479,6 +525,60 @@ SEXP vec_order_info_impl(SEXP x,
...
+
+    r_ssize start = 0;
+
+    for (r_ssize i = 0; i < n_groups; ++i) {
+      p_keys[i] = p_order->p_data[start];
+      start += p_sizes[i];
+    }
+
+    // Appearance order of the unique keys
+    struct order* p_order_appearance = new_order(n_groups);
```
DavisVaughan, line 546: The idea is to find the sorted group keys of `x` and extract their order values into `key`. And then we take that integer vector of group key order values and sort it again, which gives us a way to map back to order by appearance.

One nice thing is that we are reusing all of the memory that was already allocated for the first ordering. Because this is ordering a dense integer vector and is reusing memory, it has extremely low overhead.

PR #1757, opened 2022-12-20
Don't `memset()` with `DICT_EMPTY`

DavisVaughan:
I think it is fairly dangerous to `memset()` with `DICT_EMPTY` here. It does work as is, because `DICT_EMPTY` is set to `-1` and that is one of the two values that happens to work as you'd expect with `memset()`. The other is `0`, which we do quite a bit.

See https://stackoverflow.com/a/7202474/7394743 for more details about why `-1` actually works.

A few people noted that it only works because of how two's complement works. I doubt vctrs could ever work on a computer that doesn't abide by two's complement, but I don't love relying on things like this when we can easily avoid it.

It would also obviously break if we ever changed the `DICT_EMPTY` value to anything else.

I feel like only using `memset()` with `0` is a reasonable restriction to make

### DIFF
--- a/src/dictionary.c
+++ b/src/dictionary.c
@@ -85,7 +85,10 @@ static struct dictionary* new_dictionary_opts(SEXP x, struct dictionary_opts* op
     uint32_t size = dict_key_size(x);
 
     d->key = (R_len_t*) R_alloc(size, sizeof(R_len_t));
-    memset(d->key, DICT_EMPTY, size * sizeof(R_len_t));
+
+    for (uint32_t i = 0; i < size; ++i) {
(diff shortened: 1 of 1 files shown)


lionel- (approved):
If we ever need to do this in more places we could add `uint32_fill()` as a wrapper for `std::fill()` to avoid the raw loops.

> A few people noted that it only works because of how two's complement works. I doubt vctrs could ever work on a computer that doesn't abide by two's complement, but I don't love relying on things like this when we can easily avoid it.

2 complement is required by C23 but I agree that this coercion to uchar is unexpected and unsafe. Reading this code quickly I would probably have expected `std::fill()` generic behaviour.

PR #1771, opened 2023-01-20
Implement `vec_run_sizes()`

DavisVaughan:
Part of https://github.com/r-lib/vctrs/issues/1210

I originally called this `vec_count_runs()` but that is kind of confusing because I would actually expect that to return a scalar integer of the number of runs in `x`. I think this name is much clearer, and connects nicely to `list_sizes()`, which is another function you'd use with `vec_partition(x, sizes)`. It also nicely conveys that this returns "sizes" you'd pass to the `sizes` argument of `vec_partition()`.

It turns out we also have `vec_unrep()`, which is a generalized `rle()`:

```r
> vec_unrep(c(1, 1, 1, 2, 2, 1, 1))
  key times
1   1     3
2   2     2
3   1     2
```

`vec_run_sizes()` is identical to the `times` column from `vec_unrep()`, but is much faster if you don't need the run key (slicing data frames to get the key can make it overly expensive). I looked in ivs and I only need the key in 3 of the 6 times I use this pattern. profvis also [uses `vec_unrep()`](https://github.com/rstudio/profvis/blob/56909161249ef1b9308ccba99eb1917a13828bf0/R/parse.R#L346) but also doesn't need the `key`, so they could switch to `vec_run_sizes()` too. I think the performance benefit here is worth it even though it seems a little repetitive.

I rewrote `vec_unrep()` to use `vec_run_sizes()`, as it makes the implementation much simpler and more performant. I also added some notes in the "runs" doc page about `vec_unrep()`, since it is so similar to `rle()`.

I also prototyped `vec_partition()` (not included here but the POC is here https://github.com/DavisVaughan/vctrs/pull/22). This is the pattern i use in ivs that is improved by all this:

``` r
x <- sample(10000, 100000, replace = TRUE)
y <- sample(10000, 100000, replace = TRUE)

# `needles` are in sorted order
matches <- vec_locate_matches(x, y)

# Split `haystack` by `needles`
bench::mark(
  partition = vec_partition(matches$haystack, vec_run_sizes(matches$needles)),
  chop = vec_chop(matches$haystack, vec_group_loc(matches$needles)$loc)
)
#> # A tibble: 2 × 6
#> 1 partition    13.5ms   16.4ms      59.8    2.86MB     18.0
#> 2 chop         52.8ms   52.8ms      19.0   20.21MB     75.8
```

And here is the other use for `vec_partition()` that Hadley had awhile back:

``` r
xs <- list("a", c("b", "c"), c("d", "e", "f"))

x <- list_unchop(xs)
x <- toupper(x)
x
#> [1] "A" "B" "C" "D" "E" "F"

sizes <- list_sizes(xs)
vec_partition(x, sizes)
#> [[1]]
#> [1] "A"
#> 
#> [[2]]
#> [1] "B" "C"
#> 
#> [[3]]
#> [1] "D" "E" "F"
```


src/runs.c
```diff
@@ -90,6 +90,41 @@ r_obj* vec_identify_runs(r_obj* x) {
...
+r_obj* ffi_vec_run_sizes(r_obj* x) {
+  return vec_run_sizes(x);
+}
+
+r_obj* vec_run_sizes(r_obj* x) {
+  const bool start = false;
+  r_obj* where = KEEP(vec_detect_run_bounds0(x, start));
+  const bool* v_where = r_raw_cbegin(where);
+
+  const r_ssize size = r_length(where) / sizeof(bool);
```
lionel-, line 102: Should this info be returned in a struct? Since the raw vector only acts as a shelter, it seems unnatural to return it as the primary output.

src/runs.c
```diff
@@ -90,6 +90,41 @@ r_obj* vec_identify_runs(r_obj* x) {
...
+
+  r_ssize n = 0;
+  for (r_ssize i = 0; i < size; ++i) {
+    n += v_where[i];
+  }
```
lionel-, line 107: Extract in `bool_sum()`? Avoiding raw loops generally makes the code easier to follow.

src/runs.c
```diff
@@ -90,6 +90,41 @@ r_obj* vec_identify_runs(r_obj* x) {
 
 // -----------------------------------------------------------------------------
 
+r_obj* ffi_vec_run_sizes(r_obj* x) {
+  return vec_run_sizes(x);
+}
+
+r_obj* vec_run_sizes(r_obj* x) {
+  const bool start = false;
+  r_obj* where = KEEP(vec_detect_run_bounds0(x, start));
```
lionel-, line 99: Should this be called `vec_detect_run_start0()`? It'd then be immediately clear what it's doing.

Also should the bool version have `_bool` suffix? The `0` suffix doesn't convey information if you don't already know what it means in this case.

src/runs.c
```diff
@@ -90,6 +90,41 @@ r_obj* vec_identify_runs(r_obj* x) {
...
+  }
+
+  r_obj* out = KEEP(r_alloc_integer(n));
+  int* v_out = r_int_begin(out);
+  r_ssize j = 0;
+
+  int count = 1;
+
+  for (r_ssize i = 0; i < size; ++i) {
+    const bool end = v_where[i];
```
lionel-, line 116: I think `where` should be called `starts`.

And `end` should be called `start` instead? Or am I missing something?

lionel- (approved):
I like the name `vec_run_sizes()`.

I'm not sure about the upcoming `vec_partition()`. I think it might be more natural to reuse `vec_chop()` here because the intent of vec-chop is to partition a vector into a list of parts. This is a case where taking two exclusive arguments `indices, sizes` would make sense I think.

src/rep.c
```diff
@@ -344,88 +344,40 @@ void stop_rep_times_size(struct r_lazy call,
...
-static
-r_obj* new_unrep_data_frame(r_obj* key, r_obj* times, r_ssize size) {
   r_obj* out = KEEP(r_new_list(2));
 
-  r_list_poke(out, 0, key);
+  r_list_poke(out, 0, vec_slice_unsafe(x, loc));
   r_list_poke(out, 1, times);
 
-  r_obj* names = KEEP(r_new_character(2));
+  r_obj* names = r_new_character(2);
```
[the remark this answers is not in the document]
lionel-: sorry I missed that

src/runs.c
```diff
@@ -90,6 +90,41 @@ r_obj* vec_identify_runs(r_obj* x) {
 
 // -----------------------------------------------------------------------------
 
+r_obj* ffi_vec_run_sizes(r_obj* x) {
+  return vec_run_sizes(x);
+}
+
+r_obj* vec_run_sizes(r_obj* x) {
```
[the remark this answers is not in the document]
DavisVaughan: https://github.com/r-lib/vctrs/pull/1771/commits/c5f054c2a468508e69df43a1ce1a5b897eb55f4f adds a `vec_check_vector()` check in `vec_detect_run_bounds0()`, which is used by all run-related functions.

I've also now passed the `error_call` down through all of the run functions. This generates better error calls and will make future addition of an R level `error_call` arg easier if we decide we need that.

I think together that is "good enough" because if I call a `vec_*()` function that has an `error_call` argument (here, `vec_detect_run_bounds0()`) then I expect that it will eventually throw a good error on non-vectors and will use the `error_call` to show the right call location

src/rep.c
```diff
@@ -344,88 +344,40 @@ void stop_rep_times_size(struct r_lazy call,
...
-  return vec_unrep(x);
-}
-
-
-static
-r_obj* new_unrep_data_frame(r_obj* key, r_obj* times, r_ssize size) {
   r_obj* out = KEEP(r_new_list(2));
 
-  r_list_poke(out, 0, key);
+  r_list_poke(out, 0, vec_slice_unsafe(x, loc));
```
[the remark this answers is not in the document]
DavisVaughan: Same argument as above where since I'm calling `vec_run_sizes()` and passing it `error_call`, then I expect that to handle checking for non-vectors for me

src/runs.c
DavisVaughan: I'll do this and `bool_sum()` in another PR as it touches a few other functions

src/runs.c
DavisVaughan: It actually computes the _ends_ here because `start = false`. It is "bounds" because it can do either, and this algorithm needs the ends.

src/runs.c
DavisVaughan: https://github.com/r-lib/vctrs/pull/1771#discussion_r1084391910

src/runs.c
DavisVaughan: It should probably be an enum `VCTRS_RUN_BOUND_starts` and `VCTRS_RUN_BOUND_ends` or something like that to make it clearer

DavisVaughan:
> I'm not sure about the upcoming vec_partition(). I think it might be more natural to reuse vec_chop() here because the intent of vec-chop is to partition a vector into a list of parts. This is a case where taking two exclusive arguments indices, sizes would make sense I think.

As mentioned on Slack, I strongly agree that this is the right approach. Should also let me implement it more efficiently as I think I will be able to reuse a singly compact-seq object for all the slices.

src/runs.c
DavisVaughan: - switched to enum here https://github.com/r-lib/vctrs/pull/1774
- switched to `_bool` suffix here https://github.com/r-lib/vctrs/commit/864e92ec9a11863cdedde96385e1795759fe111b

lionel-: ah gotcha. Good call regarding the enum.

Issue #1442, opened 2021-08-29
`vec_unique` produces named output


eutwt:
I think it would be helpful to document that  `vec_unique` produces named output if the arguments are named. The `help(vec_unique)` page says ays that `vec_unique` is "Equivalent to `unique()`" and does not mention names. However, in `unique()` "No attributes are copied (so the result has no names)"  (quoted from `help(unique)`).

Examples:

``` r
library(vctrs)
inputs <- list(
  named_first = vec_c(a = 1, 1),
  named_last = vec_c(1, a = 1)
)

str(lapply(inputs, vec_unique))
#> List of 2
#>  $ named_first: Named num 1
#>   ..- attr(*, "names")= chr "a"
#>  $ named_last : Named num 1
#>   ..- attr(*, "names")= chr ""
str(lapply(inputs, unique))
#> List of 2
#>  $ named_first: num 1
#>  $ named_last : num 1
```

<sup>Created on 2021-08-29 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

lionel-:
We should either strip them or document them as order-unstable. Probably best to strip them to avoid programming mistakes?

DavisVaughan:
Reopening because dplyr relies on this behavior right now in `relocate()`, see https://github.com/r-lib/vctrs/pull/1545

DavisVaughan:
The other idea is to document `vec_unique_loc()` and `vec_unique()` as order _stable_, and just say that the unique value is the first occurrence of that value (similar to the way `vec_match()` works).

I now feel like it might be useful to be able to program with a guarantee that you get the first occurrence back (and any names that are attached to it). It has historically been useful with `vec_match()` to have this guarantee. I can see a case sort of like `relocate()` where you accept a tidyselection through `...` that might allow for renaming, and you want to keep the first occurrence (or last, by reversing first), like `verb(df, foo = bar, foo2 = bar)` - keeping only the `foo = bar`.

DavisVaughan:
And we could go a step further by adding a `which = c("first", "last")` argument to `vec_unique_loc()`, which would complement https://github.com/r-lib/vctrs/issues/1239 nicely

```
# I think:
vec_unique_loc(x, which = "last") == which(!vec_duplicate_detect(x, ignore = "last"))
```

lionel-:
[...]

PR #1774, opened 2023-01-23
Use `enum vctrs_run_bound` to clarify code

DavisVaughan:
Also switched from `start = TRUE/FALSE` to `which = "start" / "end"` since we are probably going to add `which`-like arguments to `vec_unique()` and `vec_detect_duplicate()` too. https://github.com/r-lib/vctrs/issues/1442#issuecomment-1189524851

### DIFF
--- a/src/runs.c
+++ b/src/runs.c
@@ -1,18 +1,23 @@
 #include "vctrs.h"
 
+enum vctrs_run_bound {
+  VCTRS_RUN_BOUND_start = 0,
+  VCTRS_RUN_BOUND_end = 1
+};
+
 #include "decl/runs-decl.h"
 
 // -----------------------------------------------------------------------------
 
 r_obj* ffi_vec_detect_run_bounds(r_obj* x, r_obj* ffi_start, r_obj* frame) {
   struct r_lazy error_call = { .x = frame, .env = r_null };
-  const bool start = r_arg_as_bool(ffi_start, "start");
-  return vec_detect_run_bounds(x, start, error_call);
+  const enum vctrs_run_bound which = as_run_bound(ffi_start, error_call);
+  return vec_detect_run_bounds(x, which, error_call);
 }
 
 static
-r_obj* vec_detect_run_bounds(r_obj* x, bool start, struct r_lazy error_call) {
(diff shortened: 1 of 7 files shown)


src/runs.c
```diff
@@ -99,8 +103,7 @@ r_obj* ffi_vec_run_sizes(r_obj* x, r_obj* frame) {
 }
 
 r_obj* vec_run_sizes(r_obj* x, struct r_lazy error_call) {
-  const bool start = false;
-  r_obj* ends = KEEP(vec_detect_run_bounds_bool(x, start, error_call));
+  r_obj* ends = KEEP(vec_detect_run_bounds_bool(x, VCTRS_RUN_BOUND_end, error_call));
```
DavisVaughan, line 106: Makes this more obvious

src/runs.c
```diff
@@ -303,32 +306,56 @@ void col_detect_run_bounds_bool(r_obj* x, r_ssize size, bool start, bool* v_out)
...
+
+  switch (match) {
+  case 0: out = VCTRS_RUN_BOUND_start; break;
+  case 1: out = VCTRS_RUN_BOUND_end; break;
+  default: r_stop_internal("Unknown `which` value.");
+  }
+
+  FREE(1);
+  return out;
+}
```
DavisVaughan, line 361: Yay for C level `r_arg_match()`!

tests/testthat/_snaps/runs.md
```diff
@@ -14,51 +14,51 @@
...
 ---
 
     Code
-      vec_locate_run_bounds(1, start = c(TRUE, TRUE))
+      vec_locate_run_bounds(1, which = c("foo", "bar"))
     Condition
-      Error in `vec_locate_run_bounds()`:
-      ! `start` must be `TRUE` or `FALSE`.
+      Error in `arg_match()`:
+      ! `arg` must be length 1 or a permutation of `values`.
```
DavisVaughan, line 39: This seems like it could be a user error so I opened https://github.com/r-lib/rlang/issues/1545

lionel- (approved):

PR #1775, opened 2023-01-24
Collapse start/end loops into a single loop

DavisVaughan:
@lionel- I'm rather proud of this one. I was using two loops to manage the start/end bound cases of the runs functions, but I figured out a way to merge them. This greatly reduces the amount of code in the macros and makes it so that you can focus on the easier `start` case and then the `end` case "just works" too

All of the loops now increment `i` in an increasing sequence too (rather than decreasing when doing `end`), which is way more natural


lionel-:
Nice!

PR #1776, opened 2023-01-24
Introduce `struct r_bool_vector` and use in runs helpers

DavisVaughan:
@lionel- what do you think about this? Feels pretty nice to me.


lionel- (approved):
I think that makes sense. Should it be `r_vector_bool` though, esp. if we might have more vectors in the future? Should we have a generic `r_vector_begin()` macro that would work with more vector structs as long as they have the expected fields?

DavisVaughan:
> Should we have a generic r_vector_begin() macro that would work with more vector structs as long as they have the expected fields?

I think that'll be a little easier to determine once we have one or two more of these

PR #1779, opened 2023-01-26
Typo: @returns -> @return

MichaelChirico:
Fixing in the web UI -- sorry that the PR doesn't include the generated docs


DavisVaughan:
Actually not a typo! We have loosely started preferring `@returns` but i think they are mostly interchangeable

MichaelChirico:
oh, that's great! I'm always unsure about which it is, glad both are
accepted. it was flagged by a tool looking for the typo in other languages,
I'll have to edit it to ignore roxygen comments. Thanks!

PR #1782, opened 2023-02-07
Implement a `numeric_version` equality proxy

DavisVaughan:
Closes https://github.com/tidyverse/dplyr/issues/6680

`<numeric_version>` is a strange base R type where the version numbers are stored as integer vectors within a list.

``` r
x <- numeric_version(c("1.5.2", "6.3"))
x
#> [1] '1.5.2' '6.3'

unclass(x)
#> [[1]]
#> [1] 1 5 2
#> 
#> [[2]]
#> [1] 6 3
```

The length of the list is the length of the vector, but each element of the list can vary in size.

There is pseudo-support for `NA`, which is represented by `integer()` elements of the list

``` r
# `NA` is sort of supported
y <- numeric_version(c(NA, "1_1", "3.2"), strict = FALSE)
y
#> [1] <NA>  <NA>  '3.2'

unclass(y)
#> [[1]]
#> integer(0)
#> 
#> [[2]]
#> integer(0)
#> 
#> [[3]]
#> [1] 3 2

is.na(y)
#> [1]  TRUE  TRUE FALSE
```

We need to create an equality/comparison/order proxy for this kind of thing.

This is surprisingly hard, because the number of components across versions in a single vector can vary (3 and 2 components respectively in the first example above). We can't just take the max number of components in a single vector (3 there) and pad out to that size, because if you are doing `vec_compare(x, y)` you'd have to generate the _joint_ comparison proxy between `x` and `y` - i.e. the proxy would be _data dependent_.

I've come up with an _independent_ data frame proxy that seems to work fairly well. It requires us to place a restriction on the maximum number of components allowed in a version number, which I've hardcoded to 8. It looks like:

``` r
x <- numeric_version(c("1.5.2", "6.3"))
proxy_equal_numeric_version(x)
#>   ...1 ...2 ...3 ...4 ...5 ...6 ...7 ...8
#> 1    1    5    2    0    0    0    0    0
#> 2    6    3    0    0    0    0    0    0

y <- numeric_version(c(NA, "1_1", "3.2"), strict = FALSE)
proxy_equal_numeric_version(y)
#>   ...1 ...2 ...3 ...4 ...5 ...6 ...7 ...8
#> 1   NA   NA   NA   NA   NA   NA   NA   NA
#> 2   NA   NA   NA   NA   NA   NA   NA   NA
#> 3    3    2    0    0    0    0    0    0
```

This works for all of the Version numbers returned in `available.packages()`, with the maximum there being 6 components, so I've left some wiggle room.

I tried to think of a number of ways to collapse the version number into some double value or a character string that would always be able to be compared to other character strings, but wasn't able to come up with anything easy. Base R has `.encode_numeric_version()` that encodes it as a character string, but they have the same joint proxy issue that I mentioned above for `vec_compare()`. In `Ops.numeric_version` they are forced to combine `x` and `y` before computing `.encode_numeric_version()` to get a joint proxy.

I'm hopeful this is a "good enough" solution that we won't have to ever revisit again.


tests/testthat/test-type-misc.R
```diff
@@ -29,6 +29,133 @@ test_that("`numeric_version` falls back to base methods", {
...
+    numeric_row(1, 2, 0, 0, 0, 0, 0, 0),
+    numeric_row(NA, NA, NA, NA, NA, NA, NA, NA)
+  )
+
+  expect_identical(vec_proxy_equal(x), expect)
+  expect_identical(vec_proxy_compare(x), expect)
+  expect_identical(vec_proxy_order(x), expect)
+})
+
+test_that("`numeric_version` works with functions using the equality proxy", {
```
lionel-, line 86: Also add tests comparing versions with 8 components to versions with fewer components?

lionel- (approved):
Nice!

Also needs a test in dplyr before closing?

PR #1791, opened 2023-02-21
Implement `relationship` for `vec_locate_matches()`

DavisVaughan:
Related to https://github.com/tidyverse/dplyr/issues/6731

@lionel- if you just want to do a high level review, that would be ok with me

This PR implements `relationship`, a new argument for `vec_locate_matches()` that supersedes and expands upon `multiple = "error" / "warning"`.

`relationship` is based on [database table relationships](https://www.metabase.com/learn/databases/table-relationships) and will get exposed in dplyr as well. It has a number of options:

- `"none"` (will be the default for dplyr inequality joins)
- `"one_to_one"`
- `"one_to_many"`
- `"many_to_one"`
- `"many_to_many"`
- `"warn_many_to_many"` (will be the default for dplyr equality joins, and likely rolling joins)

The dplyr default will be `NULL`, which will choose between `"none"` and `"warn_many_to_many"` automatically.

`relationship` adds a _constraint_ on the matching procedure, for example, `"many_to_one"` says that:
- Each element of `needles` can match _at most 1_ value in `haystack`
- Each element of `haystack` can match _any number_ of values in `needles`

This `"many_to_one"` option is then a perfect replacement for `multiple = "error"`, which I really like because now `multiple` is purely about optionally reducing the number of matches, not about erroring.

Notably, `relationship` is _not_ a uniqueness check on the inputs up front. It is instead a check on the actual matched results. I think this better aligns with the SQL definitions. It means that you can do `relationship = "one_to_one"` and have duplicate `needles` and not get an error if they don't match anything. If that is a problem then you are supposed to use `no_match` to handle the "unmatched `needles`" case (or `unmatched` on the dplyr side):

``` r
x <- c(1, 1, 2)

# fine
vec_locate_matches(x, 2, relationship = "one_to_one")
#>   needles haystack
#> 1       1       NA
#> 2       2       NA
#> 3       3        1

# error
vec_locate_matches(x, 1, relationship = "one_to_one")
#> Error in `vec_locate_matches()`:
#> ! Each element can match at most 1 observation.
#> ✖ The element at location 1 has multiple matches.

# handling unmatched `needles`
vec_locate_matches(x, 2, relationship = "one_to_one", no_match = "error")
#> Error in `vec_locate_matches()`:
#> ! Each element must have a match.
#> ✖ The element at location 1 does not have a match.
```

I've gone ahead and silently soft-deprecated `multiple = "error"` and `multiple = "warning"` by removing them from the documentation and by moving all their tests to a deprecation section.


R/match.R
```diff
@@ -108,9 +108,43 @@
 #'     `"last"` if you just need to detect if there is at least one match.
 #'   - `"first"` returns the first match detected in `haystack`.
 #'   - `"last"` returns the last match detected in `haystack`.
-#'   - `"warning"` throws a warning if multiple matches are detected, but
-#'     otherwise falls back to `"all"`.
-#'   - `"error"` throws an error if multiple matches are detected.
```
DavisVaughan, line 113: Removed from the docs. No formal deprecation with `deprecate_soft()` yet, I'd like to remove usage of them from dplyr and ivs first

R/match.R
```diff
@@ -108,9 +108,43 @@
...
+#'   - `"many_to_one"` expects:
+#'     - Each value in `needles` matches at most 1 value in `haystack`.
+#'     - Each value in `haystack` matches any number of values in `needles`.
+#'
+#'   - `"many_to_many"` expects:
+#'     - Each value in `needles` matches any number of values in `haystack`.
+#'     - Each value in `haystack` matches any number of values in `needles`.
+#'
+#'     This performs no checks, and is identical to `"none"`, but is provided to
+#'     allow you to be explicit about this relationship if you know it exists.
```
DavisVaughan, line 135: Made sure to note that `many_to_many` doesn't really do anything

src/match.c
```diff
@@ -1626,25 +1669,58 @@ r_obj* expand_compact_indices(const int* v_o_haystack,
...
-  bool any_multiple = false;
-  bool check_for_multiple =
+  // Check is always needed for `multiple = "all"`.
+  // This also handles `relationship` options too, since if `multiple` is
+  // `"any"`, `"first"`, or `"last"`, we can't invalidate a `relationship`.
+  bool check_multiple_needles =
     multiple == VCTRS_MULTIPLE_all ||
+    // TODO: Remove deprecated support for `multiple = "error"/"warning"`
     multiple == VCTRS_MULTIPLE_error ||
     multiple == VCTRS_MULTIPLE_warning;
```
DavisVaughan, line 1685: Figuring out if we need to check for multiple matches in `needles` is much simpler than for `haystack`.

We basically just need to check if `multiple = "all"`, regardless of the value of `relationship`.

If `multiple = "all"`, we _have_ to check for multiple matches in `needles` no matter what because we do a sort at the end if we detect multiple matches.

If `multiple %in% c("any", "first", "last")` then we _don't_ need to check for multiple matches in `needles`, because by definition they will be reduced down to 1 match through `multiple`.

src/match.c
```diff
@@ -1747,23 +1823,74 @@ r_obj* expand_compact_indices(const int* v_o_haystack,
...
+              error_call
+            );
+          case VCTRS_MULTIPLE_warning: {
+            warn_matches_multiple(
+              loc_first_multiple_needles,
+              needles_arg,
+              haystack_arg,
+              error_call
+            );
+            break;
```
DavisVaughan, line 1884: Squeezing in some backwards compatible support for `multiple = "error" / "warning"` here in the "we detected multiple matches in `needles`" case

src/match.c
```diff
@@ -1747,23 +1823,74 @@ r_obj* expand_compact_indices(const int* v_o_haystack,
...
+              haystack_arg,
+              error_call
+            );
+          }
+          break;
+        }
+        default: {
+          switch (multiple) {
+          case VCTRS_MULTIPLE_all:
+            break;
```
DavisVaughan, line 1868: This case corresponds to how `check_multiple_needles` is defined, where `multiple = "all"` is the main case where we want to know if there are multiple matches in `needles`, but we don't error or warn or anything.

src/match.c
```diff
@@ -1850,9 +2012,57 @@ r_obj* expand_compact_indices(const int* v_o_haystack,
...
+                }
+
+                // We know there are multiple and don't need to continue checking
+                check_multiple_haystack = false;
+
+                // Only continue tracking if needed for `remaining`
+                track_matches_haystack = retain_remaining_haystack;
+
+                break;
+              }
```
DavisVaughan, line 2057: Basic idea here is that as we process the `haystack` matches, we record the location in `haystack` that the match corresponds to in `v_detect_matches_haystack`. If we see that location more than once, that corresponds to a multiple match and we take the appropriate action based on `relationship`.

If the "action" is to just throw a warning, then we throw it once and after that we typically no longer need to check for multiple `haystack` matches.

The exception is if `remaining = "error" / <int>` is set, in which case we still have to track usage of all of the `haystack` locations to figure out which locations _weren't_ used at all (i.e. they are "unmatched" in dplyr join terminology). So that is what `track_matches_haystack = retain_remaining_haystack;` is for.

tests/testthat/_snaps/match.md
```diff
@@ -153,33 +181,245 @@
...
+      ! Each element can match at most 1 observation.
+      x The element at location 2 has multiple matches.
+    Code
+      (expect_error(vec_locate_matches(c(1, 1), c(1, 2), relationship = "one_to_one"))
+      )
+    Output
+      <error/vctrs_error_matches_relationship_one_to_one>
+      Error in `vec_locate_matches()`:
+      ! Each element can match at most 1 observation.
+      x The element at location 1 has multiple matches.
```
DavisVaughan, line 201: Here is one place I think it is confusing that we have `needles_arg = ""` and `haystack_arg = ""` as defaults. One of these errors is related to `needles` and one is related to `haystack` but the errors are the same.

I think we should default them to `"needles"` and `"haystack"` and then completely disallow the empty string of `""` as a possible arg name in `vec_locate_matches()`. This would also greatly simplify my error generation code on the R side

tests/testthat/_snaps/match.md
```diff
@@ -153,33 +181,245 @@
...
+
+    Code
+      (expect_warning(vec_locate_matches(c(1, 2, 1), c(1, 2, 2), relationship = "warn_many_to_many"))
+      )
+    Output
+      <warning/vctrs_warning_matches_relationship_many_to_many>
+      Warning in `vec_locate_matches()`:
+      Detected an unexpected many-to-many relationship.
+      x The element at location 2 has multiple matches.
+      x The element at location 1 has multiple matches.
```
DavisVaughan, line 235: The many-to-many warning is another place the `needles_arg` / `haystack_arg` defaults are particularly confusing. The first line relates to `needles` and the second is for `haystack`

R/match.R
```diff
@@ -108,9 +108,43 @@
...
+#'
+#'   - `"warn_many_to_many"` doesn't assume there is any known relationship, but
```
lionel-, line 137: Should `warn` be a suffix instead of a prefix? I think that's what I'd have expected to make it clearer it's a variant.

tests/testthat/_snaps/match.md
lionel-: I think taking `""` args is probably a remnant of the early days before we had developed our approach to error reporting. We probably didn't have `error_call` arguments at the time, so we were hiding the call and it made sense to hide the arguments as well.

Now that we show the call, we should probably start naming the arguments by default as you suggest.

lionel- (approved):
This seems like a nice improvement!

R/match.R
DavisVaughan: I couldn't really decide, because it's not like `"many_to_many"` errors and `"many_to_many_warn"` would just downgrade that error to a warning.

They do fairly different things:
- `"many_to_many"` is looking for _violations of_ the many-to-many relationship, i.e. many-to-many _is_ expected. But that relationship can't be violated so it ends up being an expressive no-op (it makes more sense for `"many_to_one"` and the other variants)
- `"warn_many_to_many"` is looking for _the existance of_ a many-to-many relationship, i.e. many-to-many _is not_ expected, and it warns if it finds one.

So they didn't really feel like variants to me

I'm not picky about this though, so if you still feel like the suffix is better then I'll change it, but I'll wait to hear what you think

lionel-: No strong opinion either way, we can go with your first intuition.

Issue #1598, opened 2022-07-27
[Suggestion] shifts and partitions


tzakharko:
I would like `vctrs` to add support for shift and partition operations. 

`vec_shift(x, amount, fill)` — shifts the vector by positive or negative amount and fills up the gaps with `fill`. The result has the same size as the input. This is basically `dplyr::lag()` and `dplyr::lead()` and is useful for similar applications. 

`vec_partition(x, indices)` — chops the vector into slices along the monotonously increasing sequence of indices. This is useful for inserting items at arbitrary locations in a vector. Examples:

```r
vec_partition(x, 5) ---> list(x[..5], x[6..]) 
vec_partition(x, 0) ---> list(x[0], x)
vec_partition(x, length(x)) ---> list(x, x[0])
vec_partition(x, c(2, 5, 10)) ---> list(x[..2], x[3..5], x[6..10], x[11...])
```

If the maintainers are interested in this kind of functionality I can submit a PR.

DavisVaughan:
FWIW the dplyr functions are now based on a common `shift()` helper
https://github.com/tidyverse/dplyr/blob/17c7a5d8d208f3c89d023c4735c14a74ee336dbd/R/lead-lag.R#L89

I do think that would be useful as a helper outside of dplyr, but im not sure that vctrs is the right place for that one. We've got a long term goal of moving higher level (user facing) vector functions like that to https://github.com/tidyverse/funs

---

`vec_partition()` is similar to (possibly a duplicate of) https://github.com/r-lib/vctrs/issues/1210

tzakharko:
Thanks, I missed #1210 when I was looking for similar stuff. 

Personally, I think shifts and other advanced vector operations (e.g. compact) would be a great fit for vctrs, which already offers optimised advanced SIMD-style operations such as interleave and scatter/gather (`vec_chop`).

PR #1813, opened 2023-03-07
Implement `vec_chop(sizes = )`

DavisVaughan:
Closes https://github.com/r-lib/vctrs/issues/1598
Closes https://github.com/r-lib/vctrs/issues/1210

This PR implements a new `sizes` argument to `vec_chop()`, with the following specs:
- `sizes` must be an integer vector containing only non-negative sizes (`0` is ok, but `-1` and `NA` are not)
- `sizes` must sum to `vec_size(x)`
- `sizes` can't be used alongside `indices`

`sizes` allows us to efficiently chop a vector using the results of either `vec_run_sizes()` or `list_sizes()`. It is typically faster than chopping with locations from `vec_group_loc()` if you know the input is ordered (or already grouped).

```r
set.seed(123)

x <- sample(1e5, 1e6, replace = TRUE)
x <- sort(x)

head(x, n = 20)
#>  [1] 1 1 1 1 1 1 1 1 1 1 1 2 2 2 2 2 2 2 2 3

# Split `x` on its groups
bench::mark(
  sizes = vec_chop(x, sizes = vec_run_sizes(x)),
  indices = vec_chop(x, indices = vec_group_loc(x)$loc)
)
#> # A tibble: 2 × 6
#> 1 sizes        8.37ms   10.7ms      94.9     2.1MB     9.49
#> 2 indices     52.61ms   53.1ms      18.9    20.2MB    31.4
```

With ^, typically you chop something else, but you use the run sizes of `x` to index it.

It also allows for some neat patterns with `list_sizes()` related to "unchop, do something, rechop" that were clunky before:

``` r
x <- list(c(1, 2, 1), c(3, 1), 5, double())

x_flat <- list_unchop(x)
x_flat <- x_flat + max(x_flat)

vec_chop(x_flat, sizes = list_sizes(x))
#> [[1]]
#> [1] 6 7 6
#> 
#> [[2]]
#> [1] 8 6
#> 
#> [[3]]
#> [1] 10
#> 
#> [[4]]
#> numeric(0)
```

---

I've spent a significant amount of time cleaning up the `vec_chop()` internals. I think they are now much easier to read. In particular, `vctrs_chop_info` was doing way too much:
- `vctrs_chop_info` was managing the `out` container, which it no longer does
- `vctrs_chop_info` was taking and managing the `proxy` info, which it no longer does

I've now narrowly scoped `vctrs_chop_info` to only manage the index generation process, and have renamed it to `vctrs_chop_indices`. It handles 3 different paths depending on which of the following cases we are in:
- `indices = NULL, sizes = NULL`
- `indices != NULL`
- `sizes != NULL`

That seemed to dramatically clean things up. I think if you just skim over the new `slice-chop.c` without the diffs, then that should be apparent.

### DIFF
--- a/R/slice-chop.R
+++ b/R/slice-chop.R
@@ -121,9 +121,33 @@
 #' x_flat <- list_unchop(x)
 #' x_flat <- x_flat + max(x_flat)
 #' vec_chop(x_flat, sizes = list_sizes(x))
-vec_chop <- function(x, indices = NULL, ..., sizes = NULL) {
-  check_dots_empty0(...)
+vec_chop <- function(x, ..., indices = NULL, sizes = NULL) {
+  if (!missing(...)) {
+    indices <- check_dots_chop(..., indices = indices)
+  }
   .Call(ffi_vec_chop, x, indices, sizes)
+}
+
+check_dots_chop <- function(..., indices = NULL, call = caller_env()) {
+  if (!is_null(indices)) {
+    # Definitely can't supply both `indices` and `...`
+    check_dots_empty0(..., call = call)
+  }
+
+  if (dots_n(...) != 1L) {
+    # Backwards compatible case doesn't allow for length >1 `...`.
+    # This must be an error case.
+    check_dots_empty0(..., call = call)
+  }
+
+  # TODO: Soft-deprecate this after dplyr/tidyr have updated all `vec_chop()`
+  # calls to be explicit about `indices =`
+
+  # Assume this is an old style `vec_chop(x, indices)` call, before we
+  # added the `...`
+  indices <- list(...)[[1L]]
+
+  indices
 }
 
 #' @rdname vec_chop
--- a/R/slice-chop.R
+++ b/R/slice-chop.R
@@ -24,7 +24,8 @@
 #'
 #' @param x A vector
 #' @param indices For `vec_chop()`, a list of positive integer vectors to
-#'   slice `x` with, or `NULL`. If `NULL`, `x` is split into its individual
+#'   slice `x` with, or `NULL`. Can't be used if `sizes` is already specified.
+#'   If both `indices` and `sizes` are `NULL`, `x` is split into its individual
 #'   elements, equivalent to using an `indices` of `as.list(vec_seq_along(x))`.
 #'
 #'   For `list_unchop()`, a list of positive integer vectors specifying the
@@ -32,12 +33,22 @@
 #'   the size of the corresponding index vector. The size of `indices` must
 #'   match the size of `x`. If `NULL`, `x` is combined in the order it is
 #'   provided in, which is equivalent to using [vec_c()].
+#' @param sizes An integer vector of non-negative sizes representing sequential
+#'   indices to slice `x` with, or `NULL`. Can't be used if `indices` is already
+#'   specified.
+#'
+#'   For example, `sizes = c(2, 4)` is equivalent to `indices = list(1:2, 3:6)`,
+#'   but is typically faster.
+#'
+#'   `sum(sizes)` must be equal to `vec_size(x)`, i.e. `sizes` must completely
+#'   partition `x`, but an individual size is allowed to be `0`.
 #' @param ptype If `NULL`, the default, the output type is determined by
 #'   computing the common type across all elements of `x`. Alternatively, you
 #'   can supply `ptype` to give the output a known type.
 #' @return
-#' - `vec_chop()`: A list of size `vec_size(indices)` or, if `indices == NULL`,
-#'   `vec_size(x)`.
+#' - `vec_chop()`: A list where each element has the same type as `x`. The size
+#'   of the list is equal to `vec_size(indices)`, `vec_size(sizes)`, or
+#'   `vec_size(x)` depending on whether or not `indices` or `sizes` is provided.
 #'
 #' - `list_unchop()`: A vector of type `vec_ptype_common(!!!x)`, or `ptype`, if
 #'   specified. The size is computed as `vec_size_common(!!!indices)` unless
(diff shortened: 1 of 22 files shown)


R/slice-chop.R
```diff
@@ -89,8 +105,25 @@
...
+#'
+#' # If you have a list of homogeneous vectors, sometimes it can be useful to
+#' # unchop, apply a function to the flattened vector, and then rechop according
+#' # to the original indices. This can be done efficiently with `list_sizes()`.
+#' x <- list(c(1, 2, 1), c(3, 1), 5, double())
+#' x_flat <- list_unchop(x)
+#' x_flat <- x_flat + max(x_flat)
+#' vec_chop(x_flat, sizes = list_sizes(x))
+vec_chop <- function(x, indices = NULL, ..., sizes = NULL) {
+  check_dots_empty0(...)
```
DavisVaughan, line 125: I don't think we can move `indices` at this point, but I wanted to add the empty dots in and felt it was worth it to do so anyways

src/bind.c
```diff
@@ -304,7 +304,7 @@ r_obj* as_df_row_impl(r_obj* x,
   // Remove names first as they are promoted to data frame column names
   x = KEEP(vec_set_names(x, r_null));
 
-  x = KEEP(vec_chop(x, r_null));
+  x = KEEP(vec_chop_unsafe(x, r_null, r_null));
```
DavisVaughan, line 307: The version of `vec_chop()` we use internally has been renamed to `vec_chop_unsafe()` to make it clear that it doesn't perform `vec_as_location()` or `vec_as_chop_sizes()` checks on the inputs.

Similar to `vec_slice_unsafe()`!

src/c-unchop.c
```diff
@@ -85,7 +85,7 @@ r_obj* list_unchop(r_obj* xs,
     out_size += r_length(r_list_get(indices, i));
   }
 
-  r_obj* locs = KEEP(vec_as_indices(indices, out_size, r_null));
+  r_obj* locs = KEEP(list_as_locations(indices, out_size, r_null));
```
DavisVaughan, line 88: `vec_as_indices()` took a list of location vectors and called `vec_as_location()` on each. I've renamed that to `list_as_locations()` to reflect modern naming schemes. It might not be a perfect name but I think it is much closer.

src/callables.c
```diff
@@ -19,7 +19,7 @@ SEXP exp_vec_cast(SEXP x, SEXP to) {
 }
 
 SEXP exp_vec_chop(SEXP x, SEXP indices) {
-  return vec_chop(x, indices);
+  return vec_chop_unsafe(x, indices, r_null);
 }
```
DavisVaughan, line 23: Exposed experimental `exp_vec_chop()` callable does not gain the `sizes` argument

tests/testthat/_snaps/c.md
```diff
@@ -256,13 +256,13 @@
...
     Code
       with_memory_prof(list_unchop(make_list_of(2000)))
     Output
-      [1] 220KB
+      [1] 205KB
     Code
       with_memory_prof(list_unchop(make_list_of(4000)))
     Output
-      [1] 439KB
+      [1] 408KB
```
DavisVaughan, line 267: `make_list_of()` calls `vec_chop()` so I'm not entirely surprised to see changes here, but I don't really have a clue what the changes are attributed to

tests/testthat/_snaps/slice-chop.md
```diff
@@ -1,3 +1,82 @@
...
+---
+
+    Code
+      (expect_error(vec_chop(1, indices = list(1.5)), class = "vctrs_error_subscript_type")
+      )
+    Output
+      <error/vctrs_error_subscript_type>
+      Error:
+      ! Can't subset elements.
+      x Can't convert from <double> to <integer> due to loss of precision.
```
DavisVaughan, line 18: The errors of `vec_chop()` aren't perfect, but that is partially due to not having exposed `error_call` and `error_arg` like `list_unchop()`. But that is big enough to be its own PR. I have moved a few tests towards snapshot tests rather than just `expect_error()` calls so we will eventually see the improvements when we add those arguments.

src/slice-chop.c
```diff
@@ -2,60 +2,144 @@
...
 
 static
-struct vctrs_chop_info new_chop_info(r_obj* x, r_obj* indices) {
-  struct vctrs_chop_info info;
-  info.shelter = KEEP(r_alloc_list(3));
+struct vctrs_chop_indices* new_chop_indices(r_obj* x, r_obj* indices, r_obj* sizes) {
+  r_obj* shelter = KEEP(r_alloc_list(4));
+
+  r_obj* self = r_alloc_raw(sizeof(struct vctrs_chop_indices));
+  r_list_poke(shelter, 0, self);
```
DavisVaughan, line 70: Also switched to our more standard approach of returning a pointer that manages itself

src/slice-chop.h
```diff
@@ -3,8 +3,10 @@
 
 #include "vctrs-core.h"
 
+r_obj* vec_chop(r_obj* x, r_obj* indices, r_obj* sizes);
+r_obj* vec_chop_unsafe(r_obj*, r_obj* indices, r_obj* sizes);
```
DavisVaughan, line 7: Moved `vec_chop()` from `<vctrs.h>` to `<slice-chop.h>`

And of course now `vec_chop()` is the version that does the index checks, and `vec_chop_unsafe()` is the performance variant that assumes the indices are already correct

lionel- (approved):
Looks great!

R/slice-chop.R
DavisVaughan: Good idea, done here https://github.com/r-lib/vctrs/pull/1813/commits/c1bd65e261efcba34b7813d5e4f38abd02287c85

PR #1815, opened 2023-03-13
Update rlang C lib

DavisVaughan:
- Removed reliance on C++ to build vctrs
- We did use `r_list_all_of()` in 1 place, so I removed that
- Updated snapshot test related to rlang changes
- There was a very old unused `r_new_environment()` implementation in vctrs that is now rlang's `r_alloc_environment()`, so I removed it
- There was an actually used `r_new_environment()` implementation in vctrs that is now rlang's `r_alloc_empty_environment()`, so I updated to that

### DIFF
--- a/src/type-info.c
+++ b/src/type-info.c
@@ -135,7 +135,19 @@ bool list_all_vectors(r_obj* x) {
   if (r_typeof(x) != R_TYPE_list) {
     r_stop_unexpected_type(r_typeof(x));
   }
-  return r_list_all_of(x, &obj_is_vector);
+
+  const r_ssize size = r_length(x);
+  r_obj* const* v_x = r_list_cbegin(x);
+
+  for (r_ssize i = 0; i < size; ++i) {
+    r_obj* elt = v_x[i];
+
+    if (!obj_is_vector(elt)) {
+      return false;
+    }
+  }
+
+  return true;
 }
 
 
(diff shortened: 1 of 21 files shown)


src/type-info.c
```diff
@@ -135,7 +135,19 @@ bool list_all_vectors(r_obj* x) {
   if (r_typeof(x) != R_TYPE_list) {
     r_stop_unexpected_type(r_typeof(x));
   }
-  return r_list_all_of(x, &obj_is_vector);
+
```
lionel-, line 138: This decreases code quality a little bit, so add TODO note to switch back to `r_list_all_of()`? Also it'll be easier to handle the memory barrier path for altlists in loop wrappers like `r_list_all_of()`.

NEWS.md
```diff
@@ -45,6 +45,8 @@
 * R >=3.5.0 is now explicitly required. This is in line with the tidyverse
   policy of supporting the [5 most recent versions of
   R](https://www.tidyverse.org/blog/2019/04/r-version-support/).
+  
+* vctrs no longer requires C++ to build, only C (#1815).
```
lionel-, line 49: I'd prefer not to mention that in NEWS because it's an implementation detail that doesn't concern users. In almost all cases C++ compilation just works. Also I still think it might be useful in the future for a package like vctrs to call the STL for internal steps of complex algorithms such as those implemented in memtools.

lionel- (approved):

PR #1817, opened 2023-03-16
Fix sf `c()` test now that `c.sfc()` requires identical crs

DavisVaughan:
In sf 1.0-10, they fixed https://github.com/r-spatial/sf/issues/1884, which means that sfc vectors now can't be combined with `c()` if they have different crs values, which seems reasonable to me.

I've done the minimal work to get our `c()` tests working correctly again, but I don't know too much about how these sf bits work.

I figure we'd merge this then move sf and maybe data.table to Suggests to ensure they are tested

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,7 @@
 # vctrs (development version)
 
+* Fixed a test related to `c.sfc()` changes in sf 1.0-10 (#1817).
+
 # vctrs 0.6.0
 
 * New `vec_run_sizes()` for computing the size of each run within a vector. It


tests/testthat/test-type-sf.R
```diff
@@ -213,22 +214,31 @@ test_that("`precision` and `crs` attributes of `sfc` vectors are restored", {
...
+  y = st_sfc(st_point(c(0, 0)), precision = 1e-4, crs = 3857)
+
+  out = vctrs::vec_c(x, y)
+  expect_identical(st_crs(x), st_crs(out))
+
+  # Error on different `crs` comes from sf as of 1.0-10
+  y = st_sfc(st_point(c(0, 0)), precision = 1e-4, crs = 4326)
+  expect_snapshot(error = TRUE, {
+    vctrs::vec_c(x, y)
+  })
```
DavisVaughan, line 241: I can't seem to figure out how to get rstudio to respect the tab width thing for just one file

lionel- (approved):
> I figure we'd merge this then move sf and maybe data.table to Suggests to ensure they are tested

Alternatively we just add a `skip_on_cran()` at the top of the files.

Moving data.table to Suggests has little cost though. I'm a bit worried about sf but maybe it's now alright with PPM binaries and improved caching on GHA?

DavisVaughan:
I guess adding `skip_on_cran()` and keeping everything else the same means it would pretty much be guaranteed that these tests would only ever run locally on our personal computers?

DavisVaughan:
One of the "bigger" benefits to consider for putting them in Suggests is that sf probably would have let us know about this change earlier on through their revdep check process

lionel-:
> I guess adding skip_on_cran() and keeping everything else the same means it would pretty much be guaranteed that these tests would only ever run locally on our personal computers?

yes this was the original idea behind removing from Suggests. At the time I didn't expect CRAN machines to still have sf installed in the checks environment.

PR #1828, opened 2023-03-30
Don't const-qualify parameters in declaration

MichaelChirico:
See https://clang.llvm.org/extra/clang-tidy/checks/readability/avoid-const-params-in-decls.html

### DIFF
--- a/src/decl/rep-decl.h
+++ b/src/decl/rep-decl.h
@@ -31,7 +31,7 @@ r_obj* vec_rep_each_uniform(r_obj* x,
 static
 r_obj* vec_rep_each_impl(r_obj* x,
                          r_obj* times,
-                         const r_ssize times_size,
+                         r_ssize times_size,
                          struct r_lazy error_call,
                          struct vctrs_arg* p_times_arg);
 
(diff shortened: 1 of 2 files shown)


lionel-:
It's weird that the definition doesn't match the declaration. Out of curiosity, why are you linting our code base with style checks?

MichaelChirico:
Hmm good point, I guess I'd thought the implementation should also remove `const` but that's not so.

https://stackoverflow.com/q/52916410/3576984

So actually this check might be something of a nuisance to maintain -- closing.

Our tooling flags style issues in C/C++ whenever we import/update a package, so I try to send any trivial improvements back upstream. (obviously this case was in error as it really is a style choice whether to keep the const).

Please flag if you'd rather not receive such PRs.

lionel-:
Personally I think those PRs are fine as long as they're not about pure style, they improve code quality, and they're not too frequent. @DavisVaughan might have another opinion.

lionel-:
FWIW I think the lint you're proposing here is a good idea for public headers, since const qualifiers are a distraction in the self-documenting API. But here these are internal headers.

PR #1872, opened 2023-08-16
Make `vctrs_init_altrep_lazy_character ` function signatures consistent

georgestagg:
In `src/altrep-lazy-character.c` we have,

```c
void vctrs_init_altrep_lazy_character(DllInfo* dll) {
[...]
}
```

but in `src/init.c` we have,

```c
extern r_obj* vctrs_init_altrep_lazy_character(DllInfo*);
```

This breaks vctrs 0.6.3 under webR because Wasm requires consistent function signatures across compilation units. The function doesn't seem to return anything, so I've switched the `extern` definition to use a return type of `void`.


DavisVaughan:
Thanks!

PR #1873, opened 2023-08-23
Direct inclusion of used headers

MichaelChirico:
As flagged by `clang-tidy`'s `misc-include-cleaner`:

https://clang.llvm.org/extra/clang-tidy/checks/misc/include-cleaner.html

### DIFF
--- a/src/altrep-lazy-character.c
+++ b/src/altrep-lazy-character.c
@@ -1,9 +1,22 @@
-#include "vctrs.h"
+#include <cstddef>
+
 #include "altrep.h"
 
 #if (!HAS_ALTREP)
 
+#include <R_ext/Altrep.h>
+#include <R_ext/Boolean.h>
+#include <R_ext/Print.h>
 #include <R_ext/Rdynload.h>
+#include <Rinternals.h>
+
+#include "call.h"
+#include "cnd.h"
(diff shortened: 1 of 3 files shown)


DavisVaughan:
We purposefully use `vctrs.h` as a megaheader that includes most useful headers you might want to use when developing new algorithms. It already includes `proxy.h` and `size.h`, for example.

MichaelChirico:
Thanks... I wonder if CRAN will ever move towards stricter inclusion rules.

Myself, I've become more of a fan of the stricter rules after dealing with all sorts of `Depends`/`Imports` headaches this year. The "megaheader" approach reminds me of `Depends` and all the associated hassle.

Anyway it seems the R headers themselves need some cleanup / I don't have a good environment to debug this PR right now, so I'll close.

DavisVaughan:
@lionel- and I definitely fought over whether to use a megaheader vs very strict include-what-you-use type of rules in the past 😆

lionel-:
> The "megaheader" approach reminds me of Depends and all the associated hassle.

I think a better analogy is that the one header per file approach is like creating a different package for each .R file.

Maybe this could help? https://clangd.llvm.org/design/include-cleaner#umbrella-headers

Issue #1877, opened 2023-09-21
Should the recycling rules really be `@internal`?


jennybc:
I came to vctrs looking for vctrs recycling rules and a help topic exists, but it's marked internal and therefore somewhat hard to find.

https://github.com/r-lib/vctrs/blob/8bbd8c4a69a9b3e2c42aa752c5339f949562af96/R/recycle.R#L70

I gather this is intentional (https://github.com/r-lib/vctrs/pull/1580/commits/88fd9b6646996c5e75e21889a39362b5c2f4b185), but that seems to sort of contradict other signals that suggest this topic is meant to be user-facing.

jennybc:
Here's the page we're talking about: https://vctrs.r-lib.org/reference/vector_recycling_rules.html

DavisVaughan:
It has since proved fairly useful to share, so it seems reasonable to make it external and give it a dedicated bullet in the pkgdown reference page

PR #1885, opened 2023-10-10
Make vector recycling rules its own FAQ

DavisVaughan:
Closes https://github.com/r-lib/vctrs/issues/1877

Tricky due to https://github.com/r-lib/pkgdown/issues/1876

~I was going to update our `vector_recycling_rules` page in vctrs to a true faq page with `@name` of `theory-faq-recycling` to match the other faq pages and use `@aliases vector_recycling_rules` to retain the old link. But since `vector_recycling_rules` won't get an html page since it is just an alias, that means the places in dplyr where we link to that page will be broken until dplyr rebuilds the site with the new version of vctrs, and I don't want that.~

~So instead I've kept `vector_recycling_rules` as the main `@name`.~

To work around the pkgdown issue, I used a manual redirect in `_pkgdown.yml` to ensure `vector_recycling_rules.html` always exists

Screenshot 2023-10-10 at 3 54 23 PM


olivroy:
Maybe the pkgdown redirects can help here? https://pkgdown.r-lib.org/news/index.html?q=redirects#new-features-2-0-0

R/faq-developer.R
```diff
@@ -12,6 +12,18 @@ NULL
 #' @name theory-faq-coercion
 NULL
 
+# The `@name` must be `vector_recycling_rules` to ensure a pkgdown page
+# continues to exist for `vector_recycling_rules` since other packages link to
+# this page and currently `@aliases` don't get a page on the pkgdown
+# site due to: https://github.com/r-lib/pkgdown/issues/1876
+#' FAQ - How does recycling work in vctrs and the tidyverse?
```
lionel-, line 19:
```suggestion

#' FAQ - How does recycling work in vctrs and the tidyverse?
```

lionel-: Just to clearly delineate comment from doc

lionel- (approved):

DavisVaughan:
@olivroy neat!

PR #1901, opened 2023-12-11
Update advice

olivroy:
Add the usethis way of adding standalone files in a package.


DavisVaughan:
Closing in favor of https://github.com/r-lib/rlang/pull/1711, thanks!

Issue #1823, opened 2023-03-24
Error in `vec_rank()`


mmprid01:
! Unsupported vctrs type `null`.
ℹ In file type-info.c at line 189.
ℹ This is an internal error that was detected in the vctrs package.
  Please report it at <https://github.com/r-lib/vctrs/issues> with a reprex and the full backtrace.
Backtrace:
    ▆
 1. ├─dplyr::ntile(...)
 2. │ └─dplyr::row_number(x)
 3. │   └─vctrs::vec_rank(x, ties = "sequential", incomplete = "na")
 4. └─rlang:::stop_internal_c_lib(...)
 5.   └─rlang::abort(message, call = call, .internal = TRUE, .frame = frame)

DavisVaughan:
Could you please turn this into a self-contained [reprex](http://reprex.tidyverse.org/) (short for minimal **repr**oducible **ex**ample)? It will help us help you if we can be sure we're all working with/looking at the same stuff.

If you've never heard of a reprex before, you might want to start by reading the tidyverse.org [help page](https://www.tidyverse.org/help/).

You can install reprex by running (you may already have it, though, if you have the tidyverse package installed):
```r
install.packages("reprex")
```

Thanks

PR #1916, opened 2024-03-07
Allow `NULL` in `vec_detect_complete()`

DavisVaughan:
Closes https://github.com/r-lib/vctrs/issues/1823

And better detect `NULL` columns and df-cols in data frames

### DIFF
--- a/src/complete.c
+++ b/src/complete.c
@@ -89,7 +89,8 @@ void vec_detect_complete_switch(SEXP x, R_len_t size, int* p_out) {
   case VCTRS_TYPE_raw: raw_detect_complete(x, size, p_out); break;
   case VCTRS_TYPE_list: list_detect_complete(x, size, p_out); break;
   case VCTRS_TYPE_dataframe: df_detect_complete(x, size, p_out); break;
-  case VCTRS_TYPE_scalar: r_stop_internal("Can't detect missing values in scalars.");
+  case VCTRS_TYPE_null: break;
+  case VCTRS_TYPE_scalar: stop_scalar_type(x, vec_args.empty, r_lazy_null);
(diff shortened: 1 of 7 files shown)


src/complete.c
```diff
@@ -169,12 +170,31 @@ void list_detect_complete(SEXP x, R_len_t size, int* p_out) {
...
   const SEXP* p_x = VECTOR_PTR_RO(x);
 
   for (r_ssize i = 0; i < n_cols; ++i) {
-    vec_detect_complete_switch(p_x[i], size, p_out);
+    col_detect_complete_switch(p_x[i], size, p_out);
+  }
+}
+
+static inline
+void col_detect_complete_switch(SEXP x, R_len_t size, int* p_out) {
```
DavisVaughan, line 186: It's a copy of `vec_detect_complete_switch()` with different branches for data frames and `NULL`

It mirrors how we do it in `vec_detect_missing()`

tests/testthat/_snaps/rank.md
```diff
@@ -1,3 +1,19 @@
...
+      Error:
+      ! This type is not supported by `vec_order()`.
+
+---
+
+    Code
+      vec_rank(NULL, incomplete = "na")
+    Condition
+      Error:
+      ! This type is not supported by `vec_order()`.
```
DavisVaughan, line 15: This used to throw the internal error seen in https://github.com/r-lib/vctrs/issues/1823

lionel-: Probably not worth it but would be nice if:

- Argument name was reported
- Argument value was revealed
- Error happened earlier and mentioned `vec_rank()`
- A `call` and `arg` argument would allow callers to take ownersip of the type checking

lionel- (approved):

Issue #1933, opened 2024-04-30
Non-API calls


lionel-:
Easy:

- `SET_ENCLOS()`: Used to create new environments, should be easy to fix: https://github.com/r-lib/rlang/issues/1707
- `PREXPR()/`R_PromiseExpr()`: https://github.com/r-lib/rlang/issues/1708

Unknown:

- `SET_S4_OBJECT()`/`UNSET_S4_OBJECT()`: Used for proxying / unproxying S4 objects. Can we do better?

Problematic:

- `SETLENGTH()`/`SET_TRUELENGTH()`: Used in various algorithms for expansible vectors. Also used in data.table for the same purpose. Would be nice to find a solution with R core.
- `STDVEC_DATAPTR()`: Part of the ALTREP API. Can we get this part of the experimental API that can be used in packages?

DavisVaughan:
I _think_ `STDVEC_DATAPTR()` can be replaced with `DATAPTR()` but I may be wrong

MichaelChirico:
Commenting to follow here. @ben-schwen already linked our corresponding issue.

MichaelChirico:
> SET_S4_OBJECT()/UNSET_S4_OBJECT(): Used for proxying / unproxying S4 objects. Can we do better?

Looked at this a little bit. I _think_ we can replace `unset_s4()` with `asS3()`.

I gave up on `r_mark_s4()`, the goal would be to handle that with `asS4()` on the R side.

Alternatively, we _might_ be able to use the C routine `asS4()` (untested)?

https://github.com/r-devel/r-svn/blob/main/src/main/objects.c#L1840

Jean-Romain:
> I _think_ `STDVEC_DATAPTR()` can be replaced with `DATAPTR()` but I may be wrong

@DavisVaughan tested in my package. It passes all unit tests with no NOTE on `devel`

DavisVaughan:
As of 2025-11-14, we've done a lot of work and are down to

```
     Found non-API calls to R: ‘PRVALUE’, ‘Rf_findVarInFrame3’,
       ‘SETLENGTH’, ‘SET_GROWABLE_BIT’, ‘SET_TRUELENGTH’, ‘TRUELENGTH’
```

[...]

PR #1940, opened 2024-06-15
Use API setter/unsetter for S4 objects

MichaelChirico:
Part of #1933.

This passes tests for me. I haven't explored any performance implications (this can induce a copy, right?).

### DIFF
--- a/src/utils.h
+++ b/src/utils.h
@@ -264,12 +264,12 @@ SEXP r_clone_referenced(SEXP x);
 SEXP r_call_n(SEXP fn, SEXP* tags, SEXP* cars);
 
 static inline SEXP r_mark_s4(SEXP x) {
-  SET_S4_OBJECT(x);
-  return(x);
+  x = Rf_asS4(x, (Rboolean) 1, 1);
+  return x;
 }
 static inline SEXP r_unmark_s4(SEXP x) {
-  UNSET_S4_OBJECT(x);
-  return(x);
+  x = Rf_asS4(x, (Rboolean) 0, 1);
+  return x;
 }
 
 bool r_has_name_at(SEXP names, R_len_t i);


lionel-:
>  I haven't explored any performance implications (this can induce a copy, right?).

It looks it only makes a copy when the object might be shared, which seems right.

Do you know what the `complete` argument does?

MichaelChirico:
Per [`?asS4`](https://rdrr.io/r/base/isS4.html)

> `complete` Optional, logical: whether conversion to S3 is completed. Not usually needed, but see the details section.

> `asS3` uses the value of `complete` to control whether an attempt is made to transform `object` into a valid object of the implied S3 class. If `complete` is `TRUE`, then an object from an S4 class extending an S3 class will be transformed into an S3 object with the corresponding S3 class (see `S3Part`). This includes classes extending the pseudo-classes `array` and `matrix`: such objects will have their class attribute set to `NULL`.

Can't say I fully understand the meaning there, nor does the source mean much to me:

https://github.com/r-devel/r-svn/blob/a068a0f52b9ae871212779b2c1571a999f23ef81/src/main/objects.c#L1853-L1869

Here's the three callsites in r-devel:

https://github.com/r-devel/r-svn/blob/a068a0f52b9ae871212779b2c1571a999f23ef81/src/main/arithmetic.c#L701
https://github.com/r-devel/r-svn/blob/a068a0f52b9ae871212779b2c1571a999f23ef81/src/main/objects.c#L1819
https://github.com/r-devel/r-svn/blob/a068a0f52b9ae871212779b2c1571a999f23ef81/src/library/base/R/lazyload.R#L83

I also only see a handful of usages on CRAN, one of them sets `/*complete=*/FALSE` with a comment that this was replacing `SET_S4_OBJECT()`:

https://github.com/search?q=org%3Acran+%2FasS4%5C%28%2F+lang%3AC%2B%2B&type=code

lionel-:
It looks like when complete is set it tries to return an internal slot in `R_getS4DataSlot()`?

The `mark_` functions were meant to mutate the objects. I see that this one returns its input but have you checked the call sites to verify that there is no assumption of mutation?

It seems like setting complete to 0 will be the least disruptive change though?

MichaelChirico:
> have you checked the call sites to verify that there is no assumption of mutation?

I haven't checked at all, sorry :)

I only saw #1933 linked to a related bug on {data.table} and wanted to share how we're approaching removing `{UN,}SET_S4_OBJECT`; I'll defer to your expertise on how {vctrs} is used for how best to proceed. On {data.table} side, we do not generally work well with S4 objects anyway, so don't need to consider the implications as carefully, IIUC.

PR #1946, opened 2024-07-16
Combine x and ... in min, max, range

### LINKED ISSUE #1372: Additional arguments in `min()` and `max()` methods for `vctrs_vctr`
Call `vec_c()` first?

brookslogan:
Resolves #1372.

Only call `vec_c(x, ...)` if dots are nonempty, in order to stay fast in the empty case.


brookslogan:
Sorry for the rewrite, fixed some commit metadata.

brookslogan:
Some notes:
- I think the main use case here is taking the `min` or `max` of two size-1 vectors.  If it's okay to deviate from mirroring base R expectations a bit, it would probably help to add some warnings or errors if `...` is used in other contexts, such as when it seems like `pmin` or `pmax` might be desired instead.  Happy to make these changes; please let me know if they belong.
- The [current check failure](https://github.com/r-lib/vctrs/actions/runs/9968263160/job/27543230701?pr=1946#step:5:158) doesn't seem like it's related to the changes in this PR.
- There doesn't seem to be a notable performance impact with the current approach, but the empty-dots optimization does seem important.

<details> <summary> Small-vctr benchmarks: </summary>

```r
# main
x <- new_vctr(c(1,2,4)); bench::mark(min(x), min_iterations = 1e5, max_iterations = 1e5)
#> # A tibble: 1 × 13
#>   expression      min   median `itr/sec` mem_alloc `gc/sec` n_itr  n_gc total_time result    
#>   <bch:expr> <bch:tm> <bch:tm>     <dbl> <bch:byt>    <dbl> <int> <dbl>   <bch:tm> <list>    
#> 1 min(x)       50.6µs   55.9µs    17597.        NA     8.98 99949    51      5.68s <vctrs_vc>
#> # ℹ 3 more variables: memory <list>, time <list>, gc <list>

# this PR
x <- new_vctr(c(1,2,4)); bench::mark(min(x), min_iterations = 1e5, max_iterations = 1e5)
#> # A tibble: 1 × 13
#>   expression      min   median `itr/sec` mem_alloc `gc/sec` n_itr  n_gc total_time result    
#>   <bch:expr> <bch:tm> <bch:tm>     <dbl> <bch:byt>    <dbl> <int> <dbl>   <bch:tm> <list>    
#> 1 min(x)       53.3µs   55.9µs    17608.        NA     8.98 99949    51      5.68s <vctrs_vc>
#> # ℹ 3 more variables: memory <list>, time <list>, gc <list>

# Without if-empty-dots optimization:
x <- new_vctr(c(1,2,4)); bench::mark(min(x), min_iterations = 1e5, max_iterations = 1e5)
#> # A tibble: 1 × 13
#>   expression      min   median `itr/sec` mem_alloc `gc/sec` n_itr  n_gc total_time result    
#>   <bch:expr> <bch:tm> <bch:tm>     <dbl> <bch:byt>    <dbl> <int> <dbl>   <bch:tm> <list>    
#> 1 min(x)        187µs    207µs     4737.        NA     10.4 99780   220      21.1s <vctrs_vc>
#> # ℹ 3 more variables: memory <list>, time <list>, gc <list>

# Testing with short integer vectors (ALTREP seqs or not) also yielded comparable performance.
```
</details>

brookslogan:
> I think the main use case here is taking the min or max of two [or maybe more] size-1 vectors. If it's okay to deviate from mirroring base R behavior a bit, it would probably help to add some warnings or errors if ... is used in other contexts, such as when it seems like pmin or pmax might be desired instead. Happy to make these changes; please let me know if they belong.

I've gone ahead and implemented these checks for `min` and `max` and made them produce hard errors.  I didn't do this for `range`, as I realized that the `range` generic just has `...` rather than `x, ...`, so I haven't made it produce errors like this, but have made sure that it combines the first of `...` (which in the `range.vctrs_vctr` implementation is called `x`), and any remaining args in `...`, if necessary.

brookslogan:
Rebased onto current main to see if it will help with what seemed like unrelated failing checks.  Coverage check still aborts immediately with
>  Error: Missing download info for actions/upload-artifact@v3

and I'm not sure how to resolve this.

PR #1954, opened 2024-10-22
doc: `new_rcrd()` requires a format method

JosiahParry:
Unline `new_vctr()`, `new_rcrd()` does not have a format() method. This is documented in the vignette at https://vctrs.r-lib.org/articles/s3-vector.html?q=record#record-style-objects.

However this is not discoverable from the reference page itself. 

I only found the vignette after googling the error.

I would love for there to be guidance in the manual page here.

### DIFF
--- a/R/type-rcrd.R
+++ b/R/type-rcrd.R
@@ -12,6 +12,9 @@
 #'   (same sizes), and contain uniquely named vectors (at least
 #'   one). `fields` is validated with [df_list()] to ensure uniquely
 #'   named vectors.
+#' @details
+#' A new rcrd class created with [new_rcrd()] will not have a [format()] method.
+#' New classes must implement their own [format()] method. See [Record-style objects](https://vctrs.r-lib.org/articles/s3-vector.html?q=record#record-style-objects) for more.
 #' @param ... Additional attributes
 #' @param class Name of subclass.
 #' @export


R/type-rcrd.R
```diff
@@ -12,6 +12,9 @@
 #'   (same sizes), and contain uniquely named vectors (at least
 #'   one). `fields` is validated with [df_list()] to ensure uniquely
 #'   named vectors.
+#' @details
```
lionel-, line 15: Can you move this below the description with some spacing please?

R/type-rcrd.R
```diff
@@ -12,6 +12,9 @@
...
+#' A new rcrd class created with [new_rcrd()] will not have a [format()] method.
```
lionel-, line 16: Maybe that should be a more general mention that a rcrd subclass is very bare bones and requires implementing a bunch of methods to make it useful, including `format()`, and then linking to the vignette as you did here?

R/type-rcrd.R
```diff
@@ -7,11 +7,17 @@
...
+#' @details
+#' Record-style objects created with [new_rcrd()] do not have a default
+#' [format()] method. See [Record-style objects](https://vctrs.r-lib.org/articles/s3-vector.html?q=record#record-style-objects
+#' for more details on implementing other methods.
 #'
 #' @param fields A list or a data frame. Lists must be rectangular
 #'   (same sizes), and contain uniquely named vectors (at least
 #'   one). `fields` is validated with [df_list()] to ensure uniquely
 #'   named vectors.
+#'
```
JosiahParry, line 20:
```suggestion
```

lionel-:
Thanks!

PR #1976, opened 2025-07-11
Aggressively inline in functions using a dictionary

DavisVaughan:
This is a fun one!

The TL;DR of this PR is that with some smart macro usage and inlining in `dict_hash_with()`, we can improve the performance of _all_ of our dictionary functions by up to 50%, when used on atomic vector types (or 1 column data frames). This includes:
- `vec_in()` and `vec_match()`, used all across the tidyverse
- `vec_group_loc()`, which powers `.by` and `group_by()`
- `vec_unique()`, `vec_unique_loc()`, `vec_unique_count()`, again used all across the tidyverse
- `vec_count()`
- `vec_duplicate_any/id/detect()`, 
- `vec_set_*()`, used in dplyr

---

While working on `vec_lookup()`, I noted that it's main source of performance bottleneck was in `vec_match()`. I've often been under the impression that our dictionary implementation could be improved, so I went spelunking with Instruments.

After spending some time looking at Instruments, I noted that almost _all_ of our time is in `dict_hash_with()`. Here is how we typically call `dict_hash_with()`:

https://github.com/r-lib/vctrs/blob/78d9f2b0b24131b5ce2230eb3c2c9f93620b10d9/src/dictionary.c#L426-L440

And here is its implementation

https://github.com/r-lib/vctrs/blob/78d9f2b0b24131b5ce2230eb3c2c9f93620b10d9/src/dictionary.c#L118-L151

What should stand out to you is that this is a HOT loop. It's actually a _double_ loop (where ideally we short circuit often in the inner loop when we either hit an empty bucket or recognize the value as something we've seen before). Performance wise, this means that anything we do in the inner `dict_hash_with()` loop should come under heavy scrutiny.

Instruments helped reveal two primary places we can improve:

## Inlining `dict_hash_with()` itself

Currently, `dict_hash_with()` itself is unable to be inlined because it is implemented in `dictionary.c` but exposed in `dictionary.h` because we call it in other files (totally reasonable). From what I understand, this means:
- We pay for a `dict_hash_with()` C function call at each iteration
- The compiler is unable to synergize optimizations between the outer loop and the inner loop inside `dict_hash_with()` 

## Inlining `d->p_equal_na_equal()`

We have to perform this equality check to know if we've found the same pre-existing hash, or if we have a case of a collision.

But we are truly shooting ourselves in the foot performance wise due to the fact that we store an arbitrary function pointer. This means:
- We have to pay the cost of the `d->p_equal_na_equal()` C function call at each iteration (very hot loop, many calls)
- The compiler is unable to fully inline this, which should in theory inline all the way through, say, `int_equal_na_equal()` being just `x == y`

It's worth noting that base R's `match()` is actually very smart about this, inlining aggressively for the 3 most common cases of integer, real, and string:
https://github.com/wch/r-source/blob/44a1cdd537c357a8bca3ad9626cfd8dc6b59d296/src/main/unique.c#L1234-L1286

## So how do we fix this?

In this PR, I've "smartly" broken the "generic" `dict_hash_with()` into typed variants, like `dbl_dict_hash_with()`, `int_dict_hash_with()`, etc. I've also moved it into `dictionary.h`, so it can be `static inline`.

That really is the main change. **There are no changes to the underlying implementation.** The rest is fall out of what that means.

No longer can we just call `dict_hash_with()` in our dictionary functions. We have to use our typical pattern of switch statement + macro loop to call the correct typed variant. Yes, this sucks. I hate macros as much as the next person. But here's the thing, IT'S WORTH IT. And I'll spend the rest of this PR convincing you, dear reader, that it is. It is also nothing new for us. In almost all of our generic vctrs functions we have one of these macro loops, they just suck to look at, but luckily this code is fairly static.

## An integer example

Look how much slower we are than base R with this realistic example:
- Integers
- `x` has 5 unique values, with a vector of size 10,000,000
- `table` contains the 5 unique values

``` r
library(vctrs)
set.seed(123)

x <- sample(5, 1e7, TRUE) + 0L
table <- sample(5) + 0L

bench::mark(match(x, table), iterations = 100)
#> # A tibble: 1 × 6
#> 1 match(x, table)   27.1ms   27.9ms      35.1    76.3MB     35.1

bench::mark(vec_match(x, table), iterations = 100)
#> # A tibble: 1 × 6
#> 1 vec_match(x, table)     49ms     50ms      19.5    76.3MB     19.5
```

The Instruments profile looks like this - notice the lack of inlining!

Screenshot 2025-07-11 at 11 21 45 AM

In this PR, I've "smartly" rewritten things to heavily encourage inlining, which results in:

Screenshot 2025-07-11 at 11 19 43 AM

And magically:

```r
bench::mark(vec_match(x, table), iterations = 100)
#> # A tibble: 1 × 6
#> 1 vec_match(x, table)   32.3ms   33.2ms      29.4    76.3MB     29.4
```

That's a 33% reduction in time, just from some macro usage! And much closer to base R!

It's worth noting that as the **number of unique values increases**, we beat `match()` before and after this PR, but the benefits still hold up!

- Integers
- `x` has 10,000 unique values, with a vector of size 10,000,000
- `table` contains the 10,000 unique values

``` r
cross::bench_branches(\() {
  library(vctrs)
  set.seed(123)

  x <- sample(10000, 1e7, TRUE) + 0L
  table <- sample(10000) + 0L

  bench::mark(
    match(x, table),
    vec_match(x, table),
    iterations = 20
  )
})
#> # A tibble: 4 × 7
#>   branch              expression        min  median `itr/sec` mem_alloc `gc/sec`
#>   <chr>               <bch:expr>    <bch:t> <bch:t>     <dbl> <bch:byt>    <dbl>
#> 1 main                match(x, tab… 140.2ms 141.2ms      7.08    76.5MB     7.08
#> 2 main                vec_match(x,…    53ms  54.1ms     18.5     76.5MB    18.5
#> 3 feature/dict-inline vec_match(x,…  34.3ms  34.8ms     28.5     76.5MB    28.5 
```

It's _also_ worth noting that as the **number of overall values increases**, we continue to do better there as well:

- Integers
- `x` has 10,000 unique values, with a vector of size 100,000,000
- `table` contains the 10,000 unique values

``` r
cross::bench_branches(\() {
  library(vctrs)
  set.seed(123)

  x <- sample(10000, 1e8, TRUE) + 0L
  table <- sample(10000) + 0L

  bench::mark(
    match(x, table),
    vec_match(x, table),
    iterations = 20
  )
})
#> # A tibble: 4 × 7
#>   branch              expression       min   median `itr/sec` mem_alloc `gc/sec`
#>   <chr>               <bch:expr>  <bch:tm> <bch:tm>     <dbl> <bch:byt>    <dbl>
#> 1 main                match(x, t…    1.39s     1.4s     0.714     763MB     2.86
#> 2 main                vec_match(… 525.33ms 532.82ms     1.87      763MB     5.62
#> 3 feature/dict-inline vec_match(… 334.14ms  335.5ms     2.96      763MB     8.87
```

## A double example

Doubles also benefit, but the story is a little more complicated. `dbl_equal_na_equal()` is somewhat complex and uses `dbl_classify()` which isn't currently inlined. If we inline `dbl_classify()` too, then we start to see serious benefits!

``` r
cross::bench_branches(\() {
  library(vctrs)
  set.seed(123)

  x <- sample(10000, 1e7, TRUE) + 0
  table <- sample(10000) + 0

  bench::mark(
    match(x, table),
    vec_match(x, table),
    iterations = 100
  )
})
#> # A tibble: 4 × 7
#>   branch              expression         min median `itr/sec` mem_alloc `gc/sec`
#>   <chr>               <bch:expr>      <bch:> <bch:>     <dbl> <bch:byt>    <dbl>
#> 1 main                match(x, table) 52.6ms   53ms      18.8   114.6MB    93.1 
#> 2 main                vec_match(x, t… 77.9ms 78.9ms      12.6    76.5MB     8.07
#> 3 feature/dict-inline vec_match(x, t… 64.7ms 65.8ms      15.1    76.5MB     9.66
```

Note how base R's `match()` is still faster than us. Looking at this example in Instruments reveals that `dbl_dict_hash_with()` is _still_ not inlined, because the compiler chose not to, likely because it thinks that this function is complex enough to not benefit from inlining.

Screenshot 2025-07-11 at 1 00 07 PM

But here's the thing, with gcc we can force inlining with `__attribute__((always_inline))`. If I tag `dbl_dict_hash_with` and the functions it calls with that, we get this, just as fast as base R:

```r
#> # A tibble: 2 × 6
#> 1 match(x, table)       52.2ms   52.6ms      18.5   114.6MB    13.4 
#> 2 vec_match(x, table)   52.3ms   52.6ms      18.9    76.5MB     8.49
```

I didn't go through with this yet (we need a portable version of it https://github.com/r-lib/pak/blob/80b07f7de40a9eff79279acc19d9557f8ed9d946/src/library/pkgdepends/src/tree-sitter/lib/src/stack.c#L15-L19) but we should talk about it.

The fun thing is that inlining `dbl_classify()` means that other functions benefit! Like `vec_compare()`:

``` r
cross::bench_branches(\() {
  library(vctrs)

  x <- sample(1e7) + 0
  y <- sample(1e7) + 0

  bench::mark(vec_compare(x, y, na_equal = TRUE))
})
#> # A tibble: 2 × 7
#>   branch              expression         min median `itr/sec` mem_alloc `gc/sec`
#>   <chr>               <bch:expr>      <bch:> <bch:>     <dbl> <bch:byt>    <dbl>
#> 1 feature/dict-inline vec_compare(x,… 11.4ms 11.7ms      84.9    38.2MB     89.4
#> 2 main                vec_compare(x,… 19.6ms 20.1ms      49.9    38.2MB     58.9
```

### A `dbl_equal_na_equal()` tweak

I've made a small tweak to `dbl_equal_na_equal()`, which also uses `dbl_classify()`. I've just lifted a fallthrough case into the switch branch it falls through from, and that also seemed to help performance of `vec_equal()`.

``` r
cross::bench_branches(\() {
  library(vctrs)
  set.seed(123)

  x <- sample(1e8) + 0
  y <- sample(1e8) + 0

  bench::mark(vec_equal(x, y, na_equal = TRUE))
})
#> # A tibble: 2 × 7
#>   branch              expression         min median `itr/sec` mem_alloc `gc/sec`
#>   <chr>               <bch:expr>       <bch> <bch:>     <dbl> <bch:byt>    <dbl>
#> 1 feature/dict-inline vec_equal(x, y,… 106ms  114ms      8.76     382MB     13.1
#> 2 main                vec_equal(x, y,… 133ms  133ms      7.50     382MB     22.5
```

## What about data frames?

Data frames of 1 column are unrolled into 1 atomic vector, so those benefit.

Data frames of >=2 columns don't tend to benefit too much from all this from what I can tell. The equality function for data frames is just too complex to want to be inlined, so there isn't much to gain here. But that's okay! We often use functions like `vec_match()` on low level integer proxies anyways, so having great performance there is first priority in my mind.

## What about other functions?

This isn't just about `vec_match()`. Anything that uses the dictionary benefits! Here's a `vec_unique()` example:

``` r
cross::bench_branches(\() {
  library(vctrs)
  set.seed(123)

  x <- sample(1e7, replace = TRUE) + 0L

  bench::mark(vec_unique(x), iterations = 10)
})
#> # A tibble: 2 × 7
#>   branch              expression         min median `itr/sec` mem_alloc `gc/sec`
#>   <chr>               <bch:expr>    <bch:tm> <bch:>     <dbl> <bch:byt>    <dbl>
#> 1 feature/dict-inline vec_unique(x)    195ms  198ms      4.98     278MB     4.98
#> 2 main                vec_unique(x)    297ms  302ms      3.19     278MB     3.19
```


lionel-:
wow, nice discovery!

> Currently, dict_hash_with() itself is unable to be inlined because it is implemented in dictionary.c but exposed in dictionary.h because we call it in other files

It's unable to be inlined from _other_ files right? From dictionary.c the header will only look like a forward declaration and shouldn't prevent inlining.

For other files, another possible route to get inlining that doesn't have the inconvenients of moving to macros (ease of editing and step-debuggability) or moving the definition in the header file (harder management of internal deps) is link-time optimisation (LTO).

DavisVaughan:
@lionel- unfortunately you really need the 2nd part as well, `Inlining d->p_equal_na_equal()`, to get maximal performance, and that pretty much requires the macros to avoid all indirection

PR #1978, opened 2025-07-12
Fix hard-coded link

lbm364dl:
I realize the hard-coded link's aim was to directly send to the section in the article, but I think we shouldn't do that. I guess there's no way to specify a section when using `vignette("s3-vector")`, but I still prefer to just name the section which the user should visit. If you don't like this and still prefer the hard-coded link, at least make it 

```https://vctrs.r-lib.org/articles/s3-vector.html#record-style-objects``` 

so that R CMD check doesn't fail.


DavisVaughan:
Thanks!

PR #1979, opened 2025-07-12
Update actions versions

lbm364dl:
Fix `test-coverage.yaml` action failing.
https://github.blog/changelog/2024-04-16-deprecation-notice-v3-of-the-artifact-actions/
Should get R CMD check running again together with https://github.com/r-lib/vctrs/pull/1978


DavisVaughan:
Thanks!

Issue #1980, opened 2025-07-14
`vec_c()` and `list_unchop(list(), indices = list())` should return `unspecified()` not `NULL`


DavisVaughan:
We have a somewhat odd behavior where if we know the `ptype`, we return an object of that `ptype`, otherwise we return `NULL`.

```
> vec_c()
NULL
> vec_c(.ptype = integer())
integer(0)
> vec_c(.ptype = unspecified())
<unspecified> [0]

> list_unchop(list(), indices = list())
NULL
> list_unchop(list(), indices = list(), ptype = integer())
integer(0)
```

It gets stranger when we add `list_unchop(size = )` into the mix, which now also enforces an output size. We have not noticed this before because when there are no inputs the output size has always been 0, which is consistent between either `NULL` or `unspecified()`.

```
# This is weird, I want a size 5 output
> list_unchop(list(), indices = list(), size = 5)
NULL

# Especially when you consider that if you specify ptype you get a size 5 output
> list_unchop(list(), indices = list(), ptype = integer(), size = 5)
[1] NA NA NA NA NA

# The first example should probably give this, not NULL
> list_unchop(list(), indices = list(), ptype = unspecified(), size = 5)
<unspecified> [5]
```

Just like `NULL`, an `unspecified()` should be able to combine with anything else, but it also gets to carry along its size information. So it's like a "sized identity"

@lionel- points out that we don't really want to leak `unspecified()` around very much though, so we have to be careful with this.

DavisVaughan:
Even though the most theoretically correct thing to return may be `unspecified(0)` rather than `NULL`, we typically try not to leak out `unspecified()` where possible, plus there is a lot of code out there that may rely on `NULL` as the output type.

Instead, we are going to say that if you specify `size` in `list_unchop()`, _then_ you will get `unspecified` as an output type if we can't determine the `ptype`, which makes sense because that's the only time a `size`d output really matters, so you end up with:

```r
vec_c()
#> NULL
list_unchop(list(), indices = list())
#> NULL
list_unchop(list(), indices = list(), size = 0)
#> unspecified[0] # Roughly `NULL`, good enough.
list_unchop(list(), indices = list(), size = 5)
#> unspecified[5] # Preserves size, good.
```

PR #1982, opened 2025-07-15
Add `size`, `default`, and `unmatched` to `list_unchop()`

DavisVaughan:
- [ ] Consider https://github.com/r-lib/vctrs/issues/1980#issuecomment-3074250868 before merging

Closes https://github.com/r-lib/vctrs/issues/1980

This PR adds `size` and `default` arguments to `list_unchop()`, making it more flexible and powerful as an engine that backs `case_when()` and the upcoming `vec_lookup()` idea.

It is also quite similar to `numpy.select()`
https://numpy.org/doc/stable/reference/generated/numpy.select.html

``` r
library(vctrs)

# You can specify a `size` larger than the sum of the `indices` lengths,
# allowing you to "partially fill" the new output object
list_unchop(
  x = list(1:2, 5:6),
  indices = list(1:2, 5:6),
  size = 7
)
#> [1]  1  2 NA NA  5  6 NA

# You can specify a `default` for the unfilled locations
list_unchop(
  x = list(1:2, 5:6),
  indices = list(1:2, 5:6),
  size = 7,
  default = 0L
)
#> [1] 1 2 0 0 5 6 0

# `default` may be length 1 or length `size`
list_unchop(
  x = list(1:2, 5:6),
  indices = list(1:2, 5:6),
  size = 7,
  default = 1:7
)
#> [1] 1 2 3 4 5 6 7

# You cannot use `size` without `indices`.
# This makes no sense, `x` would be laid out in sequence and you'd just have
# extra empty slots at the end.
list_unchop(
  x = list(1:2, 5:6),
  size = 7
)
#> Error in `list_unchop()`:
#> ! `indices` must be specified when `size` is specified.

# You cannot use `default` without `indices`.
# This makes no sense, `x` would be laid out in sequence and you'd never
# ever use the `default`
list_unchop(
  x = list(1:2, 5:6),
  default = 0
)
#> Error in `list_unchop()`:
#> ! `indices` must be specified when `default` is specified.

# You cannot use `default` without `size`.
# We need to know the `size` to recycle `default` against.
list_unchop(
  x = list(1:2, 5:6),
  indices = list(1:2, 5:6),
  default = 0
)
#> Error in `list_unchop()`:
#> ! `size` must be specified when `default` is specified.

# I have gone to great lengths to ensure you get good error messages
# when there are issues with `default`, particularly when computing the
# common type
list_unchop(
  x = list(1:2, 5:6),
  indices = list(1:2, 5:6),
  size = 7,
  default = "x"
)
#> Error in `list_unchop()`:
#> ! Can't combine <integer> and `default` <character>.
```


DavisVaughan (commented):
tmp

DavisVaughan:
Closed in favor of this PR using a branch against origin https://github.com/r-lib/vctrs/pull/1993

PR #1984, opened 2025-07-15
Implement `vec_recode_values()` and `vec_replace_values()`

DavisVaughan:
Branched from #1982 but I always forget I can't set a forked PR as the branch to compare against, so hold tight on looking at the diff until we merge #1982, or just look at the most recent commit.

- [ ] Tests
- [ ] Implement on the dplyr side


DavisVaughan:
Replaced by PR using origin branch https://github.com/r-lib/vctrs/pull/1996

PR #1990, opened 2025-07-31
Cast before recycling in `list_unchop()`

### LINKED ISSUE #1989: `list_unchop()` edge case issue with a single `NA` recycled to size 0
``` r
library(vctrs)

list_unchop(
  list("x", NA),
  indices = list(1L, integer())
)
#> Error in `list_unchop()`:
#> ! Can't convert `x[[2]]` <logical> to <character>.
```

We end up recycling the `NA` to size 0 after we've determined the common type (character) but before we actually do the actual casting to that common type. Once the `NA` is recycled to size 0, it looks like a type of `<logical>` rather than `<unspecified>`, and the cast fails.

This only affects `list_unchop()` where `indices` can affect the size of your `x` values (i.e. doesn't affect `vec_c()`).

DavisVaughan:
@lionel- can you please just focus on looking at the 2nd commit here? Ignore the first for now.

Closes https://github.com/r-lib/vctrs/issues/1989

Branched from https://github.com/r-lib/vctrs/pull/1982, again complicated by the fact that I'm using a fork. Diff will look better after we merge that one.

This fixes #1989 by ensuring that the `NA` is cast to `NA_character_` before it is recycled to size 0.

Casting before recycling is also a little better for memory reasons. When you pass in `1L` but the common type is `<double>` and you need to recycle the `1L` to the size of `index`, it is better to cast `1L` to `1` before recycling, rather than casting a large integer vector of `1L`s to a large double vector of `1`s after recycling.

``` r
cross::bench_branches(\() {
  library(vctrs)

  xs <- list(
    1,
    1L # Cast to double BEFORE recycling for best performance
  )
  indices <- list(
    1L,
    2:1000000 + 0L
  )

  bench::mark(list_unchop(xs, indices = indices))
})
#> # A tibble: 2 × 7
#>   branch             expression          min median `itr/sec` mem_alloc `gc/sec`
#>   <chr>              <bch:expr>       <bch:> <bch:>     <dbl> <bch:byt>    <dbl>
#> 1 fix/list-unchop-na list_unchop(xs,…    2ms 2.34ms      421.    15.3MB     720.
#> 2 main               list_unchop(xs,… 2.24ms  2.5ms      394.    19.1MB     403.
```

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,8 @@
 # vctrs (development version)
 
+* `list_unchop()` now works in an edge case with a single `NA` recycled to size
+  0 (#1989).
+
 * `list_unchop()` has gained new `size`, `default`, and `unmatched` arguments
   (#1982).
 
--- a/src/c-unchop.c
+++ b/src/c-unchop.c
@@ -219,6 +219,9 @@ r_obj* list_unchop(r_obj* xs,
   r_keep_loc out_names_pi;
   KEEP_HERE(out_names, &out_names_pi);
 
+  r_keep_loc x_pi;
+  KEEP_HERE(r_null, &x_pi);
+
   r_ssize i = 0;
 
   struct vctrs_arg* p_x_arg = new_subscript_arg(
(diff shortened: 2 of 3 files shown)


DavisVaughan:
Replaced by PR using origin branch https://github.com/r-lib/vctrs/pull/1994

PR #1993, opened 2025-08-01
Add `size`, `default`, and `unmatched` to `list_unchop()`

DavisVaughan:
- [ ] Consider https://github.com/r-lib/vctrs/issues/1980#issuecomment-3074250868 before merging

Closes https://github.com/r-lib/vctrs/issues/1980

This PR adds `size` and `default` arguments to `list_unchop()`, making it more flexible and powerful as an engine that backs `case_when()` and the upcoming `vec_lookup()` idea.

It is also quite similar to `numpy.select()`
https://numpy.org/doc/stable/reference/generated/numpy.select.html

``` r
library(vctrs)

# You can specify a `size` larger than the sum of the `indices` lengths,
# allowing you to "partially fill" the new output object
list_unchop(
  x = list(1:2, 5:6),
  indices = list(1:2, 5:6),
  size = 7
)
#> [1]  1  2 NA NA  5  6 NA

# You can specify a `default` for the unfilled locations
list_unchop(
  x = list(1:2, 5:6),
  indices = list(1:2, 5:6),
  size = 7,
  default = 0L
)
#> [1] 1 2 0 0 5 6 0

# `default` may be length 1 or length `size`
list_unchop(
  x = list(1:2, 5:6),
  indices = list(1:2, 5:6),
  size = 7,
  default = 1:7
)
#> [1] 1 2 3 4 5 6 7

# You cannot use `size` without `indices`.
# This makes no sense, `x` would be laid out in sequence and you'd just have
# extra empty slots at the end.
list_unchop(
  x = list(1:2, 5:6),
  size = 7
)
#> Error in `list_unchop()`:
#> ! `indices` must be specified when `size` is specified.

# You cannot use `default` without `indices`.
# This makes no sense, `x` would be laid out in sequence and you'd never
# ever use the `default`
list_unchop(
  x = list(1:2, 5:6),
  default = 0
)
#> Error in `list_unchop()`:
#> ! `indices` must be specified when `default` is specified.

# You cannot use `default` without `size`.
# We need to know the `size` to recycle `default` against.
list_unchop(
  x = list(1:2, 5:6),
  indices = list(1:2, 5:6),
  default = 0
)
#> Error in `list_unchop()`:
#> ! `size` must be specified when `default` is specified.

# I have gone to great lengths to ensure you get good error messages
# when there are issues with `default`, particularly when computing the
# common type
list_unchop(
  x = list(1:2, 5:6),
  indices = list(1:2, 5:6),
  size = 7,
  default = "x"
)
#> Error in `list_unchop()`:
#> ! Can't combine <integer> and `default` <character>.
```


src/c.c
```diff
@@ -274,36 +273,41 @@ r_obj* vec_c_fallback(r_obj* ptype,
                       struct r_lazy error_call) {
   r_obj* cls = KEEP(r_attrib_get(ptype, syms_fallback_class));
   bool implements_c = class_implements_base_c(cls);
-  FREE(1);
 
   if (implements_c) {
+    FREE(1);
     return vec_c_fallback_invoke(xs, name_spec, error_call);
-  } else {
```
DavisVaughan, line 281: No changes here, just using an early return in `vec_c_fallback()` to flatten a level of nesting, better matching the new `list_unchop_c_fallback()`

tests/testthat/_snaps/slice-chop.md
```diff
@@ -314,6 +314,19 @@
...
+    Code
+      x <- structure(foobar(1), attr_foo = "foo")
+      default <- structure(foobar(2), attr_foo = "bar")
+      list_unchop(list(x), indices = list(1), size = 2, default = default)
+    Condition
+      Error in `list_unchop()`:
+      ! Can't combine <vctrs_foobar> and `default` <vctrs_foobar>.
+      x Some attributes are incompatible.
+      i The author of the class should implement vctrs methods.
+      i See <https://vctrs.r-lib.org/reference/faq-error-incompatible-attributes.html>.
```
DavisVaughan, line 328: So much effort to get this error message to say `default` 😆

src/c-unchop.c
```diff
@@ -9,48 +10,107 @@ enum fallback_homogeneous {
 #include "decl/c-unchop-decl.h"
 
 
-static
 r_obj* list_unchop(r_obj* xs,
                    r_obj* indices,
+                   r_obj* default_,
```
DavisVaughan, line 15: `default` is a C keyword 🤷

src/c-unchop.c
```diff
@@ -9,48 +10,107 @@ enum fallback_homogeneous {
...
   if (indices == r_null) {
+    // Check if `default` and `size` were specified before reverting to `vec_c()`
+    if (default_ != r_null) {
+      r_abort("`indices` must be specified when `default` is specified.");
+    }
+    if (optional_r_ssize_is_some(size)) {
+      r_abort("`indices` must be specified when `size` is specified.");
+    }
+
     return vec_c(xs, ptype, name_spec, name_repair, p_error_arg, error_call);
```
DavisVaughan, line 35: I have a feeling there is going to come a time where we give some effort to combine `vec_c()` and `list_unchop()` into a single code path. They share a lot of utilities right now but the differences are _just_ annoying enough that I think it would be better if we just shoved everything into one big function to think about.

src/c-unchop.c
```diff
@@ -60,32 +120,93 @@ r_obj* list_unchop(r_obj* xs,
...
+  // Now merge `default` information into `xs` and `indices`
+  if (default_ != r_null) {
+    r_obj* default_index = KEEP(compute_default_index(indices, out_size));
+    default_ = KEEP(slice_default(default_, default_index, out_size));
+
+    xs = KEEP(push_default(xs, default_));
+    indices = KEEP(push_default_index(indices, default_index));
+
+    FREE(4);
   }
```
DavisVaughan, line 201: The general idea of how we handle `default` is to figure out where it should be inserted into the final result, i.e. the `default_index`, and to then append both (sliced) `default` and `default_index` to `xs` and `indices` respectively, allowing the rest of the implementation to proceed unchanged.

Originally I thought we'd use `default` instead of `vec_init(ptype)` for the output container, but I worry about consistency in some edge cases where `default` is some object that user provided that has some extra non-type related attributes attached to it. We don't want those in the output, so it's better to create a fresh container of type `ptype` and assign over the pieces of `default` we need. This method also doesn't work well with the fallback case, because there is no "container" that we fill there.

src/c-unchop.c
```diff
@@ -300,6 +446,318 @@ r_obj* list_unchop_fallback(r_obj* ptype,
...
+}
+
+// This is `vec_c_fallback()` with support for `default`.
+// Keep aligned with `vec_c_fallback()`!
+// TODO: Let `list_unchop()` own fallback helpers and
+// make `vec_c_fallback()` and friends thin wrappers
+// that set `default` to `r_null`, or merge `vec_c()`
+// and `list_unchop()` entirely!
+static
+r_obj* list_unchop_c_fallback(
```
DavisVaughan, line 478: Really I think that most of the fallback helpers should live in `c-unchop.c` not `c.c`, because `c-unchop.c` has the more general definition with a possible `default_` argument.

We can then just make wrappers around these to create the `c.c` versions that set `default_` to `r_null`, and then we won't have this duplication. But this was noisy enough that I didn't want to do it right now.

src/c-unchop.c
```diff
@@ -300,6 +446,318 @@ r_obj* list_unchop_fallback(r_obj* ptype,
...
+    const struct ptype2_opts ptype2_opts = {
+      .x = ptype,
+      .y = default_,
+      .p_x_arg = vec_args.empty,
+      .p_y_arg = p_default_arg,
+      .call = error_call,
+      .fallback = fallback_opts
+    };
+    int _;
+    vec_ptype2_opts(&ptype2_opts, &_);
```
DavisVaughan, line 522: Very large amount of effort went in to making sure we check the ptype against `default` here as well in a way that results in good errors using `p_default_arg`.

Can't just blindly merge `default` into `xs` ahead of time, because we want to use `p_default_arg` when type checking `default`, and `p_error_arg` when checking `xs`.

We have snapshot tests to ensure we keep emitting good errors in these weird fallback cases.

PR #1995, opened 2025-08-01
Implement `.name_spec = "inner"`

### LINKED ISSUE #1988: `name_spec` should support something like `zap_outer()`
A useful pattern is to provide outer names to `list_unchop()` for error message purposes, but you don't actually want them merged with the inner names and ending up in the resulting output.

For example, `case_when()` does something like this

```r
library(vctrs)

values <- list(
  `..1 (right)` = c(1, 2),
  `..2 (right)` = c("x", "y")
)
indices <- list(
  c(1, 2),
  c(3, 4)
)

list_unchop(
  values,
  indices = indices,
  error_arg = ""
)
#> Error in `list_unchop()`:
#> ! Can't combine `..1 (right)` <double> and `..2 (right)` <character>.
```

The problem is, when there aren't type errors this fails to due not having a `name_spec` provided

```r
library(vctrs)

values <- list(
  `..1 (right)` = c(1, 2),
  `..2 (right)` = c(3, 4)
)
indices <- list(
  c(1, 2),
  c(3, 4)
)

list_unchop(
  values,
  indices = indices,
  error_arg = ""
)
#> Error in `list_unchop()`:
#> ! Can't merge the outer name `..1 (right)` with a vector of length > 1.
```

I don't want the outer names in the output at all here, but I would like to retain the inner names if they exist.

None of the existing options for `name_spec` do what I'd like, and even if they did, they are slower than what I'd want, as I'm basically just asking that we ignore the outer names altogether and use the inner ones that exist with no extra manipulation on them. We don't need a function call or a glue spec for that.

``` r
library(vctrs)

# I don't actually want any names here
values <- list(
  `..1 (right)` = c(1, 2),
  `..2 (right)` = c(3, 4)
)
indices <- list(
  c(1, 2),
  c(3, 4)
)
list_unchop(
  values,
  indices = indices,
  error_arg = "",
  name_spec = function(outer, inner) as.character(inner)
)
#> 1 2 1 2 
#> 1 2 3 4
list_unchop(
  values,
  indices = indices,
  error_arg = "",
  name_spec = "{inner}"
)
#> 1 2 1 2 
#> 1 2 3 4
list_unchop(
  values,
  indices = indices,
  error_arg = "",
  name_spec = rlang::zap()
)
#> [1] 1 2 3 4

# I do want inner names here
values <- list(
  `..1 (right)` = c(a = 1, b = 2),
  `..2 (right)` = c(c = 3, d = 4)
)
indices <- list(
  c(1, 2),
  c(3, 4)
)
list_unchop(
  values,
  indices = indices,
  error_arg = "",
  name_spec = function(outer, inner) as.character(inner)
)
#> a b c d 
#> 1 2 3 4
list_unchop(
  values,
  indices = indices,
  error_arg = "",
  name_spec = "{inner}"
)
#> a b c d 
#> 1 2 3 4
list_unchop(
  values,
  indices = indices,
  error_arg = "",
  name_spec = rlang::zap()
)
#> [1] 1 2 3 4
```

I think what I want is something like

```r
structure(list(), "vctrs_zap_outer")
```

Which is like `rlang::zap()` but only zaps the outer names. 

The other benefit of this that I'd want to retain is that it should work with the fallback path, which `name_spec` typically cannot do (you get an error normally).

We'd need to support this in:
- vec_c
- list_unchop
- vec_rbind

DavisVaughan:
Closes #1988 
Branched from https://github.com/r-lib/vctrs/pull/1994

To ignore outer names (but use them in error messages), and retain inner names, see #1988 for rationale

Pretty easy implementation for this, and I'm very satisfied that this can function as a `.name_spec` that works for the fallback case as well. None of the others do, which can be a little frustrating sometimes when designing fully generic APIs.

---

There's only one thing to note. In `vec_rbind()` we previously disallowed this:

```r
vec_rbind(foo = data.frame(x = 1, row.names = "row"), .names_to = "id", .name_spec = zap())
```

i.e. when you set `.names_to = "id"` to consume the `foo` outer name and promote it to a column name, you can't also `zap()` the row names, which is supposed to zap both the outer and inner parts of the row names.

I think this is overly restrictive? I think we should just say "`.name_spec` only receives an outer name when `.names_to = NULL`, otherwise the outer name is consumed by the `.names_to` behavior". We don't have to restrict the behavior of `.name_spec` at all, and it is a little annoying that we do so right now.

You should be able to do the rbind call above. It's requesting:
- Promote the outer `foo` name to an element of a new `id` column
- Zap the inner `"row"` row name

I think this is a totally valid use case.

Similarly we are now going to allow all of these valid cases:

```r
# Zap `foo`, retain inner row names
vec_rbind(foo = data.frame(x = 1, row.names = "row"), .names_to = zap(), .name_spec = "inner")
# Promote `foo` to `id` column, retain inner row names
vec_rbind(foo = data.frame(x = 1, row.names = "row"), .names_to = "id", .name_spec = "inner")
# Let `.name_spec` handle `foo`, then `.name_spec` retains only inner row names
vec_rbind(foo = data.frame(x = 1, row.names = "row"), .names_to = NULL, .name_spec = "inner")
```

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,7 @@
 # vctrs (development version)
 
+* New `.name_spec = "inner"` option for `vec_c()`, `list_unchop()`, and `vec_rbind()`. This efficiently ignores all outer names, while retaining any inner names (#1988).
+
 * `list_unchop()` now works in an edge case with a single `NA` recycled to size 0 (#1989).
 
 * `list_unchop()` has gained new `size`, `default`, and `unmatched` arguments (#1982).
--- a/R/bind.R
+++ b/R/bind.R
@@ -44,16 +44,18 @@
 #'
 #'   `NULL` inputs are silently ignored. Empty (e.g. zero row) inputs
 #'   will not appear in the output, but will affect the derived `.ptype`.
-#' @param .names_to
-#' This controls what to do with input names supplied in `...`.
-#'   * By default, input names are [zapped][rlang::zap].
+#' @param .names_to This controls what to do with names on `...`:
 #'
-#'   * If a string, specifies a column where the input names will be
+#'   * By default, names on `...` are [zapped][rlang::zap] and do not appear
+#'     anywhere in the output.
+#'
+#'   * If a string, specifies a column where the names on `...` will be
 #'     copied. These names are often useful to identify rows with
 #'     their original input. If a column name is supplied and `...` is
 #'     not named, an integer column is used instead.
 #'
-#'   * If `NULL`, the input names are used as row names.
+#'   * If `NULL`, the outer names on `...` are instead merged with inner
+#'     row names on each element of `...` and are subject to `.name_spec`.
 #' @param .name_repair One of `"unique"`, `"universal"`, `"check_unique"`,
 #'   `"unique_quiet"`, or  `"universal_quiet"`. See [vec_as_names()] for the
 #'   meaning of these options.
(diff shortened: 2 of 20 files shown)


DavisVaughan:
Replaced by https://github.com/r-lib/vctrs/pull/2014

Issue #1951, opened 2024-08-23
vec_slice<- for POSIXlt does not clone


antalk2:
For a POSIXlt vector `x0`

After `x <- x0`  
`vec_slice(x,1) <- another`   
does not only change the value of `x`,   
it also changes the value of  `x0` .
Is this expected?

```{r}
library(vctrs)

# `[<-` works as expected (clones x)

x0_reference <- as.POSIXlt("2020-11-01")
x0           <- as.POSIXlt("2020-11-01")
another      <- as.POSIXlt("9020-11-01")
x <- x0

x[1] <- another
identical(x0, x0_reference) # TRUE
identical(x , x0          ) # FALSE

#
# `vec_slice<-` does not clone
#

x0_reference <- as.POSIXlt("2020-11-01")
x0           <- as.POSIXlt("2020-11-01")
another      <- as.POSIXlt("9020-11-01")
x <- x0

vec_is(x) # TRUE

vec_slice(x,1) <- another # also changes x0

identical(x0, x0_reference) # FALSE
identical(x , x0          ) # TRUE
```

DavisVaughan:
Minimal reprex

``` r
library(vctrs)

x <- as.POSIXlt("2020-11-01")

`vec_slice<-`(x, 1, NA)
#> [1] NA

x
#> [1] NA
```

<sup>Created on 2025-08-14 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

PR #2004, opened 2025-08-15
Revise ownership model to avoid modifying data frame proxies in place

DavisVaughan:
Closes https://github.com/r-lib/vctrs/issues/1951

Related to https://github.com/r-lib/vctrs/pull/1151
Related to https://github.com/r-lib/vctrs/pull/1884
Related to https://github.com/r-lib/vctrs/pull/1072
Related to https://github.com/r-lib/vctrs/pull/1062

Our ownership model has a long history of needing to be tweaked and revised over the years. This is unfortunately yet one more revision that needed to be made, because currently we modify non-data-frame objects that have a data frame proxy _in place_ 😱 with `vec_assign()`. This includes POSIXlt and vctrs_rcrd objects.

The TLDR of the problem is:
- `vec_proxy()` on a POSIXlt returns a "fresh" data frame, but note that the underlying columns aren't "fresh"
- `vec_owned(proxy)` makes it look like we recursively own that "fresh" data frame, even though we don't own the columns
- That propagates through all the column level assignment in `df_assign()`, so we overwrite the underlying columns in place 😱 

This doesn't happen on normal data frames because `vec_proxy()` doesn't return a "fresh" data frame, it returns the original one, so it doesn't look owned by `vec_owned()`.

I really didn't like `vec_owned()` which tries to dynamically figure out ownership, so I set out to remove that function entirely in favor of static ownership and it kind of expanded from there but it feels worth it.

---

The fix that I'm quite happy with is to introduce a 3 state ownership model

```c
enum vctrs_ownership {
  // No known ownership over this object
  //
  // If there are any references on this object, it will
  // be cloned before being modified
  VCTRS_OWNERSHIP_foreign,

  // Shallow ownership
  // - For atomics, we own the vector
  // - For lists (data frames), we own the list, but not the contents (columns)
  VCTRS_OWNERSHIP_shallow,

  // Deep ownership
  // We own the object recursively. Only used when we create it fully at C level.
  VCTRS_OWNERSHIP_deep
};
```

- In `vec_c()`, `vec_rbind()`, and `list_unchop()` where we create the object that we assign to, we get to use `VCTRS_OWNERSHIP_deep` which helps us override our "clone if this has references" check when assigning to the columns
- In `vec_cbind()` where we create the _container_ but the _columns_ come from the user, we get to use `VCTRS_OWNERSHIP_shallow`. We also use this in `df_map()`, same idea.
- Otherwise, when the data comes through the R boundary (FFI) we use `VCTRS_OWNERSHIP_foreign` and eventually use `r_clone_referenced()`

The most magical part of the states is what happens in `vec_bare_df_restore()` and `df_assign()` as we iterate over the cols. When we recursively restore or assign, we determine column ownership like this:

```c
    // During restoration, if we have deep ownership over `x` we can
    // propagate that ownership to the columns, otherwise we have no
    // known ownership over the columns
    enum vctrs_ownership col_ownership;
    switch (p_opts->ownership) {
    case VCTRS_OWNERSHIP_foreign: col_ownership = VCTRS_OWNERSHIP_foreign; break;
    case VCTRS_OWNERSHIP_shallow: col_ownership = VCTRS_OWNERSHIP_foreign; break;
    case VCTRS_OWNERSHIP_deep: col_ownership = VCTRS_OWNERSHIP_deep; break;
    default: r_stop_unreachable();
    }
```

Pretty simple and easy to read IMO!

---

I went and looked at benchmarks from old PRs related to this to confirm we aren't regressing anywhere.

We also have a few performance related tests that run on CI that are related to issues like this, and they also look good.

From https://github.com/r-lib/vctrs/pull/1151 with this PR

``` r
library(dplyr)

date_frame <- tibble(date = rep(lubridate::date("2020-01-01"), 100))
date_frames_to_bind <- rep(list(date_frame), 10000)
bench::mark(bind_rows(date_frames_to_bind))
#> disabled.
#> # A tibble: 1 × 6
#> 1 bind_rows(date_frames_to_bind)   48.4ms   50.9ms      19.1    9.01MB     53.4
```

From https://github.com/r-lib/vctrs/pull/1884 with this PR

``` r
library(vctrs)

make_list_of <- function(n) {
  df <- tibble::tibble(
    x = new_list_of(vec_chop(1:n), ptype = integer())
  )
  vec_chop(df)
}

xx <- make_list_of(4e3)

y <- list_unchop(xx)
y <- list_unchop(xx)
y <- list_unchop(xx)

bench::mark(list_unchop(xx), iterations = 20)
#> disabled.
#> # A tibble: 1 × 6
#> 1 list_unchop(xx)    149ms    154ms      6.42     110KB     42.7
```


src/slice-assign.c
```diff
@@ -32,10 +45,8 @@ r_obj* vec_assign_opts(r_obj* x,
   value = KEEP(vec_check_recycle(value, vec_size(index), opts.value_arg, opts.call));
 
   r_obj* proxy = KEEP(vec_proxy(x));
-  const enum vctrs_owned owned = vec_owned(proxy);
-  proxy = KEEP(vec_proxy_assign_opts(proxy, index, value, owned, &opts));
-
-  r_obj* out = vec_restore(proxy, x, owned);
+  proxy = KEEP(vec_proxy_assign_opts(proxy, index, value, &opts));
+  r_obj* out = vec_restore(proxy, x);
```
DavisVaughan, line 49: This is where my saga started:
- `vec_proxy()` on a POSIXlt returns a "fresh" data frame, but note that the underlying columns aren't "fresh"
- That data frame looks "owned" by us though, so `owned = VCTRS_OWNED_true`
- That propagates through all the column level assignment in `df_assign()`, so we overwrite the underlying columns in place 😱

src/proxy-restore.h
```diff
@@ -3,21 +3,35 @@
 
 #include "vctrs-core.h"
 
-
-r_obj* vec_restore(r_obj* x, r_obj* to, enum vctrs_owned owned);
-r_obj* vec_restore_default(r_obj* x, r_obj* to, enum vctrs_owned owned);
-
-r_obj* vec_restore_recurse(r_obj* x, r_obj* to, enum vctrs_owned owned);
```
DavisVaughan, line 10: I've changed the surface area of this API a bit:

- `vec_restore()` lost an `owned` arg and is only used for the "simple" non-recursive, non-owned case
- New `vec_restore_opts()` that we use most of the time
- `vec_restore_recurse()` is removed, use `vec_restore_opts()`

src/proxy-restore.c
```diff
@@ -198,13 +220,29 @@ r_obj* vec_bare_df_restore(r_obj* x,
...
+    }
+
+    const struct vec_restore_opts col_opts = {
+      .ownership = col_ownership,
+      .recursively_proxied = p_opts->recursively_proxied
+    };
+
     for (r_ssize i = 0; i < n_cols; ++i) {
-      r_obj* x_restored = vec_restore_recurse(v_x[i], v_to[i], owned);
+      r_obj* x_restored = vec_restore_opts(v_x[i], v_to[i], &col_opts);
```
DavisVaughan, line 240: Here's the `vec_bare_df_restore()` column loop

src/slice-assign.c
```diff
@@ -353,6 +365,27 @@ r_obj* df_assign(r_obj* x,
...
+
+  const struct vec_proxy_assign_opts col_proxy_assign_opts = {
+    .assign_names = p_opts->assign_names,
+    .ignore_outer_names = p_opts->ignore_outer_names,
+    .call = p_opts->call,
+    .x_arg = p_opts->x_arg,
+    .value_arg = p_opts->value_arg,
+    .ownership = col_ownership,
+    .recursively_proxied = p_opts->recursively_proxied
+  };
```
DavisVaughan, line 387: Here's the start of the `df_assign()` column loop

src/slice-assign.h
DavisVaughan, line 1: I've greatly simplified the API here. We haven't actually used many of these various forms of assignment, and it was cognitively overwhelming to have them all in front of me.

- Removed `vec_check_assign()`
- Removed `vec_assign_n()`
- Removed `vec_proxy_assign()`
- Removed `vec_proxy_check_assign()`

src/ownership.h
```diff
@@ -0,0 +1,18 @@
...
+#define VCTRS_OWNERSHIP_H
+
+#include "vctrs-core.h"
+#include "utils.h"
+
+// Wrapper around `r_clone_referenced()` that only attempts to clone if
+// we indicate that we don't own `x`
+static inline SEXP vec_clone_referenced(SEXP x, const enum vctrs_ownership ownership) {
+  switch (ownership) {
+  case VCTRS_OWNERSHIP_foreign: return r_clone_referenced(x);
```
lionel-, line 11: Would be nice to have a comment along the lines of: if foreign, we only clone if not referenced. The goal is to avoid cloning temporary values that are passed to vctrs functions.

And IIRC, temporary values only have 0 refcounts if we evaluate them from C via `.External2()`?

lionel- (approved):
I did not do a deep review but I'm on board with the general approach! I remember feeling uneasy about ownership in the past and with this revision I no longer do.

Except for the clone-referenced aspect, but I think we're conservative so that should be alright (we're more likely to copy something that is actually a temporary, than not copy something that isn't).

PR #2009, opened 2025-08-22
Implement `vec_assign(slice_value = TRUE)` and optimized logical `i` handling

DavisVaughan:
Two improvements in this PR

- `assignment_index_style` which allows for either integer _location_ indices or logical _condition_ indices. Each have their own optimized path inside of `vec_assign()`. This is auto detected for the user in `vec_assign()`. If it doesn't look like a simple logical condition index, we run it through `vec_as_location()` and use the location path. The logical condition index path is the new one.

- `assignment_slice_value` which says whether or not to optionally slice `value` by `i` before assignment as well.
    - If no, the default, assignment is done as expected, `x[i] <- value`.
    - If yes, assignment is done as an optimized form of `x[i] <- value[i]`, where `value[i]` is not actually materialized internally.
    - This is exposed as `vec_assign(slice_value = FALSE/TRUE)`

Note that `slice_value = TRUE` works regardless of the index style. These are orthogonal ideas. Figuring this out was a key part of this PR.

Taken together, this allows for optimal assignment in `dplyr::if_else()` and `dplyr::case_when()` style assignments where you have:
- LHSs that are logical condition indices, which we can now assign with directly rather than doing `which(i)` first
- RHSs that are length 1 or the length of the output, which can use `slice_value = TRUE` during assignment

This should allow for creation of the output with minimal extra allocations and overhead.

In a follow up PR I will create `list_multi_assign()`, which will eventually benefit from these optimizations and expose `slice_values = FALSE/TRUE`, and `vec_case_when()` and `vec_if_else()` will use that function directly. But getting these core assign details out of the way is the hard part.

---

Here's an example of the intended use case. `vec_assign()` with a logical index and a `value` that can benefit from `slice_value = TRUE`. This would be used `N` times by `dplyr::case_when()` where `N` is the number of provided cases.

``` r
library(vctrs)
set.seed(123)

x <- sample(1e7)
value <- sample(1e7)

# When `x > 100000`, put `value`, otherwise keep `x`
loc <- x > 100000

bench::mark(
  old = {
    # Improvements in `vec_assign()` logical `loc` handling
    vec_assign(x, loc, vec_slice(value, loc))
  },
  new = {
    # Additional optimization with `slice_value`
    vec_assign(x, loc, value, slice_value = TRUE)
  }
)
#> # A tibble: 2 × 6
#> 1 old          26.8ms   27.5ms      36.0   113.7MB     36.0
#> 2 new           7.1ms   7.49ms     114.     38.1MB     26.0

# This is CRAN vctrs with no optimizations
bench::mark(
  cran = {
    vec_assign(x, loc, vec_slice(value, loc))
  },
  iterations = 50
)
#> disabled.
#> # A tibble: 1 × 6
#> 1 cran         35.9ms   36.8ms      26.8     151MB     27.3
```

``` r
profmem::profmem(vec_assign(x, loc, value, slice_value = TRUE))
#> vec_assign(x, loc, value, slice_value = TRUE)
#> 
#> Memory allocations:
#> Number of 'new page' entries not displayed: 3
#>        what    bytes        calls
#> 3     alloc 40000048 vec_assign()
#> total       40000048
```

Here's an even more compelling example with a data frame, which has multiple columns so the overhead of slicing it is higher.

```r
library(vctrs)
set.seed(123)

dictionary <- data.frame(a = 1:10, b = 1:10, c = 1:10 + 0, d = letters[1:10])

x <- dplyr::slice_sample(dictionary, n = 1e7, replace = TRUE)
value <- dplyr::slice_sample(dictionary, n = 1e7, replace = TRUE)

# When `sample(1e7) > 100000`, put `value`, otherwise keep `x`
loc <- sample(1e7) > 100000

# This PR
bench::mark(
  old = {
    # Improvements in `vec_assign()` logical `loc` handling
    vec_assign(x, loc, vec_slice(value, loc))
  },
  new = {
    # Additional optimization with `slice_value`
    vec_assign(x, loc, value, slice_value = TRUE)
  },
  iterations = 20
)
#> # A tibble: 2 × 6
#> 1 old         129.5ms  130.7ms      7.65     493MB     29.1
#> 2 new          64.4ms   65.1ms     15.3      229MB     10.2


# This is CRAN vctrs with no optimizations
bench::mark(
  cran = {
    vec_assign(x, loc, vec_slice(value, loc))
  },
  iterations = 50
)
#> # A tibble: 1 × 6
#> 1 cran          147ms    148ms      6.75     531MB     169.
```

### DIFF
--- a/R/slice.R
+++ b/R/slice.R
@@ -1,21 +1,33 @@
 #' Get or set observations in a vector
 #'
 #' This provides a common interface to extracting and modifying observations
-#' for all vector types, regardless of dimensionality. It is an analog to `[`
-#' that matches [vec_size()] instead of `length()`.
+#' for all vector types, regardless of dimensionality. They are analogs to `[`
+#' and `[<-` that match [vec_size()] instead of `length()`.
 #'
 #' @inheritParams rlang::args_dots_empty
 #' @inheritParams rlang::args_error_context
 #'
 #' @param x A vector
-#' @param i An integer, character or logical vector specifying the
-#'   locations or names of the observations to get/set. Specify
-#'   `TRUE` to index all elements (as in `x[]`), or `NULL`, `FALSE` or
-#'   `integer()` to index none (as in `x[NULL]`).
-#' @param value Replacement values. `value` is cast to the type of
-#'   `x`, but only if they have a common type. See below for examples
-#'   of this rule. `value` must be recyclable to the size of `i` after `i`
-#'   has been converted to a valid integer location vector.
+#'
+#' @param i An integer, character or logical vector specifying the locations or
+#'   names of the observations to get/set. Specify `TRUE` to index all elements
+#'   (as in `x[]`), or `NULL`, `FALSE` or `integer()` to index none (as in
+#'   `x[NULL]`).
+#'
+#' @param value A vector of replacement values
+#'
+#'   `value` is cast to the type of `x`.
+#'
+#'   If `slice_value = FALSE`, `value` must be size 1 or the same size as `i`
+#'   after `i` has been converted to a positive integer location vector with
+#'   [vec_as_location()] (which may not be the same size as `i` originally).
+#'
+#'   If `slice_value = TRUE`, `value` must be size 1 or the same size as `x`.
+#'
+#' @param slice_value A boolean. If `TRUE`, the assignment proceeds as if you
+#'   had provided `vec_slice(x, i) <- vec_slice(value, i)`, but is optimized to
+#'   avoid materializing the slice of `value`.
+#'
 #' @param x_arg,value_arg Argument names for `x` and `value`. These are used
 #'   in error messages to inform the user about the locations of
 #'   incompatible types and sizes (see [stop_incompatible_type()] and
@@ -119,6 +131,31 @@
 #' # vector:
 #' x <- 1:3
 #' try(vec_slice(x, 2) <- 1.5)
+#'
+#' # Slicing `value` -------------------------------------------------
+#'
+#' # Sometimes both `x` and `value` start from objects that are the same length,
+#' # and you need to slice `value` by `i` before assigning it to `x`. This comes
+#' # up when thinking about how `base::ifelse()` and `dplyr::case_when()` work.
+#' condition <- c(TRUE, FALSE, TRUE, FALSE)
+#' yes <- 1:4
+#' no <- 5:8
+#'
+#' # Create an output container and fill it
+#' out <- vec_init(integer(), 4)
+#' out <- vec_assign(out, condition, vec_slice(yes, condition))
+#' out <- vec_assign(out, !condition, vec_slice(no, !condition))
+#' out
+#'
+#' # This is wasteful because you have to materialize the slices of `yes` and
+#' # `no` before they can be assigned, and you also have to validate `condition`
+#' # multiple times. Using `slice_value` internally performs
+#' # `vec_slice(yes, condition)` and `vec_slice(no, !condition)` for you,
+#' # but does so in a way that avoids the materialization.
+#' out <- vec_init(integer(), 4)
+#' out <- vec_assign(out, condition, yes, slice_value = TRUE)
+#' out <- vec_assign(out, !condition, no, slice_value = TRUE)
+#' out
 vec_slice <- function(x, i, ..., error_call = current_env()) {
   check_dots_empty0(...)
   .Call(ffi_slice, x, i, environment())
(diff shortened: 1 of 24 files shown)


R/slice.R
```diff
@@ -209,13 +216,25 @@ vec_assign <- function(x, i, value, ..., x_arg = "", value_arg = "") {
...
+      value <- vec_slice(value, existing)
+    }
+  } else {
+    # Convert logical condition `i` to integer location `i`,
+    # and slice `value` to match. Treats `NA` as `FALSE`.
+    i <- which(i)
+    if (value_size != 1L) {
+      # Only slice `value` if we aren't recycling it
+      value <- vec_slice(value, i)
+    }
```
DavisVaughan, line 237: In the fallback we convert to positive integer locations and slice `value` (like in my bench mark example)

src/c-unchop.c
```diff
@@ -104,6 +104,7 @@ r_obj* list_unchop(r_obj* xs,
   const struct vec_proxy_assign_opts unchop_proxy_assign_opts = {
     .ownership = VCTRS_OWNERSHIP_deep,
     .recursively_proxied = true,
+    .index_style = VCTRS_INDEX_STYLE_location,
```
DavisVaughan, line 107: This will eventually be an argument to `list_unchop()`, well `list_multi_assign()` really

src/slice-assign.c
```diff
@@ -37,13 +38,28 @@ r_obj* vec_assign_opts(r_obj* x,
...
+      x_names,
+      &location_opts
+    );
+    FREE(1);
+    break;
+  }
+  case VCTRS_INDEX_STYLE_condition: {
+    vec_check_condition_index(index, x_size, vec_args.i, assign_opts.call);
+    break;
+  }
```
DavisVaughan, line 59: It was pretty easy to share the `vec_assign()` code path and just branch in a few key places

src/slice-assign.c
```diff
@@ -250,54 +275,102 @@ r_obj* vec_proxy_assign_opts(r_obj* proxy,
...
+  CTYPE* out_data = DEREF(out);                                         \
+                                                                        \
+  for (r_ssize i = 0; i < index_size; ++i) {                            \
+    if (index_data[i] == 1) {                                           \
+      out_data[i] = value_data[VALUE_I];                                \
+    }                                                                   \
+  }                                                                     \
+                                                                        \
+  FREE(1);                                                              \
+  return out
```
DavisVaughan, line 319: Here's the core algorithm, very simple

src/slice-assign-array.c
```diff
@@ -1,271 +1,425 @@
...
+    k += VALUE_POST_SHAPE_INCREMENT;                                        \
+  }                                                                         \
+                                                                            \
+  UNPROTECT(n_protect);                                                     \
   return out
 
-#define ASSIGN_SHAPED_COMPACT(                                         \
+// -----------------------------------------------------------------------------
+
+#define ASSIGN_SHAPED_CONDITION_INDEX(                                 \
```
DavisVaughan, line 93: You can just trust me that this is right if you want, I've added tests, but it's actually much simpler than the location case, we don't even need the strides information.

src/vctrs-core.h
```diff
@@ -50,6 +50,32 @@ enum vctrs_ownership {
...
+ *   `vec_assign()`.
+ *
+ * - With logical condition vectors, `value` must be size 1 or the same size as
+ *   `index` itself, which matches `dplyr::case_when()` or `dplyr::if_else()`
+ *   style usage. These are used by `vec_assign_condition()`.
+ */
+enum vctrs_index_style {
+  VCTRS_INDEX_STYLE_location,
+  VCTRS_INDEX_STYLE_condition
+};
```
DavisVaughan, line 77: Pretty good docs and rationale, i think

src/c-unchop.c
```diff
@@ -104,6 +104,7 @@ r_obj* list_unchop(r_obj* xs,
   const struct vec_proxy_assign_opts unchop_proxy_assign_opts = {
     .ownership = VCTRS_OWNERSHIP_deep,
     .recursively_proxied = true,
+    .style = VCTRS_ASSIGNMENT_STYLE_location,
```
DavisVaughan, line 107: This will eventually be an argument to `list_unchop()`, well `list_multi_assign()` really

src/slice-assign.c
```diff
@@ -37,13 +38,28 @@ r_obj* vec_assign_opts(r_obj* x,
...
+    FREE(1);
+    break;
+  }
+  case VCTRS_ASSIGNMENT_STYLE_condition: {
+    vec_check_condition_index(index, x_size, vec_args.i, assign_opts.call);
+    break;
+  }
+  default: r_stop_unreachable();
+  }
+  KEEP(index);
```
DavisVaughan, line 62: It was pretty easy to share the `vec_assign()` code path and just branch in a few key places

src/slice-assign.h
```diff
@@ -4,10 +4,37 @@
...
+ *   `vec_assign()`.
+ *
+ * - With logical condition vectors, `value` must be size 1 or the same size as
+ *   `index` itself, which matches `dplyr::case_when()` or `dplyr::if_else()`
+ *   style usage. These are used by `vec_assign_condition()`.
+ */
+enum vctrs_assignment_style {
+  VCTRS_ASSIGNMENT_STYLE_location,
+  VCTRS_ASSIGNMENT_STYLE_condition
+};
```
DavisVaughan, line 31: Pretty good docs and rationale, i think

R/slice.R
```diff
@@ -1,21 +1,47 @@
...
+#'   `value` is cast to the type of `x`, but only if they have a common type.
+#'   See below for examples of this rule.
+#'
+#'   For `style = "location"`, `value` must be size 1 or the same size as `i`
+#'   after `i` has been converted to a positive integer location vector (which
+#'   may not be the same size as `i` originally).
+#'
+#'   For `style = "condition"`, `value` must be size 1 or the same size as
+#'   `i`.
+#'
```
lionel-, line 33: The "condition" terminology reads quite well.

So we have:

```
Index = Location | Condition
```

?

lionel- (approved):
Agreed that an extra argument works well. It makes it clearer we're really doing the same thing, only with a different indexing style.

PR #2011, opened 2025-09-02
Implement `list_combine()`

DavisVaughan:
Since the beginning, the code paths for `vec_c()` and `list_unchop()` have been _almost_ the same, but have been different enough that we left them separate. This has caused repeated headaches as we'd forget to change something in one when we changed the other.

This PR introduces `list_combine()`, a unified code path for both of these, and where development on the idea of "combine a list of vectors" will continue.

We are hoping to make `vec_c(!!!xs)` identical in performance to `list_unchop(xs)` (https://github.com/r-lib/rlang/issues/1820) so that in general we can recommend that instead of `list_unchop()`.

Then we'd recommend using `list_combine()` when you have more complicated cases where `indices` are needed - and the `indices` argument itself is going to be required here.


src/c-unchop.c
```diff
@@ -1,211 +1,15 @@
...
-enum fallback_homogeneous {
-  FALLBACK_HOMOGENEOUS_false = 0,
-  FALLBACK_HOMOGENEOUS_true
-};
-
 #include "decl/c-unchop-decl.h"
 
-
-static
-r_obj* list_unchop(r_obj* xs,
```
DavisVaughan, line 13: Both `list_unchop()` and `vec_c()` at the C level are now thin wrappers around `list_combine()`

tests/testthat/test-list-combine.R
```diff
@@ -0,0 +1,37 @@
+test_that("basic `list_combine()` works", {
```
DavisVaughan, line 1: Just a few `list_combine()` specific tests. Most new tests are for `list_unchop()`.

When we move to "officially" exporting and documenting `list_combine()`, I'll copy all tests over from `list_unchop()` and "freeze" the `list_unchop()` set of tests.

PR #2048, opened 2025-09-30
Fix two `vec_ptype_common()` error reporting index issues

DavisVaughan:
See snapshot tests. All changes there are improvements (i.e. more correct!) on the previous output. Super subtle issues here.

### DIFF
--- a/src/arg-counter.c
+++ b/src/arg-counter.c
@@ -23,17 +23,13 @@ struct counters* new_counters(r_obj* names,
   p_counters->next = 0;
 
   p_counters->names = names;
-  p_counters->names_curr = 0;
-  p_counters->names_next = 0;
 
   p_counters->curr_counter_arg_data = new_counter_arg_data(p_parent_arg,
                                                            &p_counters->curr,
-                                                           &p_counters->names,
-                                                           &p_counters->names_curr);
+                                                           &p_counters->names);
   p_counters->next_counter_arg_data = new_counter_arg_data(p_parent_arg,
                                                            &p_counters->next,
-                                                           &p_counters->names,
-                                                           &p_counters->names_next);
+                                                           &p_counters->names);
 
   p_counters->curr_counter = new_counter_arg(p_parent_arg, (void*) &p_counters->curr_counter_arg_data);
   p_counters->next_counter = new_counter_arg(p_parent_arg, (void*) &p_counters->next_counter_arg_data);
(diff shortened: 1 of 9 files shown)


src/arg-counter.c
```diff
@@ -61,7 +56,6 @@ void counters_shift(struct counters* p_counters) {
   // Swap the counters data
   SWAP(struct vctrs_arg, p_counters->curr_counter, p_counters->next_counter);
   SWAP(r_ssize*, p_counters->curr_counter_arg_data.i, p_counters->next_counter_arg_data.i);
-  SWAP(r_ssize*, p_counters->curr_counter_arg_data.names_i, p_counters->next_counter_arg_data.names_i);
 
   // Update the handles to `vctrs_arg`
   p_counters->curr_arg = (struct vctrs_arg*) &p_counters->curr_counter;
```
DavisVaughan, line 61: The first actual issue was the the line that is just below this, which is:

```
  // Update the current index
  p_counters->curr = p_counters->next;
```

was not also doing 

```
  p_counters->names_curr = p_counters->names_next;
```

so when we would index into the `names` at error reporting time, we'd have the wrong index!

I've fixed that by realizing that we don't need both `r_ssize curr/next` and `r_ssize names_curr/names_next`. They are tracking exactly the same thing, so we now just use `r_ssize curr/next` for everything. That immediately fixes the issue.

src/ptype2.c
```diff
@@ -38,11 +38,15 @@ r_obj* vec_ptype2_opts_impl(const struct ptype2_opts* opts,
...
+    // When `x` and `y` are `NULL`, keep using `x` name (1)
+    // When `x` is `NULL` but `y` isn't, switch to `y` name (0)
+    *left = y_type == VCTRS_TYPE_null;
     return vec_ptype2_from_unspecified(opts, x_type, y, y_arg);
   }
   if (y_type == VCTRS_TYPE_null) {
-    *left = x == r_null;
+    // When `x` and `y` are `NULL`, keep using `x` name (1)
+    // When `y` is `NULL` but `x` isn't, keep using `x` name (1)
+    *left = 1;
```
DavisVaughan, line 49: When I fixed the first problem, a snapshot test got worse because of another subtle issue.

When `y` is `NULL` in a `vec_ptype2(x, y)` comparison, we want to set `left = 1` to continue using the index/name of `x`. That isn't what was happening here with `*left = x == r_null`! 

Super subtle bug because you'd think these branches would be symmetric, but they are not.

tests/testthat/_snaps/bind.md
```diff
@@ -410,11 +410,11 @@
       # Integers as rows
       suppressMessages(with_memory_prof(vec_rbind_list(ints)))
     Output
-      [1] 2.77KB
+      [1] 2.74KB
```
DavisVaughan, line 413: Memory snapshots go down because we removed `names_curr` and `names_next` from the arg data structs

PR #2052, opened 2025-10-02
Detangle size check from condition index check

DavisVaughan:
It feels more like a type check, and an ambiguity resulted in a bad error message with arrays, which this less ambiguous approach fixes

### DIFF
--- a/src/if-else.c
+++ b/src/if-else.c
@@ -52,8 +52,9 @@ r_obj* vec_if_else(
   struct r_lazy error_call
 ) {
   obj_check_vector(condition, VCTRS_ALLOW_NULL_no, p_condition_arg, error_call);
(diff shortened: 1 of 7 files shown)


src/if-else.c
```diff
@@ -52,8 +52,9 @@ r_obj* vec_if_else(
   struct r_lazy error_call
 ) {
   obj_check_vector(condition, VCTRS_ALLOW_NULL_no, p_condition_arg, error_call);
+  check_condition_index(condition, p_condition_arg, error_call);
+
   const r_ssize size = r_length(condition);
```
DavisVaughan, line 57: Of note, not everyone needed the size check. `if_else()` takes the size from `condition`, so no check is required here. So that makes me feel even better about this.

PR #2059, opened 2025-10-08
Implement `list_of_transpose()`

### LINKED ISSUE #2055: Export `list_transpose()`?
It has different invariants than the purrr one, but it seems very well grounded in vctrs theory

https://github.com/DavisVaughan/cross/blob/95a27b17bfb123c8082c54eb972f703a8feccb4a/R/run.R#L249

- Can probably get away with early recycling/casting if we give `vec_interleave()` a `call` argument.
- Might not even really need a `size` argument? "All list elements must be size 1 or the same size" seems good enough? i.e. the same reason `vec_interleave()` doesn't have one.

```r
# Transpose a list of vectors
#
# Invariants:
#
# - `x` must be a list
# - Names of `x` are used in error messages, but are then discarded
# - Each element of `x` is coerced to the common type of the elements, or
#   `ptype`
# - Each element of `x` is recycled to the common size of the elements, or
#   `size`
# - Output is a list of size `size`. Each element is a vector of size `x_size`
#   and type `ptype`.
list_transpose <- function(
  x,
  ...,
  ptype = NULL,
  size = NULL,
  x_arg = caller_arg(x),
  error_call = caller_env()
) {
  obj_check_list(x, arg = x_arg, call = error_call)

  x_size <- vec_size(x)

  ptype <- vec_ptype_common(
    !!!x,
    .ptype = ptype,
    .arg = x_arg,
    .call = error_call
  )
  size <- vec_size_common(
    !!!x,
    .size = size,
    .arg = x_arg,
    .call = error_call
  )

  x <- vec_cast_common(
    !!!x,
    .to = ptype,
    .arg = x_arg,
    .call = error_call
  )
  x <- vec_recycle_common(
    !!!x,
    .size = size,
    .arg = x_arg,
    .call = error_call
  )

  # Don't want outer names after doing common type / size determination,
  # could use `.name_spec = "inner"` with new vctrs in `vec_interleave()`
  x <- unname(x)

  # Combine pieces of size `size` into one big vector via interleaving
  x <- vec_interleave(!!!x, .ptype = ptype)

  # Chop the one big vector into transposed pieces of size `x_size`
  x <- vec_chop(x, sizes = vec_rep(x_size, times = size))

  x
}
```

Ideal implementation is probably more like this. Need an internal `list_interleave()`.

```r
# TODO: Handle `NULL` list elements well? Is that even well defined?
# TODO: Handle empty `list()`. `list_interleave()` returns `NULL` but would
# be nice if this returned `unspecified()` for further use in `vec_chop()`.
# Need to handle divide by 0 size issue too.

list_transpose <- function(
  x,
  ...,
  ptype = NULL,
  x_arg = caller_arg(x),
  error_call = caller_env()
) {
  # - Does `obj_check_list()`
  # - Does `list_check_all_vectors()`
  # - Takes common size, recycles size 1 inputs
  out <- list_interleave(
    x,
    .ptype = ptype,
    .name_spec = "inner",
    .x_arg = x_arg,
    .error_call = error_call
  )

  x_size <- vec_size(x)
  out_size <- vec_size(out)
  elt_size <- out_size / x_size

  # Chop the one big vector into transposed pieces of size `x_size`
  out <- vec_chop(out, sizes = vec_rep(x_size, times = elt_size))

  out
}
```

DavisVaughan:
Closes #2055 

This feels quite useful to me, and took enough time thinking about the edge cases that I feel like it deserves its own function, even if it is quite a short wrapper around `list_interleave()` and `vec_chop()` at this point.

> [!CAUTION]  
> I'm a little worried about the name clashing with `purrr::list_transpose()`, but I'm not entirely sure what to do about that. It has different invariants than the purrr function, and I think it is a little closer to purer vctrs principles.

---

For the problem of `list(integer(), integer())` that we talked about over slack, I've realized that the correct thing for vctrs to do is to expose both `size` and `ptype` arguments. These control "the expected size and type of each element in the list".

I think these very elegantly let you recover the original object when you need to double-transpose in a generic way and have to worry about retaining the original size and type, even in the empty object case.

I think this is better than trying out some new `rectangular_list_of` object. I know it means that the object itself can't carry around the size and type info, but hopefully that's okay in terms of how people might do a double-transpose.

``` r
x <- list(integer(), integer())

# Input:
# - List size 2
# - Element size 0
# - Element type integer
# Output:
# - List size 0
# - Element size 2 (but no elements)
# - Element type integer (but no elements)
out <- list_transpose(x)
out
#> list()

# So then when you transpose again

# Input:
# - List size 0
# - Element size 0 (inferred)
# - Element type unspecified (inferred)
# Output:
# - List size 0
# - Element size 0
# - Element type unspecified
list_transpose(out)
#> list()

# Manually provide the original `size` and `ptype` to restore the original object

# Input:
# - List size 0
# - Element size 2 (provided)
# - Element type integer (provided)
# Output:
# - List size 2
# - Element size 0
# - Element type integer
list_transpose(out, size = vec_size(x), ptype = vec_ptype_common(!!!x))
#> [[1]]
#> integer(0)
#> 
#> [[2]]
#> integer(0)
```

The `size` argument is also nice when you have an expected element size, but each element is recyclable to that expected size and you happen to have all size 1 elements

``` r
x <- list(1, 2)

list_transpose(x, size = 3)
#> [[1]]
#> [1] 1 2
#> 
#> [[2]]
#> [1] 1 2
#> 
#> [[3]]
#> [1] 1 2
```


R/list-transpose.R
```diff
@@ -0,0 +1,171 @@
...
+  ...,
+  size = NULL,
+  ptype = NULL,
+  x_arg = caller_arg(x),
+  error_call = current_env()
+) {
+  check_dots_empty0(...)
+
+  # Disallow `NULL` entirely. These would break `vec_size()` invariants of
+  # `list_transpose()` if we simply drop them via `list_interleave()`.
```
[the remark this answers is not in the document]
DavisVaughan: I thought long and hard about this today and decided to add a `null` argument

- By default, we error on `NULL` elements
- If `null` is supplied, it replaces every `NULL` element with `null`
    - It must be size 1
    - It participates in common type determination alongside the elements of `x` and `ptype`


The thing that took me the longest to figure out was the restriction that `null` must be size 1 (as opposed to recycling alongside all of the elements of `x` and `size`). Without that, I kept running into various scenarios that were hard to explain and didn't make any sense. I have added some tests that call these out to remind us of why we have this restriction. In the end I think it makes sense, you can think of it like:

> `null` allows you to treat `NULL` as a size 1 element of some type

Which ends up being what you want for any use case I can think of.

``` r
library(vctrs)

# Error by default
try(list_transpose(list(1, NULL, 2:3, NULL)))
#> Error in list_transpose(list(1, NULL, 2:3, NULL)) : 
#>   `list(1, NULL, 2:3, NULL)[[2]]` must be a vector, not `NULL`.

list_transpose(list(1, NULL, 2:3, NULL), null = NA)
#> [[1]]
#> [1]  1 NA  2 NA
#> 
#> [[2]]
#> [1]  1 NA  3 NA

list_transpose(list(1, NULL, 2:3, NULL), null = 0L)
#> [[1]]
#> [1] 1 0 2 0
#> 
#> [[2]]
#> [1] 1 0 3 0

# Invariants on `null`:

# Must be size 1!
try(list_transpose(list(1, NULL, 2:3), null = 1:2))
#> Error in list_transpose(list(1, NULL, 2:3), null = 1:2) : 
#>   `null` must have size 1, not size 2.

# Does contribute to common type!
typeof(list_transpose(list(1L, NULL, 2L), null = 3)[[1]])
#> [1] "double"

# Some edge cases:

# With empty list
list_transpose(list(), null = 0L)
#> list()

# Input:
# - List size 0
# - Element size 2 (supplied)
# - Element type integer (from `null`)
# Output:
# - List size 2
# - Element size 0
# - Element type integer
list_transpose(list(), null = 0L, size = 2)
#> [[1]]
#> integer(0)
#> 
#> [[2]]
#> integer(0)

# Input:
# - List size 2
# - Element size 1 (`NULL` treated as size 1 when `null` is supplied)
# - Element type logical (from `null`)
# Output:
# - List size 1
# - Element size 2
# - Element type logical
list_transpose(list(NULL, NULL), null = NA)
#> [[1]]
#> [1] NA NA
```

R/list-transpose.R
```diff
@@ -0,0 +1,267 @@
...
+  obj_check_list(x, arg = x_arg, call = error_call)
+
+  if (is.object(x)) {
+    # The list input type should not affect the transposition process in any
+    # way. In particular, supplying a list subclass that doesn't have a
+    # `vec_cast.subclass.list` method shouldn't prevent the insertion of
+    # `list(null)` before the transposition. The fact that we must insert
+    # `list(null)` should be considered an internal detail.
+    x <- unclass(x)
+  }
```
DavisVaughan: I have tests for this, but basically I want this to work

```r
x <- structure(list(1, NULL), class = c("my_list", "list"))
list_transpose(x, null = 0)
```

But it doesn't due to `null` handling because we have to do this

``` r
vec_assign(x, vec_detect_missing(x), list(0))
#> Error in `vec_assign()`:
#> ! Can't convert <list> to <my_list>.
```

which doesn't work unless you bypass the class

We use `obj_check_list()` to check that it's a list, and after that we don't care about the class type.

R/list-transpose.R
```diff
@@ -0,0 +1,267 @@
...
+      x = null,
+      y = ptype,
+      x_arg = "null",
+      y_arg = "",
+      call = error_call
+    )
+  }
+
+  ptype
+}
```
DavisVaughan: Same trick used in `vec_recode_values()` and `vec_if_else()`

Basically I really want to avoid `vec_ptype_common(!!!x, null = null)` because that materializes a new list, and if `x` is really long then that is needlessly expensive.

PR #2086, opened 2025-10-24
Avoid `STDVEC_DATAPTR()` in ALTREP classes

DavisVaughan:
Part of https://github.com/r-lib/vctrs/issues/1933

When we are looking at the result of `R_altrep_data2()` we know we have a STRSXP that isn't ALTREP, so we can just use `STRING_PTR_RO()` on it to get at the underlying buffer.

There is some funny casting business that results from the ALTREP API being `void* Dataptr(x, bool writeable)` when it probably should have been two functions of `void* Dataptr(x)` and `const void* Dataptr_ro(x)`, but I think I asked Luke about this years ago for something else and he said what we are doing here is probably the right thing to do.


shikokuchuo:
`STDVEC_DATAPTR()` farewell 😢

Issue #2094, opened 2025-11-10
`vec_case_when()` type issue with `NA` and `default = <character>`


DavisVaughan:
``` r
vctrs::vec_case_when(
  conditions = list(TRUE),
  values = list(NA),
  default = "a"
)
#> Error in `vctrs::vec_case_when()`:
#> ! Can't combine <logical> and `default` <character>.

vctrs::vec_ptype2(NA, "a")
#> character(0)
```

I think this is because we do `vec_ptype_common()` of `values` which gives `logical()` which we then combine against `default`, but `logical()` doesn't look `unspecified()` so it errors.

PR #2100, opened 2025-11-11
Fix `list_combine(list(NA), default =)` and add `vec_ptype_common(.finalise=)`

DavisVaughan:
Closes https://github.com/r-lib/vctrs/issues/2094

Basic idea is that when we see this

```r
list_combine(
  x = list(NA),
  indices = list(1),
  size = 1,
  default = "a"
)
```

We need the common type across `!!!x` and `default`.

We "optimize" the ptype computation in `list_combine()` as

```r
vec_ptype2(
  vec_ptype_common(!!!x, .arg = x_arg),
  default,
  y_arg = default_arg
)
```

rather than as

```r
vec_ptype_common(!!!x, default_arg := default)
```

Because this avoids us having to reallocate a possibly very large list just to slip `default` in there. It also avoids forcing `default_arg`. I'm also not entirely sure how we'd use `x_arg` if we did that.

But in this particular `list_combine()` example, we get

```r
vec_ptype2(
  vec_ptype_common(x), # this returns a finalized `logical()`
  default, # this is `character()`
  y_arg = default_arg
)
# Error cant combine <logical> and <character>
```

So we actually need a way to tell `vec_ptype_common()` that we'd like to opt out of ptype finalization and promise to do it "manually" later on. Now that we've ensured that:
- `vec_ptype()` and `vec_ptype2()` _never_ finalize
- `vec_ptype_common()` _always_ finalizes

I think it makes some sense that we should have an internal way to make `vec_ptype_common()` act exactly like repeated `vec_ptype2()` for cases like this one.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,7 @@
 # vctrs (development version)
 
+* `vec_ptype_common()` has gained a `.finalise` argument that defaults to `TRUE`. Setting this to `FALSE` lets you opt out of prototype finalisation, which allows `vec_ptype_common()` to act like `vec_ptype()` and `vec_ptype2()`, which don't finalise. This can be useful in some advanced common type determination cases (#2100).
+
 * New `vec_pany()` and `vec_pall()`, parallel variants of `any()` and `all()` (in the same way that `pmin()` and `pmax()` are parallel variants of `min()` and `max()`).
 
 * The deprecated C callable for `vec_is_vector()` has been removed.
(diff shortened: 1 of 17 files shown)


DavisVaughan:
@lionel- and I decided it makes sense to expose `.finalise = TRUE` at the R level of `vec_ptype_common()` too. It would make it easier to prototype functions like the original `dplyr:::vec_case_when()`, gives us a place to link to `vec_ptype_finalise()`, and gives us an excuse in the `vec_ptype_common()` docs to mention that `vec_ptype()` and `vec_ptype2()` dont finalize, but `vec_ptype_common()` does by default, but you can opt out with `.finalise = FALSE` to make it act like `vec_ptype()` and `vec_ptype2()`.

Issue #1726, opened 2022-10-10
Consider "list vectors" like `<blob>`


DavisVaughan:
``` r
library(blob)
library(vctrs)

x <- blob(raw(1), raw(2))

# It looks like an explicit list
class(x)
#> [1] "blob"          "vctrs_list_of" "vctrs_vctr"    "list"

# Just goes through `new_list_of()`
new_blob
#> function (x = list()) 
#> {
#>     new_list_of(x, ptype = raw(), class = "blob")
#> }
#> <bytecode: 0x7fc22d499cb0>
#> <environment: namespace:blob>

# Which means this returns true even though blob is an "atomic vector" type
vec_is_list(x)
#> [1] TRUE
```

`<blob>` is a "list vector" type, i.e. it builds on a list-of but should be treated like an atomic vector. But it inherits from `"list"` explicitly, making it look like a list to vctrs.

All this means there are a few issues:

- `vec_is_list()` returns `TRUE` for blobs

- `[[<-.vctrs_list_of` isn't set up for "list vectors". i.e. we actually wanted to remove this code https://github.com/r-lib/vctrs/blob/6cf9bc5bc5ba9d6f528218b8991386e021ae3491/R/type-list-of.R#L148-L154, but blob relies on it right now so we can't do that. It relies on `x <- blob(raw(1)); x[[1]] <- NULL` inserting a `NULL` element into the blob object rather than shortening it, but I think that if blobs are supposed to work like atomic vectors then this should be an error like it would be with atomic vectors. They'd use `NA` assignment for this instead.

There are a few potential solutions:

- Allow `new_vctr(inherit_base_type = FALSE)` with list input, so that "list vectors" can be created. Then `new_list_of()` would also need this argument so blob can use it.

- Make blob use `new_rcrd()` instead. It would be a rcrd with 1 field, the list of blob objects. Then it automatically looks like an atomic vector and the list is an implementation detail. @lionel- reminded me this is actually documented too https://github.com/r-lib/vctrs/commit/29a6e0e0b38d6d36822875a1c9b26a8c3ab79cee.

krlmlr:
What's the motivation for changing blob's behavior? Why

> should (it) be treated like an atomic vector

?

krlmlr:
Is this maybe related to #1371?

lionel-:
From the point of view of vctrs, I think the hierarchy of defaults for R objects is:

1. Scalars
2. Atomic vectors
3. Recursive vectors

[...]

PR #2112, opened 2025-12-02
Teach `list_of` how to restrict both `size` and `ptype`

DavisVaughan:
@hadley just a surface level review from you please

---

So we can require a fully specified `list_of` in `list_of_transpose()` to power #2059 

This required touching pretty much every method of `list_of`. I greatly filled out the test suite for this class so that I could feel pretty confident about my changes.

There are only a few breaking changes related to this, I'm somewhat hopeful they aren't very impactful.

``` r
library(vctrs)

# Just the type (default behavior, backwards compatible)
x <- list_of(1, 2)
x
#> <list_of<double>[2]>
#> [[1]]
#> [1] 1
#> 
#> [[2]]
#> [1] 2

# Just the size (recycles on creation too)
y <- list_of(1, 2, .ptype = zap(), .size = 2L)
y
#> <list_of<any[2]>[2]>
#> [[1]]
#> [1] 1 1
#> 
#> [[2]]
#> [1] 2 2

# Both the size and type
z <- list_of(1, 2, .ptype = integer(), .size = 2L)
z
#> <list_of<integer[2]>[2]>
#> [[1]]
#> [1] 1 1
#> 
#> [[2]]
#> [1] 2 2

# Combining with self retains full type
vec_ptype_full(vec_c(x, x))
#> [1] "list_of<double>"
vec_ptype_full(vec_c(y, y))
#> [1] "list_of<any[2]>"
vec_ptype_full(vec_c(z, z))
#> [1] "list_of<integer[2]>"

# Coercion types towards the more lenient state
# (like how list_of + list falls back to list)
# (like how character + factor goes toward character)

# - y: Element type any, Element size 2
# - z: Element type int, Element size 2
# -  : Element type any, Element size 2 
vec_ptype_full(vec_c(y, z))
#> [1] "list_of<any[2]>"

# - x: Element type dbl, Element size any
# - y: Element type any, Element size 2
# -  : Element type any, Element size any, which is identical to list, so:
vec_ptype_full(vec_c(x, y))
#> [1] "list"
```

### DIFF
--- a/R/type-list-of.R
+++ b/R/type-list-of.R
@@ -90,6 +90,14 @@
 #' x <- list_of(1L, 2:3, .ptype = integer(), .size = 2)
 #' x
 #'
+#' # Setting an element to `NULL`
+#' x[2] <- list(NULL)
+#' x
+#'
+#' # Note that using `NULL` shortens the list, like a base R list
+#' x[2] <- NULL
+#' x
+#'
 #' # Combining a list_of with a list results in a list
 #' vec_c(list_of(1), list(2, "x"))
 #'
@@ -284,6 +292,7 @@ vec_ptype_abbr.vctrs_list_of <- function(x, ...) {
 as.list.vctrs_list_of <- function(x, ...) {
   list_of_unstructure(x)
 }
+
 #' @export
 as.character.vctrs_list_of <- function(x, ...) {
   # For compatibility with the RStudio Viewer. See tidyverse/tidyr#654.
--- a/man/list_of.Rd
+++ b/man/list_of.Rd
@@ -108,6 +108,14 @@ try(x[1] <- list(3:6))
 x <- list_of(1L, 2:3, .ptype = integer(), .size = 2)
 x
 
+# Setting an element to `NULL`
+x[2] <- list(NULL)
+x
+
+# Note that using `NULL` shortens the list, like a base R list
+x[2] <- NULL
+x
+
 # Combining a list_of with a list results in a list
 vec_c(list_of(1), list(2, "x"))
 


R/type-list-of.R
```diff
@@ -30,61 +119,100 @@ as_list_of <- function(x, ...) {
...
-as_list_of.vctrs_list_of <- function(x, ..., .ptype = NULL) {
-  if (!is.null(.ptype)) {
-    x <- unclass(x)
-    list_as_list_of(x, ptype = .ptype)
-  } else {
-    x
-  }
+as_list_of.vctrs_list_of <- function(x, ...) {
+  x
 }
```
DavisVaughan: Breaking change, removal of `.ptype` for this method. It will be a silent no-op if this arg was specified

R/type-list-of.R
```diff
@@ -188,21 +400,56 @@ vec_ptype2.vctrs_list_of.vctrs_list_of <- function(
...
+  }
+
+  if (is_null(x_size) || is_null(y_size)) {
+    size <- NULL
+  } else {
+    # No `vec_size2()`. This uses ALTREP to be efficient.
+    size <- tryCatch(
+      vec_size_common(seq_len(x_size), seq_len(y_size)),
+      vctrs_error_incompatible_size = function(cnd) NULL
+    )
```
DavisVaughan: This is admittedly a bit hacky...but it is fast? It never materializes the object.

NEWS.md
```diff
@@ -1,5 +1,22 @@
 # vctrs (development version)
 
+* The `[<-` and `$<-` methods for `<list_of>` no longer shorten the list when assigning `NULL` into the list. This is consistent with the existing `[[<-` method for `<list_of>`.
+
+* `as_list_of()` on an existing `<list_of>` no longer has a `.ptype` argument for changing the type on the fly, as this feels incompatible with the new system that allows restricting both the type and size. If you really need this, coerce to a bare list with `as.list()` first, then coerce back to a `<list_of>` using the `<list>` method of `as_list_of()`.
```
hadley: Does this have broader implications? Like if you have a date-time, what generic do you use to change the timezone? i.e. do we have a general philosophy of how you change one attribute while preserving the others?

R/type-list-of.R
```diff
@@ -1,26 +1,115 @@
...
+#'     errors if no vector inputs were provided.
+#'
+#'   - `.size = rlang::zap()` doesn't restrict the size.
+#'
+#' The default behavior infers the element type, but doesn't restrict the
+#' size in any way.
+#'
+#' Both `.ptype` and `.size` may be specified to restrict both the size and
+#' type of the list elements. You cannot set both of these to `rlang::zap()`,
+#' as that would be the same as a bare `list()` with no restrictions.
```
hadley: Does it error or does it just return a `list()`?

DavisVaughan:
Only 1 revdep issue from this https://github.com/pharmaR/riskmetric/issues/392

NEWS.md
```diff
@@ -1,5 +1,22 @@
 # vctrs (development version)
 
+* The `[<-` and `$<-` methods for `<list_of>` no longer shorten the list when assigning `NULL` into the list. This is consistent with the existing `[[<-` method for `<list_of>`.
```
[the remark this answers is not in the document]
DavisVaughan: Switched to `[[<-` aligning more with base R in https://github.com/r-lib/vctrs/pull/2112/commits/1d925ae25f9a9e5f4fec04f18cc104ab0e7540a5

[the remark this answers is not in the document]
DavisVaughan: Also discussed heavily in https://github.com/r-lib/vctrs/issues/1726

Issue #2127, opened 2025-12-17
Inconsistent type passed through in fallback path of `vec_ptype2()`


DavisVaughan:
Right here, notice how we pass in `x_type`. I think that is a typo, it should be `y_type`:
https://github.com/r-lib/vctrs/blob/29c5518bcb9477a69f0cb79a4dad699407d6a215/src/ptype2.c#L71

That results in this weird first result

``` r
vec_ptype_common_fallback(NULL, foobar(1))
#> numeric(0)
#> attr(,"class")
#> [1] "vctrs_foobar"

vec_ptype_common_fallback(foobar(1), NULL)
#> numeric(0)
#> attr(,"class")
#> [1] "vctrs:::common_class_fallback"
#> attr(,"fallback_class")
#> [1] "vctrs_foobar"

vec_ptype_common_fallback(unspecified(1), foobar(1))
#> numeric(0)
#> attr(,"class")
#> [1] "vctrs:::common_class_fallback"
#> attr(,"fallback_class")
#> [1] "vctrs_foobar"

vec_ptype_common_fallback(foobar(1), unspecified(1))
#> numeric(0)
#> attr(,"class")
#> [1] "vctrs:::common_class_fallback"
#> attr(,"fallback_class")
#> [1] "vctrs_foobar"
```

But if we change it to `y_type`, a whole bunch of tests break with snapshot diffs that look like

```
    Error in `list_combine()`:
-   ! Can't combine `x[[1]]` <vctrs_foobar> and `x[[2]]` <double>.
+   ! Can't combine `x[[1]]` <vctrs:::common_class_fallback> and `x[[2]]` <double>.
```

but I think what is happening here is that we never had good support for unwrapping the fallback object in our error messages, see https://github.com/r-lib/vctrs/issues/1981

PR #2128, opened 2025-12-17
Fix `vec_ptype2()` fallback path when LHS is `NULL`

DavisVaughan:
Closes https://github.com/r-lib/vctrs/issues/2127

The goal of this PR was to fix this small `x_type` -> `y_type` typo here

https://github.com/r-lib/vctrs/blob/29c5518bcb9477a69f0cb79a4dad699407d6a215/src/ptype2.c#L71

But since that is run as the very first iteration of almost every `vec_ptype_common()` call (because we start the reduction with `x = NULL, y = xs[[1]]`), it had a huge amount of fallout of weird buggy cases that relied on this.


DavisVaughan:
There are unfortunately more revdep failures we are going to have to deal with from this 

Some related to POSIXlt going towards POSIXct more often. Some related to ptype2 being called more often, I think.

lionel- (approved):

DavisVaughan:
We won't be able to do this right now due to the complexity of getting the sf methods compatible with this change, so we are going to have to wait on most of this until we can move all of the vctrs methods into sf itself first
https://github.com/r-spatial/sf/issues/2568

After that, it should hopefully be a bit easier to do this, since all of the vctrs methods will live in one place.

Ideally, sf won't need to use the vctrs fallback mechanism at all, allowing us to simplify things as we've tried to do here (i.e. by removing `vec_ptype_or_s3_fallback()` in favor of just `vec_ptype()`, avoiding the weird ptype2 computation there)

---

I'll try and pluck out some of the unrelated changes (POSIXlt methods, any extra tests, etc) into their own PR to get those in.

PR #2168, opened 2026-06-06
docs: route ptype2 incompatibility through default method

### LINKED ISSUE #1731: ptype2 howto shouldn't advise `stop_incompatible_type()`
In https://vctrs.r-lib.org/reference/howto-faq-coercion.html

Should recommend `vec_default_ptype2()` instead.

LeonidasZhak:
Fixes #1731.

## Summary
- update the simple-vector coercion how-to to return `vec_default_ptype2()` for attribute incompatibilities
- update the theory FAQ so it gives the same guidance
- keep the generated Rd files in sync

## Checks
- `Rscript -e 'tools::checkRd("man/howto-faq-coercion.Rd"); tools::checkRd("man/theory-faq-coercion.Rd")'` (only pre-existing non-ASCII Rd warnings)\n- `git diff --check`\n- `R CMD build --no-build-vignettes .`


DavisVaughan:
Assuming this is the same as https://github.com/tidymodels/broom/pull/1281#issuecomment-4642851835

PR #2169, opened 2026-06-06
docs: fix incorrect example in ?vec_duplicate

### LINKED ISSUE #1822: Typo in `?vec_duplicate`
According to the docs in the lines below;

https://github.com/r-lib/vctrs/blob/fe37f93388eef4526f9a8429146a89e258cb2160/R/dictionary.R#L129-L130

However, it gives to opposite answer to `duplicated()`:

``` r
library(vctrs)
x <- c(10, 10, 20, 30, 30, 40)

# Equivalent to `duplicated()`:
vec_duplicate_id(x) == seq_along(x)
#> [1]  TRUE FALSE  TRUE  TRUE FALSE  TRUE

duplicated(x)
#> [1] FALSE  TRUE FALSE FALSE  TRUE FALSE
```

<sup>Created on 2023-03-21 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

I think the example could be improved if the line read `vec_duplicate_id(x) != seq_along(x)` instead.

LeonidasZhak:
docs: fix incorrect example in ?vec_duplicate

The example for equivalence with `duplicated()` used `==` instead of `!=`.
`vec_duplicate_id(x) == seq_along(x)` returns TRUE for first occurrences,
while `duplicated(x)` returns TRUE for duplicates (not first occurrences).

Fixes #1822.

## Summary

Fix the example in `?vec_duplicate` that claims to show equivalence with `duplicated()`.

The example used `vec_duplicate_id(x) == seq_along(x)` but this returns `TRUE` for first occurrences (i.e., non-duplicates), while `duplicated(x)` returns `TRUE` for duplicates. The correct comparison is `!=`.

## Changes

- `R/dictionary.R`: Change `==` to `!=` in the example comment
- `man/vec_duplicate.Rd`: Regenerated from roxygen

## Verification

```r
library(vctrs)
x <- c(10, 10, 20, 30, 30, 40)
all.equal(vec_duplicate_id(x) != seq_along(x), duplicated(x))
#> TRUE
```

Fixes #1822.


DavisVaughan:
Assuming this is the same as https://github.com/tidymodels/broom/pull/1281#issuecomment-4642851835

